PR #47, opened 2016-10-31
feat: add a paper component. fixes #42

### LINKED ISSUE #42: Layer
To have consistent styling on Android and iOS we need  `Layer` component (have a better name?) which takes an `elevation` prop and renders a `View` with correct shadows on both Android and iOS.

We'll need this before working on any component with shadows.

satya164:
Since Google doesn't document the shadow values anywhere (it was documented earlier, but only upto 5dp), this is an approximation.

image

### DIFF
--- a/example/main.js
+++ b/example/main.js
@@ -6,7 +6,7 @@ import {
   NavigationProvider,
   StackNavigation,
 } from '@exponent/ex-navigation';
-import { ThemeProvider } from 'react-native-paper';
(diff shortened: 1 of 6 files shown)


ahmedlhanafy:
img_0302
screenshot_20161102-011056
@satya164 There is a noticable difference here, I think it can be more consistent

satya164:
Should be better

image

image

ahmedlhanafy:
@satya164 Looks good now 👍

PR #49, opened 2016-11-02
fix: follow line height and opacity guidelines for text

### LINKED ISSUE #23: Text
[MD Docs Link](https://material.google.com/style/typography.html#)
We will also need to load Roboto typeface for IOS

satya164:
Fixes #23 

image

### DIFF
--- a/src/components/Typography/Headline.js
+++ b/src/components/Typography/Headline.js
@@ -1,9 +1,10 @@
 /* @flow */
 
-import withTheme from '../../core/withTheme';
-import fonts from '../../styles/fonts';
 import createStyledText from './createStyledText';
 
-const Subheading = createStyledText(fonts.regular, 24);
-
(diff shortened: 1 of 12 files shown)


src/components/Typography/Headline.js
```diff
@@ -1,9 +1,10 @@
 /* @flow */
 
-import withTheme from '../../core/withTheme';
-import fonts from '../../styles/fonts';
 import createStyledText from './createStyledText';
 
-const Subheading = createStyledText(fonts.regular, 24);
-
-export default withTheme(Subheading);
+export default createStyledText('Caption', {
```
ahmedlhanafy: @satya164 This should be `Subheader`

satya164: @ahmedlhanafy sorry will send another PR. Lost these files once and screwed up while recreating

PR #52, opened 2016-11-09
Fixed Flow Errors

### LINKED ISSUE #50: Fix Flow errors

sreejithr:
I've bumped the version of flow in our project so that it matches with react-native's flow version (0.33). It fixes this weird error:

```
node_modules/react-native/Libraries/EventEmitter/EmitterSubscription.js:0
EmitterSubscription. Duplicate module provider
current provider. See: example/node_modules/fbemitter/lib/EmitterSubscription.js:0

node_modules/react-native/Libraries/EventEmitter/EventSubscription.js:0
EventSubscription. Duplicate module provider
current provider. See: example/node_modules/fbemitter/lib/EventSubscription.js:0

node_modules/react-native/Libraries/EventEmitter/EventSubscriptionVendor.js:0
EventSubscriptionVendor. Duplicate module provider
current provider. See: example/node_modules/fbemitter/lib/EventSubscriptionVendor.js:0
```

I'm new to flow. So, I haven't yet figured out why it is not picking up libraries `ex-navigation` and `react-native-vector-icons` from `example/node_modules`.

I ignored `@exponent/ex-navigation` in `.flowconfig` since the library had tons of flow errors. Didn't make sense to have those errors in our flow results.

Otherwise, the changes I've made are obvious enough I believe. Do let me know if everything is in order. Thank you.

Fixes #50.


satya164:
Thank you. You'll need to add following so that the `module not found` error is gone,

```
module.name_mapper='^exponent$' -> 'emptyObject'
module.name_mapper='^@exponent/ex-navigation$' -> 'emptyObject'
```

Same for any other module you ignored. i'll go to sleep now. Will review properly tomorrow.

src/components/Typography/createStyledText.js
```diff
@@ -30,10 +31,11 @@ export default function createStyledText<T>(fontFamily: string, fontSize: number
         children,
         style,
       } = this.props;
+      const color = theme ? theme.text : black;
```
satya164: `theme.text` should always be there. is it missing in the type definition? f yes, let's add it there

satya164:
> I haven't yet figured out why it is not picking up libraries ex-navigation and react-native-vector-icons from example/node_modules

It's because they aren't there as travis runs `npm install` only on the root directory. We might need to change the travis script to something like this so it installs for both,

```yml
language: node_js
cache:
  directories:
    - $HOME/.npm
    - $HOME/.yarn-cache
    - node_modules
node_js:
  - 'stable'
before_install:
  - npm i -g yarn
install:
  - yarn
  - cd example && yarn
script:
  - npm run flow -- check
  - npm run lint
```

sreejithr:
@satya164 I ran `npm install` inside the `examples` directory and tried. It was the same case. `Module not found` errors popping up. So even if Travis runs `npm install` inside `examples`, we'd have the same problem.

For now, I'm adding the name_mapper you proposed.

satya164:
@sreejithr That's weird. Probably we need to add those into the `devDependencies` of the main project itself.

cc @grabbou

sreejithr:
@satya164 Ya we can do that. But how about having flow setup inside the example project?

And if a `npm run-script flow` in the root directory won't run flow in subdirectories, we can make the `flow` command do `flow-bin` in both root and example directories.

satya164:
@sreejithr We'd have to set it up at 2 places then. Having flow in root project gives us errors and autocompletion when working on it.

sreejithr:
@satya164 I'll research about getting flow detect `node_modules` in subdirectories first.

satya164:
@sreejithr Thanks :)

PR #53, opened 2016-11-11
feat: add <Button /> component

satya164:
**TODO:**
- [ ] Ability to customize text color
- [ ] Disabled button
- [ ] Button with icon
- [ ] Button with loading indicator

image

image

### DIFF
--- a/src/components/Typography/Text.js
+++ b/src/components/Typography/Text.js
@@ -19,6 +19,12 @@ class Text extends Component<void, Props, void> {
     style: NativeText.propTypes.style,
   };
 
+  _root: any;
+
+  setNativeProps(...args) {
+    return this._root.setNativeProps(...args);
+  }
+
   render() {
     const {
       style,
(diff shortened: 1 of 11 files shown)


ahmedlhanafy:
@satya164 looking good i will test once I'm on my computer and merge

src/components/Typography/Text.js
```diff
@@ -19,6 +19,12 @@ class Text extends Component<void, Props, void> {
     style: NativeText.propTypes.style,
   };
 
+  _root: any;
+
+  setNativeProps(...args) {
```
ahmedlhanafy: What is the purpose of this? @satya164

satya164: @ahmedlhanafy `TouchableOpacity` and `TouchableHighlight` need it to have this as a direct child.

ahmedlhanafy: @satya164 won't they be available by extending `Component` only?

ahmedlhanafy: Cool

PR #54, opened 2016-11-11
feat: add <Card /> component

### LINKED ISSUE #5: Card
[MD Docs Link](https://material.google.com/components/cards.html#cards-usage)

satya164:
Fixes #5 

image

image

### DIFF
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -1,0 +1,108 @@
+/* @flow */
+
+import React, {
+  Component,
+  PropTypes,
+} from 'react';
+import {
+  Animated,
+  View,
+  TouchableWithoutFeedback,
+  StyleSheet,
+} from 'react-native';
+import Paper from './Paper';
+import { white } from '../styles/colors';
+import withTheme from '../core/withTheme';
+import type { Theme } from '../types/Theme';
+
+const AnimatedPaper = Animated.createAnimatedComponent(Paper);
+
+type DefaultProps = {
+  elevation: number;
+}
+
+type Props = {
+  elevation?: number;
+  children?: string;
+  onPress?: Function;
+  style?: any;
+  theme: Theme;
+}
+
+type State = {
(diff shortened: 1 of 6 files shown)


ahmedlhanafy:
@sarya164 why does this add `Button` component as well?

satya164:
@ahmedlhanafy you need to merge that first. then it won't anymore

src/components/Card.js
```diff
@@ -0,0 +1,108 @@
...
+      onPress,
+      style,
+    } = this.props;
+
+    return (
+      <AnimatedPaper elevation={this.state.elevation} style={[ styles.card, style ]}>
+        <TouchableWithoutFeedback
+          delayPressIn={0}
+          onPress={onPress}
+          onPressIn={onPress ? this._handlePressIn : undefined}
```
ahmedlhanafy: @satya164 Can you give me a use case for this please?

satya164: @ahmedlhanafy Many times cards acts as summary provide an entry point for more data, for example tapping on a card opens a screen. It's important to provide visual feedback if cards are tappable.

ahmedlhanafy: It's a nice effect but not sure if I saw something similar before

satya164: @ahmedlhanafy It's in the guidelines, called 'raised cards'

ahmedlhanafy: @satya164 I think this should be done for the desktop, not mobiles, this is the section that talks about that, correct me if I'm wrong
> On desktop, cards can have a resting elevation of 0dp and gain an elevation of 8dp on hover.

satya164: Note that I used an elevation of 6dp, not 8dp, that's because otherwise the shadow will get cut off due to to not enough margin. No way around that.

ahmedlhanafy: Okay cool

PR #58, opened 2016-11-11
feat: Add divider

ferrannp:
- It has an optional `customInset` prop too (default to 72 which aligns with the toolbar title)

For the dark theme, we will just need: `divider: color(white).alpha(0.12).rgbaString()`.

### DIFF
--- a/example/src/DividerExample.js
+++ b/example/src/DividerExample.js
@@ -1,0 +1,49 @@
+/* @flow */
+
+import React from 'react';
+import {
+  ListView,
+  StyleSheet,
+} from 'react-native';
+import {
+  Divider,
+  Subheading,
+  Colors,
+} from 'react-native-paper';
+
+const ds = new ListView.DataSource({ rowHasChanged: (r1, r2) => r1 !== r2 });
+const dataSource = ds.cloneWithRows([ 'Apple', 'Banana', 'Coconut', 'Lemon', 'Mango', 'Peach' ]);
+
+const DividerExample = () => {
+  return (
+    <ListView
+      style={styles.container}
+      dataSource={dataSource}
+      renderRow={(rowData, sectionId, rowId) => (
+          <Subheading
+            key={rowId}
+            style={styles.item}
+          >
+            {rowData}
+          </Subheading>
+        )
+      }
+      renderSeparator={(sectionId, rowId) => <Divider key={rowId}/>}
+    />
+  );
+};
+
+DividerExample.title = 'Divider';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: Colors.grey200,
+  },
+  item: {
+    paddingVertical: 8,
+    paddingHorizontal: 16,
+  },
+});
+
+export default DividerExample;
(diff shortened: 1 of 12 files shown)


example/src/DividerExample.js
```diff
@@ -0,0 +1,49 @@
...
+    />
+  );
+};
+
+DividerExample.title = 'Divider';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: Colors.grey200,
```
satya164: Let's use `rgba(0, 0, 0, .12)`

src/styles/DefaultTheme.js
```diff
@@ -17,6 +17,7 @@ export default {
     text: black,
     secondaryText: color(black).alpha(0.7).rgbaString(),
     disabled: color(black).alpha(0.5).rgbaString(),
+    divider: color(black).alpha(0.12).rgbaString(),
```
satya164: I don't think we should add this here. When we have a dark theme we should add a property called `dark` and components can decide what style to render. Otherwise there'll be too many options here.

src/types/Theme.js
```diff
@@ -8,6 +8,7 @@ export type Theme = {
     text: string;
     secondaryText: string;
     disabled: string;
+    divider: string;
```
satya164: Remove this

src/components/Divider.js
```diff
@@ -0,0 +1,46 @@
...
+
+Divider.propTypes = {
+  customInset: PropTypes.number,
+  hasInset: PropTypes.bool,
+  theme: PropTypes.object.isRequired,
+};
+
+const styles = StyleSheet.create({
+  divider: {
+    height: 1,
```
satya164: Though MD docs say 1dp, it's actually `StyleSheet.hairlineWidth`

src/components/Divider.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const Divider = (props: Props) => {
+  const { hasInset, customInset, theme } = props;
+  return (
+    <View
+      {...props}
+      style={[
+        styles.divider,
+        { backgroundColor: theme.colors.divider },
```
satya164: Hard code a background of `rgba(0, 0, 0, .12)`.

We'll need to change it when we've dark theme, but again, every component has to be changed to support dark theme.

src/components/Divider.js
```diff
@@ -0,0 +1,46 @@
...
+  StyleSheet,
+  View,
+} from 'react-native';
+
+import withTheme from '../core/withTheme';
+import type { Theme } from '../types/Theme';
+
+type Props = {
+  customInset?: number;
+  hasInset?: boolean;
```
satya164: I'd say we just have a `inset` prop which is a `boolean` and set the `marginLeft` to `72` if it's given. If people want custom margins it's easy to do with `style={{ marginLeft: 32 }}` or similar

src/components/Divider.js
```diff
@@ -0,0 +1,46 @@
...
+
+const Divider = (props: Props) => {
+  const { hasInset, customInset, theme } = props;
+  return (
+    <View
+      {...props}
+      style={[
+        styles.divider,
+        { backgroundColor: theme.colors.divider },
+        hasInset && { marginLeft: customInset || 72 },
```
satya164: Handle if style is passed through props

PR #60, opened 2016-11-12
feat: add color prop, don't add bg for non-raised button

satya164:
image

### DIFF
--- a/example/src/ButtonExample.js
+++ b/example/src/ButtonExample.js
@@ -24,23 +24,12 @@ export default class ButtonExample extends Component {
         <View style={styles.row}>
           <Button>Simple</Button>
           <Button primary>Primary</Button>
-          <Button
-            dark
-            style={styles.button}
-          >
-            Custom
-          </Button>
+          <Button color={Colors.pink500}>Custom</Button>
         </View>
         <View style={styles.row}>
           <Button raised>Raised</Button>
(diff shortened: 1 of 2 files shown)


satya164:
You can still add custom background color for non-raised buttons with`style` prop.

ahmedlhanafy:
Looking good @satya164 will review once I get home

ahmedlhanafy:
We will still need prop `fullWidth` I think

satya164:
> We will still need prop fullWidth I think

If your parent view is `flexDirection: 'column'` and has `alignItems: 'stretch'` it will become full width (which is default). It's not full width in example because I'm using `flexDirection: 'row'`. Parent can use the style `alignItems: 'space-around'` or anything other than `stretch` to not have full-width buttons.

ahmedlhanafy:
@satya164 I know that you didn't configure it and left it to the layout to be used, but I've seen this prop in different libraries makes it easier for beginners I guess

satya164:
@ahmedlhanafy I think it's better for people to get familiar with flexbox rather than learning different props for each component. We can document this, but I'm against adding a prop.

satya164:
ping @ahmedlhanafy

ahmedlhanafy:
I will be on my computer on a while

ahmedlhanafy:
@satya164 Great

PR #61, opened 2016-11-12
Update README.md

ahmedlhanafy:


ahmedlhanafy:
@satya164 the project is public so we have to add this so people don't assume that this package is ready

PR #66, opened 2016-11-20
Add Card.Cover, Card.Actions and Card.Content

satya164:
These components can adjust their styling automatically based on their siblings without any extra work on user's end. For example, the cover adjusts it's border radius, content adjusts it's padding, etc.

image


ahmedlhanafy:
@satya164 I think that the space between the action buttons is big, it needs to be a little bit smaller

ahmedlhanafy:
This is from MD docs @satya164 
screen shot 2016-11-20 at 2 44 40 pm

satya164:
@ahmedlhanafy it looks like that cause there's horizontal margin of 8 and the guidelines say that button has to be at least 88dp wide. Not sure how to handle this. Probably can decrease horizontal margin. But that has to in `<Button />` so separate PR

ahmedlhanafy:
8 in RN doesn't exactly translate to 8 dp @satya164

satya164:
@ahmedlhanafy why not? it's device independent pixel

ahmedlhanafy:
I have weird issues with it not exactly translating to what is shown in the MD docs

satya164:
@ahmedlhanafy I guess we couldn't do much about it then. anyways, I'll decrease margin for button to 4 in another PR.

ahmedlhanafy:
@satya164 perfect

sanketsahusoft:
@satya164 @ahmedlhanafy This is coming out really great! 👍

PR #68, opened 2016-11-20
feat: support theme overrides for themed components

satya164:
For example, this will allow you to do `<Button theme={{ roundess: 3 }}>Hello world</Button>`


ahmedlhanafy:
@satya164 Cool, that was needed 👍

PR #69, opened 2016-11-20
fix: adjust button and card styling

satya164:


src/components/Card/Card.js
```diff
@@ -65,7 +65,7 @@ class Card extends Component<DefaultProps, Props, State> {
 
   _handlePressIn = () => {
     Animated.timing(this.state.elevation, {
-      toValue: 6,
+      toValue: 8,
```
ahmedlhanafy: @satya164 why did you change this value?

satya164: @ahmedlhanafy it should be 8 according to guidelines. It was being clipped earlier somehow, so I had it at 6, but seems it's no longer being clipped.

ahmedlhanafy:
@satya164 I didn't check it on IOS earlier but the borders at the top are sharp
screen shot 2016-11-20 at 4 09 55 pm

satya164:
Hmm... weird, might need to wrap it in a view, gimme 5 mins.

ahmedlhanafy:
And I still think that the margin between buttons doesn't match MD docs @satya164

satya164:
@ahmedlhanafy it's not the margin, the buttons need to be at least 88dp wide according to docs. The screenshot you shared has longer text, so the gap is not visible.

ahmedlhanafy:
@satya164 Take a look at this, I don't know if this should be a different component. Or we should remove this restriction from transparent buttons 
screenshot_20161120-161507

ahmedlhanafy:
@satya164 Also the margin between the image and the title is weird on IOS

satya164:
> Also the margin between the image and the title is weird on IOS

Fixed in the theme PR

About the margin, I don't we can do much here unless you remove the min width. Even if you set margin to 0, it'll still not match the screenshot.

Sometimes google doesn't follow it's own guidelines, so hard to say based on screenshots.

ahmedlhanafy:
> About the margin, I don't we can do much here unless you remove the min width. Even if you set margin to 0, it'll still not match the screenshot.

> Sometimes google doesn't follow it's own guidelines, so hard to say based on screenshots.
I guess we can break it here then because it looks off for me, and we will need this in the dialogs as well so we can't have it looking like this

ahmedlhanafy:
@satya164 I think we will still need to `add minWidth` but a smaller one to prevent something like this
screen shot 2016-11-20 at 4 51 48 pm
It should look like this 
screen shot 2016-11-20 at 4 53 02 pm

ahmedlhanafy:
@satya164 I think we should return the old margin again which is 8 instead of 4

satya164:
what should be the value though?

satya164:
@ahmedlhanafy 4 is a better margin actually, all the issues here are because google hasn't documented this properly https://material.google.com/components/buttons.html#buttons-flat-buttons

ahmedlhanafy:
> what should be the value though?

@satya164 honestly I don't know but you can experiment with different values

ahmedlhanafy:
@satya164 Looks much better now

PR #73, opened 2016-11-20
feat: add GridView component. fixes #44

### LINKED ISSUE #44: GridView
A responsive GridView is extremely useful in apps.

satya164:
image

### DIFF
--- a/src/components/GridView.js
+++ b/src/components/GridView.js
@@ -1,0 +1,142 @@
+/* @flow */
+
+import React, {
+  PureComponent,
+  PropTypes,
+} from 'react';
+import {
+  Dimensions,
+  ListView,
+  StyleSheet,
+  View,
+} from 'react-native';
+import { grey200 } from '../styles/colors';
+
+type Layout = {
+  width: number;
+}
+
+type Props = {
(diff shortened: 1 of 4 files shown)


src/components/GridView.js
```diff
@@ -0,0 +1,142 @@
...
+import { grey200 } from '../styles/colors';
+
+type Layout = {
+  width: number;
+}
+
+type Props = {
+  dataSource: ListView.DataSource;
+  spacing: number;
+  getNumberOfColumns: (width: number) => number;
```
ahmedlhanafy: @satya164 I don't like this name that much why not only `numberOfColumns` and why not give the option of accepting a number directly, what if someone wants to have fixed number of columns

ahmedlhanafy: @satya164 Yeah I get it but I used libraries before that accept either a function or a number but I'm kinda against that, we can leave it as a function but I think the name should be `numberOfColumns`

satya164: It might not be very good looking, but if it's a function the name should resemble a function name. Otherwise the prop is not obvious. Take any prop name in React Native, if it accepts a function It's always named like that.

PR #85, opened 2016-11-25
feat: add TextInput component

ahmedlhanafy:


ahmedlhanafy:
screen shot 2016-11-26 at 12 07 24 am

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  _setRef = (c: Object) => (this._root = c);
+
+  _handleFocus = () => {
+    this.setState({ focused: true }, () => {
+      global.requestAnimationFrame(() => this.animationTimingStart.start());
+    });
+  }
+
+  _handleBlur = () => {
+    if (!this.state.text || this.state.text === '') {
```
satya164: `!this.state.text || this.state.text === ''` is redundant, if `this.state.text === ''` is empty `!this.state.text` is always gonna be true

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+      this.setState({ focused: false }, () => {
+        global.requestAnimationFrame(() => this.animationTimingEnd.start());
+      });
+    } else {
+      this.setState({ focused: false });
+    }
```
satya164: This function should just do `this.setState({ focused: false });`. All animation logic should be handled in `componentDidUpdate`

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+    this.animationTimingEnd = Animated.timing(this.animatedVal,
+      {
+        toValue: 0,
+        duration: 180,
+        easing: Easing.easing,
+      });
+
+    this.state = {
+      focused: false,
+      text: props.value || '',
```
satya164: Need to handle the case when value is updated via props with `componentWillReceiveProps`

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  bottomLineContainer: {
+    marginBottom: 4,
+    height: 1.2,
+  },
+  bottomLine: {
+    position: 'absolute',
+    left: 0,
+    right: 0,
+    top: 0,
+    height: 1.2,
```
satya164: I think this should be 2dp and 4dp, which would be `2 / PixelRatio.get()` and `4 / PixelRatio.get()`

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  label: {
+    position: 'absolute',
+    left: 1,
+    top: 15,
+    fontSize: 16,
+  },
+  input: {
+    fontSize: 16,
+    minHeight: 44,
+    marginBottom: -8,
```
satya164: The text padding looks too small here. Check the spec for measurements https://material.google.com/components/text-fields.html#text-fields-labels

example/src/TextInputExample.js
```diff
@@ -0,0 +1,89 @@
...
+
+  static title = 'TextInput';
+
+  state = {
+    text: '',
+    text2: '',
+    text3: '',
+    text4: '',
+    text5: '',
+  };
```
satya164: new line

example/src/TextInputExample.js
```diff
@@ -0,0 +1,89 @@
...
+  validateEmail(email: string) {
+    const re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+    return re.test(email);
+  }
```
satya164: new line

example/src/TextInputExample.js
```diff
@@ -0,0 +1,89 @@
...
+
+  state = {
+    text: '',
+    text2: '',
+    text3: '',
+    text4: '',
+    text5: '',
+  };
+  validateEmail(email: string) {
+    const re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
satya164: `_validateEmail = email => /\S+@\S+/.test(email);` should be enough, nobody should use such an email regex anyways

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+        </View>
+      );
+    }
+
+    return (
+            <View style={[ containerStyle ]}>
+                {floating && <AnimatedText style={[ styles.label, labelStyle ]}>
+                  {floatingLabelText}
+                </AnimatedText>}
+                <View style={{ marginHorizontal: 2 }}>
```
satya164: This is a static style, could be moved to `StyleSheet.create`

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+            {underlineShow && editable &&
+              <Animated.View
+                style={[ styles.bottomLine, bottomLineStyle ]}
+              />}
+        </View>
+      );
+    }
+
+    return (
+            <View style={[ containerStyle ]}>
```
satya164: Unnecessary array for style

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+} from 'react-native';
+import Text from './Typography/Text';
+import withTheme from '../core/withTheme';
+import { red500 } from '../styles/colors';
+
+const AnimatedText = Animated.createAnimatedComponent(Text);
+
+type Props = {
+  containerStyle?: any;
+  editable?: boolean;
```
satya164: This looks the same as normal input, so no way to distinguish a disabled field. We need a `disabled` prop which applies special styling too

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  errorText?: string;
+  errorTextColor?: string;
+  floatingLabelColor?: string;
+  floatingLabelFixed?: boolean;
```
satya164: I think we don't need this prop. it's like disabling ripple effect on a button, someone might want it, but it doesn't add any value

src/components/TextInput.js
satya164: This should call `this.props.onBlur` if passed

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  animationTimingStart: Function;
+
+  animationTimingEnd: Function;
+
+  _root: any;
+
+  _setRef = (c: Object) => (this._root = c);
+
+  _handleFocus = () => {
+    this.setState({ focused: true }, () => {
```
satya164: This should call `this.props.onFocus` if passed

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  state: State;
+
+  componentDidMount() {
+    const text = this.props.value;
+    if (text || text !== '') {
+      this._handleFocus();
+    }
+  }
+
+  animatedVal: Object;
```
satya164: Remove new lines after single line declaration, I like to separate code blocks with new lines, but for single lines, doesn't add any value

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+      text: props.value || '',
+    };
+  }
+
+  state: State;
+
+  componentDidMount() {
+    const text = this.props.value;
+    if (text || text !== '') {
+      this._handleFocus();
```
satya164: This should call `_handleFocus` since it's not a focus event. All animation logic should go to a single function called `_animate` or something, which can be called from `componentDidMount` and `componentDidUpdate`

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  isFocused(...args) {
+    return this._root.isFocused(...args);
+  }
+
+  clear(...args) {
+    return this._root.clear(...args);
+  }
+
```
satya164: I believe there are `focus` and `blur` methods too

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+    Platform,
+} from 'react-native';
+import Text from './Typography/Text';
+import withTheme from '../core/withTheme';
+import { red500 } from '../styles/colors';
+
+const AnimatedText = Animated.createAnimatedComponent(Text);
+
+type Props = {
+  containerStyle?: any;
```
ahmedlhanafy: To indicate that this isn't the `inputStyle`

example/src/TextInputExample.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: Colors.white,
+    padding: 8,
+  },
+  inputContainerStyle: {
+    marginVertical: 4,
```
satya164: Should have 8px margin on all sides

src/components/TextInput.js
ahmedlhanafy: What if user set `editable` and `disabled` I think this will be rather confusing. There should one prop that decides whether the text input is disabled or not, and since the underlying component has one why not use it

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  editable?: boolean;
+  errorText?: string;
```
ahmedlhanafy: No hint is very confusing since hint is the name of the placeholder in almost all libraries

src/components/TextInput.js
ahmedlhanafy: I don't believe so, adding the ability to fix the position of the floating label is really essential in my opinion and I think it will be needed

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+  errorTextColor?: string;
+  floatingLabelColor?: string;
+  floatingLabelFixed?: boolean;
+  floatingLabelText?: string;
+  inputStyle?: any;
+  onChangeText?: Function;
+  underlineShow?: boolean;
```
ahmedlhanafy: Yeah but why do need to have two components and animation (`scaleX`) when the user don't want it from the beginning and I think it makes the component more configurable and easy to customize instead of doing hacks

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+
+/**
+ * TextInputs allow users to input text.
+ * **Usage:**
+ * ```
+ * const MyComponent = () => (
+ *  <TextInput
+ *    placeholder='Hint text'
+ *    floatingLabelText='Email'
+ *    value={this.state.text}
```
ahmedlhanafy: My bad :/

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+ *
+ * @extends TextInput props https://facebook.github.io/react-native/docs/textinput.html#props
+ *
+ */
+
+class TextInput extends Component<DefaultProps, Props, State> {
+
+  static propTypes = {
+    /**
+    * The style of the container element.
```
ahmedlhanafy: Please check RN docs or any UI library

src/components/TextInput.js
ahmedlhanafy: I don't see them in RN docs

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+      transform: [ { scaleX: (errorText && errorText !== '') ? 1 : this.animatedVal } ],
+    };
+
+    if (Platform.OS === 'android') {
+      bottomLineComponent = underlineShow ? (
+        <Animated.View style={[ styles.bottomLineContainer, bottomLineStyle, { transform: [], height: 1.3 } ]}/>
+      ) : null;
+    } else {
+      bottomLineComponent = (
+        <View style={[ styles.bottomLineContainer, { backgroundColor: inactiveColor } ]}>
```
ahmedlhanafy: To have it colored with the gray color and once the component focuses, scale the other one with the selected color in the x direction from 0 to 1. To have the effect in Material design spec

src/components/TextInput.js
ahmedlhanafy: Which one?

example/src/TextInputExample.js
ahmedlhanafy: Check the parent component it adds padding of 8, this was done to better show off the different use cases

src/components/TextInput.js
satya164: RN docs have multiline paragraphs for prop docs, we have short single-line description.

src/components/TextInput.js
satya164: @ahmedlhanafy but why only on iOS?

src/components/TextInput.js
satya164: @ahmedlhanafy the underline height

src/components/TextInput.js
satya164: @ahmedlhanafy in which circumstances is a fixed label better than an animated one?

src/components/TextInput.js
satya164: @ahmedlhanafy then why is `inputStyle` called `inputStyle` and not style?

I don't like it very much here that we expose implementation details here. It's an implementation detail that the input has container. What if we change the implementation? The component will break then? because people have changed the style of the container?

src/components/TextInput.js
ahmedlhanafy: @satya164 What should be handled then?

src/components/TextInput.js
ahmedlhanafy: Android does have a lot of issues when it comes to this animation specifically

src/components/TextInput.js
ahmedlhanafy: How is that it should be the same height but with different colors only @satya164

src/components/TextInput.js
ahmedlhanafy: @satya164 This is exactly the measurements, check the `translateYAnimation` values it's 8dp on top and bottom of the input

src/components/TextInput.js
```diff
@@ -0,0 +1,353 @@
...
+            {underlineShow && editable &&
+              <Animated.View
+                style={[ styles.bottomLine, bottomLineStyle ]}
+              />}
+        </View>
+      );
+    }
+
+    return (
```
ahmedlhanafy: Done

example/src/TextInputExample.js
ahmedlhanafy: The container has 8px padding the 4px is between components @satya164

src/components/TextInput.js
satya164: @ahmedlhanafy If the user wants a non-animated static text input why not just use something like,

```
<Subheading style={{ color: 'blue' }}>
<TextInput placeholder='' />
```

It's not even a lot of code to do this themselves. We already provide a `Subheader` component which has exactly this purpose, adds a heading above something, a paragraph or a component. This is nothing specific to text input.

Giving options to the use is fine, but by not having the prop we're not limiting anything for the user. It's entirely possible without this prop.

> It's not about what's better

For me it's always about what's better. I don't care about anything more than use experience. Developers do weird stuff when building UIs. It's the job of UI libraries to make them good no matter what.

src/components/TextInput.js
satya164: image

src/components/TextInput.js
satya164: @ahmedlhanafy What do you mean? The MD design guidelines show that the focused input has a thicker underline

example/src/TextInputExample.js
satya164: @ahmedlhanafy yes I know that, that's not how it should be

image

src/components/TextInput.js
```diff
@@ -0,0 +1,351 @@
...
+import withTheme from '../core/withTheme';
+import { red500 } from '../styles/colors';
+import type { Theme } from '../types/Theme';
+
+
+const AnimatedText = Animated.createAnimatedComponent(Text);
+
+type Props = {
+  disabled?: boolean;
+  errorText?: string;
```
satya164: What about changing this to something else? This has lots of other use cases than just showing errors.

src/components/TextInput.js
```diff
@@ -0,0 +1,351 @@
...
+    {
+      toValue: 0,
+      duration: 180,
+      easing: Easing.easing,
+    }).start();
+
+  _handleFocus = () => {
+    this.setState({ focused: true });
+    if (this.props.onFocus) {
+      this.props.onFocus();
```
satya164: Do `onFocus` and `onBlur` get called with an event argument? In that case we need to pass that to user's prop.

src/components/TextInput.js
satya164: What issues did you face? Maybe I can help?

src/components/TextInput.js
```diff
@@ -0,0 +1,351 @@
...
+              <Animated.View
+                style={[ styles.bottomLine, bottomLineStyle ]}
+              />}
+        </View>
+      );
+    }
+
+    return (
+      <View style={[ style ]}>
+          {floating && <AnimatedText style={[ styles.label, labelStyle ]}>
```
satya164: nit: indentation is wrong

src/components/TextInput.js
ahmedlhanafy: @satya164 Can you propose one please?

example/src/TextInputExample.js
ahmedlhanafy: @satya164 `16px` padding on all sides makes it look super ugly

src/components/TextInput.js
ahmedlhanafy: @satya164 you can experiment with it and see the issue yourself

src/components/TextInput.js
ahmedlhanafy: Done

ahmedlhanafy:
screen shot 2016-12-06 at 12 26 37 am
@satya164

src/components/TextInput.js
```diff
@@ -0,0 +1,320 @@
...
+  },
+  bottomLine: {
+    position: 'absolute',
+    left: 0,
+    right: 0,
+    top: 0,
+    height: 2 / PixelRatio.get(),
+  },
+  focusLine: {
+    height: 6 / PixelRatio.get(),
```
ahmedlhanafy: @satya164 this looks kinda big and ugly in IOS

satya164:
TODO:

- [ ] Make it work with multiline text, height is hardcoded to 64 now
- [ ] Add helper text (https://material.google.com/patterns/errors.html#errors-user-input-errors)

ahmedlhanafy:
@satya164 The keyboard covers up the textinputs, also why isn't `add helper text` marked as done?

satya164:
> The keyboard covers up the textinputs

I'll add a `<KeyboardSpacer />` component

> also why isn't add helper text marked as done 

because It's not done. I think I'll just add a new `<HelperText />`, you can do `<HelperText type='error'>Some Text</HelperText>`. reasons,

1. It's useful for other form fields also, I need it in a client project at work
2. we need to set a default height for the helper text, otherwise it will shift items at bottom when it renders, having it in all inputs isn't ideal

satya164:
@ferrannp ready for review

src/components/TextInput.js
```diff
@@ -0,0 +1,348 @@
...
+ */
+
+class TextInput extends Component<DefaultProps, Props, State> {
+  static propTypes = {
+    disabled: PropTypes.bool,
+    label: PropTypes.string,
+    placeholder: PropTypes.string,
+    onChangeText: PropTypes.func,
+    multiline: PropTypes.bool,
+    numberOfLines: PropTypes.number,
```
ferrannp: Maybe we remove `propTypes` already from here too in favour of Flow? By the way, that is the diference between `multiline` and `numberOfLines` ?

src/components/TextInput.js
```diff
@@ -0,0 +1,348 @@
...
+    left: 0,
+    top: 38,
+    fontSize: 16,
+  },
+  input: {
+    minHeight: 64,
+    paddingTop: 20,
+    paddingBottom: 0,
+    marginTop: 8,
+    marginBottom: -4,
```
ferrannp: All these numbers `-4`, `38`... Are they working/looking ok for iOS too?

PR #86, opened 2016-11-25
feat: add portal component

satya164:


src/core/PortalHost.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  };
+
+  render() {
+    const { portals } = this.state;
+    return (
+      <View style={[ styles.container, this.props.style ]}>
+        {this.props.children}
+        <View style={StyleSheet.absoluteFill}>
+          {Object.keys(portals).map(key =>
```
ahmedlhanafy: @satya164 Isn't it just sorting this object `portals`?

PR #88, opened 2016-11-26
Linear progress bar.

mike866:
Implementation of linear progress bar in two states.

Determinate:
When indicators are determinate they indicate how long an operation will take when the percentage complete is detectable.

Indeterminate:
When indicators are indeterminate they request that the user wait while something finishes when it’s not necessary to indicate how long it will take.


src/index.js
```diff
@@ -12,6 +12,8 @@ export { default as Paper } from './components/Paper';
 export { default as RadioButton } from './components/RadioButton';
 export { default as TouchableRipple } from './components/TouchableRipple';
 
+export * as Progress from './components/Progress';
```
satya164: This file should not know how individual components are exported. Let's change it to

`export { default as Progress } from './components/Progress`

src/components/Progress/index.js
```diff
@@ -0,0 +1 @@
+export { default as Linear } from './Linear';
```
satya164: Can be changed to

```js
import Linear from './Linear';

export default {
  Linear,
}
```

src/components/Progress/Linear.js
```diff
@@ -0,0 +1,193 @@
+/* @flow */
```
satya164: As a convention I usually name the component (and file) as `ProgressLinear` instead of `Linear`. That way when it's shown in error messages `ProgressLinear` will be shown which is more informative.

src/components/Progress/Linear.js
```diff
@@ -0,0 +1,193 @@
...
+      style,
+      emptyColor,
+      indeterminateWidth,
+      ...otherProps
+    } = this.props;
+
+    const { width, height } = this.state;
+
+    const progressBar = {
+      backgroundColor: color,
```
satya164: It should use the accent color from theme by default.

src/components/Progress/Linear.js
```diff
@@ -0,0 +1,193 @@
...
+  View,
+} from 'react-native';
+
+import { blueA200, blue100 } from '../../styles/colors';
+
+const roundProgress = (progress: ?number): number => Math.min(Math.max((progress || 0), 0), 1);
+
+type Props = {
+  color: string;
+  emptyColor: string;
```
satya164: This is same as `style={{ backgroundColor: color }}`, right? In that case, let's remove the prop

src/components/Progress/Linear.js
```diff
@@ -0,0 +1,193 @@
...
+    indeterminateWidth: PropTypes.number,
+    /**
+     * Progress of loaded content. Rounded to match 0-1 range.
+     */
+    progress: PropTypes.number,
+    style: View.propTypes.style,
+  };
+
+  static defaultProps: Props = {
+    color: blueA200,
```
satya164: It should use the accent color from theme by default.

src/components/Progress/Linear.js
```diff
@@ -0,0 +1,193 @@
...
+    emptyColor: blue100,
```
satya164: This could be achieved by lightening the color I believe.

src/components/Progress/index.js
grabbou: Any benefits?

src/components/Progress/Linear.js
```diff
@@ -0,0 +1,193 @@
...
+import {
+  StyleSheet,
+  Animated,
+  Easing,
+  View,
+} from 'react-native';
+
+import { blueA200, blue100 } from '../../styles/colors';
+
+const roundProgress = (progress: ?number): number => Math.min(Math.max((progress || 0), 0), 1);
```
grabbou: I'd add description, took me a while to figure out. Basically it's used to return a number between bounds (0, 1), otherwise upper or lower limit.

example/src/ProgressExample.js
```diff
@@ -0,0 +1,76 @@
...
+  Progress,
+  Headline,
+  Subheading,
+} from 'react-native-paper';
+
+class ProgressExample extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.interval = setInterval(() => {
```
satya164: Can we move this to `componentDidMount` instead? Also change it to `this._interval`.

Doesn't matter here, but it's better to have side effects in `componentDidMount`, since in plain React you don't usually want them to trigger on the server while server-rendering.

example/src/ProgressExample.js
```diff
@@ -0,0 +1,76 @@
...
+      this.setState({ progress, visible });
+    }, 700);
+
+    this.state = {
+      progress: 0,
+      visible: true,
+    };
+  }
+
+  state = {
```
satya164: This is declared twice. I think better to get rid of the constructor. And move interval stuff to `componentDidMount`.

example/src/ProgressExample.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.interval);
+  }
+
+  render() {
+    return (
+      <View style={styles.container}>
+        <Headline style={styles.text}>Linear</Headline>
```
satya164: Headline, Subheading etc have a default vertical margin of 4px. I think you don't need the additional `style`

src/components/Progress/ProgressLinear.js
```diff
@@ -0,0 +1,204 @@
...
+  PureComponent,
+  PropTypes,
+} from 'react';
+import {
+  StyleSheet,
+  Animated,
+  Easing,
+  View,
+} from 'react-native';
+import color from 'color';
```
satya164: nit: remove new line between imports

src/components/Progress/ProgressLinear.js
```diff
@@ -0,0 +1,204 @@
...
+  static defaultProps = {
+    indeterminate: false,
+    progress: 0,
+    visible: true,
+  };
+
+  constructor(props: Props) {
+    super(props);
+
+    const progress: number = roundProgress(props.progress);
```
satya164: nit: remove new lines between `const` declarations

src/components/Progress/ProgressLinear.js
```diff
@@ -0,0 +1,204 @@
...
+  _startIndeterminateAnimation = () => {
+    this.state.indeterminateAnimationValue.setValue(0);
+
+    Animated.timing(this.state.indeterminateAnimationValue, {
+      toValue: 1,
+      duration: 1200,
+      easing: Easing.linear,
+      isInteraction: false,
+    }).start((animationState) => {
+      if (animationState.finished) {
```
satya164: I wish we had `Animated.loop`

src/components/Progress/ProgressLinear.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  render() {
+    const {
+      style,
+      ...otherProps
+    } = this.props;
+    const { width, height } = this.state;
+
+    const progressColor = this.props.color || this.props.theme.colors.accent;
```
satya164: Destructure `theme` above

ferrannp:
This is stale. Re-check issue: https://github.com/callstack-io/react-native-paper/issues/11.

PR #89, opened 2016-11-27
feat: add checkbox and radio animations on Android

satya164:
Border animations aren't very nice on iOS, so I disabled animations on iOS for now. Let's add animations when we find a better way.
Also native animations don't support borders as of now, so can't enables native animations either, probably can enable in future

### DIFF
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -1,9 +1,13 @@
 /* @flow */
 
 import React, {
+  Component,
   PropTypes,
 } from 'react';
 import {
+  Animated,
+  Platform,
+  View,
   StyleSheet,
 } from 'react-native';
 import color from 'color';
@@ -11,6 +15,8 @@ import Icon from './Icon';
 import TouchableRipple from './TouchableRipple';
 import withTheme from '../core/withTheme';
 import type { Theme } from '../types/Theme';
+
+const AnimatedIcon = Animated.createAnimatedComponent(Icon);
 
 type Props = {
   checked: boolean;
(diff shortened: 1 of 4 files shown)


ahmedlhanafy:
@satya164 I think that the animation is too fast it's unnoticeable, why not slow it down a little bit?

ahmedlhanafy:
@satya164 Also we need a label for these components to give the user the ability to click the label and checkbox or radiobutton will be triggered

src/components/RadioButton.js
```diff
@@ -20,78 +26,146 @@ type Props = {
...
+
+    return (
+      <TouchableRipple
+        {...this.props}
+        borderless
+        rippleColor={rippleColor}
+        onPress={disabled ? undefined : onPress}
+        style={styles.container}
+      >
+        <Animated.View style={Platform.OS === 'android' && { transform: [ { scale: this.state.scaleAnim } ] }}>
```
ahmedlhanafy: Again why are you animating the container?

src/components/Checkbox.js
```diff
@@ -20,78 +26,141 @@ type Props = {
...
+  componentWillReceiveProps(nextProps: Props) {
+    if (nextProps.checked !== this.props.checked) {
+      if (Platform.OS === 'android') {
+        Animated.parallel([
+          Animated.timing(this.state.checkAnim, {
+            toValue: nextProps.checked ? 1 : 0,
+            duration: 450,
+          }),
+          Animated.sequence([
+            Animated.timing(this.state.scaleAnim, {
```
ahmedlhanafy: I feel like animations could be faster

src/components/Checkbox.js
```diff
@@ -20,78 +26,141 @@ type Props = {
...
+
+    return (
+      <TouchableRipple
+        {...this.props}
+        borderless
+        rippleColor={rippleColor}
+        onPress={disabled ? undefined : onPress}
+        style={styles.container}
+      >
+        <Animated.View style={{ transform: [ { scale: this.state.scaleAnim } ] }}>
```
ahmedlhanafy: Why are you animating the container I've never seen this in any material design implementation?

satya164: The checkbox and radios on my phone scale down when I press them. I tried to follow the stock animations as much I could.

src/components/Checkbox.js
satya164: The stock animations aren't very fast.

ahmedlhanafy:
@satya164 Can you please give the radio buttons another shot I feel they are too slow and they don't match the one on the phone very much, checkbox are perfect though. 
Also we need to add label to these things because they are always used with them

satya164:
@ahmedlhanafy I wasn't able to find radio buttons in my phone. Can you tell me where can I check them?

> Also we need to add label to these things because they are always used with them

What about something like `<RadioButtonGroup options={[ "Item 1", "Item 2", "Item 3" ]} value={this.state.value} onChangeValue={value =>  this.setState({ value })} />`

ahmedlhanafy:
> What about something like <RadioButtonGroup options={[ "Item 1", "Item 2", "Item 3" ]} value={this.state.value} onChangeValue={value => this.setState({ value })} />
@satya164 Yeah, we need this. We can also have a label prop for it which will be used in `RadioButtonGroup` to populate items. 

> I wasn't able to find radio buttons in my phone. Can you tell me where can I check them?
Go to settings and change ringtone

satya164:
👍

satya164:
Done!

PR #90, opened 2016-11-27
feat: improvements to portal

satya164:
- layer management according to position
- preserve theme context via ThemedPortal


example/main.js
```diff
@@ -11,7 +11,11 @@ import {
   NavigationProvider,
   StackNavigation,
 } from '@exponent/ex-navigation';
-import { Colors, ThemeProvider, Drawer } from 'react-native-paper';
+import {
+  Colors,
+  Drawer,
+  Provider as PaperProvider,
```
ahmedlhanafy: Why not naming it `PaperProvider`?

src/index.js
```diff
@@ -2,7 +2,7 @@
 
 export { default as withTheme } from './core/withTheme';
 export { default as ThemeProvider } from './core/ThemeProvider';
-export { default as PortalHost } from './core/PortalHost';
+export { default as Provider } from './core/Provider';
```
ahmedlhanafy: Why didn't you export ThemedPortal?

src/components/Portal/Portal.js
```diff
@@ -4,7 +4,7 @@ import {
   PureComponent,
   PropTypes,
 } from 'react';
-import { manager } from '../core/PortalHost';
+import { manager } from './PortalHost';
 
 type Props = {
```
ahmedlhanafy: Why didn't you add `position` prop here?

src/components/Portal/ThemedPortal.js
```diff
@@ -0,0 +1,41 @@
...
+  Children,
+  PropTypes,
+} from 'react';
+import Portal from './Portal';
+import ThemeProvider from '../../core/ThemeProvider';
+import withTheme from '../../core/withTheme';
+import type { Theme } from '../../types/Theme';
+
+type Props = {
+  children?: any;
```
ahmedlhanafy: What about adding a `position` prop here as well to inform the user, without assuming that he knows that this component uses portal and passes its props to the portal?

example/main.js
satya164: I don't like the name very much (sounds like it provides a paper, lol). Just copied what Redux did. Tell me if you have a good name in mind.

src/index.js
satya164: @ahmedlhanafy I want experiment around Portal a bit more before deciding what to export.

src/components/Portal/Portal.js
satya164: Will add that

PR #93, opened 2016-11-28
Removed 'export * as' syntax in 'src/index.js'

### LINKED ISSUE #91: Unexpected Token Error in src/index.js
I just imported from react-native-paper and this happened:

```
<path-to-project>/node_modules/react-native-paper/src/index.js: Unexpected token (6:9) at index.js:6:9
```

Seems like this line is the culprit:

```
export * as Colors from './styles/colors';
```
I fixed it by adding installing [babel-plugin-transform-export-extensions](http://babeljs.io/docs/plugins/transform-export-extensions/) and adding `transform-export-extensions` babel plugin in my `.babelrc`.

Shouldn't this plugin be added as default?

### Versions
1. react-native: `0.37.0`.
2. react-native-paper: `0.0.2` (Latest `master` also has the issue)

sreejithr:
Since `export * as ...` requires addition of a custom preset, `react-native-stage-0` or `transform-export-extensions` in `.babelrc`.

Fixes #91.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,12 +1,12 @@
 /* @flow */
 
+import * as Colors from './styles/colors';
+export { Colors };
+
 export { default as withTheme } from './core/withTheme';
 export { default as ThemeProvider } from './core/ThemeProvider';
 export { default as Provider } from './core/Provider';
 export { default as DefaultTheme } from './styles/DefaultTheme';
-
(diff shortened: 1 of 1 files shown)


satya164:
Tests are failing. You need to move it to the top

sreejithr:
Moved it to top. Also merged with latest master.

PR #95, opened 2016-11-30
fix: use `flexGrow` instead of `flex` to avoid collapsing

### LINKED ISSUE #94: Card not displaying
Here's my code,

```
<Card>
  <Card.Content>
    <Title>The Ocean<Title>
    <Paragraph>Singin' in the sunshine, laughing in the rain.</Paragraph>
  </Card.Content>
</Card>
```
Doesn't get displayed. Went in the [code](https://github.com/react-native-paper/react-native-paper/blob/master/src/components/Card/Card.js#L101), saw that both the wrapping `TouchableWithoutFeedback` and the`View` inside, has `flex: 1`. `flex: 1` inside `TouchableWithoutFeedback` doesn't display content.

I removed `flex: 1` for the view inside and sure enough, everything started working. What do you guys think?

sreejithr:
Code from the example project wasn't working properly. The Card wasn't getting displayed. Fixed by removing flex: 1 from a View inside a TouchableWithoutFeedback.

Fixes #94.

### DIFF
--- a/src/components/Card/Card.js
+++ b/src/components/Card/Card.js
@@ -98,7 +98,7 @@ class Card extends Component<DefaultProps, Props, State> {
           onPressOut={onPress ? this._handlePressOut : undefined}
           style={styles.container}
         >
-          <View style={styles.container}>
+          <View>
             {Children.map(children, (child, index) =>
               React.cloneElement(child, {
                 index,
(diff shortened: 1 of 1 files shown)


satya164:
@sreejithr can you change it to `flexGrow: 1` instead? I want it to grow to fit the card if you set a custom height / width.

sreejithr:
@satya164 Nice. I wasn't aware of the `flexGrow` style. I thought `flex` served the purpose of grow and shrink in RN. My bad.

PR #96, opened 2016-12-05
chore: use component-docs for documentation

satya164:


ahmedlhanafy:
@satya164 Why are you removing all these files?

satya164:
Because they are not needed anymore https://github.com/satya164/component-docs

PR #97, opened 2017-01-17
SearchBar component

ferrannp:
First version of the component:

- It does not include list with suggestions
- It does not include integration with Toolbar (we do not have this one yet)
- It should have cursor color and highlighted color using the `accent` color. Right now, it is not possible to change the cursor color so just changing the highlighted would look weird.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -193,7 +193,7 @@
     "no-new-symbol": 2,
     "no-spaced-func": 1,
     "no-ternary": 0,
-    "no-trailing-spaces": [1, { "skipBlankLines": true }],
+    "no-trailing-spaces": 1,
     "no-underscore-dangle": 0,
     "no-unneeded-ternary": 1,
     "object-curly-spacing": 0,
--- a/.eslintrc
+++ b/.eslintrc
@@ -193,7 +193,7 @@
     "no-new-symbol": 2,
     "no-spaced-func": 1,
     "no-ternary": 0,
-    "no-trailing-spaces": 1,
+    "no-trailing-spaces": [1, { "skipBlankLines": true }],
     "no-underscore-dangle": 0,
     "no-unneeded-ternary": 1,
     "object-curly-spacing": 0,
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -1,0 +1,123 @@
+/* @flow */
+
+import React, {
+  Component,
+  PropTypes,
+} from 'react';
+import {
+  StyleSheet,
+  TextInput,
+} from 'react-native';
+
+import withTheme from '../core/withTheme';
+import Icon from './Icon';
+import TouchableRipple from './TouchableRipple';
+import Paper from './Paper';
+import type { Theme } from '../types/Theme';
+
+type Props = {
+  placeholder: string;
+  value: string;
+  onChangeSearch: (query: string) => {};
+  theme: Theme;
+}
+
+/**
+ * Search allows users to locate app content quickly
+ */
+class SearchBar extends Component<void, Props, void> {
+  
+  static propTypes = {
+    /**
+     * The string that will be rendered before text input has been entered
+     */
+    placeholder: PropTypes.string,
+    /**
+     * The value to show for the text input.
+     */
+    value: PropTypes.string.isRequired,
+    /**
+     * Callback that is called when the text input's text changes
+     */
+    onChangeSearch: PropTypes.func.isRequired,
+    theme: PropTypes.object.isRequired,
+  };
+  
+  constructor(props: Props) {
+    super(props);
+    this._styles = getStyles(this.props.theme);
+  }
+  
+  _styles = {};
+  
+  _handleClearPress = () => {
+    this.props.onChangeSearch('');
+  };
+  
+  render() {
+    const {
+      placeholder,
+      value,
+      theme,
+    } = this.props;
+    
+    const styles = this._styles;
+    
+    return (
+      <Paper
+        elevation={4}
+        style={styles.container}
+      >
+        <Icon
+          style={[ styles.iconWrapper, styles.icon ]}
+          name='search'
+          size={24}
+        />
+        <TextInput
+          style={styles.input}
+          placeholder={placeholder}
+          value={value}
+          placeholderTextColor={theme.colors.hint}
+          underlineColorAndroid='transparent'
+          onChangeText={this.props.onChangeSearch}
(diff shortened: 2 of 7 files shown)


.eslintrc
```diff
@@ -193,7 +193,7 @@
     "no-new-symbol": 2,
     "no-spaced-func": 1,
     "no-ternary": 0,
-    "no-trailing-spaces": 1,
+    "no-trailing-spaces": [1, { "skipBlankLines": true }],
```
satya164: revert this one

src/styles/DefaultTheme.js
```diff
@@ -18,6 +18,8 @@ export default {
     text: black,
     secondaryText: color(black).alpha(0.7).rgbaString(),
     disabled: color(black).alpha(0.5).rgbaString(),
+    hint: color(black).alpha(0.38).rgbaString(),
+    icon: color(black).alpha(0.54).rgbaString(),
```
satya164: I'd remove icon from here. icons can be a lot of different colors.

src/styles/DefaultTheme.js
```diff
@@ -18,6 +18,8 @@ export default {
     text: black,
     secondaryText: color(black).alpha(0.7).rgbaString(),
     disabled: color(black).alpha(0.5).rgbaString(),
+    hint: color(black).alpha(0.38).rgbaString(),
```
satya164: chnage to `placeholder`

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+    flexDirection: 'row',
+    justifyContent: 'center',
+    margin: 8,
+  },
+  input: {
+    flex: 1,
+    fontSize: 18,
+  },
+  icon: {
+    alignSelf: 'center',
```
satya164: Instead of `alignSelf`, set `alignItems: 'center'` on the parent

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+            />
+          </TouchableRipple> :
+          null
+        }
+      </Paper>
+    );
+  }
+}
+
+const getStyles = (theme: Theme) => StyleSheet.create({
```
satya164: Don't use a function to get styles. It'll mean that even static styles without any variables will be recreated when the function is called again.

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+    justifyContent: 'center',
+    margin: 8,
+  },
+  input: {
+    flex: 1,
+    fontSize: 18,
+  },
+  icon: {
+    alignSelf: 'center',
+    color: theme.colors.icon,
```
satya164: Just use the style inline instead.

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Callback that is called when the text input's text changes
+     */
+    onChangeSearch: PropTypes.func.isRequired,
+    theme: PropTypes.object.isRequired,
+  };
+  
+  constructor(props: Props) {
+    super(props);
+    this._styles = getStyles(this.props.theme);
```
satya164: remove. this won't update when theme changes.

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+import withTheme from '../core/withTheme';
+import Icon from './Icon';
+import TouchableRipple from './TouchableRipple';
+import Paper from './Paper';
+import type { Theme } from '../types/Theme';
+
+type Props = {
+  placeholder: string;
+  value: string;
+  onChangeSearch: (query: string) => {};
```
satya164: Probably should change the name to `onChangeText`

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+} from 'react-native';
+
+import withTheme from '../core/withTheme';
+import Icon from './Icon';
+import TouchableRipple from './TouchableRipple';
+import Paper from './Paper';
+import type { Theme } from '../types/Theme';
+
+type Props = {
+  placeholder: string;
```
satya164: `placeholder` should be optional

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+  value: string;
+  onChangeSearch: (query: string) => {};
+  theme: Theme;
+}
+
+/**
+ * Search allows users to locate app content quickly
```
satya164: This should describe the component.

e.g. - `Searchbar is a simple input box where users can type search queries`

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+ */
+class SearchBar extends Component<void, Props, void> {
+  
+  static propTypes = {
+    /**
+     * The string that will be rendered before text input has been entered
```
satya164: `Hint text shown when the input is empty`

src/components/SearchBar.js
```diff
@@ -0,0 +1,123 @@
...
+     */
+    placeholder: PropTypes.string,
+    /**
+     * The value to show for the text input.
```
satya164: `The value of the text input`

src/styles/DefaultTheme.js
ferrannp: Not many colors in this case. For light theme, icons are greyish (54%) and for dark theme they are white (100%). See: 

screen shot 2017-01-19 at 00 41 58

src/components/SearchBar.js
```diff
@@ -0,0 +1,111 @@
...
+import withTheme from '../core/withTheme';
+import Icon from './Icon';
+import TouchableRipple from './TouchableRipple';
+import Paper from './Paper';
+import type { Theme } from '../types/Theme';
+
+type Props = {
+  placeholder?: string;
+  value: string;
+  onChangeText: (query: string) => {};
```
satya164: This should probably be `(query: string) => void` since you don't return an object.

src/styles/DefaultTheme.js
```diff
@@ -18,6 +18,8 @@ export default {
     text: black,
     secondaryText: color(black).alpha(0.7).rgbaString(),
     disabled: color(black).alpha(0.5).rgbaString(),
+    placeholder: color(black).alpha(0.38).rgbaString(),
+    icon: color(black).alpha(0.54).rgbaString(),
```
satya164: remove icon from theme

PR #98, opened 2017-01-20
Add dialog component

ahmedlhanafy:


src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+    const {
+      children,
+      dismissable,
+      title,
+      titleColor,
+      style,
+    } = this.props;
+    const childrenArray = Children.toArray(children);
+    const actionBtnsChildren = childrenArray.filter(child => child.type === Card.Actions);
+    const restOfChildren = childrenArray.filter(child => child.type !== Card.Actions);
```
ahmedlhanafy: @satya164 we need to make `Card.Actions` into their own component

ahmedlhanafy:
screen shot 2017-01-21 at 1 44 01 am
screen shot 2017-01-21 at 1 44 19 am

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+        </Animated.View>
+      </ThemedPortal>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    minHeight: 96,
+    maxHeight: deviceHeight / 2,
```
ahmedlhanafy: @satya164 I haven't found the `maxHeight ` of dialogs anywhere

src/index.js
```diff
@@ -7,10 +7,12 @@ export { default as withTheme } from './core/withTheme';
 export { default as ThemeProvider } from './core/ThemeProvider';
 export { default as Provider } from './core/Provider';
 export { default as DefaultTheme } from './styles/DefaultTheme';
+export { default as ThemedPortal } from './components/Portal/ThemedPortal';
```
satya164: remove this

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+
+    const {
+      children,
+      dismissable,
+      title,
+      titleColor,
+      style,
+    } = this.props;
+    const childrenArray = Children.toArray(children);
+    const actionBtnsChildren = childrenArray.filter(child => child.type === Card.Actions);
```
satya164: Shouldn't need to do any of these once we have `Dialog.Title`, `Dialog.Content` and `Dialog.Actions`

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+    BackAndroid.removeEventListener('hardwareBackPress', this._handleBack);
+    Animated.timing(
+      this.state.opacity,
+      {
+        toValue: 0,
+        duration: 300,
+        easing: Easing.easing,
+      }
+    ).start(() => {
+      if (this.props.visible && this.props.onRequestClose) {
```
satya164: Have you tried calling `this.props.onRequestClose` when trying to hide dialog? then this animation can be run on `componentDidUpdate`. Don't think there'll be performance implications, and will simplify code.

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+ */
+
+export default class Dialog extends PureComponent<DefaultProps, Props, State> {
+
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    /**
+     * Determines whether clicking outside the dialog dismiss it, true by default
+     */
+    dismissable: PropTypes.bool,
```
satya164: If `onRequestClose` fires before the animation then this prop is redundant, same as `onRequestClose={() => false}`

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+    dismissable: true,
+    titleColor: black,
+    visible: false,
+  }
+
+
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      opacity: new Animated.Value(props.visible ? 1 : 0),
```
satya164: It'll be better to move out the animation + back button handling to a separate `Modal` component

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+    visible: PropTypes.bool,
+  };
+
+  static defaultProps = {
+    dismissable: true,
+    titleColor: black,
+    visible: false,
+  }
+
+
```
satya164: nit: remove extra new line

example/src/DialogExample.js
```diff
@@ -0,0 +1,172 @@
...
+
+  renderDialogWithLongText = () => {
+    const { visible1 } = this.state;
+    return (
+      <Dialog
+        onRequestClose={() => this.setState({ visible1: false })}
+        title='Alert'
+        visible={visible1}
+      >
+        <Paragraph>Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum
```
satya164: It'd be better to have actual readable text instead of boring lorem isum

src/components/Dialog.js
ahmedlhanafy: 👍

src/components/Dialog.js
ahmedlhanafy: No, I think they should have maximum height because otherwise, they height might exceed the height of the screen

src/index.js
ahmedlhanafy: Do we need the model component for anything other than the dialog?

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+  PureComponent,
+  PropTypes,
+  Children,
+} from 'react';
+import {
+  Animated,
+  View,
+  Easing,
+  StyleSheet,
+  Dimensions,
```
ahmedlhanafy: Use `onLayout` instead?

src/components/Dialog.js
ahmedlhanafy: I think you're right, but isn't nice to add a prop for that just for the ease of use?

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+     */
+    onRequestClose: PropTypes.func.isRequired,
+    /**
+     * The title of the dialog
+     */
+    title: PropTypes.string,
+    /**
+     * The color of the title
+     */
+    titleColor: PropTypes.string,
```
ahmedlhanafy: Yup will do that `Title`, `Content` and `Actions`

src/components/Dialog.js
ahmedlhanafy: Prettier to the rescue 😄

src/components/Dialog.js
ahmedlhanafy: Do we need the `Modal` component for anything other than the dialog?

src/components/Dialog.js
ahmedlhanafy: Will try it

src/components/Dialog.js
ahmedlhanafy: 👍

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+        <Animated.View
+          style={[ { opacity: this.state.opacity }, styles.wrapper ]}
+        >
+          <View style={[ StyleSheet.absoluteFill, { backgroundColor: 'rgba(0, 0, 0, .5)' } ]}/>
+          {dismissable && <TouchableWithoutFeedback onPress={this._hideDialog}>
+              <View style={StyleSheet.absoluteFill}/>
+          </TouchableWithoutFeedback>}
+          <AnimatedPaper elevation={24} style={[ { opacity: this.state.opacity }, styles.container, style ]}>
+              {title && <Title style={[ styles.title, { color: titleColor } ]}>{title}</Title>}
+              <ScrollView style={styles.scrollViewContainer}>
```
ahmedlhanafy: 👍

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+          </AnimatedPaper>
+        </Animated.View>
+      </ThemedPortal>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    minHeight: 96,
```
ahmedlhanafy: I don't know for sure but it should have a minimum value for sure, shouldn't it?

src/components/Dialog.js
```diff
@@ -0,0 +1,254 @@
...
+    maxHeight: deviceHeight / 2,
+    width: deviceWidth - 52,
```
ahmedlhanafy: 👍

example/src/DialogExample.js
ahmedlhanafy: 👍

src/components/Dialog.js
ahmedlhanafy: I think this makes more sense and it simpler, tried the other one and didn't like it that much

src/components/Dialog.js
satya164: @ahmedlhanafy don't think it's needed. we want it to take as much space it needs to. if the space is bigger, there'll be extra space at bottom which won't be pretty

src/components/Dialog.js
satya164: @ahmedlhanafy `maxHeight: deviceHeight / 2` is pretty small. it should be able to take the full height and width it can, that is the device's height and width. it's upto the user to put it in a scrollview if the content get's bigger than the screen height.

src/index.js
satya164: Yes

src/components/Dialog.js
satya164: We shouldn't need any of it. Flex should work fine

src/components/Dialog.js
satya164: Maybe, but I'm not a fan of unnecessary props

src/components/Dialog.js
satya164: Yes, it's a generic Modal component which people can use. Dialog is material themed dialog

example/src/DialogExample.js
```diff
@@ -0,0 +1,226 @@
...
+  state = {
+    visible1: false,
+    visible2: false,
+    visible3: false,
+    visible4: false,
+    visible5: false,
+    checked: 0,
+  };
+
+  renderDialogWithLongText = () => {
```
satya164: nit: prefix these methods with `_`

example/src/DialogExample.js
```diff
@@ -0,0 +1,226 @@
...
+  Dialog,
+} from 'react-native-paper';
+
+const isIOS = Platform.OS === 'ios';
+
+export default class DialogExample extends Component {
+  static title = 'Dialog';
+
+  state = {
+    visible1: false,
```
satya164: nit: can you split each type of dialog to a separate component and then export the final component with all of them? it'll make it easier to copy/paste code

src/components/Dialog/Actions.js
```diff
@@ -0,0 +1,38 @@
...
+import React, { Children } from 'react';
+import PropTypes from 'prop-types';
+import { StyleSheet, View } from 'react-native';
+
+type Props = {
+  children?: any,
+  style?: any,
+};
+
+const Actions = (props: Props) => {
```
satya164: nit: we should probably call these components `DialogActions` etc. show the name is recognizable in devtools.

src/components/Dialog/ScrollArea.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
satya164: missing flow comment

src/components/Dialog/Content.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
satya164: missing flow comment

src/components/Dialog/Title.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
satya164: missing flow comment

src/components/Dialog/Title.js
```diff
@@ -0,0 +1,39 @@
...
+import withTheme from '../../core/withTheme';
+import type { Theme } from '../../types/Theme';
+
+type Props = {
+  children: any,
+  color?: string,
+  theme: Theme,
+};
+
+const Title = ({ children, theme: { colors: { text } }, color }: Props) => (
```
satya164: Accept a `style` prop instead of a `color` prop

src/components/Dialog/Title.js
```diff
@@ -0,0 +1,39 @@
...
+  <PaperTitle style={[styles.text, { color: color || text, margin: 0 }]}>
```
satya164: `margin: 0` seems unnecessary since you have it below

src/components/Dialog/Title.js
```diff
@@ -0,0 +1,39 @@
...
+    {children}
+  </PaperTitle>
+);
+
+const styles = StyleSheet.create({
+  text: {
+    /**
+     * I subtracted 2 from these values because Text component has
```
satya164: nit: `'I subtracted' -> 'Subtract'`

src/components/Dialog/Title.js
```diff
@@ -0,0 +1,39 @@
...
+  <PaperTitle style={[styles.text, { color: color || text, margin: 0 }]}>
+    {children}
+  </PaperTitle>
+);
+
+const styles = StyleSheet.create({
+  text: {
+    /**
+     * I subtracted 2 from these values because Text component has 
+     * 2 pixels margin vertically by default
```
satya164: Where do you subtract 2 pixels?

src/components/Dialog/Title.js
```diff
@@ -0,0 +1,39 @@
...
+     */
+    margin: 0,
+    paddingTop: 22,
```
satya164: Does `paddingTop` and `paddingBottom` work? Last time I tried they didn't work on Android for text components. I'll suggest to margins instead

src/components/Dialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+import PropTypes from 'prop-types';
+import Modal from '../Modal';
+import { black, white } from '../../styles/colors';
+import Paper from '../Paper';
+import Actions from './Actions';
+import Title from './Title';
+import Content from './Content';
+import ScrollArea from './ScrollArea';
+
+const AnimatedPaper = Animated.createAnimatedComponent(Paper);
```
satya164: Since we are doing this a lot, we should add it in `Paper` as `Paper.Animated`

src/components/Dialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+  const title = childrenArray.find(child => child.type === Title);
+  const actionBtnsChildren = childrenArray.filter(
+    child => child.type === Actions
+  );
+  const restOfChildren = childrenArray.filter(
+    child => child.type !== Actions && child.type !== Title
+  );
+  let restOfChildrenWithoutTitle = restOfChildren;
+  if (!title) {
+    restOfChildrenWithoutTitle = restOfChildren.map(child => {
```
satya164: If title is not present, we'll need to add top padding only to the first element, not all elements.

src/components/Dialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+
+Dialog.Actions = Actions;
+Dialog.Title = Title;
+Dialog.Content = Content;
+Dialog.ScrollArea = ScrollArea;
+
+Dialog.propTypes = {
+  children: PropTypes.node.isRequired,
+  /**
+     * Determines whether clicking outside the dialog dismiss it, true by default
```
satya164: nit: comment is badly formatted

src/components/Dialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+     */
+  dismissable: PropTypes.bool,
+  /**
+     * Callback that is called when the user dismisses the dialog
```
satya164: nit: comment is badly formatted

src/components/Dialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+     */
+  onRequestClose: PropTypes.func.isRequired,
+  style: PropTypes.object,
+  /**
+     * Determines Whether the dialog is visible
```
satya164: nit: comment is badly formatted

src/components/Dialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+  titleColor: black,
+  visible: false,
+};
+
+export default Dialog;
+
+const styles = StyleSheet.create({
+  container: {
+    /**
+     * This is a fix for Android because overflow: visible isn't supported.
```
satya164: not: This doesn't say what does it fix. You should mention that it prevents shadow from being clipped.

src/components/Modal.js
```diff
@@ -0,0 +1,178 @@
...
+    BackAndroid.removeEventListener('hardwareBackPress', this._handleBack);
+    Animated.timing(this.state.opacity, {
+      toValue: 0,
+      duration: 280,
+      easing: Easing.easing,
+    }).start(() => {
+      if (this.props.visible && this.props.onRequestClose) {
+        this.props.onRequestClose();
+      }
+      global.requestAnimationFrame(() => {
```
satya164: Unfortunately I've found this causes issue with large trees. Let's remove it for now.

example/src/DialogExample.js
ahmedlhanafy: Like the idea will do it 👍

src/components/Modal.js
ahmedlhanafy: What should be the solution then? @satya164

src/components/Dialog/index.js
ahmedlhanafy: You can see that I do check if the child type is Content only

src/components/Dialog/Title.js
ahmedlhanafy: Totally working

satya164: @ahmedlhanafy yea, but maybe it'll break in older versions.

src/components/Dialog/index.js
satya164: even then you don't want to add padding to all content elements based on title exists or not. only the content element after title should get the padding

src/components/Modal.js
satya164: I don't think there is a good solution right now

ahmedlhanafy: So we will leave it to go out of sync?! @satya164

satya164: I guess no better solution right now

src/components/Dialog/ScrollArea.js
ahmedlhanafy: added

src/components/Dialog/Content.js
ahmedlhanafy: added

src/components/Dialog/Title.js
ahmedlhanafy: added

src/components/Modal.js
satya164: Please remove this

src/components/Modal.js
```diff
@@ -0,0 +1,179 @@
...
+
+type State = {
+  opacity: Animated.Value,
+  rendered: boolean,
+};
+
+/**
+ * The Modal component is a simple way to present content above an enclosing view.
+ *
+ * export default class MyComponent extends Component {
```
satya164: Wrap the example in backticks with the language tag

src/components/Dialog/index.js
```diff
@@ -0,0 +1,142 @@
...
+  dismissable?: boolean,
+  onRequestClose?: Function,
+  style?: any,
+  visible: boolean,
+};
+
+/**
+ * Dialogs inform users about a specific task and may contain critical information, require decisions, or involve multiple tasks.
+ *
+ * export default class MyComponent extends Component {
```
satya164: wrap the example in backticks with language tag

src/components/Modal.js
```diff
@@ -0,0 +1,178 @@
...
+      if (this.props.visible && this.props.onRequestClose) {
+        this.props.onRequestClose();
+      }
+      if (this.props.visible) {
+        this._showModal();
+      } else {
+        this.setState({
+          rendered: false,
+        });
+      }
```
crutchcorn, line 140: I am currently migrating this component (well, the modern-day version of it) to a functional component and have gotten everything else to work as-intended. However, I am running into issues with this LOC (as `visible` is not updating due to closure within `useCallback`). 

I'm wondering if the authors can help me with some context here: What is this LOC doing and why is it needed?

crutchcorn: Apologies - I finally figured it out after commenting 😅 For others, the TLDR is that `rendered` interacts with `render()` to do an early return when the value is `false`. This means that when the prop is updated, it removes these items from the "DOM". 

This same behavior can be matched in `useCallback` by having a ref that's set on every render for `visible`:

```
  const visibleRef = React.useRef(visible);

  visibleRef.current = visible;
```

This isn't an ideal solution, but this code is difficult to port to FC without a bit of grimeyness (IMO). Not to insult the component - it's a beautiful class component, just ports to FC difficultly

crutchcorn: @satya164 do you mean to `react-native-paper`? Is the migration to FC something that's wanting to be done upstream? I think I've finally gotten the behavior to match 1:1 in all of the examples I can find, so it should be good-to-go as an upstream PR, I just wasn't sure if that's something the team wanted

satya164: @crutchcorn yea we are slowly migrating to function components, so would be awesome if you can send a PR

crutchcorn: #2557 has been opened! :)

PR #103, opened 2017-05-16
[WIP] - Update dependencies

ferrannp:
https://github.com/callstack-io/react-native-paper/issues/101

- [ ] Update to `flow-bin 0.40.0` and fix `.flowconfig

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,10 +23,10 @@
   "homepage": "https://github.com/ahmedlhanafy/react-native-paper#readme",
   "peerDependencies": {
     "react": "*",
-    "react-native": "*",
-    "react-native-vector-icons": "*"
+    "react-native": "*"
   },
   "devDependencies": {
+    "@expo/vector-icons": "^4.0.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -23,20 +23,19 @@
   "homepage": "https://github.com/ahmedlhanafy/react-native-paper#readme",
   "peerDependencies": {
     "react": "*",
-    "react-native": "*",
-    "react-native-vector-icons": "*"
```
ahmedlhanafy: I think @satya164 was doing it that to make the Icon available in non-exponent apps as well

src/components/Icon.js
```diff
@@ -1,4 +1,4 @@
 /* @flow */
 
-export { default as default } from 'react-native-vector-icons/MaterialIcons';
```
ahmedlhanafy: Same as above

ferrannp:
Let's do it in #104.

PR #109, opened 2017-06-26
feat: add a FAB component

satya164:
simulator screen shot 26-jun-2017 7 07 53 pm

### DIFF
--- a/src/components/FAB.js
+++ b/src/components/FAB.js
@@ -1,0 +1,101 @@
+/* @flow */
+
+import color from 'color';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { StyleSheet } from 'react-native';
+import Paper from './Paper';
+import Icon from './Icon';
+import TouchableRipple from './TouchableRipple';
+import { white } from '../styles/colors';
+import withTheme from '../core/withTheme';
+import type { Theme } from '../types/Theme';
+
+type Props = {
+  small?: boolean,
+  dark?: boolean,
+  icon: string,
+  color?: string,
+  onPress?: Function,
+  theme: Theme,
+  style?: any,
+};
+
+/**
+ * A floating action button represents the primary action in an application
+ */
+const FAB = (props: Props) => {
+  const { small, dark, icon, color: iconColor, onPress, theme, style } = props;
+  const backgroundColor = theme.colors.accent;
+  const isDark = typeof dark === 'boolean'
+    ? dark
+    : !color(backgroundColor).light();
+  const textColor = iconColor || (isDark ? white : 'rgba(0, 0, 0, .54)');
+  const rippleColor = color(textColor).alpha(0.32).rgbaString();
+
+  return (
+    <Paper
+      {...props}
+      style={[
+        { backgroundColor },
+        styles.content,
+        small ? styles.small : styles.standard,
+        style,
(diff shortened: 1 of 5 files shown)


src/components/FAB.js
```diff
@@ -0,0 +1,96 @@
...
+import withTheme from '../core/withTheme';
+import type { Theme } from '../types/Theme';
+
+type Props = {
+  small?: boolean,
+  dark?: boolean,
+  icon: string,
+  onPress?: Function,
+  theme: Theme,
+  style?: any,
```
ahmedlhanafy: I think we have to have `iconColor`, I've seen it so many times when the icon was colored differently and I think it's quite limiting if we don't do it

ahmedlhanafy:
@satya164 pretty simple and neat, good job 👍

src/components/FAB.js
satya164: added .a `color` prop for that

src/components/FAB.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+/**
+ * A floating action button represents the primary action in an application
+ */
+const FAB = (props: Props) => {
+  const { small, dark, icon, color: iconColor, onPress, theme, style } = props;
+  const backgroundColor = theme.colors.accent;
+  const isDark = typeof dark === 'boolean'
```
ferrannp: Doesn't flow type says its a `boolean` always ?

src/components/FAB.js
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <Paper
+      {...props}
+      style={[
+        { backgroundColor },
+        styles.content,
+        small ? styles.small : styles.standard,
+        style,
+      ]}
+      elevation={12}
```
ferrannp: Wow that's big one, is the default? Should we be able to edit it?

src/components/FAB.js
```diff
@@ -0,0 +1,101 @@
...
+        rippleColor={rippleColor}
+        style={[styles.content, small ? styles.small : styles.standard]}
+      >
+        <Icon name={icon} size={24} style={{ color: textColor }} />
+      </TouchableRipple>
+    </Paper>
+  );
+};
+
+FAB.propTypes = {
```
ferrannp: I thought we said we will move away from `propTypes` ?

src/components/FAB.js
satya164: @ferrannp it's `?boolean`, so can be undefined, null etc

src/components/FAB.js
satya164: @ferrannp that's the value for FAB as per guidelines https://material.io/guidelines/material-design/elevation-shadows.html#elevation-shadows-elevation-android

src/components/FAB.js
satya164: @ferrannp we haven't yet

src/components/FAB.js
ferrannp: Not same as `dark?: ?boolean,`. It cannot be null they way is now. I think if it can be undefined or boolean, it make might sense to check it like `!!dark`.

src/components/FAB.js
ferrannp: Not good :(

PR #112, opened 2017-07-03
Toolbar

### LINKED ISSUE #20: Toolbar
[MD Docs Link](https://material.google.com/components/toolbars.html#)

ferrannp:
Fixes #20.


src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,105 @@
...
+class Toolbar extends Component<DefaultProps, Props, void> {
+  static propTypes = {
+    /**
+     * Color used for the background
+     */
+    backgroundColor: PropTypes.string,
+    /**
+     * Toolbar content
+     */
+    children: PropTypes.oneOfType([
```
satya164: I believe `PropTypes.node` already includes an array

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,105 @@
...
+      PropTypes.node,
+      PropTypes.arrayOf(PropTypes.node),
+    ]),
+    /**
+     * Toolbar height
+     */
+    height: PropTypes.number,
```
satya164: Remove. People can pass a `style` prop with height

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,105 @@
...
+    style: View.propTypes.style,
+    /**
+     * Whether Toolbar needs to adapt to a translucent StatusBar or not
+     */
+    translucent: PropTypes.bool,
```
satya164: We should name this something else. This prop makes it look like it makes the header translucent, which is not the case.

e.g. - `statusBarHeight` to pass a custom status bar height.

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,105 @@
...
+
+    return (
+      <Paper
+        elevation={4}
+        style={[
+          toolbarStyle,
+          translucent && { paddingTop: StatusBar.currentHeight },
+        ]}
+      >
+        <View
```
satya164: Is this inner view necessary? Can we remove it?

src/components/Toolbar/ToolbarAction.js
```diff
@@ -0,0 +1,59 @@
...
+import color from 'color';
+
+import { black, white } from '../../styles/colors';
+import TouchableRipple from '../TouchableRipple';
+import Icon from '../Icon';
+
+type Props = {
+  dark?: boolean,
+  icon: string,
+  onPress?: () => void,
```
satya164: I think `Function` is fine here

src/components/Toolbar/ToolbarAction.js
```diff
@@ -0,0 +1,59 @@
...
+
+    const iconColor = dark ? white : color(black).alpha(0.54).rgbaString();
+    const rippleColor = color(iconColor).alpha(0.32).rgbaString();
+
+    return (
+      <TouchableRipple
+        borderless
+        onPress={onPress}
+        rippleColor={rippleColor}
+        style={style}
```
satya164: We should spread rest of the props here

example/src/ToolbarExample.js
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  render() {
+    const { showSubTitle } = this.state;
+
+    return (
+      <View style={styles.container}>
+        <Toolbar translucent>
+          <Toolbar.Action
+            dark
```
satya164: `Toolbar` should pass this prop to it's children to handle this automatically

src/components/Toolbar/ToolbarContent.js
```diff
@@ -0,0 +1,108 @@
...
+import Text from '../Typography/Text';
+
+import withTheme from '../../core/withTheme';
+import { white } from '../../styles/colors';
+
+import type { Theme } from '../../types/Theme';
+
+type Props = {
+  dark?: boolean,
+  subTitle?: string,
```
satya164: "Subtitle" is a single word, so we shouldn't use camel case here. Change to `subtitle`

src/components/Toolbar/ToolbarContent.js
```diff
@@ -0,0 +1,108 @@
...
+  title: string,
```
satya164: We should probably allow React elements for title and subtitle too

src/components/Toolbar/ToolbarContent.js
```diff
@@ -0,0 +1,108 @@
...
+    const { colors, fonts } = theme;
+    const { text: primaryText, secondaryText } = colors;
+    const fontFamilyMedium = fonts['medium'];
+
+    const titleColor = dark ? white : primaryText;
+    const subTitleColor = dark
+      ? color(white).alpha(0.7).rgbaString()
+      : secondaryText;
+    const titleStyles = [
+      styles.text,
```
satya164: Remove: non-existent style

src/components/Toolbar/ToolbarContent.js
```diff
@@ -0,0 +1,108 @@
...
+      subTitle,
+      subTitleStyle,
+      style,
+      titleStyle,
+      theme,
+      title,
+    } = this.props;
+    const { colors, fonts } = theme;
+    const { text: primaryText, secondaryText } = colors;
+    const fontFamilyMedium = fonts['medium'];
```
satya164: why not `fonts.medium`?

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,103 @@
...
+      height,
+      style,
+      theme,
+      translucent,
+    } = this.props;
+    const { colors } = theme;
+
+    const toolbarHeight = Platform.OS === 'ios' ? 44 : 56;
+    const statusBarHeight =
+      Platform.OS === 'ios' ? 20 : StatusBar.currentHeight;
```
satya164: Status bar height can change, though we don't have a way to listen for changes

example/src/ToolbarExample.js
```diff
@@ -0,0 +1,63 @@
+/* @flow */
+
+import React, { Component } from 'react';
+import { View, StyleSheet } from 'react-native';
+import { Colors, Button, Toolbar } from 'react-native-paper';
+
+export default class ToolbarExample extends Component {
+  static title = 'Toolbar';
+  static navigationOptions = {
+    header: null,
```
satya164: I'd just render this below the React Navigation's toolbar since our toolbar can't replace React Navigation's toolbar as is, also switching between screens will look ugly with this

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,103 @@
...
+      // TODO make height orientation aware ???
+      height: (height || toolbarHeight) + (translucent ? statusBarHeight : 0),
+    };
+
+    return (
+      <Paper
+        elevation={4}
+        style={[toolbarStyle, translucent && { paddingTop: statusBarHeight }]}
+      >
+        <View
```
satya164: Is it possible to remove this inner view and have just 1 container?

src/components/Toolbar/Toolbar.js
ferrannp: Holding on this.

src/components/Toolbar/ToolbarAction.js
ferrannp: I think it is better like that, if you were to invoke it manually (in code or test), you are saying should be `this.props.onPress()` instead of passing anything to `onPress`.

example/src/ToolbarExample.js
ferrannp: It is gonna be bad to render two toolbars, specially because the one from React Navigation will take the place of status bar. See:

screen shot 2017-07-04 at 15 34 27

ferrannp: Btw react navigation toolbar is wrong, ours follow better Material as title needs to be `72dp` from the left.

example/src/ToolbarExample.js
ferrannp: What is best way to do this? Cloning? Also... Will it be possible the case you want the icon to be dark but text dark ? Or that is too much ?

example/src/ToolbarExample.js
satya164: @ferrannp we should add some margin/padding to avoid this :)

example/src/ToolbarExample.js
satya164: @ferrannp yeah, `cloneElement` should do it. if the user has passed it already, we shouldn't pass it

ahmedlhanafy:
img_2157
I don't like the highlight on the icons, maybe we can make them circles to imitate ripples? @ferrannp

ferrannp:
@ahmedlhanafy I guess that is something about the `Touchable` so probably should be another PR.

example/src/ToolbarExample.js
ferrannp: @satya164 still the Toolbar we did will look bar because it will look is bigger than it should.

satya164:
@ferrannp you would need to set it in the `Toolbar.Action` component, also it needs a height and width

ferrannp:
@satya164 reason to not have `height` and `width` on `TouchableRipple` here is to make the circle of the icon smaller (like it is in native Android Toolbar). We can give it for iOS... But do we want it square or in circle like @ahmedlhanafy suggested?

On iOS we could do something like `height: 44, width: 44, borderRadius: 44/2`.

example/src/ToolbarExample.js
satya164: > Toolbar we did will look bar because it will look is bigger

sorry I don't understand

satya164:
@ferrannp I don't understand what you mean by that. without height and width the height of the touchable is too small (as per the screenshot) to be convenient to tap.

Also whatever we do for touchable on iOS, we should also do it for pre-lollipop Android.

ferrannp:
@satya164 @ahmedlhanafy you can try it out !

src/components/Toolbar/ToolbarContent.js
```diff
@@ -0,0 +1,122 @@
...
+class ToolbarContent extends Component<void, Props, void> {
+  static propTypes = {
+    /**
+     * Theme color for the text, a dark toolbar will render light text and vice-versa
+     */
+    dark: PropTypes.bool,
+    /**
+     * If Toolbar has a left icon, this prop will align the title accordingly
+     */
+    hasStartIcon: PropTypes.bool,
```
satya164: We shouldn't need a prop for this. We should check if the first element is `Toolbar.Action` and pass a `style` prop with `horizontalMargin` of `6` from the `toolbar`.

Though I'm wondering if icons have a `6dp` horizontal margin and that's why it looks like there's a higher padding in toolbar.

src/components/Toolbar/ToolbarContent.js
```diff
@@ -0,0 +1,122 @@
...
+          </Text>}
+      </View>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    paddingHorizontal: Platform.OS === 'ios' ? 16 : 16,
```
satya164: Unnecessary check? You're setting it to `16` either way

src/components/Toolbar/ToolbarContent.js
```diff
@@ -0,0 +1,122 @@
...
+      titleStyle,
+    ];
+
+    return (
+      <View
+        style={[
+          styles.container,
+          {
+            paddingHorizontal:
+              Platform.OS === 'ios' ? 16 : hasStartIcon ? 22 : 16,
```
satya164: Why is it different on iOS?

src/components/Toolbar/ToolbarAction.js
```diff
@@ -0,0 +1,75 @@
...
+    style: View.propTypes.style,
+  };
+
+  render() {
+    const { dark, icon, onPress, style, ...rest } = this.props;
+
+    const iconColor = dark ? white : color(black).alpha(0.54).rgbaString();
+    const rippleColor = color(iconColor).alpha(0.32).rgbaString();
+
+    const touchableStyle =
```
satya164: These are static styles. Move below with `StyleSheet.create`

src/components/Toolbar/ToolbarAction.js
```diff
@@ -0,0 +1,75 @@
...
+    const iconStyle = { alignSelf: 'center' };
+
+    return (
+      <TouchableRipple
+        borderless
+        onPress={onPress}
+        rippleColor={rippleColor}
+        hitSlop={
+          Platform.OS === 'android'
+            ? { top: 8, left: 8, bottom: 8, right: 8 }
```
satya164: If the height of the touchable is `28`, then we can set `hitSlop` to `14` since `28 + (14 * 2) = 56`, and toolbar's height is `56`

example/src/ToolbarExample.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  render() {
+    const { showSubtitle } = this.state;
+
+    return (
+      <View style={styles.container}>
+        <Toolbar statusBarIsTranslucent>
+          <Toolbar.Action
+            dark
```
satya164: We should remove the `dark` prop from here, should be automatic

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,85 @@
...
+        (statusBarIsTranslucent ? statusBarHeight : 0),
+    };
+
+    return (
+      <Paper
+        elevation={4}
+        style={[
+          toolbarStyle,
+          statusBarIsTranslucent && { paddingTop: statusBarHeight },
+          styles.toolbar,
```
satya164: Pass `style` from props in the end

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,85 @@
...
+  render() {
+    const { children, style, theme, statusBarIsTranslucent } = this.props;
+    const { colors } = theme;
+
+    const toolbarHeight = Platform.OS === 'ios' ? 44 : 56;
+    const statusBarHeight =
+      Platform.OS === 'ios' ? 20 : StatusBar.currentHeight;
+
+    const toolbarStyle = {
+      backgroundColor: (style && style.backgroundColor) || colors.primary,
```
satya164: Set `backgroundColor: colors.primary` and it'll be overridden if `this.props.style` has a background color

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,85 @@
...
+      // TODO make height orientation aware ???
+      height:
+        ((style && style.height) || toolbarHeight) +
```
satya164: `style` might be an object or an integer or an array. Use `StyleSheet.flatten(style)` to get an object to access style values

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,85 @@
...
+    };
+
+    return (
+      <Paper
+        elevation={4}
+        style={[
+          toolbarStyle,
+          statusBarIsTranslucent && { paddingTop: statusBarHeight },
+          styles.toolbar,
+        ]}
```
satya164: Pass through `{...rest}` props

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,85 @@
...
+import Paper from '../Paper';
+import ToolbarContent from './ToolbarContent';
+import ToolbarAction from './ToolbarAction';
+
+import type { Theme } from '../../types/Theme';
+
+type Props = {
+  children?: any,
+  style?: any,
+  statusBarIsTranslucent?: boolean,
```
satya164: As we talked earlier, we need a `statusBarHeight` prop to configure it

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,85 @@
...
+    return (
+      <Paper
+        elevation={4}
+        style={[
+          toolbarStyle,
+          statusBarIsTranslucent && { paddingTop: statusBarHeight },
+          styles.toolbar,
+        ]}
+      >
+        {children}
```
satya164: Use `Children.map` to loop over children and pass the `dark` prop after checking if `backgroundColor` is dark

src/components/Toolbar/ToolbarAction.js
```diff
@@ -0,0 +1,75 @@
...
+            height: 44,
+            width: 44,
+            borderRadius: 44 / 2,
+            justifyContent: 'center',
+          }
+        : {
+            height: 28,
+            width: 28,
+            justifyContent: 'center',
+            marginHorizontal: 8,
```
satya164: I think there should be `14dp` of horizontal margin

src/components/Toolbar/ToolbarAction.js
```diff
@@ -0,0 +1,75 @@
...
+          };
+    const iconStyle = { alignSelf: 'center' };
```
satya164: Why not `alignItems: 'center'` on the touchable?

src/components/Toolbar/ToolbarAction.js
```diff
@@ -0,0 +1,75 @@
...
+        rippleColor={rippleColor}
+        hitSlop={
+          Platform.OS === 'android'
+            ? { top: 8, left: 8, bottom: 8, right: 8 }
+            : null
+        }
+        style={[touchableStyle, style]}
+        {...rest}
+      >
+        <Icon color={iconColor} name={icon} size={24} style={iconStyle} />
```
satya164: Can you pass the `color` in `style` instead?

src/components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,85 @@
...
+      </Paper>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  toolbar: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    paddingHorizontal: Platform.OS === 'ios' ? 0 : 6,
```
satya164: Why different padding on iOS?

satya164:
Ok, so according to our conversation and measurements:

```css
Toolbar { padding: 4 }
ToolbarAction { height: 28; minWidth: 16; maxWidth: 28; padding: 2; margin: 10; } /* hitSlop: 10 */
ToolbarActionIOS { height: 44; minWidth: 32; maxWidth: 44; padding: 2; margin: 2; } /* hitSlop: 2 */
ToolbarTitle { paddingHorizontal: 12 }
ToolbarTitle:not(:first-of-type) { marginHorizontal: 8 }
```

src/components/Toolbar/ToolbarAction.js
ferrannp: Following convention, color is `const iconColor = dark ? white : color(black).alpha(0.54).rgbaString();`. You want to allow to override this too ?

src/components/Toolbar/ToolbarAction.js
ferrannp: This is solved.

example/src/ToolbarExample.js
ferrannp: This is a matter of taste, I prefer to leave it as it is and replace all `react-navigation` toolbars as soon as possible.

satya164:
@ferrannp what about this? https://github.com/callstack-io/react-native-paper/pull/112#discussion_r125474081

ferrannp:
@satya164 ok we'll make that `onPress` to be just a `Function`.

PR #115, opened 2017-07-04
fix: use TouchableHighlight on Android < Lollipop

satya164:
Earlier, older Android versions used a ugly blue highlight for touchable items. This commit changes it to a `TouchableHighlight`, which uses the appropriate highlight color. Merged it with the iOS version for now since they have the same behaviour.


ahmedlhanafy:
@satya164 Good work 👍

PR #118, opened 2017-07-04
chore: add Git Hook to validate commit messages

rauliyohmc:


scripts/validateCommitMessage.js
```diff
@@ -0,0 +1,58 @@
+const fs = require('fs');
```
satya164: Use Flow?

scripts/validateCommitMessage.js
```diff
@@ -0,0 +1,58 @@
...
+const TYPES = {
+  feat: true,
+  fix: true,
+  docs: true,
+  style: true,
+  refactor: true,
+  perf: true,
+  test: true,
+  chore: true,
+  revert: true,
```
satya164: breaking too

scripts/validateCommitMessage.js
```diff
@@ -0,0 +1,58 @@
...
+  return true;
+}
+
+function firstLineFromBuffer(buffer) {
+  return buffer.toString().split('\n').shift();
+}
+
+const commitMsgFile = process.argv[2];
+
+fs.readFile(commitMsgFile, function(err, buffer) {
```
satya164: maybe `fs.readFileSync` is better since it's a script

rauliyohmc: Yeah, changed

scripts/validateCommitMessage.js
rauliyohmc: Added

scripts/validateCommitMessage.js
rauliyohmc: Should I? There are only 2 functions to annotate, one with `string` type and `buffer` type, which is a node object

satya164: @rauliyohmc flow provides autocomplete

rauliyohmc: Cool, added

PR #120, opened 2017-07-04
chore: upgrade to React Native 0.45

satya164:


src/components/Card/Card.js
```diff
@@ -54,7 +54,7 @@ class Card extends Component<DefaultProps, Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      elevation: new Animated.Value(props.elevation),
+      elevation: new Animated.Value(this.props.elevation),
```
ahmedlhanafy: @satya164 why did you change this?

satya164: No reason. Was trying to fix flow and this was left when committing.

PR #121, opened 2017-07-06
SearchBar - Left icon can be used as a button

ferrannp:
You can see two uses cases on the example, back button and drawer.

### DIFF
--- a/example/src/SearchBarExample.js
+++ b/example/src/SearchBarExample.js
@@ -2,10 +2,10 @@
 
 import React, { Component } from 'react';
 import { StyleSheet, View } from 'react-native';
-import { Caption, SearchBar } from 'react-native-paper';
+import { Colors, Caption, SearchBar } from 'react-native-paper';
 
 export default class SearchExample extends Component {
-  static title = 'SearchBar';
+  static title = 'Search bar';
 
   state = {
     firstQuery: '',
@@ -15,7 +15,7 @@ export default class SearchExample extends Component {
 
   render() {
     return (
-      <View>
+      <View style={styles.container}>
         <SearchBar
           placeholder="Search"
           onChangeText={query => this.setState({ firstQuery: query })}
@@ -27,14 +27,14 @@ export default class SearchExample extends Component {
           onChangeText={query => this.setState({ secondQuery: query })}
           value={this.state.secondQuery}
           onIconPress={() => this.props.navigation.goBack()}
-          iconName="arrow-back"
+          icon="arrow-back"
         />
         <SearchBar
           placeholder="Search"
           onChangeText={query => this.setState({ thirdQuery: query })}
           value={this.state.thirdQuery}
-          onIconPress={() => {}}
-          iconName="menu"
+          onIconPress={/* In real code, this will open the drawer */ () => {}}
+          icon="menu"
         />
       </View>
     );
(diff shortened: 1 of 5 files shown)


example/src/SearchBarExample.js
```diff
@@ -1,24 +1,49 @@
 /* @flow */
 
 import React, { Component } from 'react';
-import { SearchBar } from 'react-native-paper';
+import { StyleSheet, View } from 'react-native';
+import { Caption, SearchBar } from 'react-native-paper';
 
 export default class SearchExample extends Component {
-  static title = 'Searchbar';
+  static title = 'SearchBar';
```
satya164: Since this is a title and not a variable name, let's use `Search bar` instead of camel case

example/src/SearchBarExample.js
```diff
@@ -1,24 +1,49 @@
...
-  _handleChangeSearch = query => this.setState({ query });
-
   render() {
     return (
-      <SearchBar
-        placeholder="Search"
-        onChangeText={this._handleChangeSearch}
-        value={this.state.query}
-      />
+      <View>
```
satya164: Can you add a background color to this view?

example/src/SearchBarExample.js
```diff
@@ -1,24 +1,49 @@
...
+          onChangeText={query => this.setState({ secondQuery: query })}
+          value={this.state.secondQuery}
+          onIconPress={() => this.props.navigation.goBack()}
+          iconName="arrow-back"
+        />
+        <SearchBar
+          placeholder="Search"
+          onChangeText={query => this.setState({ thirdQuery: query })}
+          value={this.state.thirdQuery}
+          onIconPress={() => {}}
```
satya164: Empty handlers are confusing imo, add a comment inside or remove it

src/components/SearchBar.js
```diff
@@ -3,18 +3,21 @@
...
 import Icon from './Icon';
-import TouchableRipple from './TouchableRipple';
+import TouchableIcon from './TouchableIcon';
 import Paper from './Paper';
 import { white } from '../styles/colors';
 import type { Theme } from '../types/Theme';
 
 type Props = {
   placeholder?: string,
+  iconName?: string,
```
satya164: We use just `icon` as the prop name everywhere to refer to icon name, so `iconName` is inconsistent

src/components/SearchBar.js
```diff
@@ -56,11 +75,19 @@ class SearchBar extends Component<void, Props, void> {
         elevation={4}
         style={[{ borderRadius: roundness }, styles.container, style]}
       >
-        <Icon
-          style={[styles.icon, { color: iconColor }]}
-          name="search"
-          size={24}
-        />
+        {onIconPress
+          ? <TouchableIcon
```
satya164: We probably shouldn't wrap the icon in a touchable if there is no `onPress`

ferrannp: Yup, if there is no on press, we render just `Icon`.

satya164: Cool!

src/index.js
```diff
@@ -15,6 +15,7 @@ export { default as Checkbox } from './components/Checkbox';
 export { default as Paper } from './components/Paper';
 export { default as RadioButton } from './components/RadioButton';
 export { default as TouchableRipple } from './components/TouchableRipple';
+export { default as TouchableIcon } from './components/TouchableIcon';
```
satya164: Let's not export this. We'll have a `ToggleButton` component which is more or less the same when not raised https://material.io/guidelines/components/buttons.html#buttons-toggle-buttons

PR #123, opened 2017-07-14
Drawer removal

ahmedlhanafy:
This closes https://github.com/callstack-io/react-native-paper/issues/100


example/main.js
```diff
@@ -2,70 +2,65 @@
...
-    );
   }
 }
 
+const App = DrawerNavigator(
+  { Home: { screen: RootNavigator } },
+  { contentComponent: () => <DrawerItems /> }
+);
+
 const styles = StyleSheet.create({
```
satya164: This seems not to be used anywhere

src/index.js
```diff
@@ -24,7 +24,8 @@ export { default as Title } from './components/Typography/Title';
 export { default as Text } from './components/Typography/Text';
 
 export { default as Divider } from './components/Divider';
-export { default as Drawer } from './components/Drawer';
+export { default as DrawerItem } from './components/DrawerItem';
+export { default as DrawerSection } from './components/DrawerSection';
```
satya164: I was hoping to group these both under a single object (maybe `DrawerContent`?), i.e. `DrawerContent.Item` and `DrawerContent.Section` etc. We'll likely add more stuff like `Cover` to match the components commonly used in drawers. So it'll be nice to group these.

satya164:
Can you run `yarn run lint -- --fix`? There seem to be prettier errors.

example/main.js
ahmedlhanafy: No, it's used by the wrapper view of the drawer content

src/index.js
ahmedlhanafy: But I think people will be dumb and use DrawerContent which we don't have??!

satya164: @ahmedlhanafy people will copy/paste from the docs, so I think they will use what we have in the docs. unless they didn't read docs, in which case they won't able to use it properly anyway

ahmedlhanafy: @satya164 still it feels weird to me, have you seen this pattern before in any other library? to expose an object which does nothing but has components inside

PR #131, opened 2017-08-10
feat: Dark Theme

ahmedlhanafy:
## Changes:
- Add dark theme
- Add new theme properties
- Update a number of components to use the new theme properties 
- Add the ability to toggle the theme from the drawer
- Add the ability to deep merge themes instead of one-level merge
- Remove `react-native-drawer` package
- Add `lodash.merge`

### Dark theme compatible components 
- FAB
- Card
- CheckBox
- GridView
- Paper
- RadioButton
- SearchBar
- DawerItem
- DrawerSection
The rest of the component involved non-trivial logic to apply the colors so it's better if their authors update them to avoid any breaking changes.

### Screenshots
screen shot 2017-08-10 at 3 07 19 pm
screen shot 2017-08-10 at 3 04 03 pm
screen shot 2017-08-10 at 3 04 15 pm
screen shot 2017-08-10 at 3 04 21 pm
screen shot 2017-08-10 at 3 04 31 pm


ferrannp:
I have some comments about the screenshots:

- Drawer focus item probably should have different color? White?
- All icons should be 100% white (not that grey from Search)

For the other stuff, it looks very good (didn't check all code).

src/components/RadioButton.js
```diff
@@ -87,7 +86,7 @@ class RadioButton extends Component<void, Props, State> {
       rippleColor = color(checkedColor)
         .clearer(0.32)
         .rgbaString();
-      radioColor = checked ? checkedColor : uncheckedColor;
+      radioColor = checked ? checkedColor : theme.colors.disabled;
```
satya164: don't think we should use disabled color for unchecked. just hardcode a new color if it's not in the theme. cc @lukewalczak

src/components/Paper.js
```diff
@@ -2,37 +2,51 @@
...
+class Paper extends Component<void, Props, void> {
   static propTypes = {
     children: PropTypes.node,
+    theme: PropTypes.object.isRequired,
     style: ViewPropTypes.style,
   };
 
   render() {
-    const { style, ...restOfProps } = this.props;
+    const { style, theme: { colors: { paper } }, ...restOfProps } = this.props;
```
satya164: this destructuring is unreadable. just use dot notation

src/components/GridView.js
```diff
@@ -119,6 +118,7 @@ export default class GridView extends PureComponent<
   _setRef = (c: Object) => (this._root = c);
 
   render() {
+    const { spacing, theme: { colors: { background } } } = this.props;
```
satya164: this destructuring is unreadable

example/src/ExampleList.js
```diff
@@ -39,44 +39,64 @@ export const examples = {
...
 
   render() {
+    const { theme: { colors: { background } } } = this.props;
     return (
-      <ListView
-        dataSource={dataSource}
-        renderRow={this._renderRow}
-        renderSeparator={this._renderSeparator}
-      />
+      <View style={{ flex: 1, backgroundColor: background }}>
```
satya164: Pass `style` to `ListView` instead of wrapping it in a view

example/src/ExampleList.js
```diff
@@ -39,44 +39,64 @@ export const examples = {
...
-      <Text style={styles.text}>{examples[id].title}</Text>
-    </TouchableRipple>
-  );
+  state = {
+    value: 0,
+  };
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.theme !== this.props.theme) {
+      /// Super dirty way to update the listview when now rows change
```
satya164: extract `renderRow` to a component and wrap in `withTheme` instead of this. that's how it should be handled in real apps to avoid losing state.

satya164:
I don't think `paper` is a good name for the theme property. The name is tied to a very specific component and the property is not specific to that component.

src/styles/DarkTheme.js
```diff
@@ -0,0 +1,28 @@
+/* @flow */
+
+import color from 'color';
+import merge from 'lodash.merge';
```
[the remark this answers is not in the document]
ahmedlhanafy: We can merge it as is and make these modifications later

ahmedlhanafy:
@satya164 Okay, then suggest another name 😄

PR #144, opened 2017-09-22
Progress bar component

lukewalczak:


src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+    Platform.OS === 'ios' ? ProgressViewIOS : ProgressBarAndroid;
+
+  const setTintColor = color => {
+    return setColor(color).alpha(0.38).rgbaString();
+  };
+
+  return (
+    <View>
+      <ProgressBarComponent
+        {...progressProps}
```
ferrannp: I do not see a benefit of doing `...` here, I just need then to look up to find what is `progessProps`. I think you can just pass them as you would do normally.

src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+  * Color of the progress bar
+  */
+  color: PropTypes.string,
+  style: PropTypes.any,
+  theme: PropTypes.object.isRequired,
+};
+
+const styles = StyleSheet.create({
+  progressBarHeight: {
+    height: 40,
```
ferrannp: Did you find this height somewhere in the docs? If not I think we could give some padding by default, something like `paddingVertical: 8`. I think it is better to leave it to the default height. Think that in Android < 5.0, progress bar is different.

src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+* Progress bar is an indicator used to present some activity in the app
+*/
+const ProgressBar = ({ progress, color, style, theme }: Props) => {
+  const ProgressBarComponent =
+    Platform.OS === 'ios' ? ProgressViewIOS : ProgressBarAndroid;
+
+  const setTintColor = color => {
+    return setColor(color).alpha(0.38).rgbaString();
```
ferrannp: Have you found this alpha somewhere? I'd be nice to document it (just a comment). If not, I guess it looks fine!

src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+  styleAttr: 'Horizontal',
+  indeterminate: false,
+};
+
+/**
+* Progress bar is an indicator used to present some activity in the app
+*/
+const ProgressBar = ({ progress, color, style, theme }: Props) => {
+  const ProgressBarComponent =
+    Platform.OS === 'ios' ? ProgressViewIOS : ProgressBarAndroid;
```
ferrannp: I am not sure if is the best to import both and then choose in this case, our instead use `.android.js` and `.ios.js`. @satya164 ?

src/components/ProgressBar.js
lukewalczak: I was trying to set the same height for both progressBar components. For sure, I have to change it for padding.

src/components/ProgressBar.js
lukewalczak: Ok, will just pass them normally.

src/components/ProgressBar.js
lukewalczak: Alpha and value come from [defaultTheme](https://github.com/callstack/react-native-paper/blob/0a203079f590eec2e33ed005fad18bd10e950c00/src/styles/DefaultTheme.js#L16)

src/components/ProgressBar.js
lukewalczak: For small component like progressBar, I think it's not worth to duplicate code and separate for two files.

ferrannp: You can use then `Platform.select`.

rauliyohmc: You can use `Platform.select` to dynamically require one component or the other according to the platform

src/components/ProgressBar.js
ferrannp: Mmm but that is default theme from placeholder in Android design... Anyway if you can't find it and looks how it should, it is ok.

src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+
+  const setTintColor = color => {
+    return setColor(color).alpha(0.38).rgbaString();
+  };
+
+  return (
+    <View>
```
satya164: let's remove the wrapper view

src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/**
+* Progress bar is an indicator used to present some activity in the app
+*/
+const ProgressBar = ({ progress, color, style, theme }: Props) => {
+  const ProgressBarComponent =
+    Platform.OS === 'ios' ? ProgressViewIOS : ProgressBarAndroid;
+
+  const setTintColor = color => {
```
satya164: I'd prefer to just have a local variable for this inside render instead of a function

src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+  theme: Theme,
+};
+
+export const progressProps = {
+  styleAttr: 'Horizontal',
+  indeterminate: false,
+};
+
+/**
+* Progress bar is an indicator used to present some activity in the app
```
satya164: let's add an example here

src/components/ProgressBar.js
```diff
@@ -0,0 +1,72 @@
...
+        color={color || theme.colors.primary}
+        style={[styles.progressBarHeight, style]}
+        trackTintColor={setTintColor(color || theme.colors.primary)}
+      />
+    </View>
+  );
+};
+
+ProgressBar.propTypes = {
+  /**
```
satya164: As we're thinking to move to flowtype, let's add the comments on flowtype
also the comment `*` should have 1 space indentation

PR #146, opened 2017-09-24
feat: Toolbar platform adaptation for iOS

### LINKED ISSUE #138: Adapt Toolbar to iOS platform adaptation
See https://material.io/guidelines/platforms/platform-adaptation.html#platform-adaptation-platform-recommendations.

Include back icon from Ionicons.

ferrannp:
Fixes #138.

Following https://material.io/guidelines/platforms/platform-adaptation.html

- [x] iOS toolbar has same height than Android
- [x] Title is centered when `leftActions < 2 &&  rightActions < 2`
- [x] Back button uses iOS default icon

## Example screenshots:
simulator screen shot - iphone 6 plus - 2017-09-25 at 16 00 49
simulator screen shot - iphone 6 plus - 2017-09-25 at 16 00 55
simulator screen shot - iphone 6 plus - 2017-09-25 at 16 01 00
simulator screen shot - iphone 6 plus - 2017-09-25 at 16 01 06

### DIFF
--- a/example/src/ToolbarExample.js
+++ b/example/src/ToolbarExample.js
@@ -4,6 +4,8 @@ import React, { Component } from 'react';
 import { View, Platform, StatusBar, StyleSheet } from 'react-native';
 import { Colors, Button, Toolbar } from 'react-native-paper';
 
+const MORE_ICON = Platform.OS === 'android' ? 'more-vert' : 'more-horiz';
+
 export default class ToolbarExample extends Component {
   static title = 'Toolbar';
   static navigationOptions = {
@@ -12,11 +14,18 @@ export default class ToolbarExample extends Component {
 
   state = {
     showLeftIcon: true,
-    showSubtitle: true,
+    showSearchIcon: Platform.OS === 'android',
+    showMoreIcon: true,
+    showSubtitle: Platform.OS === 'android',
   };
 
   render() {
-    const { showLeftIcon, showSubtitle } = this.state;
+    const {
+      showLeftIcon,
+      showSearchIcon,
+      showMoreIcon,
+      showSubtitle,
+    } = this.state;
 
     return (
       <View style={styles.container}>
@@ -25,8 +34,7 @@ export default class ToolbarExample extends Component {
           statusBarHeight={Platform.OS === 'ios' ? 20 : StatusBar.currentHeight}
         >
           {showLeftIcon && (
-            <Toolbar.Action
-              icon="arrow-back"
+            <Toolbar.BackAction
               onPress={() => this.props.navigation.goBack()}
             />
           )}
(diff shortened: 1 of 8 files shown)


src/components/Toolbar/Toolbar.js
```diff
@@ -70,6 +72,38 @@ class Toolbar extends Component<DefaultProps, Props, void> {
...
     };
 
+    const childrenArray = Children.toArray(children);
+    let isToolbarContentFound = false;
+    let leftActions = 0,
+      rightActions = 0;
+
+    if (Platform.OS === 'ios') {
+      childrenArray.forEach(child => {
+        if (!isToolbarContentFound && child.type !== ToolbarContent) {
```
satya164: Let's use `child.type.displayName` instead two avoid importing `ToolbarContent`. I'll write a babel plugin later which optimizes the imports, so you import only what you use.

src/components/Toolbar/ToolbarBackAction.js
```diff
@@ -0,0 +1,30 @@
...
+
+const getBackIcon = Platform.select({
+  ios: ({ dark }: Props) => {
+    const iconColor = dark ? white : color(black).alpha(0.54).rgbaString();
+    return <Icon name="keyboard-arrow-left" size={36} color={iconColor} />;
+  },
+  android: () => 'arrow-back',
+});
+
+type Props = {
```
satya164: we should import the `Props` type from `ToolbarAction` and use it here. Otherwise you won't get autocompletion or checks for all the types supported by `ToolbarAction`

src/components/TouchableRipple.js
```diff
@@ -89,7 +89,9 @@ export default class TouchableItem extends PureComponent<
             : color(rippleColor).clearer(0.5).rgbaString()
         }
       >
-        {Children.only(children)}
+        <View>
```
satya164: Instead of the `View` here, you should add it where you need it.

src/components/Toolbar/ToolbarBackAction.js
ferrannp: They share some (3) but not 2 of them. I am going to add comments that I forgot thought but I think we are ok here.

src/components/Toolbar/Toolbar.js
ferrannp: I'll leave it for now as we need that component to be imported anyway.

example/src/ToolbarExample.js
```diff
@@ -4,6 +4,8 @@ import React, { Component } from 'react';
 import { View, Platform, StatusBar, StyleSheet } from 'react-native';
 import { Colors, Button, Toolbar } from 'react-native-paper';
 
+const MORE_ICON = Platform.OS === 'android' ? 'more-vert' : 'more-horiz';
```
satya164: I usually do `Platform.OS === 'ios'`

PR #150, opened 2017-09-26
GridView: ListView upgraded to VirtualizedList

### LINKED ISSUE #142: Change GridView to use VirtualizedList and support orientation changes

jukben:
Close #142 

screen shot 2017-10-04 at 16 26 02
screen shot 2017-10-04 at 16 26 17

### DIFF
--- a/example/src/GridViewExample.js
+++ b/example/src/GridViewExample.js
@@ -42,13 +42,23 @@ export default class GridViewExample extends Component {
     return (
       <Card style={styles.tile}>
         <View style={styles.inner}>
-          <Text style={styles.text}>
-            {index}
-          </Text>
+          <Text style={styles.text}>{index}</Text>
         </View>
       </Card>
     );
   };
+
+  _renderItem = ({ id, name }) => {
+    return (
+      <Card style={styles.tile}>
+        <View style={styles.inner}>
+          <Text style={styles.text}>{id}</Text>
+        </View>
+      </Card>
+    );
+  };
+
+  _keyExtractor = ({ id }) => id;
 
   _getNumberOfColumns = (width: number) => {
     return Math.floor(width / CARD_SIZE);
@@ -63,6 +73,8 @@ export default class GridViewExample extends Component {
         getNumberOfColumns={this._getNumberOfColumns}
         dataSource={this.state.dataSource}
         renderRow={this._renderRow}
+        keyExtractor={this._keyExtractor}
+        renderItem={this._renderItem}
         onEndReached={this._genRows}
       />
     );
--- a/example/src/GridViewExample.js
+++ b/example/src/GridViewExample.js
@@ -45,9 +45,7 @@ export default class GridViewExample extends Component {
     );
   };
 
-  _keyExtractor = item => {
-    return item.id;
-  };
+  _keyExtractor = item => item.id;
 
   _getNumberOfColumns = (width: number) => {
     return Math.floor(width / CARD_SIZE);
@@ -55,18 +53,16 @@ export default class GridViewExample extends Component {
 
   render() {
     return (
-      <View>
-        <GridView
-          {...this.props}
-          spacing={8}
-          getNumberOfColumns={this._getNumberOfColumns}
-          data={this.state.items}
-          keyExtractor={this._keyExtractor}
-          renderItem={this._renderItem}
-          ref={ref => (this.ref = ref)}
-          onEndReached={this._genRows}
-        />
-      </View>
+      <GridView
+        {...this.props}
+        spacing={8}
+        getNumberOfColumns={this._getNumberOfColumns}
+        data={this.state.items}
+        keyExtractor={this._keyExtractor}
+        renderItem={this._renderItem}
+        ref={ref => (this.ref = ref)}
+        onEndReached={this._genRows}
+      />
     );
   }
 }
--- a/example/src/GridViewExample.js
+++ b/example/src/GridViewExample.js
@@ -1,9 +1,8 @@
-/* @flow */
+// @flow
 
 import React, { Component } from 'react';
 import { View, StyleSheet } from 'react-native';
 import { Colors, Card, Text, GridView } from 'react-native-paper';
-
 const CARD_SIZE = 160;
 
 export default class GridViewExample extends Component {
@@ -36,7 +35,7 @@ export default class GridViewExample extends Component {
     });
   };
 
-  _renderItem = (item: Object) => {
+  _renderItem = item => {
     return (
       <Card style={styles.tile}>
         <View style={styles.inner}>
@@ -46,7 +45,7 @@ export default class GridViewExample extends Component {
     );
   };
 
-  _keyExtractor = (item: Object) => {
+  _keyExtractor = item => {
     return item.id;
   };
 
--- a/example/src/GridViewExample.js
+++ b/example/src/GridViewExample.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import React, { Component } from 'react';
-import { View, StyleSheet, Button } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import { Colors, Card, Text, GridView } from 'react-native-paper';
 
 const CARD_SIZE = 160;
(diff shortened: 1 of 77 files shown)


example/src/GridViewExample.js
```diff
@@ -11,16 +11,15 @@ export default class GridViewExample extends Component {
...
 
   componentWillMount() {
-    this._genRows();
+    this.genRows();
   }
 
-  _genRows = () => {
+  ref: GridView;
+
+  genRows = () => {
```
satya164: Prefix the methods with since they aren't meant to be public `_`

src/components/GridView.js
```diff
@@ -63,41 +59,82 @@ export default class GridView extends PureComponent<
...
+        style={[
+          styles.grid,
+          { padding: this.props.spacing / 2 },
+          this.props.contentContainerStyle,
+        ]}
+      >
+        {item.map(a => {
+          const row = this.props.renderItem(a);
+
+          if (!row) {
```
satya164: we should check for `null` specifically. `undefined` or any other falsy value shouldn't be allowed

src/components/GridView.js
```diff
@@ -36,13 +37,10 @@ export default class GridView extends PureComponent<
   State
 > {
   static propTypes = {
-    dataSource: PropTypes.instanceOf(ListView.DataSource).isRequired,
+    data: PropTypes.array.isRequired,
     spacing: PropTypes.number.isRequired,
     getNumberOfColumns: PropTypes.func.isRequired,
-    renderSectionHeader: PropTypes.func,
-    renderRow: PropTypes.func.isRequired,
-    onLayout: PropTypes.func,
```
satya164: We still expose `onLayout` prop, right?

src/components/GridView.js
```diff
@@ -2,22 +2,23 @@
...
+import { Dimensions, View, StyleSheet, VirtualizedList } from 'react-native';
 import { grey200 } from '../styles/colors';
 
 type Layout = {
   width: number,
 };
 
 type Props = {
-  dataSource: ListView.DataSource,
   spacing: number,
```
satya164: Let's add some comments above the prop type definitions which will be used in the documentation.

jukben: Yeah, I wasn't aware how it works. Maybe we should describe it a little bit more for an easier start of other contributors. Maybe in another PR? I'm not really satisfied with it.

satya164: sure

example/src/GridViewExample.js
jukben: Even in the example, right? All of them are private. In the `source` it makes sense for me, but I really don't see the point why we need keep this "style" also in examples?

src/components/GridView.js
jukben: I don't get it.

`renderItem: any => ?React$Element<*>,`

if I rewrite this into `if (row === null) {}` what should I do in case of `undefined`?

src/components/GridView.js
```diff
@@ -63,41 +76,74 @@ export default class GridView extends PureComponent<
...
+          return React.cloneElement(row, {
+            key: keyExtractor(a),
+            style: [row.props.style, style],
+          });
+        })}
+      </View>
+    );
   };
 
   _handleLayout = (e: any) => {
```
ferrannp: Inside this function, can you round the values with `Math.round` when comparing? I worked with this in a project and the equal was inexact with so many decimals.

src/components/GridView.js
satya164: `renderItem` should always return a React element. `undefined` isn't valid as a react element, same for things like returning `string`, `number` etc.

example/src/GridViewExample.js
satya164: unnecessary to change conventions according to the scenario. if a convention works in 2 different scenario, no need to change it.

src/components/GridView.js
jukben: Yeah, ok, I'll do.

src/components/GridView.js
jukben: Ok, I will fix the type for it, make sense.

jukben:
@ferrannp Yeah It's probably due to re-rendering I don't know if I solved the issue properly.

jukben:
I'm kinda stuck here. Could someone help a little?

jukben:
Ready to check! @satya164

jukben:
@satya164 I've used Animated.Value as you suggested. I'm still struggling with the Flow, but maybe we could merge it first.

example/src/GridViewExample.js
```diff
@@ -1,25 +1,23 @@
-/* @flow */
+// @flow
```
satya164: We use `/* @flow */` everywhere, so let's use it for consistency

example/src/GridViewExample.js
```diff
@@ -1,25 +1,23 @@
...
 
 import React, { Component } from 'react';
 import { View, StyleSheet } from 'react-native';
 import { Colors, Card, Text, GridView } from 'react-native-paper';
-
```
satya164: let's leave the space

example/src/GridViewExample.js
```diff
@@ -29,40 +27,46 @@ export default class GridViewExample extends Component {
...
       <Card style={styles.tile}>
         <View style={styles.inner}>
-          <Text style={styles.text}>{index}</Text>
+          <Text style={styles.text}>{item.id}</Text>
         </View>
       </Card>
     );
   };
 
+  _keyExtractor = item => {
```
satya164: Maybe make it shorter like `item => item.id`

example/src/GridViewExample.js
```diff
@@ -29,40 +27,46 @@ export default class GridViewExample extends Component {
...
-      <GridView
-        {...this.props}
-        removeClippedSubviews={false}
-        spacing={8}
-        getNumberOfColumns={this._getNumberOfColumns}
-        dataSource={this.state.dataSource}
-        renderRow={this._renderRow}
-        onEndReached={this._genRows}
-      />
+      <View>
```
satya164: Why is the wrapper `View` needed?

src/components/GridView.js
```diff
@@ -51,96 +76,95 @@ export default class GridView extends PureComponent<
...
+      layout: { width: layoutWidth },
     });
   };
 
-  _setRef = (c: Object) => (this._root = c);
+  _getItemCount = (data: Array<any>) => data.length;
+
+  _getItem = (data, index) => data[index];
+
+  _getWidthOfOneItem = () => {
```
satya164: This function doesn't seem to be used anywhere

example/src/GridViewExample.js
jukben: Sry, I forget about him when I tested scroll methods. :)

jukben:
Another round @satya164. Thanks 🙏

example/src/GridViewExample.js
```diff
@@ -56,11 +55,12 @@ export default class GridViewExample extends Component {
...
         {...this.props}
-        removeClippedSubviews={false}
         spacing={8}
         getNumberOfColumns={this._getNumberOfColumns}
-        dataSource={this.state.dataSource}
-        renderRow={this._renderRow}
+        data={this.state.items}
+        keyExtractor={this._keyExtractor}
+        renderItem={this._renderItem}
+        ref={ref => (this.ref = ref)}
```
satya164: Do we need this `ref`?

satya164:
Can you post a screenshot? Earlier the horizontal spacing didn't match the vertical spacing. Both should be the same.

example/src/GridViewExample.js
jukben: Not anymore. Sorry, I really miss the proper ESLint :/

satya164: ESLint won't catch this

jukben:
screen shot 2017-10-04 at 16 26 02
screen shot 2017-10-04 at 16 26 17

satya164:
Let's fix the conflicts and merge this

Issue #134, opened 2017-09-14
Checkbox as part of a list item / row / etc


ferrannp:
The pattern here is that the Checkbox is not clickable but the whole row, then the Checkbox copies the touching state from the row.

Right now we cannot make the Checkbox non-clickable.

PR #151, opened 2017-09-26
refactor: checkbox as a part of a list item

lukewalczak:
Solves #134 

Added:

- New example in CheckboxExample which present checkbox as a part of a list item (clickable whole row)

### DIFF
--- a/example/src/CheckboxExample.js
+++ b/example/src/CheckboxExample.js
@@ -2,7 +2,12 @@
 
 import React, { Component } from 'react';
 import { View, StyleSheet } from 'react-native';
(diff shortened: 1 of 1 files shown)


example/src/CheckboxExample.js
```diff
@@ -15,16 +20,21 @@ export default class CheckboxExample extends Component {
   render() {
     return (
       <View style={styles.container}>
+        <TouchableRipple
+          onPress={() =>
```
jukben: It will be better to avoid of binding function in the render.

lukewalczak: Sure, it's good approach, but think in example we can keep it in component.

PR #153, opened 2017-09-26
refactor: remove elevation prop, use styles instead

Krizzu:


ahmedlhanafy:
@Krizzu I feel like this might be a little bit misleading because anyone who used RN for awhile knows that `elevation` style property only works on Android

satya164:
We need it for properly animating stuff.

Issue #137, opened 2017-09-19
Adapt checkboxes, radio buttons and switches to iOS platform adaptation


ferrannp:
See https://material.io/guidelines/platforms/platform-adaptation.html#platform-adaptation-platform-recommendations

They look the same. We might consider to have wrapper for multi/single selection.

For the icon on iOS use Ionicons from vector icons.

PR #155, opened 2017-09-29
feat: Controls adaptation to IOS

lukewalczak:
Solves #137 

Changes:

- [X] Checkbox and radio buttons adapted to IOS
- [X] Added switch 
- [X] Change theme disabled color for lighter one

### DIFF
--- a/src/components/Checkbox.ios.js
+++ b/src/components/Checkbox.ios.js
@@ -10,9 +10,21 @@ import withTheme from '../core/withTheme';
 import type { Theme } from '../types/Theme';
 
 type Props = {
+  /**
+   * Whether checkbox is checked
+   */
   checked?: boolean,
+  /**
+   * Whether checkbox is disabled
+   */
   disabled?: boolean,
+  /**
+   * Function to execute on press
+   */
   onPress?: Function,
+  /**
+   * Custom color for checkbox
+   */
   color?: string,
   theme: Theme,
 };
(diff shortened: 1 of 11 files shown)


src/components/Checkbox.ios.js
```diff
@@ -0,0 +1,94 @@
...
+
+    return (
+      <TouchableRipple
+        {...this.props}
+        borderless
+        rippleColor={rippleColor}
+        onPress={disabled ? undefined : onPress}
+        style={styles.container}
+      >
+        <View style={{ height: 36 }}>
```
satya164: Let's add `width: 36` and extract out to `StyleSheet.create`

src/components/Checkbox.ios.js
```diff
@@ -0,0 +1,94 @@
...
+    onPress: PropTypes.func,
+    /**
+     * Custom color for checkbox
+     */
+    color: PropTypes.string,
+    theme: PropTypes.object.isRequired,
+  };
+
+  render() {
+    const { checked, disabled, onPress, theme } = this.props;
```
satya164: Let's have `...rest` and pass it instead of `...this.props`

src/components/Switch.js
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Disable toggling the switch
+   */
+  disabled?: boolean,
+  /**
+   * Switch value- true or false
+   */
+  value?: boolean,
+  /**
+   * Invoked with the new value when the value changes
```
satya164: This comment is out of place

src/components/Switch.js
```diff
@@ -0,0 +1,86 @@
...
+      disabled,
+      onValueChange,
+      color,
+      theme,
+      ...props
+    } = this.props;
+
+    const checkedColor = color || theme.colors.accent;
+    const trackTintColor =
+      Platform.OS === 'android'
```
satya164: I usually prefer to have `Platform.OS === 'ios'` so that Android is default on other platforms if we ever support them. e.g. uwp and web

PR #158, opened 2017-10-04
Fix TouchableRipple throwing because of `setNativeProps`

ahmedlhanafy:
This fixes https://github.com/callstack/react-native-paper/issues/157

### DIFF
--- a/src/components/TouchableRipple.js
+++ b/src/components/TouchableRipple.js
@@ -47,6 +47,13 @@ export default class TouchableItem extends PureComponent<
     borderless: false,
     rippleColor: 'rgba(0, 0, 0, .32)',
   };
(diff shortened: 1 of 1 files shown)


satya164:
I'm not sure this is the proper fix. This changes the behaviour on iOS by wrapping the child in an additional View, which means that if you want to have different styling, say stretch the element to full screen with `flex: 1`, it won't work.

ahmedlhanafy:
@satya164 Hmmm 🤔 So what do you propose?

satya164:
Dunno. I'll take a look.

lukewalczak:
We can wrap icon in view. https://github.com/callstack/react-native-paper/pull/163

PR #160, opened 2017-10-11
Tab Navigation

Krizzu:
Yo!
This is a friendly wrapper over `rn-material-bottom-navigation` which follows Google Material designs.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -39,7 +39,9 @@
 
 ; Ignore some modules we don't need to parse
 .*/node_modules/prettier/.*
-.*/node_modules/eslint.*
+.*/node_modules/eslint/.*
+.*/node_modules/react-native-material-bottom-navigation/lib/.*
+
 
 [include]
 
--- a/src/components/BottomTabs/BottomTabs.js
+++ b/src/components/BottomTabs/BottomTabs.js
@@ -1,0 +1,135 @@
+/* @flow */
+
+import React from 'react';
+import { StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import withTheme from '../../core/withTheme';
+import BottomNavigation, { Tab } from 'react-native-material-bottom-navigation';
+import Icon from 'react-native-vector-icons/MaterialIcons';
+import type { Theme } from '../../types/Theme';
+
+type TabType = {
+  /*
+   * Label for tab tab
+   */
+  label: string,
+
+  /*
+   * Name of the icon from Material Icons
+   * https://material.io/icons/
+   */
+  icon: string,
+};
+
+type Props = {
+  /*
+   * Wheter we color background of the navigation
+   * If true, primary color is used, while icons and ripple is white
+   * otherwise, color icons and ripple with primary color, keeping bg white
+   */
+  colorNavigationBackground: boolean,
+
+  /*
+   * Decides how tabs should look like
+   */
+  tabs: TabType[],
+
+  /*
+   * Invoked with `newTabIndex` and `oldTabIndex` when tab is switched
+   */
+  onTabChange: (newTabIndex: number, oldTabIndex: number) => mixed,
+  theme: Theme,
+};
+
+class BottomTabs extends React.Component<void, Props, void> {
+  static propTypes = {
+    tabs: PropTypes.arrayOf(
+      PropTypes.shape({
+        label: PropTypes.string,
+        icon: PropTypes.string,
+      })
+    ).isRequired,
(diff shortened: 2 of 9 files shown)


src/components/BottomTabs/BottomTabs.js
```diff
@@ -0,0 +1,135 @@
...
+  componentWillReceiveProps(nextProps) {
+    this._warnAboutTabNumber(nextProps);
+  }
+
+  /*
+   * We use Bottom Navigation only if tab number is between 3 and 5
+   */
+  _warnAboutTabNumber = ({ tabs }) => {
+    if (tabs.length < 3 || tabs.length > 5) {
+      throw new Error('Number of tabs in BottomTabs must be between 3 and 5.');
```
ahmedlhanafy: Why does it have to be from 3 to 5?

Krizzu: G designs says `Three to five top-level destinations`

ferrannp: I feel this should be `tabs.length > 5`. Because I think it is fine to use only 2. Moreover, not sure if we should throw an error or a warning say, hey! this might look bad if you use more than 5 options.

.flowconfig
```diff
@@ -39,7 +39,9 @@
 
 ; Ignore some modules we don't need to parse
 .*/node_modules/prettier/.*
-.*/node_modules/eslint.*
+.*/node_modules/eslint/.*
+.*/node_modules/react-native-material-bottom-navigation/lib/.*
```
ferrannp: Mmm I saw the library uses Flow, would be nice to take advantage of it. Do you need to ignore the whole library or we can be more specific?

src/components/BottomTabs/BottomTabs.js
```diff
@@ -0,0 +1,135 @@
...
+import { StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import withTheme from '../../core/withTheme';
+import BottomNavigation, { Tab } from 'react-native-material-bottom-navigation';
+import Icon from 'react-native-vector-icons/MaterialIcons';
+import type { Theme } from '../../types/Theme';
+
+type TabType = {
+  /*
+   * Label for tab tab
```
ferrannp: Duplicated word `tab`.

src/components/BottomTabs/BottomTabs.js
```diff
@@ -0,0 +1,135 @@
...
+   * Wheter we color background of the navigation
+   * If true, primary color is used, while icons and ripple is white
+   * otherwise, color icons and ripple with primary color, keeping bg white
+   */
+  colorNavigationBackground: boolean,
+
+  /*
+   * Decides how tabs should look like
+   */
+  tabs: TabType[],
```
ferrannp: Mmm not sure if we should go this way or have instead a `BottomTabItem` component to use as children. @satya164 ?

src/components/BottomTabs/BottomTabs.js
```diff
@@ -0,0 +1,135 @@
...
+
+  /*
+   * Invoked with `newTabIndex` and `oldTabIndex` when tab is switched
+   */
+  onTabChange: (newTabIndex: number, oldTabIndex: number) => mixed,
+  theme: Theme,
+};
+
+class BottomTabs extends React.Component<void, Props, void> {
+  static propTypes = {
```
ferrannp: I'd remove `propTypes` now that we use Flow.

src/components/BottomTabs/BottomTabs.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+  };
+
+  /*
+   * Create a Tab based on passed `tabs` props
+   */
+  _tabFactory = (tab: TabType, key: ?any) => {
+    const { theme, colorNavigationBackground } = this.props;
+    const activeColor = colorNavigationBackground
+      ? '#fff'
```
ferrannp: Can we use a constant instead? From colors.

.flowconfig
Krizzu: This lib has some inner flow issues, that's why I disabled it.

src/components/BottomTabs/BottomTabs.js
Krizzu: Yup 👍

src/components/BottomTabs/BottomTabs.js
Krizzu: Done.

Krizzu:
Closing this one down, as we'll create Bottom Tabs from scratch

PR #183, opened 2017-10-18
feat: add ripple example

### LINKED ISSUE #168: Add example for ripple

lukewalczak:
close #168


example/src/RippleExample.js
```diff
@@ -1,11 +1,38 @@
...
+  };
 
   render() {
-    return null;
+    const { theme: { colors: { background } } } = this.props;
+    return (
+      <TouchableRipple
+        style={[styles.container, { backgroundColor: background }]}
+        onPress={() => {}}
+        borderless
```
ferrannp: I would make the `borderless` example to be an icon, then you can add another example without `borderless`. Just missing here more than one example.

lukewalczak: Updated

Issue #175, opened 2017-10-18
Add Helper text and Error text to TextInput


ferrannp:
Follow https://material.io/guidelines/components/text-fields.html

PezCoder:
I would like to work on this.
I'm new to this project so I'll read up the contribution doc & [textInputDoc](https://callstack.github.io/react-native-paper/textinput.html) first to understand & will create a PR

PR #184, opened 2017-10-18
TextInput: Helper Text & error text

PezCoder:
#### What ? 
Add Helper text & error text based on 
https://material.io/guidelines/components/text-fields.html

Issue: https://github.com/callstack/react-native-paper/issues/175

#### Style specifications for helperText & errorText
screen shot 2017-10-19 at 4 12 32 pm

screen shot 2017-10-19 at 4 17 19 pm

#### Things to clear up 
- [ ] bottomLineContainer has a margin Bottom of 4px, why is it needed ? 
As i'm adding helper text below that i'm maintaining a 8px margin on top of it by having 4px on top of helper i.e (4px + 4px = 8px) & 4px below helper to maintain an overall margin-bottom below the TextInput that was previously put by bottomLineContainer.
- [ ] This is the behaviour of error state:
1. Whenever hasError is true, error state will show up
2. HelperText will be replaced by errorText until the error is fixed i.e hasError is false
3. HelperText, label & borderLine will have the color mentioned in the guidelines for invalid input
4. For disabled inputs, there won't be possibility for having an error state, hence even when hasError is provided & true. The styles of `disabled` will take priority over `error` styles.

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -60,12 +60,21 @@ type Props = {
    * Text that gives context about a field’s input
    */
   helperText?: string,
+  /**
+   * Represents error state, true represents an invalid input
+   */
+  hasError?: boolean,
+  /**
+   * Text to show for error state, visible only when hasError prop is true
+   */
+  errorText?: string,
   style?: any,
   theme: Theme,
 };
 
 type DefaultProps = {
   disabled: boolean,
+  hasError: boolean,
 };
 
 type State = {
@@ -112,11 +121,14 @@ class TextInput extends Component<DefaultProps, Props, State> {
     style: ViewPropTypes.style,
     value: PropTypes.string,
     helperText: PropTypes.string,
+    hasError: PropTypes.bool,
+    errorText: PropTypes.string,
     theme: PropTypes.object.isRequired,
   };
 
   static defaultProps = {
     disabled: false,
+    hasError: false,
   };
 
   constructor(props: Props) {
@@ -220,6 +232,8 @@ class TextInput extends Component<DefaultProps, Props, State> {
       label,
       underlineColor,
       helperText,
+      hasError,
+      errorText,
       style,
       theme,
       ...rest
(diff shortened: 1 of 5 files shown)


src/components/TextInput.js
```diff
@@ -56,12 +56,25 @@ type Props = {
...
+  /**
+   * Text that gives context about a field’s input
+   */
+  helperText?: string,
+  /**
+   * Represents error state, true represents an invalid input
+   */
+  hasError?: boolean,
+  /**
+   * Text to show for error state, visible only when hasError prop is true
```
ferrannp: I would have only one prop `error`, if is `''` which is falsy value means no error, if has text so there is an error and you can show it.

src/styles/DefaultTheme.js
```diff
@@ -23,6 +31,10 @@ export default {
     placeholder: color(black)
       .alpha(0.38)
       .rgbaString(),
+    helperText: color(black)
+      .alpha(0.54)
+      .rgbaString(),
+    error: redA400,
```
ferrannp: We need those in `DarkTheme` too.

ferrannp:
For your first point of things to clear up cc @satya164.

PezCoder:
Updated the PR description with "Style descriptions for both helperText & errorText

@ferrannp Need your input on a few things: 
1. For disabled input, the helper text/error won't show up. 
I assumed this, as there wasn't any documentation for their styles in the disabled section, Also I think those are to help user type something & if the field is disabled it doesn't make sense to have them.
2. I'm a bit confused on how the focus of the invalid field should work, i.e if a user tries to leave, 
    1. should we keep the previous invalid field focused, for the user to correct with change of colors
    2. or let the user go wherever he click but change the colors

src/components/TextInput.js
PezCoder: Yeah, this will definitely reduce a variable but think of it like this from client's perspective who is consuming our component.
1. By the current scenario, an `errorText` is defined just once when declaring the component itself & it can become valid or invalid multiple times in the future i.e a boolean is the correct way to represent that valid state.
2. The `errorText` is defined with the component i.e I don't have to read any validation logic to know what the component will display as an error message but this won't be the case if I make it one i.e the validation logic will also have to contain what to show as error.

What do you think ? Is a there an advantage of making it one ?

PezCoder:
Self reviewed & tested everything from my side.

PezCoder:
@satya164 @ferrannp Can we please finish the review for this ?
I'm quite new to open source community so pretty excited to get the merged. 😁

satya164:
Sorry, I haven't gotten time to review. I'll do it this week.

satya164:
Hey @PezCoder. Sorry for the long delay in review. So I wanted to have a separate `HelperText` component so that it could be reused in other places than text input. In future, we could provide a form component which handles errors etc. for all of the fields inside it.

PezCoder:
@satya164 Sorry but I mightn't be able to work on this sooner, maybe somebody else will pick it up from here.
Closing this but can be used as a reference if needed.

shujathkhan:
@satya164  Any progress on helperText for error validation messages? If not, what would the advisable approach be for handling helperText?

Issue #169, opened 2017-10-17
Make some tweaks to Dark theme


ferrannp:
Check Button example.

felixaa:
I'd love to help out! What needs to be done here exactly?

ferrannp:
@felixaa awesome! Just checkout the example app with the dark theme activated. There are some missing points (you will see in the `Button example`, not everything is correct). We just need to make sure that all examples look correctly using the dark theme.

felixaa:
I'm not really into using expo tbh. But I tried viewing the app using the QR. Didn't see any dark/light-theme toggle?🤔

Harkindey:
@felixaa working on this now.

Harkindey:
@felixaa i peeped into this, but i'm finding it hard to identify differences between this two 
<img width="376" alt="screen shot 2017-10-21 at 7 21 16 am" src="https://user-images.githubusercontent.com/14913673/31848778-98d57db2-b630-11e7-9a2b-95806f12be0d.png">
<img width="376" alt="screen shot 2017-10-21 at 7 21 33 am" src="https://user-images.githubusercontent.com/14913673/31848779-99094ebc-b630-11e7-9517-3f19e98ff76e.png">

SO CAN YOU POINT OUT WHAT ISN'T WORKING WELL OR WHAT YOU MEAN BY THAT.

satya164:
#230

PR #203, opened 2017-11-06
fix: correct components using dark theme

lukewalczak:
Solve #169


src/components/Divider.js
```diff
@@ -27,20 +32,37 @@ type Props = {
  * ```
  */
 const Divider = (props: Props) => {
-  const { inset, style } = props;
+  const { inset, style, theme } = props;
+  const { dark: isDarkTheme } = theme;
+  console.log(isDarkTheme);
```
satya164: remove

Issue #205, opened 2017-11-07
Set `js` language inside of Component's example for proper highlight


jukben:
Sice we updated Components-docs for a better syntax, there is sadly no way how to preserve "auto" fallback.

PR #206, opened 2017-11-07
docs: set js inside component for proper highlight

lukewalczak:
Solve #205


ferrannp:
Why divider doesn't work and RadioButton shows wrong, is that a bug @satya164 ?

lukewalczak:
Sure!

Issue #166, opened 2017-10-16
Create a row component for Checkboxes and Radio Buttons


ferrannp:
If we want to use a checkbox or a radio button with a label, common pattern is that the whole `row` is clickable. We should have a component in place for it.

Example where this behaviour is needed is in the Dialog with Radio Buttons example.

felixaa:
I'd love to jump on this if possibru!🙏🏽

Trancever:
Created pull request for this issue - #219

PR #214, opened 2017-12-26
feat: Implemented radio group component.

Trancever:
Added implementation of RadioGroup component and example of how to use it. 
I have problem with styling of this component. Should I wrap it in withTheme HOC? Currently, i made possibility to pass style prop and it will be set to the View element that wraps Radio Buttons.


satya164:
Hey @Trancever! Thanks for the PR.

What do you think about something like this for the API:

```js
<RadioGroup value={this.state.value} onValueChange={value => this.setState({ value })}>
  <RadioButton value="first" />
  <RadioButton value="second" />
  <RadioButton value="third" />
</RadioGroup>
```

I think it's better than using index, since index can be confusing (e.g. what happens when one element is `null`, does it count or no) and using a string for value makes it easier to spot the current one (with index you'll need to count).

Regarding the implementation, I think we would need to use context instead of `cloneElement`. Because many times we'd need to add a label or some other stuff, and we'll write code like this:

```js
<View>
  <RadioButton />
  <Text>First one</Text>
</View>
```

Due to `RadioButton` being nested inside a `View`, our `cloneElement` won't work.

In addition to this, I think it'll be nice to also provide a `RadioGroupItem` component which you can use like:

```js
<RadioGroup value={this.state.value} onValueChange={value => this.setState({ value })}>
  <RadioGroupItem label="First" value="first" />
  <RadioGroupItem label="Second" value="second" />
  <RadioGroupItem label="Third" value="third" />
</RadioGroup>
```

But that can be a separate PR :)

Trancever:
Hey @satya164 , Thanks for review. 
I thought about using value prop instead of indexes but currently, RadioButton does not accept value prop. Maybe if we could change a RadioButton component to accept value prop and make it pass value as an argument to onPress function when radio button is clicked? What do you think? 

Regarding to RadioGroupItem, I created PR already, because there was an issue for that component - #166

src/components/RadioGroup.js
```diff
@@ -0,0 +1,71 @@
+/* @flow */
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Broadcast } from 'react-broadcast';
```
satya164: We are using a `create-react-context` in other places, so let's switch to it. It has a similar API, so shouldn't be an issue.

src/components/RadioGroup.js
```diff
@@ -0,0 +1,71 @@
...
+
+type Props = {
+  /**
+   * Array of RadioButton elements
```
satya164: Change to "React elements containing radio buttons"

src/components/RadioGroup.js
```diff
@@ -0,0 +1,71 @@
...
+   */
+  children?: React.Node,
```
satya164: Probably`children` shouldn't be optional

src/components/RadioGroup.js
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Function to execute on selection change
+   */
+  onValueChange: Function,
```
satya164: Change type to `(value: string) => mixed`

src/components/withRadioGroup.js
```diff
@@ -0,0 +1,67 @@
...
+
+function withRadioGroup(
+  Component: React.ComponentType<any>
+): React.ComponentType<any> {
+  return class extends React.Component<Props> {
+    render() {
+      const { value } = this.props;
+
+      return (
+        <Subscriber channel={channel}>
```
satya164: Let's remove the HOC and move this code to `RadioButton` so that we don't have to wrap radio button when we use it

src/components/RadioGroup.js
```diff
@@ -0,0 +1,79 @@
...
+   * Function to execute on selection change
+   */
+  onValueChange: (value: string) => mixed,
+  /**
+   * Value of currently selected Radio
+   */
+  value: string,
+};
+
+type Ctx = {
```
satya164: Let's rename the type to `RadioContext`

src/components/RadioGroup.js
```diff
@@ -0,0 +1,79 @@
...
+ *       </RadioGroup>
+ *     )
+ *   }
+ * }
+ *```
+ */
+
+export const RadioGroupContext: Context<Ctx> = createReactContext({
+  value: '',
+  passed: false,
```
satya164: Instead of empty string as value and `passed` property, I'd change this to:

```js
const RadioGroupContext: Context<?RadioContext> = createReactContext(null);
```

Then inside the `RadioButton` components, you check if you get a `null` from the context to decide if there's a provider in parent or not.

src/components/RadioGroup.js
```diff
@@ -0,0 +1,79 @@
...
+ *     value: 'first',
+ *   };
+ *
+ *   render() {
+ *     return(
+ *       <RadioGroup
+ *         onValueChange={value => this.setState({ value })}
+ *         value={this.state.value}
+ *       >
+ *         <View>
```
satya164: Can you put something like `<Text>First</Text>` before the button? So when you copy/paste, it's more obvious.

src/components/RadioButton.js
```diff
@@ -117,54 +122,75 @@ class RadioButton extends React.Component<Props, State> {
...
+
+          return (
+            <TouchableRipple
+              {...rest}
+              borderless
+              rippleColor={rippleColor}
+              onPress={
+                disabled
+                  ? undefined
+                  : passed ? () => onValueChange(this.props.value) : onPress
```
satya164: I'd make it like:

```
(e) => {
  onValueChange(this.props.value);
  onPress && onPress(e);
}
```

So that the `onPress` still gets called if provided.

example/src/RadioButtonExample.js
```diff
@@ -62,11 +62,11 @@ class RadioButtonExample extends React.Component<Props, State> {
         </TouchableRipple>
         <View style={styles.row}>
           <Paragraph>Checked (Disabled)</Paragraph>
-          <RadioButton checked disabled />
+          <RadioButton value="checked disabled" checked disabled />
```
satya164: Let's rename the value to something like `value="first"` to avoid confusion. First time I saw it I thought the value is "checked"

mayteio:
👏would a similar API for checkboxes make sense? I have a multi-select dialog that I'd like to implement similar functionality within. Radios don't make sense here.

PR #218, opened 2018-01-21
Feat: FAB animations with optional labels

### LINKED ISSUE #210: FAB options
https://material.io/guidelines/components/buttons-floating-action-button.html#buttons-floating-action-button-transitions

Some animation fun.

components-fab-flyouts-do

lukewalczak:
Fixes #210


example/src/AnimatedFABExample.js
```diff
@@ -0,0 +1,50 @@
+/* @flow */
+
+import * as React from 'react';
+import { View, StyleSheet } from 'react-native';
+import { AnimatedFAB, withTheme, Button } from 'react-native-paper';
```
satya164: Let's change the name to `FABGroup`

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+  label?: string,
+  onPress: () => void,
+};
+
+type State = {
+  animatedValue: Animated.Value,
+  isOpen: boolean,
+};
+
+class AnimatedFABExample extends React.Component<Props, State> {
```
satya164: Wrong name :D

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+  constructor(props) {
```
satya164: Can get rid of the constructor

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+
+type Props = {
+  items: Array<Item>,
+  theme: Theme,
+};
+
+type Item = {
+  icon: string,
+  label?: string,
+  onPress: () => void,
```
satya164: Change the signature to `() => mixed`

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+import Card from './Card/Card';
+import ThemedPortal from './Portal/ThemedPortal';
+
+import withTheme from '../core/withTheme';
+import type { Theme } from '../types';
+
+import FAB from './FAB';
+
+type Props = {
+  items: Array<Item>,
```
satya164: We should provide 2 more props:

```js
open: boolean,
onStateChange: (state: { open: boolean }) => mixed,
```

This will allow users to use it as a controlled component, which means:
1. They can open it programmatically
2. They can control status bar styles in response to the state change (see below)

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+    this.state = {
+      animatedValue: new Animated.Value(0),
+      isOpen: false,
+    };
+  }
+
+  componentWillUpdate(_, nextState) {
+    Platform.OS === 'ios' && nextState.isOpen
+      ? StatusBar.setBarStyle('dark-content')
+      : StatusBar.setBarStyle('light-content');
```
satya164: I'm not sure if we should do this. `light-content` works in our example app, because our app has a dark toolbar. However, in apps with a light toolbar, this will erroneously set the status bar content to light.

Let's remove this from here and provide hooks so the user can do it in their app manually (see above). We can document the approach and do it in the example.

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  _handleOpen = () =>
+    this.setState({ isOpen: true }, () => {
+      Animated.timing(this.state.animatedValue, {
```
satya164: Let's move the logic from here and from `_handleDismiss` to `componentDidUpdate`:

```js
componentDidUpdate(prevProps, prevState) {
  if (prevState.isOpen === this.state.isOpen) {
    return;
  }

  const { isOpen } = this.state;

  Animated.timing(this.state.animatedValue, {
    toValue: isOpen ? 1 : 0,
    duration: isOpen ? 300 : 200,
    easing: Easing.out(Easing.sin),
    useNativeDriver: true,
  }).start();
}
```

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+        toValue: 1,
+        duration: 300,
+        easing: Easing.sin,
+        useNativeDriver: true,
+      }).start();
+    });
+
+  _handleDismiss = () =>
```
satya164: Rename the method to `_handleClose` to complement `_handleOpen`

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+    this.setState({ isOpen: false }, () => {
+      Animated.timing(this.state.animatedValue, {
+        toValue: 0,
+        duration: 200,
+        easing: Easing.sin,
+        useNativeDriver: true,
+      }).start();
+    });
+
+  toggleModal = () => {
```
satya164: Prefix the method with `_`

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+                    />
+                  </View>
+                </Animated.View>
+              );
+            })}
+          </View>
+          <FAB
+            dark
+            onPress={this.toggleModal}
+            icon="add"
```
satya164: We should accept an `icon` prop so the user can customize it.

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+                      ]}
+                      onPress={it.onPress}
+                    />
+                  </View>
+                </Animated.View>
+              );
+            })}
+          </View>
+          <FAB
+            dark
```
satya164: `FAB` should automatically detect this from its background color. Let's remove it from here.

src/components/AnimatedFAB.js
```diff
@@ -0,0 +1,231 @@
...
+    justifyContent: 'flex-end',
+  },
+  fab: {
+    marginHorizontal: 16,
+    marginBottom: 16,
+    marginTop: 8,
+  },
+  backdrop: {
+    ...StyleSheet.absoluteFillObject,
+    backgroundColor: 'rgba(255, 255, 255, .9)',
```
satya164: This color should depend on the theme.

PR #227, opened 2018-02-01
 docs: add screenshots of FAB, Paper, ProgressBar and Switch components

kpsroka:


src/components/Switch.js
```diff
@@ -31,6 +31,17 @@ type Props = {
 /**
  * Switch is a visual toggle between two mutually exclusive states—on and off
  *
+ * <div class="screenshots">
+ *   <div>
+ *     <img src="screenshots/switch-android.png" />
+ *     <span>Android</span>
```
satya164: We should add some CSS for a caption

satya164: We should add some CSS for captions

src/components/FAB.js
```diff
@@ -39,6 +39,10 @@ type Props = {
 /**
  * A floating action button represents the primary action in an application
  *
+ * <div class="screenshots">
+ *   <img src="screenshots/fab.png" />
```
satya164: I think we should split items into separate screenshots. For example, paper and switch look noisy with many screenshots in a small dimension https://281-71323749-gh.circle-artifacts.com/0/docs/paper.html

docs/assets/styles.css
```diff
@@ -3,14 +3,22 @@
   margin: 0 -4px;
 }
 
-.screenshots > img {
-  width: 280px;
+.screenshots img {
+  max-width: 280px;
```
satya164: Using `max-width` makes the screenshot sizes non-deterministic. For example you can see that the switch and paper screenshots look smaller than the button screenshots. Let's change it back to absolute width.

docs/assets/styles.css
```diff
@@ -3,14 +3,22 @@
...
   height: auto;
+}
+
+.screenshots > div {
+  display: inline-flex;
+  flex-direction: column;
+}
+
+.screenshots > * {
   margin: 4px;
```
satya164: The margin in button screenshots and switch screenshots seem different. Can you check that?

src/components/Switch.js
kpsroka: What style do you suggest? Perhaps italics, smaller font?

src/components/FAB.js
kpsroka: So, for the switch examples you prefer 8 screenshots with all states, or should I drop some states from snapshots?

docs/assets/styles.css
kpsroka: The only difference I see a margin between the screenshots and the paragraph above. Where do you see the difference?

docs/assets/styles.css
kpsroka: I can't see what you see :/

src/components/FAB.js
satya164: I think 3 screenshots with 2 items in each of them should be fine!

src/components/Switch.js
satya164: Let's copy from here https://material.io/guidelines/components/bottom-navigation.html#bottom-navigation-usage :p

docs/assets/styles.css
satya164: Oh, I think it looks that way because of the small font size in some screenshots. Have you resized them? I kept the size 1:1 when creating the screenshots.

docs/assets/styles.css
satya164: Yes, that's the margin I'm talking about

docs/assets/styles.css
kpsroka: I updated the screenshots with Pixel 2 emulator (for Android). Now they should look more similar to the button ones.

docs/assets/styles.css
kpsroka: That was caused by change to the "Usage" label below. I updated that labels now.

src/components/FAB.js
kpsroka: Done.

src/components/Switch.js
kpsroka: Done.

src/components/Typography/Paragraph.js
```diff
@@ -31,7 +31,7 @@ export default Paragraph;
 
 const styles = StyleSheet.create({
   text: {
-    fontSize: 14,
+    fontSize: 20,
```
satya164: Is this intentional?

kpsroka: No, reverted.

PR #231, opened 2018-02-05
fix: disabling Touchables if onPress was not set to allow event handling in parent components (#217)

### LINKED ISSUE #217: DrawerItem should have pointerEvents="none" when onPress not set
As in the title, when having DrawerItem embedded within Drawer (react-navigation), I don't want to set `onPress` and let the built-in handlers do their thing. 

Right now, DrawerItem registers Touchable that captures touches and automatically prevents the built-in handler from being called.

Expected behaviour:

I would expect DrawerItem to either not have Touchable component or have pointerEvents none when `onPress` is undefined.

kpsroka:
Fixes #217.


example/src/ButtonExample.js
```diff
@@ -27,24 +27,33 @@ class ButtonExample extends React.Component<Props, State> {
     };
     const source = require('../assets/chameleon.jpg');
     const { theme: { colors: { background } } } = this.props;
+    const dummyFunction = () => {}; // There's no ripple effect without a valid onPress prop.
```
ferrannp: I would prefer to stick with inline function in all elements like `<Button onPress={(} => {}}>Simple</Button>` if not my brain needs to figure out all the time, wait, what is `dummyFunction` ?

src/components/TouchableRipple.js
```diff
@@ -75,13 +75,16 @@ export default class TouchableRipple extends React.Component<Props, void> {
       ...rest
     } = this.props;
 
+    const disabled = this.props.onPress === undefined;
```
ferrannp: We need to be able to pass `disabled` as a prop too. This should be: `this.props.disabled || !this.props.onPress`.

example/src/ButtonExample.js
kpsroka: Done.

src/components/TouchableRipple.js
kpsroka: Done. I also extracted the disabled prop from the `...rest` spread, so we don't pass it twice.

kpsroka:
PTAL :)

PR #236, opened 2018-02-08
docs: add screenshots for Card and Typography

satya164:


src/components/Card/Card.js
```diff
@@ -40,6 +40,11 @@ type State = {
 /**
  * A card is a sheet of material that serves as an entry point to more detailed information.
  *
+ * <div class="screenshots">
+ *   <img style="width: 480px" src="screenshots/card-1.png" />
+ *   <img style="width: 480px" src="screenshots/card-2.png" />
```
ferrannp: Can we make it a bit smaller? They do not fit next to each other on my MBP 13".  400px would do.

Issue #237, opened 2018-02-08
Divider color should be inside theme (to be able to override it too)


ferrannp:
Right now the `dark` theme does not work for it. See: https://github.com/callstack/react-native-paper/blob/master/src/components/Divider.js#L53-L66

satya164:
@ferrannp can you elaborate what the issue is with a code sample and screenshot? divider works fine for me with the dark theme in the example app.

Trancever:
For me also divider works fine in both themes in the example app and I am able to override it. Haven't checked it carefully before submitting PR.. :D

ferrannp:
Yeah I guess my problem has been that the divider in dark theme for iOS is almost not visible for my taste here:

<img width="425" alt="screen shot 2018-02-10 at 19 06 43" src="https://user-images.githubusercontent.com/774577/36065060-9fff63ce-0e95-11e8-9e89-3d50b960665b.png">

So I wanted to override the color like:

<img width="423" alt="screen shot 2018-02-10 at 19 07 09" src="https://user-images.githubusercontent.com/774577/36065063-a80bd52a-0e95-11e8-8921-d359ffb0d894.png">

However, I'd like to have all colors as part of the `theme`. 

Maybe can we extract the nested `theme` so it's easier to override/extend the type? Like `type Theme = ThemeColors & ThemeFonts...` ?

For me it would be fine to have colors that material guide specify like for dividers and icons directly into the theme but if you don't want that, I'm fine too.

satya164:
Let's check it on Monday

ferrannp:
Caused by using older version.

PR #238, opened 2018-02-09
fix: moves divider color to theme.

Trancever:
It's a fix for #237 .

Moves divider color from StyleSheet to theme's colors.

### DIFF
--- a/src/components/Divider.js
+++ b/src/components/Divider.js
@@ -1,11 +1,9 @@
 /* @flow */
 
 import * as React from 'react';
-import color from 'color';
 import { StyleSheet, View } from 'react-native';
 import withTheme from '../core/withTheme';
 import type { Theme } from '../types';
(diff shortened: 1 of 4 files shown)


satya164:
I avoid moving component specific styles to theme because while it makes it easy to modify the theme to update the components, it also puts tons of configs in the theme. Now instead of tweaking a few colors and getting everything adapted to it, you need to update a lot of stuff for each theme. Besides, if someone wants this behaviour, they can always wrap the component in a custom one. Individual components should allow customizing their colors and also automatically adapt themselves to light and dark themes.

I'm not exactly sure what's the issue in #237 because divider seems fine in dark theme for me in the example app.

ferrannp:
Ok I need to check why divider was not looking good in my app with dark theme but we won't add it into theme on paper (you can do it in your own app).

PR #239, opened 2018-02-10
add bootstrap script to run yarn

manjula91:
Add bootstrap script to run yarn in example and docs folder

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,6 @@
     "flow": "flow",
     "lint": "eslint .",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "bootstrap": "cd example && yarn && cd ../docs && yarn"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
     "precommit": "yarn run lint && yarn run flow",
     "flow": "flow",
     "lint": "eslint .",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "bootstrap": "cd example && yarn && cd ../docs && yarn"
```
satya164: I think we can do `yarn --cwd example && yarn --cwd docs` which looks nicer. We also need to add it to CONTRIBUTING.md

manjula91: Cool

PR #240, opened 2018-02-10
tests: add tests commit type

manjula91:


scripts/validate-commit-message.js
```diff
@@ -16,6 +16,7 @@ const TYPES = {
   chore: true,
   revert: true,
   breaking: true,
+  tests: true
```
satya164: Hey, we already have `test` above, so we don't need it

manjula91:
yo cool i dint notice may be we should add it to the contributing.md as
well ;)

manjula91:
@satya164 added a spell check and test to contributing.md

PR #243, opened 2018-02-16
feat: implement helper and error text support in TextInput

### LINKED ISSUE #175: Add Helper text and Error text to TextInput

kpsroka:
Please take a look at the example app for a demo of how it looks like/works. For now, I didn't want to refactor this into a separate element -- we should do that once we know what it should look like in other components. One not implemented feature is to have an icon in the error text. The [material docs](https://material.io/guidelines/components/text-fields.html#) use error-outline, but we might consider allowing any (or any material?) icon source.

(Resolves #175)


kpsroka:
@satya164 Can you please take a look at this?

src/components/TextInput.js
```diff
@@ -26,6 +26,18 @@ type Props = {
...
    */
   placeholder?: string,
+  /**
+   * Helper text to display under the input.
+   */
+  helperText?: string,
+  /**
+   * Whether to style the TextInput with error style.
+   */
+  hasError: boolean,
```
satya164: I think instead of this, a `validate` prop might be better,

```
validate={text => text.length < 3 ? new Error('Please enter at least 3 characters') : null}
```

There are 2 advantages: first, it can be used with both controlled and uncontrolled inputs, currently, you need to maintain a state in your component to do validation. second, it reduces duplication and makes impossible states impossible, e.g.-

```
hasError={(() => {
  if (this.state.text.length < 3) {
    return true;
  }
})()}

```

satya164: I think instead of this, a `validate` prop might be better,

```js
validate={text => text.length < 3 ? new Error('Please enter at least 3 characters') : null}
```

There are 2 advantages: first, it can be used with both controlled and uncontrolled inputs, currently, you need to maintain a state in your component to do validation. second, it reduces duplication since the logic of both checking whether there is an error and the error message are in the same place, and makes impossible states impossible, e.g. the user cannot pass `hasError={true}` without an error text.

src/components/TextInput.js
```diff
@@ -208,43 +231,73 @@ class TextInput extends React.Component<Props, State> {
...
+      value && hasError
+        ? this.state.errorShown.interpolate({
+            inputRange: [0, 0.5, 1],
+            outputRange: [0, 4, 0],
+          })
+        : 0;
+
+    /* Move label to top if value is set */
+    const labelTranslateY = value
       ? -22
```
satya164: Let's extract this number to a constant to the top of the file

src/components/TextInput.js
```diff
@@ -208,43 +231,73 @@ class TextInput extends React.Component<Props, State> {
...
       : this.state.focused.interpolate({
           inputRange: [0, 1],
           outputRange: [0, -22],
         });
-    const fontSize = value
+
+    const labelFontSize = value
       ? 12
```
satya164: Extract to a constant

src/components/TextInput.js
```diff
@@ -301,12 +380,44 @@ class TextInput extends React.Component<Props, State> {
         />
         <View pointerEvents="none" style={styles.bottomLineContainer}>
           <View
-            style={[styles.bottomLine, { backgroundColor: inactiveColor }]}
+            style={[
+              styles.bottomLine,
+              { backgroundColor: (hasError && errorColor) || inactiveColor },
```
satya164: `hasError ? errorColor : inactiveColor` looks better

src/components/TextInput.js
```diff
@@ -301,12 +380,44 @@ class TextInput extends React.Component<Props, State> {
...
+            ]}
           />
           <Animated.View
-            style={[styles.bottomLine, styles.focusLine, bottomLineStyle]}
+            style={[
+              styles.bottomLine,
+              styles.focusLine,
+              bottomLineStyle(bottomLineColor, this.state.focused),
```
satya164: Let's rename it to something like `getBottomLineStyle`

src/components/TextInput.js
```diff
@@ -301,12 +380,44 @@ class TextInput extends React.Component<Props, State> {
...
+              styles.focusLine,
+              bottomLineStyle(
+                errorColor,
+                // $FlowFixMe$
+                Animated.multiply(this.state.focused, this.state.errorShown)
+              ),
+            ]}
           />
         </View>
+        {(helperText || errorText) && (
```
satya164: This will return the `helperText`/`errorText` if they are empty string and `{(helperText || errorText) && (...)}` will evaluate to `''`, RN will throw an error saying that it's not possible to render raw string outside text. let's change this to a ternary

src/styles/DarkTheme.js
```diff
@@ -22,6 +22,11 @@ const DarkTheme: Theme = {
       .alpha(0.38)
       .rgb()
       .string(),
+    helperText: color(white)
```
satya164: We don't put component specific styles in theme. Let's move these to the component

src/components/TextInput.js
kpsroka: Done.

src/components/TextInput.js
kpsroka: Done.

src/components/TextInput.js
kpsroka: Done.

src/styles/DarkTheme.js
kpsroka: I removed helperText from here, however I think that the 'errorText' is generic enough and helpful enough (for custom theming) to warrant its inclusion here.

src/components/TextInput.js
kpsroka: Done.

kpsroka:
How would you like the extracted component to be used? Let's assume we have a `HelperText` component. We have the following options:
1. We let users use `HelperText` without restrictions, then they need to wrap it in some other `View` that aligns their components with `HelperText`. So, there's a lot of repetition on the client side, since there's a very common use case of having them under `TextInput`. Also, the users need to handle potential resizing of the `HelperText`, which can lead to error-prone code.
    1. We can add a prop to `TextInput` (and other common cases) to use a custom component in the underline area, and manage potential issues with resizing as well. This is quite permissive, and can lead to users abusing this and ending up writing crazy issues.  
1. We keep `HelperText` internal, and either:
    1. spread the props to `TextInput` (and other form components), or
    1. introduce a `helperTextProps` prop to pass on to `HelperText` (and disable `HelperText` altogether).

I'm in favor of the last option. This makes sure that users don't need to spend time on possibly error-prone code. On the other hand, it encapsulates the `HelperText` from other components, so that breaking changes in `HelperText` are restricted to users of that feature.

satya164:
> but we can add a Form component later which handles validation for various components and reduces boilerplate

kpsroka:
Hi @satya164 ,
As requested, I split most of the functionality into a separate component. The changes in `TextInput` are to support red coloring and wiggle animation of minimized label. I don't think introducing the 'verify' prop in `TextInput` is a good think -- the upcoming `FormInput` seems like a better place for such functionality.

PTAL.

src/components/TextInput.js
kpsroka: Done.

src/components/TextInput.js
kpsroka: I think this should wait for the `FormInput` helper.

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+  /**
+   * @optional
+   * Helper text to display.
+   */
+  helperText?: string,
+  /**
+   * @optional
+   * Whether to style the TextInput with error style.
+   */
+  hasError?: boolean,
```
satya164: I think this is a better API:

```js
type Props = {
  type: 'error' | 'info',
  visible: boolean,
  children: React.Node,
}
```

Then it will be more flexible, and in future, we could add more types like `warning`, `success` etc. The animation can happen when `visible` prop changes.

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      errorShown: new Animated.Value(props.hasError ? 1 : 0),
+    };
+  }
+
+  state: State;
```
satya164: This is not needed

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+ *   }
+ * }
+ * ```
+ */
+class HelperText extends React.Component<Props, State> {
+  static defaultProps = {
+    hasError: false,
+  };
+
+  constructor(props) {
```
satya164: Can remove the constructor, this should be enough:

```js
state = {
  errorShown: new Animated.Value(this.props.hasError ? 1 : 0),
}
```

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+  render() {
+    const { helperText, hasError, errorText, style, theme, color } = this.props;
+    const { colors, dark } = theme;
+    const { errorText: errorTextColor } = colors;
+
+    const helperTextColor =
+      color || (hasError && errorTextColor) || this.getHelperTextColor(dark);
+
+    const underlineArea = {
+      height: Animated.multiply(
```
satya164: We should avoid animating height. Height animations are slow and they can't use native driver. Also we don't know the height anyway, so we won't be able to do it. We should just animate the transform.

I'd imagine that the space for helper text doesn't grow, but it's there. When the helper text becomes visible, it scales up.

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+    clearTimeout(this._timer);
+  }
+
+  _timer: any;
+  _root: any;
+  _setRef: any = (c: Object) => {
+    this._root = c;
+  };
+
+  _animateFocus = (animatedValue: Animated.Value) => {
```
satya164: Since this animated value is always the same, no need to accept it as an argument, right?

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+          .alpha(0.7)
+          .rgb()
+          .string()
+      : colorModule(black)
+          .alpha(0.54)
+          .rgb()
+          .string();
+  }
+
+  renderUnderlineText(text?: string, containerStyle: Object, color: string) {
```
satya164: nit: The name `renderUnderlineText` doesn't sound right. Also let's and `_` prefix.

src/components/TextInput.js
```diff
@@ -13,6 +13,11 @@ import type { Theme } from '../types';
 
 const AnimatedText = Animated.createAnimatedComponent(Text);
 
+const minimizedLabelYOffset = -22;
```
satya164: We usually use `UPPER_SNAKE_CASE` for constants

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+  _animateBlur = (animatedValue: Animated.Value) => {
+    Animated.timing(animatedValue, {
+      toValue: 0,
+      duration: 180,
+    }).start();
+  };
+
+  getHelperTextColor(dark: boolean) {
+    return dark
+      ? colorModule(white)
```
satya164: Let's do:

```js
color(theme.colors.text)
  .alpha(dark ? 0.7 : 0.54)
  .rgb()
  .string();
```

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+          .alpha(0.54)
+          .rgb()
+          .string();
+  }
+
+  renderUnderlineText(text?: string, containerStyle: Object, color: string) {
+    return (
+      text && (
+        <Animated.View style={containerStyle}>
+          {text && <Text style={[styles.helperText, { color }]}>{text}</Text>}
```
satya164: Redundant check

src/components/TextInput.js
```diff
@@ -26,6 +31,10 @@ type Props = {
    * Placeholder for the input.
    */
   placeholder?: string,
+  /**
+   * Whether to style the TextInput with error style.
+   */
+  hasError: boolean,
```
satya164: Maybe just `error: boolean` since we usually skip prefixes everywhere else, like `disabled`, `multiline`

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+        ) : null}
+      </View>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  helperText: {
+    fontSize: 12,
+    marginTop: 4,
```
satya164: Maybe this should be `marginVertical: 4`?

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+  /**
+   * @optional
+   * Text to replace the helper text with on error. If none set, will use the helperText value.
+   */
+  errorText?: string,
+  /**
+   * @optional
+   * Text color to use.
+   */
+  color?: string,
```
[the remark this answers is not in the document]
kpsroka: Which theme color should be used here? The [MD specs](https://material.io/guidelines/components/text-fields.html) differentiate between the placeholder and helper text colors. Should we introduce a helper text color in the theme?

src/components/HelperText.js
kpsroka: What effect would the `type` prop have? Maybe it's better to have `style` instead?

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+    this.state = {
+      errorShown: new Animated.Value(props.hasError ? 1 : 0),
+    };
+  }
+
+  state: State;
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.hasError !== this.props.hasError) {
+      (nextProps.hasError ? this._animateFocus : this._animateBlur)(
```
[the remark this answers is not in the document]
kpsroka: ✅

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+        this.state.errorShown
+      );
+    }
+  }
+
+  componentWillUnmount() {
+    clearTimeout(this._timer);
```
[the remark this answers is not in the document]
kpsroka: ✅

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  _timer: any;
+  _root: any;
+  _setRef: any = (c: Object) => {
```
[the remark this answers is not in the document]
kpsroka: ✅

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
+/* @flow */
+
+import * as React from 'react';
+import colorModule from 'color';
```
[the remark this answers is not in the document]
kpsroka: ✅

src/components/HelperText.js
kpsroka: ✅

src/components/HelperText.js
kpsroka: ✅

src/components/HelperText.js
kpsroka: ✅

src/components/TextInput.js
kpsroka: ✅

src/components/HelperText.js
kpsroka: ✅

src/components/HelperText.js
kpsroka: ✅

src/types.js
```diff
@@ -8,9 +8,11 @@ export type Theme = {
     background: string,
     paper: string,
     accent: string,
+    error: string,
     text: string,
     disabled: string,
     placeholder: string,
+    errorText: string,
```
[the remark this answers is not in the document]
kpsroka: ✅

src/components/HelperText.js
```diff
@@ -0,0 +1,224 @@
...
+    super(props);
+
+    this.state = {
+      errorShown: new Animated.Value(props.hasError ? 1 : 0),
+    };
+  }
+
+  state: State;
+
+  componentWillReceiveProps(nextProps) {
```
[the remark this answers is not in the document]
kpsroka: ✅

src/components/HelperText.js
kpsroka: ✅

kpsroka:
I applied the requested changes, let's take a look at it together on Tuesday and merge.

src/components/TextInput.js
```diff
@@ -110,15 +120,25 @@ type State = {
...
   state = {
     focused: new Animated.Value(0),
     placeholder: '',
+    errorShown: new Animated.Value(this.props.hasError ? 1 : 0),
     value: this.props.value,
   };
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.hasError !== this.props.hasError) {
+      (nextProps.hasError ? this._animateFocus : this._animateBlur)(
```
kpsroka: use if statement here

src/components/TextInput.js
```diff
@@ -244,38 +265,51 @@ class TextInput extends React.Component<Props, State> {
...
+        { translateX: labelTranslateX },
+        { translateY: labelTranslateY },
       ],
     };
 
-    const bottomLineStyle = {
-      backgroundColor: bottomLineColor,
-      transform: [{ scaleX: this.state.focused }],
-      opacity: this.state.focused.interpolate({
+    const getBottomLineStyle = (
```
kpsroka: extract this function to a method

kpsroka: Done

kpsroka:
@satya164 I applied the changes that we discussed today, but I kept the `hasError` prop name.

PR #248, opened 2018-02-22
docs: make screenshots for Dialog

ferrannp:


satya164:
Added your commit to this https://github.com/callstack/react-native-paper/pull/249

Make further commits there and then we'll merge.

PR #250, opened 2018-02-23
chore(eslint): use eslint-config-callstack-io

thymikee:


src/components/Button.js
```diff
@@ -141,16 +141,12 @@ class Button extends React.Component<Props, State> {
...
-        if (buttonColor) {
-          backgroundColor = buttonColor;
-        } else {
-          if (primary) {
-            backgroundColor = colors.primary;
-          } else {
-            backgroundColor = theme.dark ? '#535354' : white;
-          }
-        }
+        backgroundColor = theme.dark ? '#535354' : white;
```
thymikee: This was autofixed by eslint

.eslintrc
```diff
@@ -1,148 +1,30 @@
...
-    "react/jsx-key": "error",
-    "react/jsx-no-bind": "off",
-    "react/jsx-no-duplicate-props": "error",
-    "react/jsx-no-literals": "off",
-    "react/jsx-no-undef": "error",
-    "react/jsx-pascal-case": "off",
-    "react/jsx-sort-props": "off",
-    "react/jsx-uses-react": "error",
-    "react/jsx-uses-vars": "error",
+    "import/first": "off",
```
satya164: Do we need to disable this rule? I'd prefer if we didn't.

docs/pages/src/Home.js
```diff
@@ -131,28 +131,35 @@ export default class Home extends React.Component<{}> {
...
               </a>{' '}
               to try it out
             </div>
             <a
               href="https://expo.io/@satya164/react-native-paper-example"
               target="_blank"
+              rel="noopener noreferrer"
             >
-              <img {...styles(qrcode)} src="images/expo-qr.png" />
+              <img {...styles(qrcode)} src="images/expo-qr.png" alt="" />
```
satya164: Can you add the alt tag as "Expo QR Code"?

.eslintrc
thymikee: Yea, because we would need to mix type imports with regular ones

PR #260, opened 2018-03-02
docs: add the missing CardCover props

gawrysiak:
Added the missing CardCover props documentation. It's now linking to react-native Image component.


satya164:
Thank you for the PR!

PR #262, opened 2018-03-03
fix: align card actions to the left

gawrysiak:
This is a suggestion to change the CardActions align from right to left. Additionally, it adds a "right" prop to easily go back to the original behavior, without changing any CSS.

### Motivation

It seems like the original [Material Design guidelines](https://material.io/guidelines/components/cards.html) aligns the actions to the left by default. This is as well the standard behavior for other libs, like [material-ui](http://www.material-ui.com/#/components/card).


src/components/Card/CardActions.js
```diff
@@ -29,29 +33,36 @@ type Props = {
...
-const styles = StyleSheet.create({
-  container: {
-    flexDirection: 'row',
-    alignItems: 'center',
-    justifyContent: 'flex-end',
-    padding: 4,
-  },
-});
+  render() {
+    const styles = StyleSheet.create({
```
satya164: Don't create stylesheets in render, move it to outside. Use inline styles for styles based on props.

src/components/Card/CardActions.js
```diff
@@ -8,6 +8,10 @@ type Props = {
    * Content of the `CardActions`.
    */
   children: React.Node,
+  /**
+   * Should actions be aligned to the right.
+   */
+  right?: boolean,
```
satya164: I don't think we want to provide a prop which doesn't align with material design guidelines. Let's align them to left by default and remove this prop.

PR #274, opened 2018-03-16
feat: Implement controlled ListDialog.

### LINKED ISSUE #247: Create different kind of Dialogs ready to use
Instead of having to do everything we do on the example, we need...

- ListDialog (single selection vs multiple selection)

Trancever:
Partially fixes #247 .
Add implementation for both single select and multi select ListDialog.


ferrannp:
I blocked it because we need to update docs library for static components.

PR #277, opened 2018-03-19
Indirectly import ProgressViewIOS / ProgressBarAndroid to fix web support

brentvatne:
### Motivation

ProgressViewIOS / ProgressBarAndroid aren't included in react-native-web, so if we import them in the ProgressBar component then we break web support.

### Test plan

I ran tests and the example app.

### DIFF
--- a/src/components/ProgressBar.js
+++ b/src/components/ProgressBar.js
@@ -1,13 +1,9 @@
 /* @flow */
 
 import * as React from 'react';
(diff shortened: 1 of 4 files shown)


satya164:
Merged in https://github.com/callstack/react-native-paper/commit/3259fbbf8e56e4cd1c87f329bef945fd5a103c05

PR #278, opened 2018-03-19
feat: add ProgressDialog component.

Trancever:
Partially solves #247 
Implement ProgressDialog component.


example/src/Dialogs/index.js
```diff
@@ -1,9 +1,7 @@
 /* @flow */
 
 export { default as DialogWithCustomColors } from './DialogWithCustomColors';
-export {
-  default as DialogWithLoadingIndicator,
-} from './DialogWithLoadingIndicator';
+export { default as ProgressDialog } from './ProgressDialog';
```
ferrannp: Let's not export it and expose it like `Dialog.Loading` and so on.

src/components/Dialog/ProgressDialog.js
```diff
@@ -0,0 +1,134 @@
...
+import Paragraph from '../Typography/Paragraph';
+import Dialog from './Dialog';
+import DialogTitle from './DialogTitle';
+import DialogContent from './DialogContent';
+
+type Props = {
+  /**
+   * Determines whether clicking outside the dialog dismiss it.
+   */
+  dismissable?: boolean,
```
ferrannp: We don't need to repeat the `Dialogs` props here. You can export the Props type from `Dialog` and then just do `type Props = DialogProps & { ... }`.

src/components/Dialog/ProgressDialog.js
```diff
@@ -0,0 +1,134 @@
...
+ *     return (
+ *       <View>
+ *         <Button onPress={this._showDialog}>Show Dialog</Button>
+ *         <ProgressDialog
+ *           color={Colors.indigo500}
+ *           onDismiss={this._hideDialog}
+ *           visible={visible}
+ *           size={isIOS ? 'large' : 48}
+ *           text="Loading....."
+ *           title="Progress Dialog"
```
ferrannp: Can we just keep the relevant part here? Like the `ProgressDialog` component? I would remove all that logic, state, buttons and so on.

src/index.js
```diff
@@ -31,6 +31,7 @@ export { default as FAB } from './components/FAB';
 export { default as Modal } from './components/Modal';
 export { default as Paper } from './components/Paper';
 export { default as ProgressBar } from './components/ProgressBar/ProgressBar';
+export { default as ProgressDialog } from './components/Dialog/ProgressDialog';
```
ferrannp: We would remove it from here too with `Dialog.Loading`.

example/src/Dialogs/index.js
Trancever: It's just example app. I use Dialog.Progress inside ProgressDialog which is example component.

Trancever:
@satya164 Added // $FlowFixMe for now inside example when we use Dialog.Progress.

Maybe we can do something like that to make flow working as we want :D
```
const Comp = withTheme(Dialog);
Comp.Progress = ProgressDialog;
export default Comp;
```

What do you think?

example/src/Dialogs/index.js
ferrannp: A bit confusing, can't you just use `Dialog.Progress` instead? I do not see anything special in wrapping it inside into another component.

ferrannp:
I blocked it because we need to update docs library for static components.

Issue #290, opened 2018-03-28
`withTheme` throws uncaught errors when called with arrow function SFCs


you-fail-me:
### Current behaviour

I'm getting uncaught errors coming from `withTheme`, more exactly from `isClassComponent` helper. If I wrap an SFC, written with an arrow function. The error is that it can't read property `render` of undefined. The reason is that arrow functions don't have a prototype (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions ).

### Expected behaviour

Expected behavior is the absence of uncaught errors when working with arrow functions

### Code sample

```
const component = () => <div />;
const failingComponent = withTheme(component);
```

### What have you tried

For now my workaround is to either use class or monkey-patch component's `prototype` property so `withTheme` doesn't fail.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | 11.2
| react-native          | 0.54.0
| react-native-paper    | 1.1.1
| node                  | 9.4.0
| npm or yarn           | yarn 1.3.2

* also using `haul` and typescript with `awesome-typescript-loader` btw

I don't mind making a PR, just confirm you recognize the issue and look forward to merging that asap and I'll make it

satya164:
Will be awesome if you can send a PR :D

PR #292, opened 2018-03-29
added nested property check to withTheme

you-fail-me:
### Motivation

Fixes https://github.com/callstack/react-native-paper/issues/290

### Test plan

Wrapping an arrow function SFC with `withTheme` used to throw `Cannot read property "render" of undefined" on a project with haul and awesome-typescript-loader. With additional nested property check it no longer throws


satya164:
Thank you for the pull request!

satya164:
Can you fix eslint? The CI is failing. Run `yarn lint --fix` to fix it automatically.

Issue #291, opened 2018-03-28
Icon component:  Invalid props.style key `color` supplied to `View`


pleportz:
### Current behaviour

In certain cases, the `View` component used in `src/components/Icon` is responsible for the following warning : `Failed prop type: Invalid props.style key `color` supplied to `View``
Example : when using the `FAB` component with a custom Icon like below

```
import { Icon } from 'myComponents';
 
<FAB
        icon={<Icon name="chevron-left" size={13} color="#ffffff" />}
        onPress={() => null}
/>
```

### Expected behaviour

There should not be the following warning : `Warning: Failed prop type: Invalid props.style key `color` supplied to `View``

### What I tried

Would it be ok to make a contribution ? I tried to push a branch with a fix suggestion but I got the following message : `ERROR: Permission to callstack/react-native-paper.git denied`

Trancever:
@pleportz Your probably tried to push a branch to callstack/react-native-paper directly. Instead you should first fork this repository, push changes to your forked repo and then create pull request to this repository. Here is the link to paper's contributing guide - https://github.com/callstack/react-native-paper/blob/master/CONTRIBUTING.md
Follow the instructions and you will be able to make contribution.

pleportz:
@Trancever Thank you I'm going to try that :)

PR #294, opened 2018-04-01
fix: View in Icon component should not receive color as props.style key

pleportz:
### Motivation

This fix suggestion is linked to [issue 191](https://github.com/callstack/react-native-paper/issues/291).
It aims at suppressing `Warning: Failed prop type: Invalid props.style key 'color' supplied to 'View'` generated by the Icon component in certain cases.

### Example case

When using the `FAB` component with a custom Icon like below

```
import { Icon } from 'myComponents';
 
<FAB
        icon={<Icon name="chevron-left" size={13} color="#ffffff" />}
        onPress={() => null}
/>
```

In my case I am using the [react-native-vector-icons](https://oblador.github.io/react-native-vector-icons/) lib to generate icons.

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -38,6 +38,10 @@ const Icon = ({ name, ...props }: Props) => {
       />
     );
   }
+  const style = { ...props.style };
+  if (style) {
+    delete style.color;
+  }
   return (
     <View
       {...props}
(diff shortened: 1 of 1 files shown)


satya164:
Hey! I sent a PR #299 which should fix this issue. Can you check it and lemme know?

ferrannp:
Thanks for your PR but we'll use https://github.com/callstack/react-native-paper/pull/299 as problem comes from other components.

Issue #270, opened 2018-03-06
Use pure component in BottomNavigation example


ferrannp:

Rafaell416:
Hi, @ferrannp  could you please give a little bit more information about the issue? 

do you mean doing this ? 

`
import { PureComponent } from 'react';

export default class Example extends PureComponent {
  render() {
    return <SomeComponent someProp={props.someProp}/>
  }
}
})

`

ferrannp:
Hello @Rafaell416, not really, what we meant here is to follow the BottomNavigation docs https://callstack.github.io/react-native-paper/bottom-navigation.html.

Specifically this part:

```js
_renderScene = BottomNavigation.SceneMap({
  music: MusicRoute,
  albums: AlbumsRoute,
  recents: RecentsRoute,
});
```

Which is explained here https://callstack.github.io/react-native-paper/bottom-navigation.html#renderScene.

Let me know if you need something more! :)

PR #295, opened 2018-04-01
refactor: add components for bottom navigation example

kaynanguarany:
attempt to solve #270 .
Add class components to bottom navigation routes.


kaynanguarany:
@ferrannp I created one component per route, it should be done this way?

satya164:
Already done.

PR #301, opened 2018-04-05
feat: List component

gawrysiak:
List component proposal. Following the [MD guidelines](https://material.io/guidelines/components/lists.html), it contains some common elements:
- text
- secondaryText
- icon
- avatar


src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
...
+import Subheading from './Typography/Subheading';
+import withTheme from '../core/withTheme';
+import type { Theme } from '../types';
+import type { IconSource } from './Icon';
+
+type Props = {
+  /**
+   * The primary text of the item.
+   */
+  text: string | React.Node,
```
satya164: Probably `title` is a better name,

I think `React.Node` also includes `string`

src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Secondary (optional) text for the item.
+   */
+  secondaryText?: string | React.Node,
```
satya164: Probably `description` is a better name.

src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
...
+   */
+  avatar?: React.Node,
+  /**
+   * Function to execute on press.
+   */
+  onPress?: Function,
+  /**
+   * Custom color for the drawer text and icon.
+   */
+  color?: string,
```
satya164: This prop doesn't seem to be used

src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
...
+}: Props) => {
+  const { colors } = theme;
+  const secondaryTextColor = color(colors.text)
+    .alpha(0.54)
+    .rgb()
+    .string();
+  return (
+    <TouchableRipple {...props}>
+      <View
+        style={[
```
satya164: We should merge the passed style and add these styles directly to `TouchableRipple`. Then the user can override them easily.

src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
...
+          styles.wrapper,
+          {
+            backgroundColor: 'transparent',
+            paddingHorizontal: 16,
+          },
+        ]}
+      >
+        {(avatar || icon) && (
```
satya164: We normally use `condition ? <SomeThing /> : null` instead of `condition && <SomeThing />` because it could sometimes produce unexpected results.

src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
...
+              }}
+            >
+              {secondaryText}
+            </Text>
+          )}
+        </View>
+        {avatar && icon ? (
+          <View
+            style={{
+              maxWidth: 56,
```
satya164: Probably you mean `width: 56` instead of `maxWidth: 56`?
I think a `margin: 16` in the `Icon` should be sufficient and this wrapper view can be removed

src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
...
+        style={[
+          styles.wrapper,
+          {
+            backgroundColor: 'transparent',
+            paddingHorizontal: 16,
+          },
+        ]}
+      >
+        {(avatar || icon) && (
+          <View style={{ width: 56 }}>
```
satya164: `margin: 16` on the icon should be sufficient. also let's change it to ternary

src/components/ListItem.js
```diff
@@ -0,0 +1,127 @@
+/* @flow */
+
+import color from 'color';
+import * as React from 'react';
+import { View, Text, StyleSheet } from 'react-native';
```
satya164: `import { Text } from './Typography/Text`

gawrysiak:
Hi @satya164 ! Thank you for the good feedback. And sorry for outdating all the comments, I've renamed some files (ListSection) to keep it clean.

I was able to cover most of them, with some slight differences:
- https://github.com/callstack/react-native-paper/pull/301/files#diff-5095fb1f06ad95700e956bb17e736afcR72 - I wasn't really able to move the styles to TouchableRipple since it requires a single child and a non-styled View inside breaks the flex direction. It works similar with the current [DrawerItem](https://github.com/callstack/react-native-paper/blob/master/src/components/DrawerItem.js#L81) as this was the base I used. Extra help would be appreciated,
- https://github.com/callstack/react-native-paper/pull/301/files#diff-5095fb1f06ad95700e956bb17e736afcR75 this one I had to keep just marginRight, since the Item already has some extra margin for text-only items.

I'm trying the expo app now and the end result appears to be the same.

Also, I wasn't able to reproduce the screenshots as it seems like they're all 640x480. Mine are sometimes bigger with no frame around it. Please let me know if you have a procedure for them.

satya164:
Merged in 7b2948f570faa2a6c8dbb32a1f7ba6a08d8531f4

PR #302, opened 2018-04-06
feat: prop to specify capitalization on button text

JohnKanev:
A simple prop that can specify if the text in a button should be capitalized or not.

The prop will be:
upperCase?: boolean // default value true

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -44,13 +44,13 @@ type Props = {
    */
   icon?: IconSource,
   /**
+   * Whether to capitalize the text.
+   */
+  uppercase?: boolean,
+  /**
    * Custom text color for flat button, or background color for raised button.
    */
   color?: string,
-  /**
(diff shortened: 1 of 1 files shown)


ferrannp:
@JohnKanev can you fix the build so we can merge this?

Hunsu:
Why the default value is true? This can break existing users. And do we really need this? I think we should keep the components simple but allow people to extends them to fulfill their use cases. For this one you just create your own components that passes the capitalized text to this buttons.

Trancever:
@Hunsu All the buttons in material design guidelines are uppercase, that's why the default value is true.

JohnKanev:
Sorry guys I have been on a long vacation.

The truth is that I made this pull request is because most of my clients ask for everything to not be upcase because of some psychology studies.

I will create a new pull request later on no problem.

Issue #43, opened 2016-10-26
Portal


satya164:
## Description

Elements like FAB and Snackbar render on a different plane above the content in z-axis. Many other elements such as bottom sheets, modals etc. need to be rendered above content too. It'll be useful to have a single Portal component to handle these cases.

To determine which elements exist on which plane, we need to refer to their elevation - https://material.google.com/material-design/elevation-shadows.html#elevation-shadows-elevation-android

## Use cases

- Elements with same elevation should be grouped in the same layer along the z-axis (e.g. `FAB`, `Snackbar`)
- Grouped elements should move together, e.g. when a `Snackbar` slides in from button, the `FAB` should move up

### Props

The `Portal` component will accept two props, `elevation` and `layered`.

The `elevation` prop determines the elevation of the layer in the z-axis. There will be predefined elevations for internal use according to material design guidelines:

```js
const Elevations = {
  DIALOG: 24, // Dialogs and Pickers
  DRAWER: 16, // Drawer, Bottom sheet
  FAB_RAISED: 12, // Floating action button (raised)
  CARD_RAISED: 8, // Card (raised), Button (raised), Bottom navigation, Menu
  FAB_RESTING: 6, // Floating action button, Snackbar
  APPBAR: 4, // Floating action button, Snackbar
  CARD_RESTING: 2, // Card, Button, Search bar
}
```

This is not the full list of elevations, rather the ones we need. Naming is up-to-debate. This will be internal for now, so we can change them later.

The `layered` prop is a `boolean` which determines whether the item is grouped with other layered items in a single layer. Grouped items are rendered under the same `View` and positioned using flexbox, so items will appear one below another. When the `layered` prop is `false` or not passed, the items will be positioned absolutely instead. Non-layered items will render below layered elements.

### Usage

```js
<Portal elevation={Elevations.FAB_RESTING} layered>
  <View style={styles.snackbar}>
    Hey! I'm a `Snackbar` :)
  </View>
</Portal>
```

### Animations

The main motivation behind this is animations, i.e. when you show a new `Snackbar`, it should slide in from the bottom and move everything else up (`FAB`). The easiest way seems to be to handle the animation inside `Portal`. This is how it'll work:

1. On initial render, render everything normally.
2. When a new element is rendered later, render it offscreen and measure the `height` (`{ position: 'absolute', left: 0, right: 0, bottom: -9999 }`).
3. Now render the element normally, but also apply a `translateY` of `height` to all the elements. This should have no visual difference.
[...]

PR #305, opened 2018-04-09
feat: Snackbar component

### LINKED ISSUE #16: Snackbars
[MD Docs Link](https://material.google.com/components/snackbars-toasts.html#)

Trancever:
Closes #16


vladikoff:
I am excited for this 😄

src/styles/colors.js
```diff
@@ -271,6 +271,7 @@ export const grey500 = '#9e9e9e';
 export const grey600 = '#757575';
 export const grey700 = '#616161';
 export const grey800 = '#424242';
+export const grey850 = '#323232';
```
satya164: Since our `colors` file is just material design colors, maybe move this to the component? We do that for other colors.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * const styles = StyleSheet.create({
+ *   container: {
+ *     flex: 1,
+ *     justifyContent: 'space-between',
+ *   },
+ * });
+ * ```
+ */
+class SnackBar extends React.Component<Props, State> {
```
satya164: Let's name it `Snackbar` since it's a single word, like `Toolbar`

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+
+type Props = {
+  /**
+   * Whether snackbar is currently visible.
+   */
+  visible: boolean,
+  /**
+   * Text that will be displayed inside SnackBar.
+   */
+  content: string,
```
satya164: Maybe use the `children` prop for this?

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Object that determines button text and callback for button press. It should contains following properties:
+   * - `text` - Content of the action button
+   * - `onPress` - Callback that is called when action button is pressed, user needs to update the `visible` prop.
+   * - `color` - Color of the action button
+   */
+  action?: {
```
satya164: Maybe we want to support multiple actions?

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+    text: string,
+    onPress: () => mixed,
+    color?: string,
```
satya164: Probably unnecessary since people can already override the accent color with the theme prop.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
+// @flow
```
satya164: nit: use `/* @flow */`

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+   * - `onPress` - Callback that is called when action button is pressed, user needs to update the `visible` prop.
+   * - `color` - Color of the action button
+   */
+  action?: {
+    text: string,
+    onPress: () => mixed,
+    color?: string,
+  },
+  /**
+   * Time after which onDismiss callback will be called.
```
satya164: Let's say, `The duration for which the Snackbar is shown`.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+   * It can take following values:
+   * - `'indefinite'` - SnackBar will hide only when user tap it.
+   * - `'short'` - SnackBar will hide after 2 seconds.
+   * - `'long'` - SnackBar will hide after 3.5 seconds.
+   */
+  duration?: 'indefinite' | 'short' | 'long',
```
satya164: Not too keen on accepting string values for the duration. I'd rather accept a `number` which represents duration in milliseconds.

Is it possible to have `Snackbar.DURATION_LONG`? Let's try to figure it out how we can do it.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Callback called when Snackbar is dismissed, user needs to update the `visible` prop.
+   */
+  onDismiss: () => any,
```
satya164: use `mixed`

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+  contentColor?: string,
```
satya164: Let's remove this prop. People shouldn't usually override it, if they really need to, they can use a `Text` for the content.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+  backgroundColor?: string,
```
satya164: Users should be able to pass a `style` prop to override the background color and other stuff.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+ *   },
+ * });
+ * ```
+ */
+class SnackBar extends React.Component<Props, State> {
+  static defaultProps = {
+    duration: DURATION_LONG,
+  };
+
+  constructor(props) {
```
satya164: Unnecessary contsructor

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    alignItems: 'center',
+  },
+  content: {
+    marginLeft: 24,
+    marginVertical: 14,
+    flexWrap: 'wrap',
+    flex: 1,
+    fontSize: 14,
```
satya164: Shouldn't need this

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+        toValue: 0,
+        duration: SNACKBAR_ANIMATION_DURATION,
+        useNativeDriver: true,
+      }),
+      Animated.timing(this.state.yPosition, {
+        toValue: 48,
+        duration: SNACKBAR_ANIMATION_DURATION,
+        useNativeDriver: true,
+      }),
+    ]).start(this.props.onDismiss);
```
satya164: Since we're calling `hide` on `componentDidUpdate`, we shouldn't call `this.props.onDismiss` again.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+      Animated.timing(this.state.yPosition, {
+        toValue: 0,
+        duration: SNACKBAR_ANIMATION_DURATION,
+        useNativeDriver: true,
+      }),
+    ]).start(() => {
+      const { duration } = this.props;
+      if (duration !== 'indefinite') {
+        this.hideTimeout = setTimeout(
+          this.hide,
```
satya164: Let's call `this.props.onDismiss` here instead of `hide`. `componentDidUpdate` will trigger the animation.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      opacity: new Animated.Value(0),
+      yPosition: new Animated.Value(48),
+    };
+  }
+
+  hideTimeout: number;
```
satya164: Use `_` prefix for private properties and methods.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+              styles.container,
+              {
+                opacity: this.state.opacity.interpolate({
+                  inputRange: [0, 0.8, 1],
+                  outputRange: [0, 0.2, 1],
+                }),
+              },
+            ]}
+          >
+            <Animated.Text
```
satya164: This shouldn't need `Animated`

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+            {action ? (
+              <View
+                style={{
+                  marginRight: buttonRightMargin,
+                }}
+              >
+                <TouchableWithoutFeedback
+                  onPress={() => {
+                    this.hide();
+                    action.onPress();
```
satya164: Maybe call `onPress` first, then `hide`

src/components/SnackBar.js
Trancever: MD docs say that only one action is allowed, so that's why i did it like that. Do we really want to support more than one?

ferrannp: Yeah 0-1 actions is ok.

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+                  }}
+                >
+                  <View>
+                    <Text style={{ color: action.color || colors.accent }}>
+                      {action.text}
```
ferrannp: This needs to be uppercase. You can use our own `Button` with `compact` (I think).

src/components/SnackBar.js
```diff
@@ -0,0 +1,270 @@
...
+      theme: { fonts, colors },
+      contentColor,
+      backgroundColor,
+    } = this.props;
+
+    const buttonRightMargin = action ? 24 : 0;
+    const contentRightMargin = action ? 0 : 24;
+
+    return (
+      <Animated.View
```
ferrannp: Can you wrap everything with `ThemedPortal`

src/components/SnackBar.js
Trancever: I used our button and forgot to add toUpperCase when I changed it. I decided not to use button from paper cause it has padding and margin by default and styling was not so clear.

src/components/SnackBar.js
```diff
@@ -0,0 +1,269 @@
...
+    ]).start(() => {
+      const { duration } = this.props;
+      if (duration !== Snackbar.DURATION_INDEFINITE) {
+        this._hideTimeout = setTimeout(this.props.onDismiss, duration);
+      }
+    });
+  };
+
+  _hide = () => {
+    if (this._hideTimeout) {
```
satya164: Let's clear this on unmount too

satya164:
@vladikoff did you follow the getting started guide? you need a `<Provider />` at root of your app.

vladikoff:
> did you follow the getting started guide? you need a <Provider /> at root of your app.

Yeah I have the rest of paper working. Just trying to add a snackbar from this Pull Request 

```js
  render () {
    return (
      <Provider store={store}>
        <App/>
      </Provider>
    )
  }
```

Removing `<ThemedPortal>` from Snackbar.js helps remove the red crash but the Snackbar is not visible at the moment.

vladikoff:
@satya164 oh I have `import { Provider } from 'react-redux';` not `Provider as PaperProvider` should I use the PaperProvider?

vladikoff:
Nvm, let me revisit https://callstack.github.io/react-native-paper/getting-started.html again , sorry

vladikoff:
Cool it's working here! I just need to move my fab up now :) 

However, note that if you set the initial state to:

```js
  constructor(props) {
    super(props);
    this.props = props;
    this.state = {
      visible: true // <---------- TRUE here
    }
```

then it will not show the Snackbar when the view renders for the first time, not sure if bug or something with my setup. Thanks again!

Trancever:
@vladikoff Cool that it's working! We are planning to add support for automatically moving FAB whenever Snackbar shows so stay tuned 😄

vladikoff:
Sorry for a stupid question but what controls the stacking order (like css z-index) for this?

For example if I have a drawer, how do I make sure the Snackbar doesn't overlay it? 

Example:

satya164:
> what controls the stacking order (like css z-index) for this

It's controlled by the `Provider` component. If you place the drawer inside `Provider`, the drawer will be below the Snack and vice-versa.

vladikoff:
> It's controlled by the Provider component. If you place the drawer inside Provider, the drawer will be below the Snack and vice-versa.

Ah right, that helps! I tried doing the animation of the FAB and the Snackbar together but it is very choppy. Probably because it is not using the `Animated.parallel` call

Trancever:
@vladikoff There is a description in a Portal issue #43 which says how those components (Snackbar, FAB, etc.) should  behave together and how we would like to implement it. You can check it out.

PR #308, opened 2018-04-11
feat: save memory usage for inactive routes in bottom navigation

satya164:
When routes are inactive, position them offscreen and use `removeClippedSubviews` to optimize memory usage. The basic approach is this:

- Use animated values to specify which routes are offscreen, so we can imperatively set the value and avoid unnecessary re-renders
- Before animation starts, clear offset of previous and new route, so we see the cross-fade animation
- When animation ends, set offset of inactive screens to a large value to position it offscreen

cc @brentvatne

### DIFF
--- a/src/components/BottomNavigation.js
+++ b/src/components/BottomNavigation.js
@@ -169,6 +169,12 @@ type State = {
    */
   shifts: Animated.Value[],
   /**
+   * The top offset for each tab item to position it offscreen.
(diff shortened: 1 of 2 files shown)


brentvatne:
lgtm! better to use this than setState for sure :)

jukben:
Love the approach! 🙌

PR #312, opened 2018-04-16
fix: support static properties in withTheme HOC

satya164:
Currently, there is no way to preserve types of static properties of React components when they are wrapped in the `withTheme` HOC. This PR adds a HOC named `withStatics` which helps to add static properties to a component

API:

```js
export default withStatics(withTheme(Toast), {
  DURATION_LONG: 3500
});
```

The document generator won't handle such static properties. So we'll need to write those manually.

### DIFF
--- a/example/src/RootNavigator.js
+++ b/example/src/RootNavigator.js
@@ -23,6 +23,7 @@ const routes = Object.keys(examples)
           <ToolbarContent title={(Comp: any).title} />
         </Toolbar>
       ),
+      /* $FlowFixMe */
       ...(typeof Comp.navigationOptions === 'function'
         ? Comp.navigationOptions(props)
         : Comp.navigationOptions),
--- a/src/core/withTheme.js
+++ b/src/core/withTheme.js
@@ -31,14 +31,13 @@ const REACT_METHODS = [
   'updateComponent',
 ];
 
-const isClassComponent = (Component: Function) =>
-  !!(Component && Component.prototype && Component.prototype.render);
+const isClassComponent = (Component: any) =>
+  Boolean(Component.prototype && Component.prototype.isReactComponent);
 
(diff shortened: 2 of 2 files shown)


satya164:
Alternative approach:

When wrapping the component with `withTheme`, we can annotate the component to have statics:

```js
import withTheme, { type ComponentType } from '../withTheme';

export default withTheme(Toast, ComponentType<Props<*>, {
  DURATION_LONG: number
}>);
```

The actual values for static properties can be defined normally inside the class.

satya164:
Alternative approach:

Add another HOC named `withStatics` instead of packing the functionality into `withTheme`. Then do something like:

```js
export default withStatics(withTheme(Toast), {
  DURATION_LONG: 3500
});
```

Though I haven't been able to make it work yet.

src/core/withStatics.js
```diff
@@ -0,0 +1,11 @@
+/* @flow */
+
+export default function withStatics<C, S: {}>(Comp: C, statics: S): C & S {
```
Andarist: there is quite popular package that solves this problem - https://github.com/mridgway/hoist-non-react-statics

its not a HOC though, it's just a thing that u use in your HOC implementation

dunno though how it plays with types

satya164: We already use that package in our HOC. I don't think it works with Flow :(

Andarist: That's a bummer then, have u tried flow-typed definitions for it?

satya164: The type definitions throw an error when I try to use it!

satya164:
Yay! We don't need a new API for that. Pushed a fix which supports static properties without extra work: https://twitter.com/tadeuzagallo/status/985891764060827649

PR #313, opened 2018-04-16
uncheckedColor prop for Checkbox and RadioButton 

NazarYablonskiy:
#### What existing problem does the pull request solve?
It was impossible to set the color of Checkbox/RadioButton in the unchecked state.

#### Can you solve the issue with a different approach?
No, most likely only this way.

#### Test plan
List the steps with which we can test this change. Provide screenshots if this changes anything visual.
just pass uncheckedColor="white" to the Checkbox/RadioButton component.

result:
img_20180415_114757

### DIFF
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -108,9 +108,9 @@ class Checkbox extends React.Component<Props, State> {
   render() {
     const { checked, disabled, onPress, theme, ...rest } = this.props;
     const checkedColor = this.props.color || theme.colors.accent;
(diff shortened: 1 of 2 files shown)


satya164:
Does `<Checkbox theme={{ colors: { text: white }}} />` solve your use case?

NazarYablonskiy:
@satya164 no, because of this `.alpha(theme.dark ? 0.7 : 0.54)`

NazarYablonskiy:
thanks!

PR #315, opened 2018-04-16
Update SearchBar.js

shcheuk:
icon props is fixed to be 'search' only if onIconPress is not defined. It should change according to the string  provided to icon props

### DIFF
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -115,7 +115,7 @@ class SearchBar extends React.Component<Props> {
             name={icon || 'search'}
           />
         ) : (
-          <Icon style={styles.icon} name="search" size={24} color={iconColor} />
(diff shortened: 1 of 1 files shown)


ferrannp:
@shcheuk it looks good. Can you fix lint?

shcheuk:
image
These?

gawrysiak:
@shcheuk check out https://circleci.com/gh/callstack/react-native-paper/1536?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link for details

shcheuk:
￼￼￼￼image
line 12,13,14 seems not related to this fix. They seems all related to using .ts syntax in .js

shcheuk:
oh thx, will do tmr.

satya164:
@shcheuk run `yarn lint --fix` in a terminal and it should fix the errors. the above errors are from typescript which is not relevant since we don't use typescript.

PR #316, opened 2018-04-16
WIP: Feat/more flexible text input

NazarYablonskiy:
#### What existing problem does the pull request solve?
It was impossible to change style(padding in my case) of the NativeTextInput.

#### Can you solve the issue with a different approach?
No.

#### Test plan
List the steps with which we can test this change. Provide screenshots if this changes anything visual.
just pass textInputStyle={{paddingTop: 0}} to the TextInput component.

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -56,6 +56,10 @@ type Props = {
   value?: string,
   style?: any,
   /**
+   * Style of native text input.
+   */
+  textInputStyle?: any,
+  /**
    * @optional
    */
   theme: Theme,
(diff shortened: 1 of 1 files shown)


satya164:
I'd prefer not to add props to customize internals of a component since it prevents us from changing them later as it'll break people's code. Can you post a screenshot of what you want to achieve and we can figure out a way to do that?

NazarYablonskiy:
@satya164 I'd like to reduce `paddingTop`. Also, I'd like to manually control style of the underline (the second text input)

screen shot 2018-04-16 at 14 00 40

NazarYablonskiy:
@satya164 awesome! eta?

satya164:
No ETA, but we'll try to get it merged ASAP. Meanwhile, you could use https://github.com/ds300/patch-package to apply your changes.

PR #323, opened 2018-04-17
fix: "hijacked" scroll on iOS Safari

thymikee:
### Motivation

On iOS Safari there's lack of momentum scroll, which feels weird

### Test plan

website deployment and device testing


thymikee:
Confirmed it works well now 👌

PR #326, opened 2018-04-18
update SearchBar.js according to TextInput.js

shcheuk:
added focus(),clear()blur(),isFocused(),setNativeProps() methods to SearchBar.js, exactly the same way as the TextInput.js....
although I prefer adding textInputRef props to expose the TextInput ref to the SearchBar..
onChangeText no longer a required props for the clear button to work.

### Motivation
SearchBar is kind of a  textInput groomed with one additional icon and clear button. Basic  Functionality like focus(), blur() is better exposed to user. a very common use cases is when user click on the search icon, the textinput get focused. something like onIconPress=> this.ref.searchbar.focus().
this issue can be solved with a different approach, that is expose the underlying nativeTextInput ref to SearchBar. But the TextInput already doing it this way, so ... consistent.

### DIFF
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -71,8 +71,30 @@ type Props = {
  */
 class SearchBar extends React.Component<Props> {
   _handleClearPress = () => {
-    this.props.onChangeText('');
+    this.clear();
   };
+  _root: any;
+  _setRef: any = (c: Object) => {
+    this._root = c;
+  };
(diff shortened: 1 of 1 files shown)


shcheuk:
Not much test has been done, but it works is my use case - click the icons and focus()/clear() called to the underlying textinput
demo

satya164:
Thank you for the PR!

PR #327, opened 2018-04-18
feat: make vector-icons optional. fixes #303

### LINKED ISSUE #303: Remove peer dependency on vector icons
Since react navigation is using his library now under the hood, npm started throwing a warning at me that vector icons is missing as a peer dependency. Is there any way you could make the dependency on vector icons optional. 

In my app I render all my icons as SVGs directly. Vector icons isn’t bad but maybe it should be the users choice wether that’s how they want to render their icons. 

I think expo projects already include vector icons but won’t have it as an explicit dependency either so they will get the warning as well.

satya164:


src/components/Icon.js
```diff
@@ -15,7 +23,7 @@ export type Props = {
 
 const Icon = ({ name, ...props }: Props) => {
   if (typeof name === 'string') {
-    return <MaterialIcons {...props} name={name} />;
+    return MaterialIcons ? <MaterialIcons {...props} name={name} /> : null;
```
thymikee: What about adding a warning? Something like:
```js
const error = Error(); // for stack trace
error.message = 'Prop "name" was passed to <Icon /> component, but `react-native-vector-icons` is not present in your dependencies. To remove this warning don't pass "name" to <Icon /> or install `react-native-vector-icons`.
console.warn(error)
```

thymikee:
Awesome DX work @satya164!

PR #329, opened 2018-04-19
feat: ListControl component

gawrysiak:
Adds a ListControl component, implementing [Material Design Guidelines](https://material.io/guidelines/components/lists-controls.html).

### Motivation

Although this could maybe work as a replacement for the current ListItem, I believe the Material Design Guidelines for [lists](https://material.io/guidelines/components/lists.html) and [lists: controls](https://material.io/guidelines/components/lists-controls.html) were split as they cover different use cases and provides a different set of features.

Material Design Guidelines differences:
- list controls are used to indicate state or provide actions, while list items simply display rows/items
- list controls require an action, while it's not required for list items
- list controls don't usually support text only rows, while list items do
- list controls provide extra features such as expand/collapse

Paper differences:
- list control provides a more flexible API instead of the simple avatar/icon system introduced by ListItem
- list controls can have children
- the list control contents are always using the extra left padding since they have primary actions or they're used as subitems

### Notes

I haven't provided any screenshots this time, since my ListItem screens were garbage anyway (thanks for fixing it @satya164).

ListControl could be further extended with more features, e.g. reorder, leave-behinds.

In case you don't like the component idea, would it be okay to change the ListItem API and move towards primary & secondary action instead of avatar/icon to support more use cases?

### Test plan

There's a snapshot test available. Example added to the Expo app as well.

### Preview

fullsizeoutput_2a6


satya164:
So on a superficial level, there are lots of similarities between this and `ListItem`. In fact, most of the things here could be acheived by using `ListItem`.

Here is what I suggest:

- Provide `onAvatarPress` and `onIconPress` props for `ListItem`: This should solve most of the use cases.
- Expand/Collapse is a special case: This is the only case which accepts `children`, changes color and icon depending on active/inactive. So it should be a separate component, `ListAccordion` or something.
- Children of `ListAccordion`: This is the only case where the children have extra space on left. It can use `Children.map` over it's children to pass a `marginLeft` style.

gawrysiak:
I really like the `ListAccordion` idea, I'm gonna re-do that definitely.

Regarding the `ListItem` props, the names feel quite unfortunate when covering more use cases, but like you said, the other solution duplicates the components.

> Provide onAvatarPress and onIconPress props for ListItem: This should solve most of the use cases.

Do we need the `onAvatarPress` and `onIconPress` props? It should be already possible to do it like `<ListItem avatar={<Checkbox onPress={} />} />`

> We'd also want ListItem, Card or any other component to have the ability to re-order, swipe etc. So I think it'll be better to provide more components like Swipeable etc. which can wrap any component to provide this feature.

This would be great :+1:

gawrysiak:
@satya164 could you please check if the [ListAccordion PR](https://github.com/callstack/react-native-paper/pull/331) is any good? As for this pull request I guess it could be closed, with maybe some extra issue in case of ListItem changes.

PR #332, opened 2018-04-20
feat: ContentStyle prop for Button

alexmbp:
### Motivation

Sometimes in project we need more customisation for component, that probably will make it not following guidelines. Full width buttons / custom height buttons - are good samples for it. So I've added extra prop to allow adding extra styles to buttons content.
Also I was thinking about height / fullWidth / fullSize prop. But think it will be overengineering.

### Test plan

I've added usage  example and also screenshot is here
[Examples Screenshot](https://user-images.githubusercontent.com/17987276/39042739-4226d1bc-4494-11e8-8841-696979ab0a25.jpeg)

### DIFF
--- a/example/src/ButtonExample.js
+++ b/example/src/ButtonExample.js
@@ -103,6 +103,20 @@ class ButtonExample extends React.Component<Props, State> {
             Custom component
           </Button>
         </View>
+        <View style={styles.row}>
+          <Button
+            raised
+            primary
+            style={{ flex: 1, height: 50 }}
+            contentStyle={{
+              flex: 1,
+              alignItems: 'center',
+            }}
+            onPress={() => {}}
+          >
+            Custom styles
+          </Button>
+        </View>
       </View>
     );
   }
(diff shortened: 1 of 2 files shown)


ferrannp:
What do you think @satya164 ?

satya164:
Thanks for the PR.

We don't want to expose styles for internal implementation details. Is your issue solved by just adding `flex: 1` to `styles.content` by default?

alexmbp:
Hello, @satya164 ! Well. It doesn't I still need to fix
justifyContent: 'center' <- to move spinner and label together.

Also, setting flex: 1 by default will break 'default' behaviour.

@ferrannp Hello! We have discussed TS/Flow a bit at React Native Camp.

alexmbp:
Hey! Any suggestions here? Probably prop for height?

satya164:
If you can figure out a way to make the touchable fill it's parent, we'd accept a PR for that.

ferrannp:
Closing in favour of https://github.com/callstack/react-native-paper/pull/336.

PR #333, opened 2018-04-20
fix: hide checkbox on iOS instead of removing it

ferrannp:
The way is done now, it removes the view completely so the space is not filled anymore (which is weird from check<->uncheck).

### DIFF
--- a/src/components/Checkbox.ios.js
+++ b/src/components/Checkbox.ios.js
@@ -64,16 +64,14 @@ class Checkbox extends React.Component<Props> {
         onPress={disabled ? undefined : onPress}
         style={styles.container}
       >
-        <View style={styles.iconContainer}>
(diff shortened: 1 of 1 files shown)


src/components/Checkbox.ios.js
```diff
@@ -64,16 +64,14 @@ class Checkbox extends React.Component<Props> {
...
-          {checked && (
-            <Icon
-              allowFontScaling={false}
-              name={checked && 'done'}
-              size={24}
-              color={checkedColor}
-              style={styles.icon}
-            />
-          )}
+        <View style={[styles.iconContainer, { opacity: checked ? 1 : 0 }]}>
```
satya164: Can you remove `styles.iconContainer`? Seems it's not needed anymore

PR #337, opened 2018-04-26
Docs: Add Babel example with env setting from CRNA

peterpme:
I generated a new project with CRNA. My babelrc file had an `env` key so I thought this clarification could help beginners.

Thanks!


satya164:
Thank you for the PR!

Issue #335, opened 2018-04-21
Not possible to change ToolbarAction icon color


vladikoff:
### Current behaviour

Currently it's not possible to match the color of text with the ToolbarAction icon color.

### Expected behaviour

Match icon with text color

### Code sample

the code from: https://github.com/callstack/react-native-paper/blob/91bf179df0d68b60af4095c9a06c98c9998542e0/src/components/Toolbar/ToolbarAction.js#L37

```js
   const iconColor = dark
      ? white
      : color(black)
          .alpha(0.54)
          .rgb()
          .string();
    const rippleColor = color(iconColor)
      .alpha(0.32)
      .rgb()
      .string();
```

### Screenshots (if applicable)

Example to match blue of the Title with the menu icon: 
![image](https://user-images.githubusercontent.com/128755/39079009-bf85ef8e-44e0-11e8-9fca-2accec3067a6.png)

### Your Environment

| software              | version
| --------------------- | -------
| android        | 7,8, etc
| react-native          | 0.54
| react-native-paper    | ^1.2.5
| node                  | 8
| npm or yarn           | 5

satya164:
We should accept a `color` prop. Wanna send a PR?

PR #345, opened 2018-04-30
fix: add custom icon color for ToolbarAction

Taym95:
Add the ability to customize icon color #335 

### Test plan
Test it on the example app

### DIFF
--- a/src/components/Toolbar/ToolbarAction.js
+++ b/src/components/Toolbar/ToolbarAction.js
@@ -17,9 +17,9 @@ type Props = {
    */
   dark?: boolean,
   /**
-   * Custom icon color for Toolbar action.
+   *  Icon color for Toolbar action.
    */
-  customColor?: string,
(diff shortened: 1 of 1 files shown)


src/components/Toolbar/ToolbarAction.js
```diff
@@ -16,6 +16,10 @@ type Props = {
    * Theme color for the action icon, a dark action icon will render a light icon and vice-versa.
    */
   dark?: boolean,
+  /**
+   * Custom icon color for Toolbar action.
+   */
+  customColor?: string,
```
satya164: Can you change it to just `color`?

Taym95: Yeah, I wanted to use `color` but it was already declared in upper scope!

satya164:
Merged in d04d09dadbf238e8f406340b431e4b51c48a3f2f

PR #369, opened 2018-05-13
breaking: Chips redesign

### LINKED ISSUE #357: Chips: Redesign
- [x] Accept "avatar" prop for contact chips and change styling of "icon" prop
- [x] Implement chips with outline
- [x] Implement states: selected, disabled, pressed

image
image
image

https://material.io/design/components/chips.html

gawrysiak:
Chip redesign proposal for 2.0.

screen shot 2018-05-13 at 02 50 56

### Motivation

Fixes #357.

### Test plan

Snapshot tests updated. Run `yarn test`.


src/components/Chip.js
```diff
@@ -19,6 +19,22 @@ type Props = {
...
+   */
+  selected?: boolean,
+  /**
+   * Disables the chip. `onPress` function won't execute.
+   */
+  disabled?: boolean,
+  /**
+   * Displays the chip as pressed.
+   */
+  pressed?: boolean,
```
satya164: Instead of `pressed` prop, we should automatically apply the styles on `onPressIn` and remove them in `onPressOut` (only when `onPress` prop is provided).

src/components/Chip.js
```diff
@@ -27,6 +43,10 @@ type Props = {
    * Function to execute on delete. The delete button appears only when this prop is specified.
    */
   onDelete?: () => mixed,
+  /**
+   * Displays chip with outline.
+   */
+  outlined?: boolean,
```
satya164: Instead of `outlined` prop, can you add a `mode` prop similar to button? The values can be `flat` and `outlined`, and default would be `flat`.

src/components/Chip.js
```diff
@@ -53,37 +73,94 @@ type Props = {
...
+            style,
+            pressed || selected
+              ? {
+                  backgroundColor: pressedColor,
+                }
+              : null,
+          ]}
+        >
+          {icon || selected ? (
+            /* $FlowFixMe */ <Icon
```
satya164: Icon margin doesn't seem right:

image

should look something like:

image

Should be same as `close` button padding, which should be reduced a bit:

image

src/components/Chip.js
```diff
@@ -53,37 +73,94 @@ type Props = {
...
+              source={selected ? 'done' : icon}
+              color={iconColor}
+              size={20}
+            />
+          ) : null}
+          {avatar
```
satya164: Probably shouldn't allow both `icon` and `avatar`to be specified

satya164: For `selected` state with avatar, a check mark should appear as overlay on the avatar

image

src/components/Chip.js
gawrysiak: yeah, I was looking into the button. Couldn't figure out the default mode name. Flat is good :+1:

gawrysiak:
Updated:

screen shot 2018-05-14 at 00 33 01

8BallBomBom:
Selected with image looks like it has a padding issue on the left due to longer text?

gawrysiak:
@8BallBomBom you're right, nice catch. It has been fixed in a later commit (59f52ae).

inspireui:
Hi @gawrysiak, could you suggest how to style the actived state of Chip (with difference text color), similar this screenshot - https://user-images.githubusercontent.com/1174278/39865846-895aaf44-544e-11e8-9097-95dd3d67a525.png 

Thank you!

Issue #339, opened 2018-04-27
Checkbox animation on android


KroekerNick:
### Current behaviour

![icon](https://user-images.githubusercontent.com/5189530/39360934-a5cec6b6-4a39-11e8-96e4-dbccccd7aea2.gif)

### Expected behaviour

![icon2](https://user-images.githubusercontent.com/5189530/39360964-cc2974c8-4a39-11e8-8902-05c056708ef6.gif)

Color is not important. From the application Material Design Demo

### Code sample

from the documentation

### Screenshots (if applicable)

above

### Your Environment

| software              | version
| --------------------- | -------
| android        | 7, 8
| react-native          | 0.55.3
| react-native-paper    | ^1.4.1
| node                  | 8
| npm or yarn           | 5

ferrannp:
This issue is being stale for a year.

PR #376, opened 2018-05-16
checkbox deep ripple

ritesh-malav:
https://github.com/callstack/react-native-paper/issues/339
this will make the checkbox go give deep ripple but i am not sure how check icon can be handeled in selected state.


ferrannp:
Can you elaborate? Maybe some gifs of what it was, how is now and why we are doing it?

Issue #383, opened 2018-05-24
Cannot read property 'DURATION_LONG' of undefined


jaulz:
### Current behaviour
The following line triggers an error no matter how I try to transpile it:
https://github.com/callstack/react-native-paper/blob/master/src/components/Snackbar.js#L125

The transpiled Babel version (for  `['transform-class-properties', { spec: false }]`) looks like this for me:
`
_class.defaultProps = {
    duration: undefined.DURATION_LONG
}
`

Note that I try to use this library with `react-native-web`.

### Expected behaviour
Static properties should be transpiled correctly.

### Code sample
n/a

### Screenshots (if applicable)
n/a

### What have you tried

I tried to transpile the files in different ways with different devtools. Any idea which babel-plugin is necessary to make it work? Thanks for any hints!

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | n/a
| react-native          | 0.55
| react-native-paper    | 1.8.0
| node                  | 9.11.1
| yarn           | 0.27.5

jaulz:
I just referenced a pull request which would solve the issue easily.

PR #384, opened 2018-05-24
Avoid "this" in static properties

jaulz:
### Motivation

Solves the issue https://github.com/callstack/react-native-paper/issues/383 with simple means.

### Test plan

No special tests necessary.


satya164:
Merged in c8472ce6fce9c95f7e1d3f04d09c8edba38d1d11

PR #396, opened 2018-05-28
feat: export TouchableIcon

jaulz:
### Motivation

TouchableIcon was not exported so far and could not be used.

### Test plan

n/a


satya164:
Thanks for the PR. We want to export an icon button component (https://material.io/develop/web/components/buttons/icon-toggle-buttons/), but the `TouchableIcon` component in its current state is only intended for internal use as it's specific to toolbar and searchbar. i.e. the dimensions are hardcoded to match toolbar's buttons. The `ToolbarAction` component is basically the same thing right now.

PR #398, opened 2018-05-29
fix: make Modal nestable

jaulz:
### Motivation

Nesting modals currently cause an exception because the portal manager is not passed on.

### Test plan

```
<Dialog visible={true}>
  <Dialog visible={true}>
  </Dialog>
</Dialog>
```


satya164:
It looks good on a quick look but I'll test it later to make sure it doesn't introduce any regressions.

PR #400, opened 2018-05-29
feat: export Icon

jaulz:
### Motivation

Export Icon component as well.

### Test plan

n/a

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,7 @@ export { default as Divider } from './components/Divider';
 export { default as DrawerSection } from './components/Drawer/DrawerSection';
 export { default as FAB } from './components/FAB/FAB';
 export { default as HelperText } from './components/HelperText';
+export { default as Icon } from './components/Icon';
 export { default as ListSection } from './components/List/ListSection';
 export { default as Modal } from './components/Modal';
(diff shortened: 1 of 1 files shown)


gawrysiak:
The `Icon` component uses react-native-vector-icons/MaterialIcons. So generally:

```
import { Icon } from 'react-native-paper';

<Icon name="add" />
```

should be same as:

```
import Icon from 'react-native-vector-icons/MaterialIcons';

<Icon name="add" />
```

Importing it from paper feels a bit easier, but then it limits the icons just to `MaterialIcons`. There's some other packages that are pretty cool, like `MaterialCommunityIcons`. Might be not the best idea to create a wrapper that cripples the original vector icons features.

Can't really say if this was the original idea, it bothered me as well for some time. Hopefully someone can clarify it.

satya164:
@gawrysiak is right. You can just import `MaterialIcons` from `react-native-vector-icons` which we use internally or use another package. You have it as a dependency anyway since you need to install it to use Paper.

The icon component we have is for internal use and the additional functionality it provides (such as custom image and custom element) are great for accepting an `icon` prop in our components, but those features don't make sense when you are making your own component.

PR #405, opened 2018-05-30
BREAKING: rename Toolbar to Appbar and add Appbar.Header

satya164:
`Appbar` can be used for both top and bottom bars. `Appbar.Header` replaces the current `Toolbar` and handles status bar padding.

I didn't add an `Appbar.FAB` component because it's extremely tricky to handle it properly due to overflow issues on Android. However it's simple enough to position it outside the `Appbar`.

image


src/components/Appbar/AppbarFAB.js
```diff
@@ -0,0 +1,43 @@
+/* @flow */
```
gawrysiak: the file looks unused

src/components/Appbar/AppbarContent.js
```diff
@@ -0,0 +1,106 @@
...
+
+type Props = {
+  /**
+   * CUstom color for the text.
+   */
+  color?: string,
+  /**
+   * Text for the title.
+   */
+  title: string | React.Node,
```
gawrysiak: can be just React.Node

src/components/Appbar/AppbarContent.js
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * Style for the title.
+   */
+  titleStyle?: any,
+  /**
+   * Text for the subtitle.
+   */
+  subtitle?: string | React.Node,
```
gawrysiak: can be just React.Node

gawrysiak:
Great stuff, I really like the bottom bar. Can't see any more issues.

PR #409, opened 2018-06-04
feat: elevation for Modal and Dialog

jaulz:
### Motivation

Allow multiple groups of portals. Elevation property was already implemented for Portal but not usable in Modal in Dialog.

### Test plan

n/a

### DIFF
--- a/src/components/Dialog/Dialog.js
+++ b/src/components/Dialog/Dialog.js
@@ -15,6 +15,10 @@ const AnimatedPaper = Animated.createAnimatedComponent(Paper);
 
 type Props = {
   /**
+   * Elevation of the element in the z-axis
+   */
+  elevation?: number,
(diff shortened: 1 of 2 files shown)


satya164:
I've removed this in `material-next` to reduce complexity. Can you give an example of the scenario where you'd need this feature? Currently the modal rendered last should be on top.

jaulz:
So there won't be any possibility to set the elevation (i.e. z-index)? In my case I would like to have an authentication form always on top of everything. It's rendered when the user is not signed in or when the session times out while he uses the application. In the latter case it could be that another Modal is open in parallel and that would be rendered on top.

satya164:
@jaulz does setting a z-index not work? if it doesn't, open an issue and we can see how to make it work.

PR #410, opened 2018-06-04
fix: ref set to null on updates

jaulz:
### Motivation

The ref function is called twice on updates (see https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs). This behaviour causes the variable `this._manager` sometimes to be null during rendering and in that case actions are pushed to the queue again. Since the queue is only processed once (during componentDidMount), somes changes will never be visible. The simple solution is to bind the ref-function as mentioned in the link above.

### Test plan

Dynamically change `visible` parameter of Modal. The Modal won't be visible on the second time.


satya164:
BTW good catch. Didn't know this!

jaulz:
Better? :)

PR #413, opened 2018-06-06
fix: empty string causes error

jaulz:
### Motivation

When `description` is an empty string, the empty string is rendered instead of the `Text` node.

### Test plan

Pass empty string for `description` property.


Trancever:
Thanks for the PR. It looks good to me but let's wait what @satya164 will say :)

PR #414, opened 2018-06-06
feat: IconButton

jaulz:
### Motivation

Sorry for creating this one again but I stumbled again the same requirement that it's really helpful to have an Icon component that accepts a string (i.e. icon name) or alternatively a source object. Having the same as a touchable component makes it even better :-) It keeps the styling consistent across my app.

### Test plan

n/a


satya164:
I think exporting `TouchableIcon` is fine, though we should rename it to `IconButton` or something before exporting.

Regarding exporting `Icon`, I'm not convinced that it's useful for normal projects which aren't libraries. Let's discuss it on our discord.

gawrysiak:
I agree with @satya164 on the standard `Icon`. Being able to display an object or component instead is not really what I'd normally expect from a Material Design Icon component. That's a side effect, doesn't really bring any direct value to the original vector icons lib.

ferrannp:
@jaulz did you rename?

jaulz:
If you are okay, I can for sure refactor and rename it. Is `IconButton` the final decision?

jaulz:
@ferrannp done :)

jaulz:
Yep, will do so in the coming days.

jaulz:
@ferrannp can you please check? Though I'm not sure why the documentation is not updated since I can see it locally.

satya164:
Thanks for the PR. I'll take care of merging.

satya164:
Merged in 0108b23d33f9d215f83414f7aa289bb8d7a0bc83

PR #415, opened 2018-06-07
Platform-specific RadioButton and CheckBox

jaulz:
### Motivation

Animations for RadioButton and CheckBox are turned off for any other platform than Android so far. Since there's anyway an separate iOS component (via extension) we can remove the check and at the same time provide the animations for web platform as well.

### Test plan

n/a


satya164:
Actually I'm thinking if we should export these components as `CheckboxAndroid` and `RadioButtonAndroid` so people can use it on other platforms like iOS if they really want it. In that case, it'll be better to turn the animation off on iOS with `Platform.OS === 'ios` instead of `Platform.OS !== 'android` :)

BTW are you on our discord? You should join it if you aren't!

jaulz:
Good idea. That would also follow the conventions of React Native components. Shall I come up with an implementation or are you already working on it?

P.S.: just joined your Slack channel :-)

Trancever:
We switched from slack to discord recently, here is the link https://discord.gg/mYACqX

ocarreterom:
How about to implement this pattern?
`RadioButtonIOS` -> iOS style
`RadioButtonAndroid` -> Android (and default for web) style
`RadioButton` -> if `Platform.os === 'ios'` return `RadioButtonIOS` else return `RadioButtonAndroid`

ferrannp:
@jaulz & @satya164 is this PR still valid?

jaulz:
@ferrannp yes, it is still valid. These changes will allow you to choose from platform specific components which gives you for example the opportunity to use `CheckboxAndroid` on iOS. If you simply require `Checkbox` or `RadioButton` it will automatically use the correct component for the platform but falls back to Android (e.g., for web platform)

satya164:
It was merged in https://github.com/callstack/react-native-paper/commit/b9631580b5db79ce84d6b360a177d4c37414b36f

jaulz:
@satya164 thanks! Though it seems you missed the exports in index.js :-)

PR #422, opened 2018-06-13
fix: elevation on every platform

jaulz:
### Motivation

On any other platform than Android there's no shadow since elevation property is passed to View/SafeAreaView.

### Test plan

n/a


satya164:
Thanks for the PR. The lack of elevation shadow on the Appbar on iOS is intentional. Headers on iOS don't usually have a drop shadow. The material design docs don't add a shadow either: https://material.io/develop/ios/components/app-bars/

Issue #344, opened 2018-04-28
Error: SelectionColor isn't Working in TextInput Component


syedabuthahirm:
### Current behavior

Selection Color doesn't work on TextInput Component

### Code sample

https://snack.expo.io/@syedabuthahirm/selectioncolor---react-native-paper

I found this link trigger this behavior
https://github.com/callstack/react-native-paper/blob/27bc0e67dc0d28ee96401b7d0411cd71102e7a0b/src/components/TextInput.js#L297

adi23arora:
@ferrannp `primaryColor` is used as `labelColor` and `bottomLineColor`. So, when we are using a dark color as the `primaryColor`, then text selection highlight color will also be dark, which makes the dark text in the TextInput invisible.
To solve this, we need to set the `selectionColor` separately or set it to the `accentColor`.

adi23arora:
@ferrannp Please reopen this issue. I can send a fix for this if required.

wack17s:
#484

methodbox:
> `selectionColor` is taken from the primary color of your theme. If you want to change it, you can for example do:
> 
> ```js
> <TextInput
>   theme={{ colors: { primary: "red" }}}
>   value={this.state.inputValue}
>   onChangeText={this._handleTextChange}
>   style={{ width: 200, height: 44, padding: 8 }}
> />
> ```

This might be the worst solution to a concept in React I've seen.

Your solution is that I need to wrap my ENTIRE app in your component in order to facilitate the simple modification of a prop on a component like ```TextInput```???

That is basically the definition of the imperative approach.

Unfortunately, while I think you guys have some nifty components, while trying to simply modify something as fundamental as ```selectionColor``` I ended up on this issue and this is enough for me to uninstall until you're ready for production.

I'd say I'd like to contribute but if the approach taken was that to modify a single prop you need to apply an app-wide theme I'm not sure I want to go down that rabbit hole.

HasanYousef:
Why did you close this issue? It's still now working

Issue #401, opened 2018-05-30
Bug: Unable to set a SelectionColor Property in TextInput Component


syedabuthahirm:
### Current behaviour

SelectionColor Property Doesn't work on TextInput Component. Currently, it's checking error and error color variables from props.
Reference Links:
[Variable Declartion in TextInput Component](https://github.com/callstack/react-native-paper/blob/8974cd14c2265a7ad2c071d8e3bc8296d4549883/src/components/TextInput.js#L315)
[Variable Used in TextInput Component](https://github.com/callstack/react-native-paper/blob/8974cd14c2265a7ad2c071d8e3bc8296d4549883/src/components/TextInput.js#L373)

### Expected behaviour

It Should Accept any color

### Code sample

```js
 <TextInput
      label='Email'
      selectionColor={'#F8DE7E'}
      value={this.state.text}
      onChangeText={text => console.log(text) }
 />
```

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android
| react-native          | 0.54.0
| react-native-paper    | 1.9.1
| node                  | 8.11.2
| npm or yarn           | 4.2.0

ferrannp:
Related to https://github.com/callstack/react-native-paper/pull/425.

PR #425, opened 2018-06-20
fix: set text input selection color to accent color

adi23arora:
This fixes https://github.com/callstack/react-native-paper/issues/344 or https://github.com/callstack/react-native-paper/issues/401

### Motivation

With this PR, the `selectionColor` is set to the `accent` color. `selectionColor` is used to change the `TextInput` cursor and text highlight color. 
Previously, it was set to `primary` color which makes it difficult to view the highlighted text if the primary color is dark. So, to separate it from the `primary` color used for the label and underline color, we can use `accent` color for the same.

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -304,6 +304,7 @@ class TextInput extends React.Component<Props, State> {
     const fontFamily = fonts.regular;
     const {
       primary: primaryColor,
+      accent: accentColor,
       disabled: inactiveColor,
       error: errorColor,
     } = colors;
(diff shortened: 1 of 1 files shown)


adi23arora:
@satya164 Please review and suggest changes if any?

syedabuthahirm:
But it does show red color when we have error. We have check for that also.

adi23arora:
@syedabuthahirm Yes but color used for the `errorColor` could also interfere with the dark color text in your `TextInput`. That `errorColor` is pre-set and isn't customizable, so we might have problems with the red color or any dark color text.
Hope you are getting the point.

PR #427, opened 2018-06-21
Fix bug that blocks touch at the bottom of the screen

bosung90:
If you have a tabbar or anything at the bottom, and you move the snackbar up, snackbar will block all pointer events at the bottom of the screen.
`pointerEvents="box-none"` on both SafeAreaView and Animated.View will fix this issue but still allow you to add buttons within Snackbar component

### DIFF
--- a/src/components/Snackbar.js
+++ b/src/components/Snackbar.js
@@ -174,8 +174,9 @@ class Snackbar extends React.Component<Props, State> {
 
     return (
       <ThemedPortal>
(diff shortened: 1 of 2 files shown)


bosung90:
Can someone update snapshot for me?

jaulz:
@bosung90 any issues when updating the snapshots?

bosung90:
@ferrannp Can you check?

PR #430, opened 2018-06-22
Add disabled props

bosung90:

### DIFF
--- a/src/components/FAB.js
+++ b/src/components/FAB.js
@@ -32,6 +32,10 @@ type Props = {
    */
   dark?: boolean,
   /**
+   * Is Button disabled?.
+   */
+  disabled?: boolean,
+  /**
    * Custom color for the `FAB`.
    */
   color?: string,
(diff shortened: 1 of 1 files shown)


ferrannp:
@bosung90 is there a reason to have a FAB disabled? Have you found this in material guideliness?

bosung90:
@ferrannp Yes, I use this FAB button as "confirm purchase" button. I need to disable this button while server confirms that the payment has been successfully processed. I didn't check the material guidelines but it is a button, so I don't see a reason why disabled props should not be passed through?

jaulz:
@bosung90 do you need help?

bosung90:
@jaulz Thanks! I am actually really busy this week. I was thinking of working on it next weekend but if you can help it would be wonderful. I just found out that the text component in line 109 also needs 
`pointerEvents="none"` because the text will block the touch events underneath. Thanks!

jaulz:
Since I could not push to your repository please see this PR: https://github.com/callstack/react-native-paper/pull/445

ferrannp:
Close in favour of #445.

PR #431, opened 2018-06-25
fix: prevent focus and blur if disabled

jaulz:
### Motivation

Focus and blur events are still triggered even if the input is disabled

### Test plan

Set disabled to true and see how you can still focus and blur an input field.

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -228,7 +228,7 @@ class TextInput extends React.Component<Props, State> {
     if (this.props.disabled) {
       return;
     }
-    
+
     this.setState({ focused: true });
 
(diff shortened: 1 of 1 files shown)


ferrannp:
I think the problem here is that we are not passing `disabled` prop.

jaulz:
You mean for the TextInput itself? Here it is passed:
https://github.com/jaulz/react-native-paper/blob/36066f7b2e9fedced859bb6159a70d5ebdb02857/src/components/TextInput.js#L380

Even if the input is disabled, you are able to focus the input

satya164:
There's no `disabled` prop in `TextInput` @ferrannp

PR #435, opened 2018-06-29
feat: allow disabling of AppbarAction

jaulz:
### Motivation

Currently, AppbarActions cannot be disabled. This allows to pass on the `disabled` property.

### Test plan

n/a


satya164:
Merged in 1e3ddc4

PR #444, opened 2018-07-06
feat: indeterminate Checkbox

jaulz:
### Motivation

Implement indeterminate Checkboxes according to:
https://material.io/design/components/selection-controls.html#checkboxes

### Test plan

image

### DIFF
--- a/example/src/CheckboxExample.js
+++ b/example/src/CheckboxExample.js
@@ -18,6 +18,7 @@ type Props = {
 type State = {
   checkedNormal: boolean,
   checkedCustom: boolean,
+  indeterminate: boolean,
 };
 
 class CheckboxExample extends React.Component<Props, State> {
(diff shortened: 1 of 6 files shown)


jaulz:
Hm, I'm not sure if that's a good idea from a developer perspective since he needs to handle strings instead of truthy/falsey values.

satya164:
It's a minor inconvenience, but better than an inconsistent API IMO. Having two separate props which control the same thing also means that people who actually use the `indeterminate` will have to deal with 2 props in all places which can be cumbersome, and if someone wants to add this behavior to existing app, it'll require more refactoring.

jaulz:
@satya164 just implemented the changes as suggested :) The indeterminate status for radio buttons is by the way not part of the specs (https://material.io/design/components/selection-controls.html#radio-buttons).

ferrannp:
@satya164 can you re-check?

jaulz:
@satya164 any chance to have a look at it for the 2.0 release?

satya164:
merged in 3b09171aff76baddcb86f0c46685a913a4b7946d

PR #446, opened 2018-07-09
feat: DataTable component

jaulz:
### Motivation

New component Data Table according to https://material.io/design/components/data-tables.html.

### Test plan

image


src/components/DataTable/DataTableCell.js
```diff
@@ -0,0 +1,94 @@
...
+
+import React from 'react';
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
+import Text from './../Typography/Text';
+
+type Props = {
+  /**
+   * Content of the `DataTableCell`.
+   */
+  children: string | React.Node,
```
gawrysiak: can be just React.Node (https://flow.org/en/docs/react/types/#toc-react-node)

src/components/DataTable/DataTableCell.js
```diff
@@ -0,0 +1,94 @@
+/* @flow */
+
+import React from 'react';
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
+import Text from './../Typography/Text';
```
gawrysiak: `'../Typography/Text'` should be enough. Same for other imports. That's minor though :)

src/components/DataTable/DataTablePagination.js
```diff
@@ -0,0 +1,121 @@
...
+  componentDidUpdate() {
+    const { count, onChangePage, page, rowsPerPage } = this.props;
+    const newLastPage = Math.max(0, Math.ceil(count / rowsPerPage) - 1);
+
+    if (page > newLastPage) {
+      onChangePage(newLastPage);
+    }
+  }
+
+  render(): Object {
```
gawrysiak: is the Object type here needed?

src/components/DataTable/DataTablePagination.js
```diff
@@ -0,0 +1,121 @@
...
+  onChangePage: () => mixed,
+  /**
+   * Customize the displayed rows label.
+   */
+  labelDisplayedRows?: () => mixed,
+  /**
+   * Customize the rows per page label. Invoked with a `{ from, to, count, page }`
+   * object.
+   */
+  labelRowsPerPage?: () => mixed,
```
gawrysiak: string here maybe?

src/components/DataTable/DataTable.js
```diff
@@ -0,0 +1,91 @@
...
+import { View } from 'react-native';
+import DataTableBody from './DataTableBody';
+import DataTableCell from './DataTableCell';
+import DataTableHead from './DataTableHead';
+import DataTableHeaderCell from './DataTableHeaderCell';
+import DataTablePagination from './DataTablePagination';
+import DataTableRow from './DataTableRow';
+
+type Props = {
+  style?: any,
```
gawrysiak: missing children prop

src/components/DataTable/DataTableBody.js
```diff
@@ -0,0 +1,18 @@
+/* @flow */
+
+import React from 'react';
+import { View } from 'react-native';
+
+type Props = {
+  style?: any,
```
gawrysiak: missing children prop

src/components/DataTable/DataTableHeaderCell.js
```diff
@@ -0,0 +1,141 @@
+/* @flow */
+
+import React from 'react';
```
gawrysiak: try import * as React, breaks flow otherwise

src/components/DataTable/DataTableCell.js
```diff
@@ -0,0 +1,94 @@
+/* @flow */
+
+import React from 'react';
```
gawrysiak: try import * as React, breaks flow

src/components/DataTable/DataTableHeaderCell.js
```diff
@@ -0,0 +1,141 @@
...
+};
+
+class DataTableHeaderCell extends React.PureComponent<Props, State> {
+  static displayName = 'DataTable.HeaderCell';
+
+  static defaultProps = {
+    size: 1,
+  };
+
+  constructor(props) {
```
gawrysiak: flow wants `props: Props` here

src/components/DataTable/DataTableHeaderCell.js
```diff
@@ -0,0 +1,141 @@
...
+  render() {
+    const { children, numeric, onPress, size, sortDirection } = this.props;
+
+    const spin = this.state.spinAnim.interpolate({
+      inputRange: [0, 1],
+      outputRange: ['0deg', '180deg'],
+    });
+    const icon = sortDirection ? (
+      <Animated.View style={[styles.icon, { transform: [{ rotate: spin }] }]}>
+        <Icon source="arrow-downward" size={16} />
```
gawrysiak: color is required for the icon

src/components/DataTable/DataTableRow.js
```diff
@@ -0,0 +1,42 @@
+/* @flow */
+
+import React from 'react';
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
+import { grey100 } from './../../styles/colors';
+
+type Props = {
+  /**
```
gawrysiak: missing children prop

jaulz:
@gawrysiak thanks for the comments! Will do the changes and update you. Sorry for the flow issues but I had to skip them locally since it runs into an infinite loop locally. Any idea why this is the case?

gawrysiak:
@jaulz yeah, other than the comments, there's quite a long list of reported flow issues. Visually looks nice though. @satya164 can probably give you some more API feedback.

As for the flow issue, it's worth a try to run flow through the local node_modules instead of the globally installed one. It should be possible with most of the plugins, e.g. VSC:
screen shot 2018-07-09 at 21 10 08
or Nuclide:
screen shot 2018-07-09 at 21 09 53

Then look into your processes and kill any flow servers that shouldn't be running (when editor is off).

As for the terminal, normally killing the global process and running `yarn flow` should do.

jaulz:
Okay, fixes are in. Though there is one left regarding theming, e.g.:
> Cannot create `DataTable.HeaderCell` element because property `theme` is missing in props [1] but exists in `Props` [2].

Any idea why this occurs? Compared my component to the existing ones and can't find the difference... shouldn't `withTheme` provide the theme parameter?

satya164:
Not sure. Will need to check later.

ferrannp:
How is this PR going @satya164 ?

jaulz:
@satya164 just pushed an update so it's properly passing the build

giautm:
Waitting for it. 👍

jaulz:
@giautm thanks for the hint! Though, I would rather wait for feedback from the team before I do any more changes...
@ferrannp @satya164 any comments from your side? I would be happy to incorporate any requested changes.

giautm:
@jaulz : no, `{}` make an object with property `children`. That will cause:
image

```jsx
        {typeof children === 'string' || typeof children === 'number' ? (
          <Text
            style={[
              styles.cell,
              numeric ? styles.numeric : styles.text,
              this.props.style,
            ]}
            numberOfLines={1}
          >
            {children}
          </Text>
        ) : (
          children // <---- no need `{}` right here.
        )}
```

src/components/DataTable/DataTableCell.js
```diff
@@ -0,0 +1,94 @@
...
+              styles.cell,
+              numeric ? styles.numeric : styles.text,
+              this.props.style,
+            ]}
+            numberOfLines={1}
+          >
+            {children}
+          </Text>
+        ) : (
+          { children }
```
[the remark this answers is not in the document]
jaulz: Sorry, actually I meant to wait for feedback for this PR in general. As long as I don't know if this PR will be considered in this way at all it does not make sense to fix issues, does it?
Once I receive green light from your side, I am more than happy to fix issues...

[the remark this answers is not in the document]
Trancever: Ok, I understand :) For me this PR looks great, I like the api of components. The only thing that those components misses for me are styles based on theme. There is no way to change theme to dark, we have to pass style prop to every component, am I right?

@satya164 Could you take a look on this PR?

src/components/DataTable/DataTablePagination.js
```diff
@@ -0,0 +1,130 @@
...
+    color: '#00000099',
+  },
+
+  rowsPerPage: {
+    marginRight: 30,
+  },
+
+  displayedRows: {
+    marginLeft: 30,
+    marginRight: 44,
```
giautm: I think we should not use the fixed margin like this. That cause some render issues on the device with the small screen width. (ex: iPhone SE.)
image

src/components/DataTable/DataTablePagination.js
```diff
@@ -0,0 +1,130 @@
...
+      labelRowsPerPage,
+      theme,
+    } = this.props;
+    const from: number = count === 0 ? 0 : page * rowsPerPage + 1;
+    const to: number = Math.min(count, (page + 1) * rowsPerPage);
+
+    return (
+      <View style={[styles.container, this.props.style]}>
+        <Text style={[styles.text, styles.rowsPerPage]} numberOfLines={1}>
+          {labelRowsPerPage} {rowsPerPage}
```
giautm: Should remove the spacing between `{labelRowsPerPage}` and `{rowsPerPage}`. And set default props of `labelRowsPerPage` to `'Rows per page: '`

giautm: Or simply use the custom render right here. 😄

ferrannp:
What's the status of this one @jaulz ?

jaulz:
@ferrannp if @satya164 could take a look at the component first and confirms then I could adapt the component to the dark theme as well.

satya164:
I pushed some changes in https://github.com/callstack/react-native-paper/tree/%40satya164/data-table, but it still needs typescript definitions. ANybody up for adding the defs?

jaulz:
@satya164 will take it up next week, thanks!

jaulz:
@satya164 types are created in your branch. shall we create a separate PR or merge the changes in this one here?

satya164:
@jaulz thanks! merging the changes to this branch works for me :)

src/components/DataTable/DataTableCell.js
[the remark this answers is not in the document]
jaulz: Will check it tomorrow :blush: it's anyway not yet merged...

jaulz:
@satya164 thanks! Though, I just created another PR https://github.com/callstack/react-native-paper/pull/684 which I was working on in parallel and allows not only text and numbers in cells.

Issue #436, opened 2018-06-29
TextInput's behavior diverging from Material's guidelines 


knobandre:
### Current behaviour

TextInput's states are not behaving quite like Material's. When the component is on the "idle & filled" state, the label and bottom line are still with focus effect.

### Expected behaviour

When the component is on the "idle & filled" state, the label and bottom line should be grey.
For reference, check the text field's guidelines [here](https://material.io/archive/guidelines/components/text-fields.html#text-fields-states).

### Screenshots

![screenshot_20180629-154416](https://user-images.githubusercontent.com/15720180/42109433-0246be0c-7bb4-11e8-9e76-5a9d9a9a2b5e.png)

### Code Sample

Not needed as the screenshot was taken from the expo demo.

### Your Environment

| software              | version
| --------------------- | -------
| android        | 6.0
| react-native          | 0.55.3
| react-native-paper    | 1.10.4
| node                  | 8.10.0
| npm or yarn           | 5.6.0

PR #454, opened 2018-07-14
TextInput: Only highlight bottom line when focused

krizpoon:
### Motivation

The TextInput should only be highlighted when focused or in error state. #436 

### Test plan

1. Focus the text input. The label color and the bottom line color should be the primary theme color and the bottom line highlighted.
2. Blur the text input. The color should be the disabled color and the bottom line is not highlighted.
3. Enter some text and blur the text input. The color should be the disabled color and the bottom line is not highlighted.
4. Set text input in error state, the color should be the error color and the bottom line is highlighted.


satya164:
merged in a19ffa09ce89ad219a017c1e62b9226667742887

Issue #458, opened 2018-07-18
[Feature Request] Modification to ListItem


ballenwillis:
Maybe there's a better place for feature requests, but I couldn't find anything. I would like to have a component like this if it's possible. I couldn't find anything in your kitchen sink app.

![image](https://user-images.githubusercontent.com/14190965/42886359-f91ecc7a-8a68-11e8-8b16-711c9ddf09f7.png)

ballenwillis:
This can easily happen by giving the `listItem` a `paddingLeft: 0` to overwrite `paddingLeft: 64`.

ferrannp:
Commented on #459.

PR #459, opened 2018-07-18
feat: Applying indent implemented

ballenwillis:
### Motivation
I was asked to create a component with drop-down functionality equivalent to `ListAccordion`, but without the children indented. I opened up a very small issue (that will probably get deleted) [here](#458 ). 

I could just add `paddingLeft: 0` to my `ListItem` component on my end, but I thought this would be a better solution if it were just implemented straight into React Native Paper since it was relatively simple and easier to understand why `indented={false}` is in my code more than `{paddingLeft: 0}`.

### Test plan
Run the example app and set the `indented` prop to false for a `ListAccordian` component. One example is [here](https://github.com/callstack/react-native-paper/blob/master/example/src/ListAccordionExample.js#L30). It changes the children of `ListAccordion` to not be indented.

### DIFF
--- a/src/components/List/ListAccordion.js
+++ b/src/components/List/ListAccordion.js
@@ -22,6 +22,10 @@ type Props = {
    * Icon to display for the `ListAccordion`.
    */
   icon?: React.Node,
+  /**
+   * Indicates whether children should be indented or not.
+   */
+  indented?: boolean,
 
   /**
    * Content of the section.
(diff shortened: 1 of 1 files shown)


ferrannp:
We would like to stick to `styles` or we can end up adding too many specific cases that people need. You can for example do:

```js
const ListItem = ({ indented, ...rest}) => {
  return <PaperComponent style={[rest.style, indented && { paddingLeft: 0 }]} {...rest} />
}
```

Issue #450, opened 2018-07-10
TextInput does not show the string from beginning


knobandre:
### Current behaviour

When the string is larger than the width of TextInput and the component is mounted or lose it's focus, the component does not display the string from the very beginning. Rather, the component starts displaying the string by the end.

### Expected behaviour

When the component is mounted or lose it's focus, the component should display the string starting from left to right.  

### Code sample

Not needed as the screenshot was taken from the expo demo.

### Screenshots
Example string: 'Jonathan erina joseph zeppelin speedwagon jonathan erina'. Note that when the component loses it's focus, the string keeps being displayed from right to left.

![screenshot_20180710-092647](https://user-images.githubusercontent.com/15720180/42510378-13f21826-8425-11e8-8f81-7118e0a77381.png)

### What have you tried

Checked if there's a prop to control such behavior, none was found.

### Your Environment

| software              | version
| --------------------- | -------
| android        | 6.0
| react-native          | 0.55.3
| react-native-paper    | 1.10.4
| node                  | 8.10.0
| npm or yarn           | 5.6.0

jaulz:
@knobandre not sure if the team will add it to the library but I just created a pull request where you can see how you could possibly solve this issue.

kpsroka:
Hi @knobandre,
Could you point me to some official material design documentation that mentions this behaviour? I've tried to find some support for it, but couldn't find it. [Google's web demo](https://material-components.github.io/material-components-web-catalog/#/component/text-field) implements it, but it's not consistent with the prod apps (see screencap from Google Maps app below).
![google-maps-text-input-rewind](https://user-images.githubusercontent.com/24893014/43460992-8ea264dc-94d2-11e8-982c-b8c4b8445525.gif).

knobandre:
Hi kpsroka,

I could not find any Material documentation about this behavior. After some discussion, some analysts in my team agreed that this would be the desired behavior for our application. 
But since we don't have any Material docs to reinforce this behavior, we're gonna make use of the multiline prop to achieve a more acceptable behavior.

Thank you for your time, appreciate it.

PR #463, opened 2018-07-20
fix: show beginning of text in input

jaulz:
### Motivation

When the string is larger than the width of TextInput and the component is mounted or lose it's focus, the component does not display the string from the very beginning. Rather, the component starts displaying the string by the end.
See issue https://github.com/callstack/react-native-paper/issues/450

### Test plan

Enter text longer than the screen width and afterwards unfocus TextInput. The beginning of the text in the input should be visible. Once you focus the TextInput again, the selection is at the very end.

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -79,6 +79,10 @@ type State = {
   focused: boolean,
   placeholder: ?string,
   value: ?string,
+  selection: ?{
+    start: number,
+    end: number,
+  },
 };
 
 /**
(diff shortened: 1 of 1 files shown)


jaulz:
Maybe we should think about making this behaviour optional or what do you think?

ferrannp:
@kpsroka as you are assigned to #450. Can you check this one?

kpsroka:
Hi @jaulz!
As you can read in the referenced ticket, we decided there's no support in material docs for this kind of behaviour, and we shouldn't include it in paper. Thank you for your time and code -- while we'll not include it in the library, it still might be usable by people who want to implement it on their own.

PR #475, opened 2018-07-31
feat: add background color prop on FAB group action item

tkorakas:
### Motivation
Adds optional background color property to action item of FAB group component.

### Test plan

image

### DIFF
--- a/src/components/FABGroup.js
+++ b/src/components/FABGroup.js
@@ -25,14 +25,14 @@ type Props = {
    * - `icon`: icon to display (required)
    * - `label`: optional label text
    * - `color`: custom icon color of the action item
-   * - `backgroundColor`: custom icon background color of the action item
+   * - `style`: pass additional styles, for example, backgroundColor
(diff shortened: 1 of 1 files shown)


src/components/FABGroup.js
```diff
@@ -25,12 +25,14 @@ type Props = {
...
    * - `label`: optional label text
    * - `color`: custom icon color of the action item
+   * - `backgroundColor`: custom icon background color of the action item
    * - `onPress`: callback that is called when `FAB` is pressed (required)
    */
   actions: Array<{
     icon: string,
     label?: string,
     color?: string,
+    backgroundColor?: string,
```
satya164: Thanks for the PR! Can you change it `style` instead?

tkorakas: I made the change you suggested, let me know if you need anything else.

satya164:
Merged in https://github.com/callstack/react-native-paper/commit/8b9176b44bb01a1eef497a403b0304bc389c9aee

PR #482, opened 2018-08-03
docs: minor english fix in snackbar actions doc

grsabreu:
### Motivation
Minor english fix. `contain` should be used normally when followed by `should`, `following` also asks for an article.


Trancever:
Hi @grsabreu . Thanks for contribution.

Issue #9, opened 2016-10-19
Menu and Dropdown


ahmedlhanafy:
[MD Docs Link](https://material.google.com/components/menus.html#menus-usage)

marktkimball:
Is anything happening with this component?  I love this component library, but without the drop down menu to use with the "More" icon in a Toolbar, it seems slightly gimped.  What are people doing to display a "more menu" from the Toolbar?

Trancever:
@marktkimball @kpsroka is currently working on that component, so it should be released in the near future.

satya164:
If you're on Android, you can currently use the undocumented popup menu in React Native:

```js
UIManager.showPopupMenu(
  ReactNative.findNodeHandle(
    this._anchor // Ref to the item that'll act as anchor to the popup menu
  ),
  ['One', 'Two', 'Three'],
  error => { /* handle error */ },
  (
    action, // "itemSelected", "dismissed"
    index // index of item that's selected
  ) => { /* handle action /* },
);
```

marsonmao:
Any progress on this component?

ferrannp:
@marsonmao not yet.

ahce:
This [project](https://github.com/n4kz/react-native-material-dropdown) is a reference for this

cltsang:
It's being implemented in #485

mgunther:
Look to react-native-material-menu...

PR #485, opened 2018-08-08
feat: Menu component

iyadthayyil:
### Motivation
Menu component with auto positioning and RTL support:
https://material.io/design/components/menus.html
fixes: #9 

### Test plan
menu-2


ferrannp:
@iyadthayyil are you still working on this or you need some help? :)

src/components/Menu/AnimatedSurface.js
```diff
@@ -0,0 +1,81 @@
...
+ *     height: 80,
+ *     width: 80,
+ *     alignItems: 'center',
+ *     justifyContent: 'center',
+ *     elevation: 4,
+ *   },
+ * });
+ * ```
+ */
+class Surface extends React.Component<Props> {
```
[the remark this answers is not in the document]
iyadthayyil: It's a quick hack to overcome an animation  problem I faced when creating a new animated view from the existing `Surface` component. This was the original way I did it:
```javascript
const AnimatedSurface = Animated.createAnimatedComponent(Surface);
```
But this seems to render very choppy animations, would like to know if there is a way to just use `Surface` without this animation issue.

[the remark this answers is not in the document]
satya164: @iyadthayyil We can change `View` to `Animated.View` here: https://github.com/callstack/react-native-paper/blob/master/src/components/Surface.js#L65

Then we wouldn't need `Animated.createAnimatedComponent(Surface)`

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,305 @@
...
+  children: React.Node,
+  style?: any,
+  /**
+   * @optional
+   */
+  theme: Theme,
+};
+
+/**
+ * Menus display a list of choices on temporary surfaces. Their placement varies based on the element that opens them.
```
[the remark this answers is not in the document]
iyadthayyil: Ok will do that.

iyadthayyil:
Ok, sorry guys, I have been busy with uni work, will try to get this ready soon.

raajnadar:
Any ETA for this component?? This component looks beautiful

thevikas:
knock knock

geminiyellow:
please move forward, we need menu

waquidvp:
Hey guys, I want to help merge this PR. Can I rebase from 'upstream' master or is that not needed?

waquidvp:
Hey guys, I have made some changes to follow the new way of doing things compared to when the PR was first created. Part of this was to add the typescript typings here: `typings/components/Menu.d.ts`, this is my first time doing some typescript so would be cool if someone could have a look and give some feedback.

ferrannp:
Hey @waquidvp thanks for coming back to this. I guess you work together with @iyadthayyil ?

Would you mind adding: Screenshots and code for docs? You can see how to do it from other components.

I am also wondering if we could animate scale instead animating `height` and `width`, that would allow use to use `useNativeDriver`. However we can merge first this and then focus on that.

waquidvp:
There were already some screenshots and some code examples. They just weren't in the right place and so wasn't showing up in the generated docs. I have moved it to the right place and the docs look ok now.

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,305 @@
...
+
+type Props = {
+  /**
+   * Whether the Menu is currently visible.
+   */
+  visible: boolean,
+  /**
+   * The button to trigger the menu.
+   */
+  button: React.Node,
```
ferrannp: I am wondering if we should call this prop`button`. As far as I know these menus can appear in different kind of actions (like in forms). Should we call it `anchor` as is more generic and it is what they use in the docs? Does it make sense?

waquidvp:
So I finished the changing all the things that @kpsroka mentioned, now waiting on @iyadthayyil to make a decision on the `button` to `anchor` change.

waquidvp:
So changed `button` to `anchor`, is there anything else we need to fix or do before merging this?

typings/components/Menu.d.ts
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { StyleProp, ViewStyle, ViewProps } from 'react-native';
+import { ThemeShape, IconSource } from '../types';
+
+export interface MenuProps {
+  visible: boolean;
+  anchor: React.ReactNode;
+  onDismiss: () => any;
```
ferrannp: Should be `() => void`.

typings/components/Menu.d.ts
```diff
@@ -0,0 +1,25 @@
...
+  children: React.ReactNode;
+  style?: any;
+  theme?: ThemeShape;
+}
+
+export interface ItemProps {
+  title: React.ReactNode;
+  icon?: IconSource;
+  disabled?: boolean;
+  onPress?: () => any;
```
ferrannp: Should be `() => void`.

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,305 @@
...
+    };
+
+    const animationStarted = menuState === 'animating';
+    const modalVisible = menuState === 'shown' || animationStarted || visible;
+
+    return (
+      <View ref={this._setContainerRef} collapsable={false}>
+        <View onLayout={this._onAnchorLayout}>{anchor}</View>
+
+        <Modal visible={modalVisible} onRequestClose={this._hide} transparent>
```
satya164: Don't use `Modal` from `react-native`. Use `Portal` from `react-native-paper`

typings/components/Menu.d.ts
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { StyleProp, ViewStyle, ViewProps } from 'react-native';
+import { ThemeShape, IconSource } from '../types';
+
+export interface MenuProps {
+  visible: boolean;
+  anchor: React.ReactNode;
+  onDismiss: () => any;
+  children: React.ReactNode;
+  style?: any;
```
satya164: Use `StyleProp<ViewStyle>`

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,305 @@
...
+      left: I18nManager.isRTL ? {} : left,
+      top,
+    };
+
+    const animationStarted = menuState === 'animating';
+    const modalVisible = menuState === 'shown' || animationStarted || visible;
+
+    return (
+      <View ref={this._setContainerRef} collapsable={false}>
+        <View onLayout={this._onAnchorLayout}>{anchor}</View>
```
satya164: Is the wrapper view around anchor necessary? feels like this `onLayout` could be placed on the parent view as well

satya164: Also, can we measure the layout when we first need to show the menu instead of mount? Will be better for perf.

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,305 @@
...
+  menuHeight: number,
+  anchorWidth: number,
+  anchorHeight: number,
+  opacityAnimation: Animated.Value,
+  menuSizeAnimation: Animated.ValueXY,
+  menuState: 'hidden' | 'animating' | 'shown',
+};
+
+const ANIMATION_DURATION = 300;
+const EASING = Easing.bezier(0.4, 0, 0.2, 1);
```
satya164: Can you add a note on where you got these numbers? Doesn't any existing easing preset in RN match these?

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,305 @@
...
+const SCREEN_INDENT = 8;
```
satya164: Can you add a comment on what this is used for?

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,305 @@
...
+    menuSizeAnimation: new Animated.ValueXY({ x: 0, y: 0 }),
+  };
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.visible !== this.props.visible) {
+      this._updateVisibility();
+    }
+  }
+
+  _container: any;
```
satya164: Should be `?View`

src/components/Menu/Menu.js
waquidvp: The 'Standard easing' section of this page states that numbers that we are using, and from what I have looked at, there isn't any identical ones in RN: https://material.io/design/motion/speed.html#easing

src/components/Menu/Menu.js
waquidvp: Done.

satya164:
Pushed the requested changes here: https://github.com/callstack/react-native-paper/compare/pr/485

Seems when using our `Portal`, the menu is under the statusbar. So we need to handle it and then merge.

waquidvp:
To solve the status bar issue, should a combination `SafeAreaView` and `StatusBar.currentHeight` do, or is there any other we handle it in this library?

satya164:
> To solve the status bar issue, should a combination SafeAreaView and StatusBar.currentHeight do, or is there any other we handle it in this library?

Let's do it for now. We can refactor to extract this logic later.

src/components/Menu/Menu.js
waquidvp: @satya164 thanks for adding this, I have merged your change to this branch.

typings/components/Menu.d.ts
waquidvp: Done

src/components/Menu/Menu.js
waquidvp: I have changed it to remove the inner view and just add the `onLayout` on the outer view

ferrannp:
@satya164 can we have your approval here for merging it?

geminiyellow:
ohhh! amazing! thanks all.

RichardLindhout:
Great work @iyadthayyil! Thanks everyone!

Zefau:
Can the Menu component be used as a Dropdown for TextInput as shown https://material.io/design/components/menus.html#exposed-dropdown-menu ?

Replacing the `anchor` attribute with TextInput (and using `onFocus` instead of `onPress`) won't work.

mihaidaviddev:
@Zefau I also needed to do something similar and I ended up using focus and blur events. It's working on android, but for ios is working just in the simulator. On real ios device the surface has a bug related to height. Any ideea what might be the problem ? This is how the component looks like

`import React, { Component } from "react";
import { Keyboard, ScrollView } from "react-native";
import { Menu, TextInput, Theme } from "react-native-paper";
import PaperMenu from "./PaperMenu";

export interface TextInputAutocompleteProps {
  width: number;
  items: string[];
  value: string;
  theme: Theme;
  label: string;
  onChange: (value: string) => void;
}

interface TextInputAutocompleteState {
  filteredItems: string[];
  value: string;
  showMenu: boolean;
}

export default class TextInputAutocomplete extends Component<
  TextInputAutocompleteProps,
  TextInputAutocompleteState
> {
  state = {
    filteredItems: [] as string[],
    value: "",
    showMenu: false
  };
  componentDidMount = () => this.updateStateFromProps(this.props);

  componentWillReceiveProps = (props: TextInputAutocompleteProps) =>
    this.updateStateFromProps(props);

  updateStateFromProps = (props: TextInputAutocompleteProps) =>
    this.setState({ value: props.value });

  filterItems = value => {
    const filteredItems =
      value.trim() !== ""
        ? this.props.items.filter(
            item =>
              item !== value &&
              item.toUpperCase().startsWith(value.toUpperCase())
          )
        : [];
    const showMenu = filteredItems.length > 0;
    this.setState({
      showMenu,
      filteredItems
    });
  };

  onTextInputFocus = () => this.filterItems(this.state.value);

  onTextInputBlur = () => this.setState({ showMenu: false });

  onMenuDismiss = () => {
    this.setState({ showMenu: false });
    Keyboard.dismiss();
  };

  onItemPress = (item: string) => {
    this.onMenuDismiss();
    this.setState({ value: item });
    this.props.onChange(item);
  };

  onTextInputChange = value => {
    this.setState({ value });
    this.filterItems(value);
  };

  render() {
    const menuItems = this.state.filteredItems.map((item, i) => {
      const onPress = () => this.onItemPress(item);
      return (
        <Menu.Item
          style={{ minWidth: this.props.width, maxWidth: this.props.width }}
          onPress={onPress}
          key={i}
          title={item}
        />
      );
    });
    return (
      <PaperMenu
        style={{ marginTop: 50, width: this.props.width }}
        visible={this.state.showMenu}
        onDismiss={this.onMenuDismiss}
        anchor={
          <TextInput
            label={this.props.label}
            mode="outlined"
            theme={this.props.theme}
            value={this.state.value}
            onChangeText={this.onTextInputChange}
            onFocus={this.onTextInputFocus}
            onBlur={this.onTextInputBlur}
          />
        }
      >
        <ScrollView
          scrollEnabled={true}
          showsVerticalScrollIndicator={true}
          keyboardShouldPersistTaps="always"
          style={{ height: 135 }}
        >
          {menuItems}
        </ScrollView>
      </PaperMenu>
    );
  }
}
`
The PaperMenu is a fork of Menu component to temporary fix the bug I described here
[https://github.com/callstack/react-native-paper/issues/970](url)

RichardLindhout:
On iOS Statusbar.currentHeight does not work so ending up with a undefined top.

mihaidaviddev:
Yes..I use a default value of 40 for now..so i removed StatusBar completely..the issue I have now is with the height of the menu..not the position

raajnadar:
Please create issue ticket instead of posting here. This component is very new so you will see lot of issues.

PR #497, opened 2018-08-17
fix: FAB elevation to match material design

### LINKED ISSUE #495: [alpha] FAB elevation on press
### Current behaviour

When we press the FAB the elevation is always the same (12).

### Expected behaviour

The elevation should be 6 by default and should change to 12 while pressing. Check [this](https://material.io/design/environment/elevation.html#default-elevations)!

### Your Environment

| react-native-paper    | 2.0.0-alpha.7

I would be happy to contribute to this tonight. Can I do a PR with the changes? It would look like what was made for [Button](https://github.com/callstack/react-native-paper/blob/material-next/src/components/Button.js).

luissmg:
Fixes #495

### DIFF
--- a/src/components/FAB/FAB.js
+++ b/src/components/FAB/FAB.js
@@ -48,6 +48,10 @@ type Props = {
   theme: Theme,
 };
 
+type State = {
+  elevation: Animated.Value,
+};
+
 /**
  * A floating action button represents the primary action in an application.
  *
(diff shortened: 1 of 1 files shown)


kpsroka:
@luissmg Thanks for this PR. I think it's going to be useful if there's also an `elevation` prop, similarly to how it is done in the [Card component](https://github.com/callstack/react-native-paper/blob/a6508dcfe2dc9290e9eb588cfa48e3d0cd970ff5/src/components/Card/Card.js#L78?). Can you add that as well?

luissmg:
@kpsroka Sure, I could do that easily but shouldn't the values be the ones from MD?

What should it do if the elevation value passed in is higher then the value when you press?

satya164:
Yeah, probably doesn't make sense to have an elevation prop. For card, it's a leftover from when we accepted elevation as a prop. I'll remove it in material-next.

luissmg:
@satya164 Thanks for merging this, glad I could contribute!

satya164:
I reverted this since it was throwing an error with `FAB.Group`. If you manage to fix it, feel free to send another PR!

luissmg:
I will try to look into it this week!

luissmg:
@satya164 could you be more specific about the error?

satya164:
it's about native driver, if you tap the fab group button you'll see it

luissmg:
Thanks for the feedback!

PR #499, opened 2018-08-17
feat: TextInput RTL support

Esemesek:
- [x] - [Working RTL](https://i.imgur.com/sXYWKZi.png)
- [x] - [Working LTR](https://i.imgur.com/SjV0k8Z.png)

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -6,6 +6,7 @@ import {
   Animated,
   TextInput as NativeTextInput,
   StyleSheet,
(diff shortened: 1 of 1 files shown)


HassanHeydariNasab:
RTL doesn't mean that you flip everything :/

RTL languages should be written from right to left and LTR languages should be written from left to right. almost nobody wants to write English from right to left and align it to the right. Nobody wants to stick Arabic text to the left of TextInput.

PR #503, opened 2018-08-24
feat: RTL toggle

iyadthayyil:
### Motivation
The expo app on IOS is always LTR even if the device is RTL, this feature helps paper's example app to display in RTL on IOS devices.

### Issues on Android
- ~~mainDrawer is not filliping, i think it is react-navigation related issue~~
- It doesn't flip to LTR if the device is currently RTL
 

### Test plan
untit11led

### DIFF
--- a/example/App.js
+++ b/example/App.js
@@ -1,8 +1,8 @@
 /* @flow */
 
-import { KeepAwake } from 'expo';
+import { KeepAwake, Util } from 'expo';
 import * as React from 'react';
-import { StatusBar } from 'react-native';
+import { StatusBar, I18nManager, AsyncStorage, Platform } from 'react-native';
 import {
   Provider as PaperProvider,
   DarkTheme,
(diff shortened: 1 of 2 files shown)


ferrannp:
Hi @iyadthayyil, thanks for the PR. I noticed some issues. 

On my Pixel 2 (wrong side):

screenshot_20180828-132234

On iPhone X simulator (first label is truncated):

simulator screen shot - iphone x - 2018-08-28 at 13 22 38

Would you mind taking a look to them?

iyadthayyil:
Thanks @ferrannp  for replaying,
I fixed the first issue, the second one looks weird because it is working fine for me when i tested it on iPhone 8.

ferrannp:
@iyadthayyil mmm we will check. By the way, can you make a rebase or another PR? All the `backdrop` stuff does not feel right here.

jaulz:
I observed the truncated labels already before on iOS but it happened on a random basis and I could not really reproduce it so far. Hence, I think it might not be related to this change...

ferrannp:
Will check truncated labels in a separated issue then. Thanks for heads up @jaulz.

@iyadthayyil It looks like test were broken in `material-next`. I fixed that, if you rebase again, you should have CI ok.

satya164:
Merged

PR #506, opened 2018-08-27
fix: justify content centered in Modals

jaulz:
### Motivation

Since #491 the content is not centered automatically anymore because the children take up 100% and `justifyContent` on the parent container has no effect.

### Test plan

n/a


jaulz:
That was quick, thanks @satya164! :)

PR #507, opened 2018-08-27
fix: allow Modal's background color to be defined via theme

jaulz:
### Motivation

Currently, the `Modal`'s background color is fixed but in certain cases it would be nice to turn off the background or have a different background color. If we move the `Modal`'s background color to the theme anyone is able to adjust it to their needs. 

### Test plan

n/a


src/styles/DefaultTheme.js
```diff
@@ -22,6 +22,10 @@ export default {
       .alpha(0.38)
       .rgb()
       .string(),
+    modal: color(black)
```
satya164: I don't like to have component-specific properties in the theme. This property is useful, but I would prefer it to be generic such as `backdrop` and use it in components that have a backdrop (currently `Modal` and `FABGroup`)

jaulz:
Yep, I like the input. Just did the relevant changes...
Unfortunately, `flow` does not work on this machine but I will later check if flow runs through as well.

jaulz:
@satya164 okay, seems to be fine from my side. Could you please check as well?

ocarreterom:
Many Material components use backdrops. Should it be moved to a component?
Something like [Overlay](https://material.angular.io/cdk/overlay/overview) in Angular Materrial.

Backdrop style reference: https://github.com/angular/material2/blob/4639a874c7d787d4a47f878b28e680d04ad7da62/src/cdk/overlay/_overlay.scss#L9-L13

jaulz:
Isn't that what the Modal is supposed to be?

PR #510, opened 2018-08-28
fix: broken menu button of example app

iyadthayyil:


ferrannp:
@iyadthayyil Can you rebase with `material-next` for broken tests?

PR #523, opened 2018-09-01
Typescript type definitions

### LINKED ISSUE #505: Typescript Type Definitions
### Current behaviour

There is no type definitions file `index.d.ts` shipping with react-native-paper, or no DefinitelyTyped module @types/react-native-paper. This causes TypeScript users to not have type checking for the react-native-paper library.

### Expected behaviour

As a popular library, I would expect the react-native-paper to have type definitions and can work on them if it is wanted by the community.

### Your Environment

| ios or android        | *
| react-native          | *
| react-native-paper    | *
| node                  | *
| npm or yarn           | *

It would be nice if such a nice library had TypeScript type definitions, they also work for flow. I will start working on them when I find time, anybody can help if they want. It will be in the standard format other JS libraries create type definitions, a single .d.ts file containing all the components. 

Note: If the file is not wanted in the library, I can also submit it to DefinitelyTyped, however I think that small libraries containing their .d.ts file instead of having seperate @types/bla-bla libraries is much nicer and doesn't pollute package.json's as much.

Kausta:
### Motivation

The library was missing Typescript Type Definitions as described in #505. This solves that by adding a index.d.ts file based on @doomsower's one and source code flow types/comments.
It would be nice for any typescript users using the library.

### Test plan

There is nothing to test in the Javascript/Runtime part since this type file doesn't change any runtime/add no code. However, when the library updates/props change, they may be needed to be updated in index.d.ts. (Only way to avoid this would be having the library in Typescript and auto generating definitions file, and since it is not, we need to manually update)

Closes #505.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ project.xcworkspace
 .idea
 .gradle
 local.properties
+*.iml
 
 # node.js
 #
@@ -53,3 +54,4 @@ android/keystores/debug.keystore
 # Build
 #
 dist/
+
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,2325 @@
+/**
+ * Typescript type definitions for react-native-paper v1.11.1
+ * Adapted from sources and https://gist.github.com/doomsower/4fe0d7d45e9b2b463377567ae89df1bc
+ * Requires @types/react, @types/react-native
+ */
+
+declare module 'react-native-paper' {
+  import * as React from 'react';
+  import * as ReactNative from 'react-native';
+
+  // Helper for Partial Theme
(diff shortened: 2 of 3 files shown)


Kausta:
Note that circleci failed because I managed to submit the pull request while there is a Minor Service Outage, https://status.circleci.com/

satya164:
Thanks so much for the PR!

Ideally, I would prefer to add the definitions in 2.0, which is the `material-next` branch. There are some API changes there, so the definitions need to be updated. Here is the migration guide listing the changes (WIP): https://github.com/callstack/react-native-paper/wiki/Migration-guide-for-2.0

Really appreciate it if you can help with it :D

Issue #452, opened 2018-07-11
Button with icon in the right side


yaronlevi:
Sometimes it is needed to put the icon in the Button component right to text, instead of left.
For example when using Buttons in a pagination screen (prev, next).

Tried setting `styled={{flexDirection:'row-reverse'}}` via the style prop but the style props is being applied to the outer AnimatedPaper element and not the inner view which sets the direction.

How can make so that the icon will appear right of the text?

gawrysiak:
I've been looking into the guidelines and there's no such examples. If you try to google "material design pagination" and switch to images, it's not how it's normally done as well.

On the other side, your request is pretty basic and there's several ways to add support for it - maybe with an `iconRight` prop or a more flexible API, e.g.
```
<Button>
  <Text>Prev</Text>
  <Icon name="prev" />
</Button>
```
If you look at the official [web material component](https://material.io/develop/web/components/buttons/), it's how they did it.

Should it be done? There's definitely other MD guidelines areas that aren't yet supported at this point and they should be maybe prioritized. But ultimately, it's up to the Callstack guys to decide.

What I normally do in such cases - simply copy the paper's component as your internal component and modify it to suit your needs. Paper grows rapidly and allows more every day, but the possible customization options are almost unlimited, so it will never cover them all.

Noitidart:
I am also trying to get an icon in the right, I do the row-reverse on the `contentStyle` prop but the padding gets messed up:

![image](https://user-images.githubusercontent.com/6372489/63651223-5833d080-c707-11e9-907b-16e39d41258f.png)

Rajan-Savaliya:
contentStyle={{flexDirection: 'row-reverse'}}
 work for me

PR #535, opened 2018-09-06
feat(typography): Implement the Typography component. Fix #532

### LINKED ISSUE #532: Feature request: Typography component
It would be great to be able to have a component for typography, something similar to https://material-ui.com/style/typography/

screen shot 2018-09-05 at 15 29 51

The props in their docs are awesome too:
https://material-ui.com/api/typography/

Thanks so much!

SkyzohKey:
This PullRequest implements the requested Typography component from #532. it borrows a lot of props from the material-ui API so it's simple for people to use it.

### Motivation

Give developers an easier way to declare texts. This way they just specify the variant instead of choosing between `Title`, `Caption`, etc.

### Test plan

I actually didn't tested it because I fuckin' write this Component from inside the Conference room, yay! :tongue:. Please tell me what changes are needed, I'll be proud to implement changes ;)

### DIFF
--- a/src/components/Typography/Typography.js
+++ b/src/components/Typography/Typography.js
@@ -1,0 +1,108 @@
+/* @flow */
+import React, { Component } from 'react'
+
+import { withTheme } from '../../core/theming';
+import type { Theme } from '../../types'
+
+import StyledText from './StyledText';
+
+export type Variant = 'display4' | 'display3' | 'display2' | 'display1' | 'headline' | 'title' | 'subheading' | 'body1' | 'body2' | 'caption'
+export type Align = 'left' | 'center' | 'right' | 'justify'
+
+type Props = {
+  theme: Theme,
+  align?: Align,
+  variant?: Variant,
+  color?: string,
+  gutterBottom?: boolean,
+  headlineMapping?: object,
+  noWrap?: boolean,
+  paragraph?: boolean,
+  style?: any,
+}
+
+export class Typography extends Component<Props> {
+  render () {
+    const {
+      theme, align, color, gutterBottom, headlineMapping,
+      noWrap, paragraph, variant, style, ...rest
+    } = this.props;
+
+    return (
+      <StyledText
+        numberOfLines={noWrap ? null : 1}
+        style={[
+          styles[variant],
+          {
+            fontFamily: theme.fonts.regular,
+            color: color != null ? color : theme.colors.text,
+            textAlign: align != null ? align : 'inherit',
+            marginBottom: gutterBottom || paragraph ? '0.35em' : 'auto',
+          },
+          style
+        ]}
+        {...rest}
+      />
+    )
+  }
+}
+
+export default withTheme(Typography);
+
+const styles = StyleSheet.create({
+  display4: {
+    fontSize: "7rem",
+    fontWeight: 300,
+    lineHeight: "1.14286em",
+    letterSpacing: "-0.04em",
+    marginLeft: "-0.04em",
+  },
+  display3: {
+    fontSize: "3.5rem",
+    fontWeight: 400,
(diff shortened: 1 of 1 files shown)


jaulz:
Looks nice, however, if we go for a `Typography` component at all I would suggest to import the existing components and render them. What do you think?

gawrysiak:
Personally I'd love to see the types featured by [MD guidelines](https://material.io/design/typography/the-type-system.html), with some customization options. So instead of just theming the fonts for regular, medium, light, thin, we should be able to decide if e.g. which font to use for "Subtitle 1", what's the size & color.

satya164:
Hey! Thanks for the PR. As you know, we already have few components for typography. We'd prefer not to add multiple components which do the same thing.

Why do you prefer `variant` prop over those components? I think `<Title>Some title</Title>` is shorter and more readable than `<Typography variant="title">Some title</Typography>`.

Regarding rest of the props, `align`, `color`, `gutterButtom` etc. are re-mapping of some style props which we tend to avoid because people are already faimilar with the style prop. Similar thoughts on the `noWrap` prop which is remapping of `numberOfLines`.

Also, I'm not sure if your code will run, because `em` unit is not supported in React Native AFAIK, and `fontWeight` needs to be a string ('400'). Also we need to make sure if the font weights are actually available by default on supported platforms, or the app will crash (iOS) or look inconsistent (Android).

@gawrysiak for your use case, I think currently the best way is to create your own component which passes some styles to the base component you want to use and use it everywhere.

gawrysiak:
@satya164 that's basically what I did - created a couple of different text components with proper styles for the project. But gets worse when using components with Text dependencies inside. Some examples:
- DrawerItem uses paper's typography, with the font fixed to medium,
- List.Item uses paper's Text component,
- Button uses text with medium,

Font sizes aren't possible to adjust at all, since they're sometimes simply fixed and the font I'm using just needs the letters smaller. For some cases I'm creating a local copy of the component.

If you lookup the "theming" sections, the new MD guidelines encourage designers to play around with the fonts, colors and sizes. I'm not really so much into styling myself, just got caught into the hype. I'm generally handling it somehow, but there's ways to do that cleaner than having a full component copy just to change a single font size. So, not really an issue or request, but if you ever think to extend this somehow, that's a big :+1:

satya164:
@gawrysiak yeah, I'm not sure it's possible to provide that much customization without significant maintenance overhead. Maybe someone can build bindings with something like https://shoutem.github.io/docs/ui-toolkit/theme which will make this use case possible. I'm not a big fan of this approach because it cannot be statically typed, also it has same issues as global CSS.

gawrysiak:
@satya164 As for the fonts, what I've been thinking is to actually simplify the current components. Remove any internal styling for Text in components like Button, List.Item and others and try to completely rely on pre-defined Text/Body1/Body2/Captions etc. Then if we just allow to adjust all the different text components with theming, this should do for 99% cases.

While the new MD allows quite a lot, it's pretty clear on which font component to use. As an example, Chip uses Body2 and that's it. You're allowed to modify Body2, but it's still Body2.
mio-design 2fassets 2f1xkwm2vrtpzva-gm1ln8xqw2eee7tr2uy 2fchipchoice-shrine-type

Could you help me with the link you sent? Turns 404 on me.

gawrysiak:
@satya164 just checked the link - that's a very nice idea. Maybe not even using the whole thing, but a "styles" prop (instead of just container "style") would have covered the visual changes allowed by MD and even more. Tasks like https://github.com/callstack/react-native-paper/issues/452 could have been easily fixed by simply overriding the styles.content. That's worth looking into.

I'll definitely try with the typography RFC. Font family, sizing, letter spacing should be possible. Just not sure if colors are possible to handle per component, but gonna check that as well.

Issue #531, opened 2018-09-05
Feature request: Avatar component


justinlawrence:
It would be great to be able to have a component for avatars, something similar to https://material-ui.com/demos/avatars/

Image avatars:
![screen shot 2018-09-05 at 15 32 16](https://user-images.githubusercontent.com/1521183/45100191-ebec4580-b120-11e8-8bd9-7b7211232bb9.png)

Icon avatars:
![screen shot 2018-09-05 at 15 32 20](https://user-images.githubusercontent.com/1521183/45100222-fc9cbb80-b120-11e8-8a05-14ad3a9e51ec.png)

Letter avatars:
![screen shot 2018-09-05 at 15 32 25](https://user-images.githubusercontent.com/1521183/45100236-032b3300-b121-11e8-85f3-d81e604b9ea8.png)

The props in their docs are great:
https://material-ui.com/api/avatar/

Pleeeeassee :)

Amazing work, everyone!

jaulz:
I created something like that for my application. Maybe I will create a PR for that but otherwise you are of course invited to create a PR as well 😊

gawrysiak:
Merged in [b68fac7](https://github.com/callstack/react-native-paper/commit/b68fac70b0b77a1338ff80ed69e8c1cb293c6337)

PR #536, opened 2018-09-06
feat: Avatar component

jaulz:
### Motivation

Simple avatar component as mentioned in issue #531 

### Test plan

image


satya164:
Maybe instead of `children`, `icon` and `image` props on a single `Avatar` component, maybe it'll be better to have `<Avatar.Text label="JH" />`, `<Avatar.Icon icon="folder" />` and `<Avatar.Image source={require('./user.png')} />`. Because otherwise it's not obvious what will happen when someone passes both `icon` and `image` etc.

jaulz:
@satya164 thanks for the input. Can you please check again? 😊

nahoc-petal:
Thank you! This looks really good.

nahoc-petal:
Would it be possible to implement a border prop or elevation shadow?

jaulz:
Should be possible via the style prop I guess. Or which props with which effect do you have in mind?

nahoc-petal:
You're right, I didn't even think of the style prop. This has been a long day...

BenWildeman:
> Maybe instead of `children`, `icon` and `image` props on a single `Avatar` component, maybe it'll be better to have `<Avatar.Text label="JH" />`, `<Avatar.Icon icon="folder" />` and `<Avatar.Image source={require('./user.png')} />`. Because otherwise it's not obvious what will happen when someone passes both `icon` and `image` etc.

I would argue that it would make sense that the avatar component should have a text prop as a fallback for when images don't render for whatever reason instead of this turning blank

aqeelbe:
Please resolve the conflicts and  merge this feature . we need this feature in our app.

jaulz:
@ferrannp I updated the types and now all checks passed 😊

jaulz:
@ferrannp any updates on this?

jaulz:
@Trancever please check now

Trancever:
It looks good to me. @satya164 Could you take a look and check api?

brunohkbx:
@satya164 could you please take a look?

satya164:
Merged in b68fac70b0b77a1338ff80ed69e8c1cb293c6337

PR #539, opened 2018-09-10
Feat: Badge component

nahoc-petal:
### Motivation

Simple badge component, can be used alone, added on text, button or icon button.

### Test plan

https://material.angular.io/components/badge/examples


jaulz:
My feeling is that we should not enhance the `Button` component with new properties but instead use something like this, i.e. use the children prop:
```javascript
<Badge position="top-right" color={colors.accent} content={26}>
  <Button>Notifications</Button>
</Badge>
```

Note that I also renamed "count" to "content" since we are not always dealing with numbers. What do you think?

PR #542, opened 2018-09-11
docs: Update RadioButton docs with new version 2.0.1

Moreno97:
### Motivation

Update documentation with new version `2.0.1` of `react-native-paper`. RadioButton props are quite different from version 1 to 2, so docs needed an update.
### Test plan

See docs.

### DIFF
--- a/src/components/RadioButton.js
+++ b/src/components/RadioButton.js
@@ -79,12 +79,12 @@ type Props = {
  *       <View>
  *         <RadioButton
  *           value="first"
- *           status={checked === 'first' && 'checked'}
+ *           status={checked === 'first' ? 'checked' : 'unchecked'}
  *           onPress={() => { this.setState({ checked: 'first' }); }}
  *         />
  *         <RadioButton
  *           value="second"
- *           status={checked === 'second' && 'checked'}
+ *           status={checked === 'second' ? 'checked' : 'unchecked'}
  *           onPress={() => { this.setState({ checked: 'second' }); }}
  *         />
  *       </View>
--- a/src/components/RadioButton.js
+++ b/src/components/RadioButton.js
@@ -79,13 +79,13 @@ type Props = {
  *       <View>
  *         <RadioButton
  *           value="first"
- *           checked={checked === 'first'}
- *           onPress={() => { this.setState({ checked: 'firstOption' }); }}
+ *           status={checked === 'first' && 'checked'}
+ *           onPress={() => { this.setState({ checked: 'first' }); }}
  *         />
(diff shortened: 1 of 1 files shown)


src/components/RadioButton.js
```diff
@@ -79,13 +79,13 @@ type Props = {
...
- *           checked={checked === 'first'}
- *           onPress={() => { this.setState({ checked: 'firstOption' }); }}
+ *           status={checked === 'first' && 'checked'}
+ *           onPress={() => { this.setState({ checked: 'first' }); }}
  *         />
  *         <RadioButton
  *           value="second"
- *           checked={checked === 'second'}
- *           onPress={() => { this.setState({ checked: 'secondOption' }); }}
+ *           checked={checked === 'second' && 'unchecked'}
```
alvaromb: `checked` should be `status`.

alvaromb: In case `checked === 'second'` evaluates to `true`, you'll end up with an `'unchecked'` status. 

Should be changed to:

```js
checked={checked === 'second' && 'checked'}
```

src/components/RadioButton.js
```diff
@@ -79,13 +79,13 @@ type Props = {
  *       <View>
  *         <RadioButton
  *           value="first"
- *           checked={checked === 'first'}
- *           onPress={() => { this.setState({ checked: 'firstOption' }); }}
+ *           status={checked === 'first' && 'checked'}
```
satya164: This needs to be `status={checked === 'first' ? 'checked' : 'unchecked'}`

src/components/RadioButton.js
```diff
@@ -79,13 +79,13 @@ type Props = {
...
+ *           onPress={() => { this.setState({ checked: 'first' }); }}
  *         />
  *         <RadioButton
  *           value="second"
- *           checked={checked === 'second'}
- *           onPress={() => { this.setState({ checked: 'secondOption' }); }}
+ *           status={checked === 'second' && 'checked'}
```
satya164: This needs to be `status={checked === 'second' ? 'checked' : 'unchecked'}`

alvaromb: @satya164 it does read better, but there's no `'unchecked'` use in the actual source code, see https://github.com/callstack/react-native-paper/blob/master/src/components/RadioButtonAndroid.js#L111

Current PR would work, even without `'unchecked'`, dunno about flow typing though.

Moreno97: I pushed new changes, now is like @satya164 code because of better reading.

Issue #354, opened 2018-05-08
Custom input inside TextInput


andreluizbellon:
### Current behaviour
There's no way to pass a custom input to be rendered inside TextInput.
### Expected behaviour
There should be a way to pass a custom input to be rendered, for ex: `react-native-text-input-mask`
### Code sample
```javascript
<TextInput
        label='Email'
        value={this.state.text}
        onChangeText={text => this.setState({ text })} 
        customTextInput={CustomInput} />
```

PR #547, opened 2018-09-14
feat: add custom render to TextInput component

Esemesek:
### Motivation
Fix for #354 .

### Test plan

```javascript
  <View>
    <TextInput
      render={props => (
        <TextInputMask {...props} mask="+[00] [000] [000] [000]" />
      )}
      placeholder="+12 345 678 910"
    />
    <TextInput placeholder="Normal text input" />
  </View>
```
ezgif com-video-to-gif

Sorry for the poor quality gif. First time actually doing this C:

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -24,6 +24,22 @@ const LABEL_WIGGLE_X_OFFSET = 4;
 const FOCUS_ANIMATION_DURATION = 150;
 const BLUR_ANIMATION_DURATION = 180;
 
+type RenderProps = {
+  ref: NativeTextInput => void,
+  onChangeText: string => void,
+  placeholder: ?string,
+  placeholderTextColor: string,
+  editable?: boolean,
+  selectionColor: string,
+  onFocus: () => mixed,
+  onBlur: () => mixed,
+  underlineColorAndroid: string,
+  style: any,
+  multiline?: boolean,
+  numberOfLines?: number,
+  value?: string,
+};
+
 type Props = {
   /**
    * Mode of the TextInput.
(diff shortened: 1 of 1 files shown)


src/components/TextInput.js
```diff
@@ -73,6 +73,13 @@ type Props = {
    * Callback that is called when the text input is blurred.
    */
   onBlur?: () => mixed,
+  /**
+   * @optional
```
satya164: Don't need optional here

src/components/TextInput.js
```diff
@@ -73,6 +73,13 @@ type Props = {
...
+   * Custom render function to replace rendering native TextInput.
+   */
+  render: <T: React.ElementProps<NativeTextInput>>(
```
satya164: I didn't know about this: `React.ElementProps<NativeTextInput>`. We should also use it like `type Props = React.ElementProps<NativeTextInput> & { ... }`, and in more components which wrap native ones. But in separate PR.

src/components/TextInput.js
```diff
@@ -73,6 +73,13 @@ type Props = {
    * Callback that is called when the text input is blurred.
    */
   onBlur?: () => mixed,
+  /**
+   * @optional
+   * Custom render function to replace rendering native TextInput.
```
satya164: Maybe something like this will be clearer:

  Callback to render a custom input component such as `react-native-text-input-mask` instead of the default `TextInput` component from `react-native`.

  Example:

  ```js
  <TextInput
    label="Email"
    value={this.state.text}
    render={props =>
      <TextInputMask
        {...props}
        mask="+[00] [000] [000] [000]"
      />
    }
  />
  ```

satya164:
Merged in 768094119c38b5a37261e84e06faaf1ef11c77bc

PR #543, opened 2018-09-12
feat: allow configuring fontSize of TextInput label

vieiralucas:
### Motivation
This PR aims to allow users to customize the fontSize of a `TextInput`'s label. Should FIX #348 

### Test plan

I added a TextInput to the example app showing this feature.

### Why this PR is Work In Progress?
I know this implementation has a problem because `react-native-paper` currently sends the `style` prop to a `View` component which is not supposed to receive `fontSize`.
So I don't know how to overcome this issue, do you have any suggestions?

Thanks ❤️

PR #552, opened 2018-09-18
feat: Ability to adjust the button height

### LINKED ISSUE #352: Custom height for Button
When i set custom button height with 
```
        <Button
            style={{
              marginTop: 20,
              height: 50,
              marginLeft: 0,
              marginRight: 0,
              justifyContent: 'center',
            }}
            loading={loading}
            raised
            compact
            primary
            onPress={this.onSubmit}
          >
            {t('signIn')}
          </Button>
```

i see on touch start

screen shot 2018-05-07 at 3

is there any way to do 
Hover effect all over button ?

gawrysiak:
### Motivation

Another way to fix #352. Uses the TextInput font size approach (#543). Takes the container height and uses it for the content view.

It's not really a clean way to handle customization, but on the other side there's no point in using the height style attribute otherwise. So I'd like to check if this one goes through.

### Test plan

Updated the snapshot tests. `yarn test` works fine.

Sample button:

```
<Button onPress={() => {}} style={{ height: 80 }}>
  Default
</Button>
```

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -239,7 +239,12 @@ class Button extends React.Component<Props, State> {
           rippleColor={rippleColor}
           style={touchableStyle}
         >
-          <View style={styles.content}>
+          <View
+            style={[
+              styles.content,
+              { height: style && style.height ? style.height : 'auto' },
+            ]}
+          >
             {icon && loading !== true ? (
               <View style={styles.icon}>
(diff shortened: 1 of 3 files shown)


Trancever:
@gawrysiak It looks like content of the button is not vertically aligned with this approach, am I right?

PR #565, opened 2018-09-30
fix: searchbar RTL support

iyadthayyil:
### Motivation
searchbar RTL support
whatsapp image 2018-09-30 at 5 35 18 pm


satya164:
there should be some padding in the beginning before the placeholder, no?

iyadthayyil:
Yup, it should be like this

whatsapp image 2018-09-30 at 9 34 48 pm

ferrannp:
@iyadthayyil is this ready?

Issue #124, opened 2017-07-24
Toggle buttons


ferrannp:
https://material.io/design/components/buttons.html#toggle-button

alvaromb:
@Moreno97 do you feel confident for tackling this one?

Moreno97:
> @Moreno97 do you feel confident for tackling this one?

Yeah, already working on it @ferrannp

PR #575, opened 2018-10-05
feat: ToggleButton component

Moreno97:
### Motivation

Implement Toggle Button component to the library, specified on the Material Design guidelines. Solves #124 

### Test plan

Go to Examples List and select `Toggle Button` to test the new feature.


alvaromb:
@Moreno97 fix ts tests/

jaulz:
Great, that's definitely something which is missing.
Not sure if it's feasible but would it maybe make sense to reuse the `Button` and just pass different styles depending on whether it's toggled or not? 
And regarding the `Group` it might be good to follow the same approach as in `RadioButtonGroup` (i.e., use a context).

alvaromb:
@jaulz we had a discussion about using context and we found a bug with the current implementation of the Radio button (which we're going to report today). First time I looked into the Radio button implementation seemed odd the use of the context for that component. After testing some edge cases we found out that, when defining several groups, all of them gets rendered when only one changes. This is due to the nature of context, which sends updates to all of the components that are consuming it.

This is the reason why we favoured another implementation of the group with a render prop.

satya164:
> Not sure if it's feasible but would it maybe make sense to reuse the Button

Probably `IconButton`

> And regarding the Group it might be good to follow the same approach as in RadioButtonGroup (i.e., use a context).

👍

> After testing some edge cases we found out that, when defining several groups, all of them gets rendered when only one changes

Can you give an example? You mean all radio buttons nested inside a single group?

alvaromb:
I'm preparing an example with a repro of the issue, will send you the code later @satya164.

alvaromb:
Hi @satya164.

I have the reproduction of the issue ready to upload to Github, [here you have the repo](https://github.com/alvaromb/paper-radio-issue). However, I have mixed feelings about it because it's kinda hacky/edge case. Let me explain you first why we reverted our first implementation with context of the toggle group button.

This is an excerpt of the React docs about Context:

> (...) sometimes the same data needs to be accessible by many components in the tree, and at different nesting levels. Context lets you “broadcast” such data, and changes to it, to all components below. Common examples where using context might be simpler than the alternatives include managing the current locale, theme, or a data cache.

The first statement makes it clear that Context should be used to disseminate data into several components around the app, not for passing props to the most inner children. This is what the current implementation of `RadioButton.Group` is doing.

Besides of this theoretical or philosophical issue, the grouping works because the `Context.Consumer` gets the info from the closest `Context.Provider` in the tree, as per context spec. Anytime, for any reasons whatsoever, that you end up having unrelated Radio Buttons or Toggle Buttons inside a grouped buttons, those buttons will also trigger context updates. This can be solved by always wrapping the groups into the `Grouped` components, but IMHO we're not using the context API properly if we follow strictly what the docs says about it. I can only see future problems for very complex forms with the current approach.

For all of this, I suggested @Moreno97 to change the context implementation of grouped toggle buttons to a render prop, which seems more correct from my point of view. Also, detaching context from the implementation solves potential issues that could arise in the future.

If you think we should keep the context implementation, we will revert it to be similar to the actual radio button implementation. If not, I'll provide a new PR with a switch from context to render prop in the radio button.

satya164:
It's highly unlikely that somebody would have unrelated radio/toggle buttons inside a group, and even if they have, it's expected behavior that they'll belong to the group, so I wouldn't worry about it.

Regardless, it's better if we keep the same API across components for familiarity. So it'll be better to use context for this too.

alvaromb:
@Moreno97 please insert again context into Toggle Group and drop render prop.

alvaromb:
I'm curious @satya164, is there any reason that I missed out that you guys used context for this component grouping?

satya164:
because using context made the API significantly simpler to use. many other libraries also provide a similar API, such as https://material-ui.com/demos/selection-controls/

Moreno97:
@satya164 changed to Context API, feel free to merge.

alvaromb:
Thanks for the reply @satya164. I'm still not convinced about context here (MUI's doesn't use it after a quick search in the repo), but we're happy to follow your suggestion. Ping me if you need further help with this PR.

jaulz:
> `IconButton` is not more customizable than `TouchableRipple`. Probably the best option is `Button`, but is causing problems with children style, so I think `TouchableRipple` is fine.

@Moreno97 what do you mean in detail? My feeling is that the code in `ToggleButton` is more or less a duplicate of `Button` which would make it more difficult to keep both in sync in the long term. Can we somehow extend `Button` to accomodate these changes?

Moreno97:
> > `IconButton` is not more customizable than `TouchableRipple`. Probably the best option is `Button`, but is causing problems with children style, so I think `TouchableRipple` is fine.
> 
> @Moreno97 what do you mean in detail? My feeling is that the code in `ToggleButton` is more or less a duplicate of `Button` which would make it more difficult to keep both in sync in the long term. Can we somehow extend `Button` to accomodate these changes?

You're right, is more or less a duplicate, but`Button` has different styles that with the current API it is impossible to customize `ToggleButton` to fit correctly.

jaulz:
@Moreno97 what I did not see before is that the examples at https://material.io/design/components/buttons.html#toggle-button only show icons and no labels for `ToggleButtons`. Also the anatomy section only specifies icons: https://material.io/design/components/buttons.html#anatomy. Hence, @satya164 is right and we could use the `IconButton` which is already a quite flat component which we could adapt via `style` prop. What do you think?

Moreno97:
> @Moreno97 what I did not see before is that the examples at https://material.io/design/components/buttons.html#toggle-button only show icons and no labels for `ToggleButtons`. Also the anatomy section only specifies icons: https://material.io/design/components/buttons.html#anatomy. Hence, @satya164 is right and we could use the `IconButton` which is already a quite flat component which we could adapt via `style` prop. What do you think?

So, you mean to change `TouchableRipple` for `IconButton`? The purpose to use label was for customize more the Touchable, but ok, I can change that and use only children 👍

jaulz:
Not sure what you mean... my point was that we could use `IconButton` since the specification anyway specifies only icons so we do not need labels at all. At the same we can reduce duplications. By taking a quick glance at `IconButton` it seems that we can pass the `style` prop and adjust the styles to the `ToggleButton` needs.

Moreno97:
> Not sure what you mean... my point was that we could use `IconButton` since the specification anyway specifies only icons so we do not need labels at all. At the same we can reduce duplications. By taking a quick glance at `IconButton` it seems that we can pass the `style` prop and adjust the styles to the `ToggleButton` needs.

Ok, I understand what are you saying about the label, but what I'm saying is how can we achieve the custom toggle button with images like material design specification (see the image from my upper comment). With IconButton is not possible, that's why I decided to use `TouchableRipple`.

satya164:
> how can we achieve the custom toggle button with images like material design specification
> With IconButton is not possible

Why isn't it possible with `IconButton`? It's just 2 icons in 2 states, no?

Moreno97:
@satya164 @jaulz changed to `IconButton`, review please.

Moreno97:
@satya164 @jaulz changed to IconButton. Review, please.

src/components/ToggleButton.js
```diff
@@ -0,0 +1,208 @@
...
+      <ToggleGroupContext.Consumer>
+        {(context: ?{ value: string, onValueChange: Function }) => {
+          const checked: ?boolean =
+            (context && context.value === value) || selected;
+
+          if (disabled) {
+            backgroundColor = white;
+          } else if (checked) {
+            textColor = white;
+            backgroundColor = colors.primary;
```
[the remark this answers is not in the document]
Moreno97: Shouldn´t be `textColor`?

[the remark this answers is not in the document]
jaulz: Basically yes. Please also keep in mind that there is a dark mode and hence we cannot always use `white`. You can see something similar here: https://github.com/callstack/react-native-paper/blob/master/src/components/Button.js#L193

src/components/ToggleGroup.js
```diff
@@ -0,0 +1,77 @@
+/* @flow */
+
+import * as React from 'react';
+import createReactContext, { type Context } from 'create-react-context';
+
+type Props = {
```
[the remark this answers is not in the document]
jaulz: Right, good catch :-)

Moreno97:
@jaulz new changes, review please! :)

src/components/ToggleButton.js
```diff
@@ -0,0 +1,203 @@
...
+              .string();
+          } else if (buttonColor) {
+            textColor = buttonColor;
+          }
+
+          return (
+            <IconButton
+              icon={icon}
+              onPress={() => {
+                if (typeof this.props.onPress === 'function') {
```
jaulz: I think we can change it to:
```javascript
this.props.onPress && this.props.onPress(status);
```

example/src/ToggleButtonExample.js
```diff
@@ -0,0 +1,170 @@
...
+  render() {
+    const {
+      theme: {
+        colors: { background },
+      },
+    } = this.props;
+    return (
+      <View style={[styles.container, { backgroundColor: background }]}>
+        <View style={styles.toggleGroup}>
+          <ToggleButton.Group
```
jaulz: Can we maybe change this example to non-exclusive? Then we can show both possibilities here.

Moreno97: Yo mean include an example with no groupped toggle button?

jaulz: Yep 😊

src/components/ToggleButton.js
```diff
@@ -0,0 +1,203 @@
...
+                  this.props.onPress(status);
+                }
+                return checked
```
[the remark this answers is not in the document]
Moreno97: Need to think about that change, because is not working for some strange reason...

src/components/ToggleButton.js
```diff
@@ -0,0 +1,203 @@
...
+  /**
+   * Color of the toggle button.
+   */
+  color?: string,
+  /**
+   * @optional
+   */
+  theme: Theme,
+  /**
+   * Whether the button is disabled.
```
jaulz: Just a minor thing: could you please write `ToggleButton` instead of "button"?

src/components/ToggleButton.js
```diff
@@ -0,0 +1,203 @@
...
+   */
+  disabled?: boolean,
+  /**
+   * Accessibility label for the button. This is read by the screen reader when the user taps the button.
+   */
+  accessibilityLabel?: string,
+  /**
+   * Function to execute on press toggle button.
```
jaulz: Just a minor thing: could you please write `ToggleButton` instead of "toggle button"?

src/components/ToggleButton.js
```diff
@@ -0,0 +1,203 @@
...
+   */
+  onPress?: (value: ?string) => mixed,
+  /**
+   * Value of the toggle button.
```
jaulz: Just a minor thing: could you please write `ToggleButton` instead of "toggle button"?

src/components/ToggleButton.js
```diff
@@ -0,0 +1,203 @@
...
+   */
+  value?: string,
+  /**
+   * The actual button that is selected from the ToggleButton.Group
```
jaulz: This description needs to be adjusted to the new `status` property.

src/components/ToggleButton.js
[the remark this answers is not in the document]
jaulz: No worries... we are almost there 😊

satya164:
I'll review it next week.

src/components/ToggleButton.js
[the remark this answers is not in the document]
jaulz: Hm, I think the representation here looks a bit confusing. Anyway, I think you got it almost now. Though the negation of `checked` is not correct, is it?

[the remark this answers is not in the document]
Moreno97: Yeah, it is. Is working fine, here we should change the value to `unchecked` if the ToggleButton state is `checked`.

Moreno97:
@jaulz discussions resolved!

jaulz:
@Moreno97 great, thanks, I like it! Unfortunately, I will only be able to test it by next week but in the meantime I did a few cosmetics.
@satya164 I will let you know once I tested it from a functional perspective so you can have a second look as well 😊

jaulz:
@Moreno97 I just added some more contrast and moved the component to a sub folder (since it consists of two components). From my side everything looks fine and I look forward to the feedback from @satya164 to see it being merged soon 😊

satya164:
Sorry, it's holidays season, so I haven't gotten much time to review. Will review it soon.

satya164:
I pushed the final changes here: https://github.com/callstack/react-native-paper/tree/pr/575

Let's add typescript definitions and then I can merge.

jaulz:
I added types in this commit https://github.com/callstack/react-native-paper/commit/c22b42a26bdb6471499eae3dcc2759610d069326 in the same branch. 
@Moreno97 can you please merge the branch with this PR?

Moreno97:
> I added types in this commit [c22b42a](https://github.com/callstack/react-native-paper/commit/c22b42a26bdb6471499eae3dcc2759610d069326) in the same branch.
> @Moreno97 can you please merge the branch with this PR?

Sure! In having time I will

jaulz:
@satya164 changes are merged into this PR :blush:

Moreno97:
Thanks @jaulz !!!

PR #579, opened 2018-10-06
fix: Fix component types

### LINKED ISSUE #578: Typescript - having react-native base component props generates compiling issues
### Current behaviour

Using base props from react-native-paper components are not compiling as they don't  exist in the current definition

```
[ts] Property 'keyboardType' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes & Readonly<{ children?: ReactNode; }> & Readonly'.
```

### Expected behaviour

It should compile with components having base props set with react-native-paper as defined in the spec

### Code sample

```
import { TextInput } from 'react-native-paper';

  state = {
    text: ''
  };

  render(){
    return (
      <TextInput
        label='Email'
        value={this.state.text}
        autoCapitalize="none"
        keyboardType="number-pad"
      />
    );
  }
}
```

Added autoCapitalize and keyboardType props that are available in react-native TextInput component

### What have you tried

``` TextInput.d.ts
import { TextInputProps } from "@types/react-native" // added this line
import { ThemeShape } from '../types';

export interface PaperTextInputProps extends TextInputProps { // added extends TextInputProps and renamed interface
  mode?: 'flat' | 'outlined';
  disabled?: boolean;
  label?: string;
  placeholder?: string;
  error?: boolean;
  onChangeText?: Function;
  underlineColor?: string;
  multiline?: boolean;
  numberOfLines?: number;
  onFocus?: () => any;
  onBlur?: () => any;
  render?: (props: RenderProps) => React.ReactNode;
  value?: string;
  style?: any;
  theme?: ThemeShape;
}

// TODO: Adjust this types to Typescript
export interface RenderProps {
  ref: (ref: any) => void;
  onChangeText: (text: string) => void;
  placeholder?: string;
  placeholderTextColor: string;
  editable?: boolean;
  selectionColor: string;
  onFocus: () => any;
  onBlur: () => any;
  underlineColorAndroid: string;
  style: any;
  multiline?: boolean;
  numberOfLines?: number;
  value?: string;
}

export declare class TextInput extends React.Component<PaperTextInputProps> {} // exporting renamed interface
```
this is working as I have devDependencies @types/react-native but I'm not sure this is a viable solution otherwise

### Your Environment

software | version
-- | --
ios or android | both
react-native | 0.57.2
react-native-paper | 2.1.1
node | 10.11.0
npm or yarn | 6.4.1
typescript | 3.1.1
tsc | 3.0.3

Esemesek:
Fixes #578

### DIFF
--- a/typings/components/TextInput.d.ts
+++ b/typings/components/TextInput.d.ts
@@ -1,13 +1,14 @@
 import * as React from 'react';
+import { TextInputProps as NativeTextInputProps } from 'react-native';
 import { ThemeShape } from '../types';
 
-export interface TextInputProps {
+export interface TextInputProps extends NativeTextInputProps {
   mode?: 'flat' | 'outlined';
   disabled?: boolean;
   label?: string;
   placeholder?: string;
   error?: boolean;
-  onChangeText?: Function;
+  onChangeText?: (text: string) => void;
   underlineColor?: string;
   multiline?: boolean;
   numberOfLines?: number;
@@ -19,8 +20,7 @@ export interface TextInputProps {
   theme?: ThemeShape;
 }
 
-// TODO: Adjust this types to Typescript
-export interface RenderProps {
+export interface RenderProps extends NativeTextInputProps {
   ref: (ref: any) => void;
   onChangeText: (text: string) => void;
   placeholder?: string;
--- a/typings/components/Typography.d.ts
+++ b/typings/components/Typography.d.ts
@@ -1,7 +1,8 @@
 import * as React from 'react';
+import { TextProps } from 'react-native';
 import { ThemeShape } from '../types';
 
-interface TypographyProps {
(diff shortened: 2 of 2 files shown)


typings/components/TextInput.d.ts
```diff
@@ -1,13 +1,14 @@
...
 
-export interface TextInputProps {
+export interface TextInputProps extends NativeTextInputProps {
   mode?: 'flat' | 'outlined';
   disabled?: boolean;
   label?: string;
   placeholder?: string;
   error?: boolean;
-  onChangeText?: Function;
+  onChangeText?: (text: string) => void;
```
satya164: Will this pass if user does something like:

```js
onChangeText = async () => {
  // do some request
}
```

Since it returns promise and not `void`? In Flow, I usually type them as `mixed` to avoid that.

PR #580, opened 2018-10-07
chore: Update expo to latest version.

Trancever:
This PR upgrades expo in example app to latest version.

Actually, there is no issue for that, but It's good to use latest stable version of libraries.


Trancever:
Expo 30 still uses react-native version 0.55.4

PR #598, opened 2018-10-13
fix: Always show error outline when including when unfocused (#595)

### LINKED ISSUE #595: TextInput error not red on blur
TextInput with error true has red borders only if its focused

### Current behaviour

No red borders on blur

### Expected behaviour
If error passed it should be always with red borders

### Code sample

```
<View>
          <TextInput
            value=" "
            error
            mode="outlined"
            placeholder=" "
            label="Test"
            style={{ backgroundColor: '#ffffff', paddingLeft: 10 }}
          />
          <HelperText type="error" visible>
            Email address is invalid!
          </HelperText>
        </View>
````

screen shot 2018-10-12 at 6 16 59 pm
screen shot 2018-10-12 at 6 16 52 pm

### Your Environment

| react-native-paper    | 2.1.2

### PR DESCRIPTION
**Pull request by Aidurber**
### Motivation
TextInput with outline in an error state should correctly indicate an error when unfocused. This would also close #595 

### Test plan

It could also be possible to add a different prop `keepOutline` (not 100% on the name) instead of relying on the error state. However I'm unsure how that might be useful. 

focused-cropped
unfocused-cropped

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -484,7 +484,8 @@ class TextInput extends React.Component<Props, State> {
               {
                 borderRadius: theme.roundness,
                 borderWidth: this.state.focused ? 2 : 1,
-                borderColor: this.state.focused ? activeColor : outlineColor,
+                borderColor:
+                  this.state.focused || error ? activeColor : outlineColor,
               },
             ]}
           />
@@ -581,7 +582,7 @@ class TextInput extends React.Component<Props, State> {
                   color: activeColor,
                   opacity: this.state.labeled.interpolate({
                     inputRange: [0, 1],
-                    outputRange: [this.state.focused ? 1 : 0, 0],
+                    outputRange: [this.state.focused || error ? 1 : 0, 0],
                   }),
                 },
               ]}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Aidurber** [comment]
Looking at the material design spec for outline text inputs this is the correct functionality: https://material.io/design/components/text-fields.html#outlined-text-field

File: src/components/TextInput.js [hunk 1]
```diff
@@ -484,7 +484,8 @@ class TextInput extends React.Component<Props, State> {
               {
                 borderRadius: theme.roundness,
                 borderWidth: this.state.focused ? 2 : 1,
```
satya164: I think it should also have the `borderWidth` of `2` when it has an error?

src/components/TextInput.js
```diff
@@ -484,7 +484,8 @@ class TextInput extends React.Component<Props, State> {
...
-                borderColor: this.state.focused ? activeColor : outlineColor,
+                borderColor:
+                  this.state.focused || error ? activeColor : outlineColor,
```
satya164: Since we need the `this.state.focused || error` check everywhere, can you move it to a variable `hasActiveOutline` or something and use it instead?

Issue #606, opened 2018-10-17
Portal component doesn't crash app


denisviklov:
### Current behaviour

Throw an error:
TypeError: Cannot read property 'mount' of null

This error is located at:
    in PortalConsumer (at Portal.js:53)
    in Portal (created by Context.Consumer)
    in ThemeProvider (created by Context.Consumer)
    in withTheme(Portal) (at topbar/index.js:18)
    in RCTView (at View.js:44)
    in Topbar (at src/index.js:23)
    in App (at renderApplication.js:34)
    in RCTView (at View.js:44)
    in RCTView (at View.js:44)
    in AppContainer (at renderApplication.js:33)
PortalConsumer.componentDidMount
    PortalConsumer.js:14:35
PortalConsumer.proxiedComponentDidMount
    createPrototypeProxy.js:61:39
commitLifeCycles
    ReactNativeRenderer-dev.js:12144:19
commitAllLifeCycles
    ReactNativeRenderer-dev.js:13495:6
Object.invokeGuardedCallbackImpl
    ReactNativeRenderer-dev.js:92:9
invokeGuardedCallback
    ReactNativeRenderer-dev.js:306:30
commitRoot
    ReactNativeRenderer-dev.js:13701:6
completeRoot
    ReactNativeRenderer-dev.js:15269:2
performWorkOnRoot
    ReactNativeRenderer-dev.js:15197:8
performWork
    ReactNativeRenderer-dev.js:15090:6

### Expected behaviour

Work

### Code sample

import * as React from 'react';
import { Appbar, Modal, Portal, Text, Button, Dialog, Paragraph } from 'react-native-paper';
import { StyleSheet, View } from 'react-native';

export default class MyComponent extends React.Component {
  state = {
    visible: false,
  };

  _showDialog = () => this.setState({ visible: true });

  _hideDialog = () => this.setState({ visible: false });

  render() {
    return (
      <View>
        <Button onPress={this._showDialog}>Show Dialog</Button>
        <Portal>
          <Dialog
             visible={this.state.visible}
             onDismiss={this._hideDialog}>
            <Dialog.Title>Alert</Dialog.Title>
            <Dialog.Content>
              <Paragraph>This is simple dialog</Paragraph>
            </Dialog.Content>
            <Dialog.Actions>
              <Button onPress={this._hideDialog}>Done</Button>
            </Dialog.Actions>
          </Dialog>
        </Portal>
      </View>
    );
  }
}

const styles = StyleSheet.create({
[...]

PR #607, opened 2018-10-17
fix: dialog example.

Trancever:
Fix example usage of Dialog component. #606


satya164:
I think it doesn't make much sense to have `<Provider />` in the same component as `<Portal />`. I think we should:

1. Show a better error message with a link to the docs if we don't find a `<PortalHost />`
2. In the script which adds "Open in Snack" button, add a regex to remove `<Portal />` from the code, so it doesn't give error or Snack

PR #614, opened 2018-10-19
fix: snackbar not actually rendering when visible=true during mounting

andreibarabas:
The Snackbar was not actually displayed if during mounting the visible property was set to true, mostly because the _show function was never called to start the animation (componentDidUpdate is not called on the initial render).

To prove it, simply run the updated test which now mock the Animated module to finalise the animation before asserting it's visibility, and you will see the previous snapshots are still valid, even though they mention that the opacity of the view is 0, thus transparent.

### Motivation

Currently, the example you have provided works because the snack bar is mounted with visible = false and it's property changes once you hit the button. But in a real life scenario, you could receive the property from a redux store instead of a button, and there you can have it visible = true.

### Test plan

- Simply set the property to visible = true and see that now the Snackbar appears and will disappear after duration is hit.

### DIFF
--- a/src/components/Snackbar.js
+++ b/src/components/Snackbar.js
@@ -128,6 +128,12 @@ class Snackbar extends React.Component<Props, State> {
     hidden: !this.props.visible,
   };
 
+  componentDidMount() {
+    if (this.props.visible) {
+      this._show();
+    }
+  }
+
   componentDidUpdate(prevProps) {
     if (prevProps.visible !== this.props.visible) {
       this._toggle();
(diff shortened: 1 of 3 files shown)


Trancever:
Hi @andreibarabas Thanks for the PR. Actually, we just merged similar PR that solves same problem yesterday, but your's adds additional test and it would be good to also have it merged. Could you resolve conflicts (keep your version of fix)?

satya164:
Merged in 8764f60370e0adfd0c07705e1cebdaf0f7e1459b

andreibarabas:
Sorry @Trancever, I was busy at work the last 2 weeks with a major release and didn't have time to do it. Thanks @satya164 for taking care of the merge.

PR #615, opened 2018-10-21
fix: ts declaration of TouchableRipple

doomsower:
### Motivation
Sometimes in my app I need to use raw TouchableRipple. However, typescript compilation fails, becuase `hitSlop` and `onLongPress` are missing. Looking at TouchableRipple source code and TouchableRipple typescript definitions, I notice that the component passes `...rest` props to underlying native components. But all these props are totally ignored in typescript definitions.

Here I'm only fixing what breaks my build, but the same situation is with Button, and (I guess) with all the other components that pass down `...rest` props. Fixing all of them is a tedious work...

### Test plan

I now use my fork of `react-native-paper` in my project, and typescript doesn't complain about touchable ripple with `hitSlop` anymore.


typings/components/TouchableRipple.d.ts
```diff
@@ -1,20 +1,19 @@
...
-  style?: any;
   theme?: ThemeShape;
 }
 
 export declare class TouchableRipple extends React.Component<
   TouchableRippleProps
 > {
   static supported: boolean;
-}
+}
```
Esemesek: Add EOF

doomsower:
> Can you check and correct other types that requires extending from react-native equivalent?

I posted my definitions for v1 somewhere here in issues, I had the extensions there. I'm afraid I don't want to do this again for v2, this is quite tedious. But If I encounter something like this in my project again, I would submit another small PR.

Issue #602, opened 2018-10-16
How to add badge in App Drawer? 


ManigandanRaamanathan:
### Current behaviour

How to add badge in App Drawer? Does React Native Paper has their own badge component? Else how to add custom badge? I'm following your snack example for the appdrawer

raajnadar:
This page is to track issues you can ask this types of question in stackoverflow or quora!

ManigandanRaamanathan:
@raajnadar  This feature is not even available in React Native Paper yet. This is more of a new required enhancement or additional support as about this feature is not even be found in their documentation nor known to others in stackoverflow or quora.

ferrannp:
Probably good to be done after #620.

raajnadar:
Here we go! https://callstack.github.io/react-native-paper/badge.html

Issue #358, opened 2018-05-10
Drawer: Redesign


satya164:
- [x] Change selected state drawer item

![image](https://user-images.githubusercontent.com/1174278/39866026-3936ed88-544f-11e8-91b8-a615947d5be5.png)

https://material.io/design/components/navigation-drawer.html

jaulz:
Can we reintroduce the color property? In my case I have a color coding (e.g., green for done etc.) and hence it was perfect to represent the color also in the drawer.

gawrysiak:
@jaulz, it's still possible to customize the color which is now theme based. If you check the `example/DrawerItems.js` file, you should be able to see this:
```
<DrawerItem
  {...props}
  key={props.key}
  theme={
    props.key === 3 ? { colors: { primary: Colors.tealA200 } } : undefined
  }
  active={this.state.drawerItemIndex === index}
  onPress={() => this._setDrawerItem(index)}
/>
```

So you can either change the primary color globally or pass it through the theme prop for customizations. While having a color prop might look a bit cleaner in this case, it's still good to avoid unnecessary duplication and API bloat. Color props are present mostly in cases when there's no way to change it with any other method.

jaulz:
You are right! Thanks for the hint with the theme... did not thought about that way :-)

ManigandanRaamanathan:
@satya164  @jaulz  How to add badge to the app drawer items?

jaulz:
@ManigandanRaamanathan not sure if that's feasible with the current set up but it might be worth to see if we can use the `children` prop instead of a `label` prop in order to be more flexible. @satya164 what do you think?

ManigandanRaamanathan:
> @ManigandanRaamanathan not sure if that's feasible with the current set up but it might be worth to see if we can use the `children` prop instead of a `label` prop in order to be more flexible. @satya164 what do you think?

What are the props available now?

Trancever:
@ManigandanRaamanathan Please, check the docs if you need information about available props.

satya164:
@jaulz maybe makes sense to support a `badge` prop? we'll need to style and position it properly anyway

jaulz:
Great, then let's take a `badge` property and if you are fine I will try to implement a `Badge` component (à la https://github.com/callstack/react-native-paper/pull/539#issuecomment-426977815) which we could reuse here and across other components.

PR #620, opened 2018-10-23
feat: Badge component

jaulz:
### Motivation

This PR introduces the `Badge` component which we could reuse in other components such as https://github.com/callstack/react-native-paper/issues/358#issuecomment-430185158. Let me know what you think :blush:

### Test plan

badge


jaulz:
Errors during build are caused by other components...

jaulz:
@jaulz yep, will have a look... sorry for the delay.

jaulz:
@ferrannp please have a look now 😊

gawrysiak:
hey @jaulz, sorry for the long wait, I've been pretty busy lately. I can see this one's assigned to me, although the API should definitely go through @satya164 as well. I've been trying the component with one of my projects and it's pretty awesome. I have some remarks though:
- would it be possible to allow null as value to not display the badge then? I'm wrapping it up with a Query component from react-apollo and I'd like to conditionally display it when needed. Now it's not really possible, an empty string still displays the badge background
- should it maybe allow numbers as well? it currently throws a flow error, but numbers is pretty much most of the use cases for the component
- the screenshots show a red background, while the default is blue - what's up with that? Not sure which one to expect :)

gawrysiak:
Some example:

I'm looking to shorten this:
```
return data.userInvitations.length > 0 ? (
  <Badge
    value={data.userInvitations.length.toString()}
>
  <MaterialIcons
    color={theme.colors.text}
    name="inbox"
    size={24}
  />
</Badge>
) : (
  <MaterialIcons
    color={theme.colors.text}
    name="inbox"
    size={24}
  />
);
```

To this:
```
return (
  <Badge
    value={data.userInvitations.length > 0 ? data.userInvitations.length : null}
  >
    <MaterialIcons
      color={theme.colors.text}
      name="inbox"
      size={24}
    />
  </Badge>
);
```

jaulz:
@gawrysiak incorporated your feedback 😊 the badge is faded in/out depending on whether the value is null or not and even numbers are accepted now. The default color is the primary color but can be changed via theme. Could you please have another look? Thanks!

gawrysiak:
@jaulz good stuff, the fade in/out is a nice touch 👍 🎉 
LGTM

@satya164 @ferrannp anything to add?

src/components/Badge.js
```diff
@@ -0,0 +1,152 @@
...
+   */
+  horizontalPosition?: 'left' | 'right',
+  /**
+   * Size of the `Badge`.
+   */
+  size?: number,
+  /**
+   * Custom background color.
+   */
+  color?: string,
```
Trancever: Can we remove this prop and let users override color with theme prop (that's how other components do it)?

src/styles/DefaultTheme.js
```diff
@@ -26,6 +26,7 @@ export default {
       .alpha(0.5)
       .rgb()
       .string(),
+    badge: '#6200ee',
```
Trancever: Is this color suggested by guidelines (if yes, please send a link)?
Maybe we could use instead  the "red" one that is shown on screenshots, cause it's more "badgy" color imo.

src/components/Badge.js
jaulz: `Button`, `IconButton`, `Switch` and `RadioButton` have a `color` prop other well. I think especially for a component like this it's handy to have such a prop because you might want to change the color in order to signal warnings etc.

satya164: Those components have a `color` prop because you couldn't customize the color in any other way. In `Button`, it can be used for background color, text or border color depending on the mode. In `IconButton`, you would want to change the icon color etc.

However, with `Badge` it's just the background color, so you will be able to change the background with `style` prop.

src/styles/DefaultTheme.js
satya164: I think it's very common for badges to have a reddish color, so it makes sense for us to use a reddish color by default as the guidelines don't have any clear directions.

This also means that it's legible over things like bottom navigation which have a primary color background by default.

We don't need to use "red", but a reddish color that match the default theme. I think the hot pink color used here is a good default: https://material.io/design/components/bottom-navigation.html#behavior

src/components/Badge.js
jaulz: Using `backgroundColor` is fine for me. It's implemented now.

src/styles/DefaultTheme.js
jaulz: I set the color to pinkish now and since it's anyway customizable via theme it nothing to discuss too much about.

ferrannp:
@jaulz can you split it into `Badge` and `Badge.Wrapper`? We need to make this work also with BottomNavigation. CC @satya164

jaulz:
@ferrannp what would be the difference between the two? is `Badge` just the red dot and `Wrapper` for the positioning?

satya164:
Yes

jaulz:
@ferrannp @satya164 please check now

src/components/Badge/Badge.js
```diff
@@ -0,0 +1,102 @@
...
+    opacity: new Animated.Value(this.props.children ? 1 : 0),
+  };
+
+  componentDidUpdate(prevProps: Props) {
+    const { children } = this.props;
+
+    if (children !== prevProps.children) {
+      Animated.timing(this.state.opacity, {
+        toValue: children ? 1 : 0,
+        duration: 150,
```
Trancever: Please add useNativeDriver: true

raajnadar:
This PR will close https://github.com/callstack/react-native-paper/issues/602 after getting merged

satya164:
1de5095d1349147e0098cd00ce3424e88eabedb6

PR #622, opened 2018-10-23
feat: provide nice error message for users that use Portal without Provider.

Trancever:
### Motivation

This PR makes clear what is the cause of app crash when user doesn't wrap his components' tree with Provider component.

### Test plan

Render Portal component without wrapping components' tree with Provider component.
Nice error message with link to our docs will be presented.

### DIFF
--- a/src/components/Portal/PortalConsumer.js
+++ b/src/components/Portal/PortalConsumer.js
@@ -18,7 +18,7 @@ export default class PortalConsumer extends React.Component<Props> {
           'https://callstack.github.io/react-native-paper/getting-started.html'
       );
     }
-    
+
     this._key = this.props.manager.mount(this.props.children);
   }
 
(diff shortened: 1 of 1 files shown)


src/components/Portal/PortalConsumer.js
```diff
@@ -11,7 +11,15 @@ type Props = {
 
 export default class PortalConsumer extends React.Component<Props> {
   componentDidMount() {
-    this._key = this.props.manager.mount(this.props.children);
+    try {
```
satya164: This might be triggered for errors in rendering the children and not just due to the absence of `Provider`. An explicit check will be better:

```suggestion
    if (!this.props.manager) {
      throw new Error(
        'Looks like you forgot to wrap your root component with `Provider` component from react-native-paper.\n\n' +
          "Please read our getting-started guide and make sure you've followed all the required steps.\n\n" +
          'https://callstack.github.io/react-native-paper/getting-started.html'
      );
    }
    
    this._key = this.props.manager.mount(this.props.children);
  ```

Trancever: Fixed

PR #626, opened 2018-10-24
fix: makes chip content to be vertically aligned.

### LINKED ISSUE #625: Out-of-the-box Chip text vertically misaligned
### Current behaviour

Text passed to chip is vertically misaligned (pushed to the bottom).

### Expected behaviour

Text should be vertically centered.

### Code sample
```js
 <ScrollView 
            horizontal={true}
            showsHorizontalScrollIndicator={false}
        >
            {categories.map(c=>(
                <Chip onPress={() => console.log('Pressed')}>Example Chip</Chip>
            ))}
</ScrollView>
```

chipmisalignment

### What have you tried

Also tried wrapping the text in a Text component, which doesn't change the behaviour, neither does using outlined mode. 
No such issue when using Button component from paper.

### Your Environment

| ios or android        | android
| react-native          | 0.50.3
| react-native-paper    | 2.0.1
| node                  |
| npm or yarn           | npm

Trancever:
### Motivation

Fix #625 

### Test plan

Previously:
44671259_590042324745413_6131087259811708928_n

Now:
44740233_293214124737498_6002635173219270656_n

### DIFF
--- a/src/components/Chip.js
+++ b/src/components/Chip.js
@@ -290,7 +290,8 @@ const styles = StyleSheet.create({
     height: 24,
     lineHeight: 24,
     textAlignVertical: 'center',
-    marginVertical: 4,
+    marginTop: 3,
+    marginBottom: 5,
   },
   avatar: {
     width: 24,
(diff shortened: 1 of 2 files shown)


satya164:
How does it look on iOS?

Trancever:
IOS: 
image

Caundy:
This should by no means be the solution to the issue, by logic this will simply move the text 1px upwards which would actually cause it to go off-center.
Managed to find the culprit (lineHeight applied to top issue on rn versions below 0.51), please disregard my ticket :)

PR #627, opened 2018-10-26
fix: connect ts declaration with @types/react-native

shenjunru:
### Motivation

Connect with "@types/react-native" ts definitions.

### Test plan

N/A


typings/components/Searchbar.d.ts
```diff
@@ -1,14 +1,12 @@
 import * as React from 'react';
+import { TextInputProps } from 'react-native';
 import { ThemeShape, IconSource } from '../types';
 
-export interface SearchbarProps {
-  placeholder?: string;
-  value: string;
+export interface SearchbarProps extends TextInputProps {
```
Esemesek: Good catch on this one.

PR #628, opened 2018-10-26
Fix: onLongPress property added to typings Card

chenglabs:
### Motivation

Missing property in Type Definition

### Test plan
Test the onLongPress property in TS

 <Card
          style={styles.card}
          onLongPress={() => {
            Alert.alert('The City is Long Pressed');
          }}
    >


chenglabs:
Hi,
Please ignore my check-in of Fix Android textinput bug. My error.
Still a newbie with pushing files in Github...

PR #630, opened 2018-10-27
Add missing onLongPress method to Card typings

chenglabs:
### Motivation
Small bug.
Missing onLongPress method in the Card Typescript typings file.


Esemesek:
This bug will be fixed by https://github.com/callstack/react-native-paper/pull/627. This error is a result of me implementing all those types based on Flow types, which are not strict and allow passing not defined props. In Typescript we actually need to extend from `@types/react-native` to make it work properly. Anyway, thanks for contribution C:

chenglabs:
Ok.
Appreciate the work you all did, great library!
Like the UI combined with react-navigation.

PR #618, opened 2018-10-22
ISSUE-616 Set props to accordion state

jimmiehansson:
### Motivation

This PR proposes a quick fix to have the user of the application set a default state for the List Accordion as per #616 

### Test plan

Run the application and provide a state for the `expanded` property

PR #638, opened 2018-11-05
Allow List.Accordion to behave as a controlled component

### LINKED ISSUE #616: Allow user to define initial expanded state of List.Accordion
### Current behaviour

Unless I am missing something in the code, currently all List.Accordion components are initialized with expanded state as false, but I think some users (myself included) would find it beneficial to have them begin opened.

### Expected behaviour

I think it would be as easy as setting the initial state as follows to allow user to define if the component should start as opened or closed
```
state = {
    expanded: this.props.expanded || false
  }
```

### Code sample

^

### LINKED ISSUE #635: Can List.Accordian start open with the user having the option to close it.
Not really a bug, but a small request. Hope that's okay. Thanks for the awesome library.

### Current behaviour

No obvious option to have a list.accordian widget start open.

### Expected behaviour

Hoping there would be an optional open property that I could set to true.

vieiralucas:
### Motivation
This PR aims to FIX #616 and FIX #635 and CLOSE #618 
It allows the List.Accordion to behave as a [controlled component](https://reactjs.org/docs/forms.html#controlled-components) by passing a `expanded` and `onPress` prop.
This also enables users to start the accordion as expanded. 

### Test plan
I added a controlled List.Accordion to the example app and also added a snapshot test.

**Thank you @jimmiehansson for starting this out in #618, I left your commits on my branch so you can take your deserved credits ❤️**

### DIFF
--- a/example/src/ListAccordionExample.js
+++ b/example/src/ListAccordionExample.js
@@ -20,7 +20,7 @@ class ListAccordionExample extends React.Component<Props, State> {
     expanded: true,
   };
 
-  _onPress = () => {
+  _handlePress = () => {
     this.setState({ expanded: !this.state.expanded });
   };
 
@@ -45,7 +45,7 @@ class ListAccordionExample extends React.Component<Props, State> {
             left={props => <List.Icon {...props} icon="folder" />}
             title="Start expanded"
             expanded={this.state.expanded}
-            onPress={this._onPress}
+            onPress={this._handlePress}
           >
             <List.Item title="List item 1" />
           </List.Accordion>
--- a/example/src/ListAccordionExample.js
+++ b/example/src/ListAccordionExample.js
@@ -9,8 +9,20 @@ type Props = {
   theme: Theme,
 };
 
-class ListAccordionExample extends React.Component<Props> {
+type State = {
+  expanded: boolean,
+};
+
+class ListAccordionExample extends React.Component<Props, State> {
   static title = 'List.Accordion';
+
+  state = {
+    expanded: true,
+  };
+
+  _onPress = () => {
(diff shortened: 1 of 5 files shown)


vieiralucas:
I tried my best to improve the docs so the user understands that he needs to pass `onPress` if he passes `expanded`.
Also I'm not sure if `onPress` is the best name for the callback

Thanks

example/src/ListAccordionExample.js
```diff
@@ -9,15 +9,28 @@ type Props = {
   theme: Theme,
```
Trancever: Add expanded and onPress to Flow types please.

src/components/List/ListAccordion.js
```diff
@@ -68,13 +104,26 @@ class ListAccordion extends React.Component<Props, State> {
...
+      } else {
+        this.props.onPress();
+      }
+    } else {
+      this.props.onPress && this.props.onPress();
+
+      this.setState(state => ({
+        expanded: !state.expanded,
+      }));
+    }
```
vieiralucas: Can I propose a little refactor for this method?
```js
const onPressIsFunction = typeof this.props.onPress === 'function';
const expandedPropIsDefined = this.props.expanded !== undefined;

if (expandedPropIsDefined) {
  if (!onPressIsFunction) {
    throw new Error(
      'The `onPress` prop must be passed when the `expanded` prop is passed to List.Accordion.'
    );
  }

  this.props.onPress();
  return;
}

if (onPressIsFunction) {
  this.props.onPress();
}

this.setState(state => ({
  expanded: !state.expanded,
}));
```

PR #645, opened 2018-11-07
Add optional contentStyle prop to bottom navigation

yacut:
### Motivation

I want use BottomNavigation component with custom scene and set custom style for content. e.g. https://github.com/archriss/react-native-snap-carousel


satya164:
What styles do you need to provide? To customize the background color, you can pass `theme={{ colors: { background: backgroundColor } }}`

yacut:
@satya164 I want to set content height or transitions

ferrannp:
@yacut can you provide an example, like a Snack where @satya164 can see what you cannot achieve?

yacut:
@satya164 @ferrannp  https://snack.expo.io/HyZSk3B6X
An example is usage of components like [react-swipeable-views-native](https://github.com/oliviertassinari/react-swipeable-views). If I use such component in `renderScene` (which required) function, it will be rendered each time the current index change. 
One possible solution for this problem could be `contentStyle` prop, so I can set content height to **0**.
Another solution could be a possibility to use `BottomNavigation` component without content renderer.

yacut:
@satya164 Thank you for explanation.

Issue #426, opened 2018-06-21
Ambiguous Module Resolution Error


iyadthayyil:
### Current behaviour

I am getting this error while trying to run example App on expo.

Forked this repo. -> Cloned to my PC ->  `yarn bootstrap` -> `yarn start` from /example -> Got this error.

### Screenshots (if applicable)

![image](https://user-images.githubusercontent.com/11161020/41695578-3c469cb4-7518-11e8-9a7b-fbb3806f28fb.png)

![image](https://user-images.githubusercontent.com/11161020/41695599-5006a302-7518-11e8-9672-20063d0edd98.png)

### What have you tried

- `yarn start -- --reset-cache`

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android 8.0
| react-native          |
| react-native-paper    |
| node                  | v8.11.1
| npm or yarn           | yarn v1.7.0

PR #654, opened 2018-11-14
Ambiguous resolution

raajnadar:
### Motivation

Fixes this issue https://github.com/callstack/react-native-paper/issues/426

Removed `glob-to-regex` module due to this the node_modules in the / and the docs/ were not ignored by the metro-config.json (Expo)

Updated `expo-cli` module to remove the okhttps spam :smile: 

### Test plan

But a new issue is thrown
`Unable to resolve "@babel/runtime/helpers/interopRequireDefault" from "..\src\index.js"`

### DIFF
--- a/example/metro.config.js
+++ b/example/metro.config.js
@@ -3,6 +3,7 @@
 const path = require('path');
 const blacklist = require('metro-config/src/defaults/blacklist');
 const pak = require('../package.json');
+const escape = require('escape-string-regexp');
 
 const dependencies = Object.keys(pak.dependencies);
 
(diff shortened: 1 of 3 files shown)


example/metro.config.js
```diff
@@ -13,8 +12,8 @@ module.exports = {
 
   resolver: {
     blacklistRE: blacklist([
-      glob(`${path.resolve(__dirname, '..')}/node_modules/*`),
-      glob(`${path.resolve(__dirname, '..')}/docs/node_modules/*`),
+      /react-native-paper\/node_modules\/(.*)/,
+      /react-native-paper\/docs\/node_modules\/(.*)/,
```
satya164: Can you try replacing these with:

```js
new RegExp(
  `^${escape(path.resolve(__dirname, '..', 'node_modules'))}\\/.*$`
),
new RegExp(
  `^${escape(path.resolve(__dirname, '..', 'docs', 'node_modules'))}\\/.*$`
)
```

Where,

```js
const escape = require('escape-string-regexp');
```

PR #659, opened 2018-11-16
chore: [WIP] Lint staged files only

raajnadar:
### Motivation

commit-msg does not accept any param instead I used HUSKY_GIT_PARAMS from the node env. It is cross platform and works very well with windows. (Unexpected issue fixed :smile: )

[Refence for commit-msd docs](https://git-scm.com/docs/githooks)

Don't run test on files that are not edited or modified. No more --no-verify to skip the pre-commit hook :smile: 

- Add --bail to stop jest test at first error
- Add --fix to auto fix any auto fix eslint errors
- Add focus-check in flow to test only edited files. (Few bugs yet to be fixed for windows)
- Flow test for only staged files
- Separate cmds for staged testing & manual testing

### Test plan

- [x] Eslint test only staged files (Works perfectly)
- [x] Jest test only staged files (Works perfectly)
- [x] Flow test only staged files (Works perfectly)
- [ ] Typescript test only staged files

@satya164 @Trancever @ferrannp  If you find something is broken please let me know!

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,8 @@
   "scripts": {
     "flow": "flow focus-check",
     "typescript": "node ./scripts/generate-ts-tests.js && tsc --noEmit --skipLibCheck --jsx react ./__ts-tests__/*",
-    "lint": "eslint .",
-    "test": "jest",
+    "lint": "eslint --fix",
+    "test": "jest --bail",
     "prepare": "node ./scripts/generate-mappings",
     "bootstrap": "yarn && yarn --cwd example && yarn --cwd docs",
     "release": "release-it"
@@ -47,7 +47,7 @@
   },
   "lint-staged": {
     "linters": {
-      "**/*.js": ["yarn lint", "yarn flow", "yarn jest", "git add"],
+      "**/*.js": ["yarn lint", "git add", "yarn jest --findRelatedTests"],
       "*": ["git add"]
     },
     "ignore": [
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "bootstrap"
   ],
   "scripts": {
-    "flow": "flow",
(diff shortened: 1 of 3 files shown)


satya164:
Hey! Thanks for the PR. I think I should clarify some things regarding the current scripts:

- We don't run `eslint --fix` as a pre-commit hook because it modifies entire file including code you didn't touch. This also means that it doesn't work well when you try to partially commit a file with `git add -p`. The new `lint-staged` has a new feature to handle this, but it's pretty new and I'm not sure if it can handle such cases properly. I think it's safer for the user to do it than letting a script modify files. Most of the time you just need to run `Format Document` in VSCode which is one step.
- We run `flow` and `eslint` on the entire codebase instead of just changed files because our codebase is pretty small, so it won't make a big difference, and changing one file will often affect others regarding types, so it's necessary to run flow on all files to ensure correct output. It also makes sure that accidentally committed errors don't stay unattended because it's not preventing anyone from committing.
- The current checks rare pretty much the same as the checks that run on the CI, which is nice, because you'll never commit something that passed locally but would fail on the CI for some reason. If it works locally, it should work on the CI.

> Add --bail to stop jest test at first error

Not sure about this one either. If you changed some things and it caused multiple tests to fail, you now have to fix one, run jest, fix next, run jest etc., whereas you could run jest once previously to see all the errors.

raajnadar:
> We don't run eslint --fix as a pre-commit hook because it modifies entire file including code you didn't touch.

Cool, I though It will only fix the staged files.

Flow is type checking in node_modules as well, is that useful?? I think it should ignore node_modules folder in the  docs, examples and root folder..!

Yeah, I agree with that `--bail` part it will be frustrating to fix one and then run the script again, fix process..

Is this PR helpful or should I close this??

PR #587, opened 2018-10-10
feat: Leading and Trailing Icon

SputnlK:
Leading and Trailing Icon with onPressIn- and out event support for manipulating secureTextEntry. Changes in conditional color properties to fit the material design guidelines

### Motivation

I needed leading icons for the input fields for a login screen. Placing an icon next to the input field doesn't look good.

### Test plan

The new Props are: leadingIcon, leadingIconPressIn, leadingIconPressOut, trailingIcon, trailingIconPressIn and trailingIconPressOut. You could test those.

screenshot_20181010-131519__01
screenshot_20181010-131533__01
screenshot_20181010-131453__01

PR #660, opened 2018-11-16
feat: Text Input type password + icon visible and visible_off

tonoli:
### Motivation

When the Text Input is a password there is no prop to declare it. 
I also added the visibility and visibility off icon when it is a password to respectively show and hide the password. 

https://material.io/design/components/text-fields.html#filled-text-field

### Test plan

```
<TextInput
            mode="flat"
            type="password"
            style={styles.inputContainerStyle}
            label="Password"
            value={this.state.password}
            onChangeText={password => this.setState({ password })}
 />
```

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -7,8 +7,10 @@ import {
   TextInput as NativeTextInput,
   StyleSheet,
   I18nManager,
+  TouchableHighlight,
 } from 'react-native';
 import { polyfill } from 'react-lifecycles-compat';
+import Icon from 'react-native-vector-icons/MaterialIcons';
 import color from 'color';
 import Text from './Typography/Text';
 import { withTheme } from '../core/theming';
@@ -74,6 +76,10 @@ type Props = React.ElementConfig<typeof NativeTextInput> & {|
    */
   underlineColor?: string,
   /**
+   * Type of the input.
+   */
+  type?: string,
+  /**
    * Whether the input can have multiple lines.
    */
   multiline?: boolean,
@@ -125,6 +131,7 @@ type State = {
   focused: boolean,
   placeholder: ?string,
   value: ?string,
+  secureTextEntry: boolean,
   labelLayout: {
     measured: boolean,
     width: number,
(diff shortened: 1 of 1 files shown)


ferrannp:
Related https://github.com/callstack/react-native-paper/pull/587.

tonoli:
It's now exactly the same but sure we should definitely work on that feature together

tonoli:
Any news about this? @Esemesek  @ferrannp

peterchibunna:
For those in need of a password input, you can create your custom component.
```
import React from 'react';
import Icon from 'react-native-vector-icons/MaterialIcons';
import {View, StyleSheet} from 'react-native';
import {TextInput} from 'react-native-paper';

export default class PasswordInput extends React.Component {

	constructor(props) {
		super(props);

		this.state = {
			icEye: 'visibility-off',
			password: true
		}
	}

	changePwdType = () => {
		let newState;
		if (this.state.password) {
			newState = {
				icEye: 'visibility',
				password: false
			}
		} else {
			newState = {
				icEye: 'visibility-off',
				password: true
			}
		}

		// set new state value
		this.setState(newState)
	};


	render() {
		return (
			<View>
				<TextInput {...this.props} secureTextEntry={this.state.password}/>
				<Icon
					style={styles.icon}
					name={this.state.icEye}
					size={this.props.iconSize}
					color={this.props.iconColor}
					onPress={this.changePwdType}
				/>
			</View>
		);
	}
}


export const styles = StyleSheet.create({

	icon: {
		position: 'absolute',
		top: 20,
		right: 10
	}

});

PasswordInput.defaultProps = {
	iconSize: 25,
	label: 'Password',
};
```

You still have control over your custom component. Introducing this feature might attract other features we don't want to be dealing with at the moment.

tonoli:
Hi guys! So I can close this PR, if I understood well you don't want to add this kind of prop to the Text input directly, but instead you need to create a custom component for it?

tonoli:
I close the PR then! Thank you @peterchibunna for you code example! 

If I understood well you don't want to add this kind of prop to the Text input directly, but instead you need to create a custom component for it!

PR #661, opened 2018-11-19
Pass down testID from Chip props to Touchable for Detox tests.

KevinGulj:
### Motivation
Make Detox (https://github.com/wix/Detox) testing available for Chips by passing down testID to Touchable.

### Test plan

Now you can test it with Detox.


ferrannp:
@KevinGulj if you pass `testID` to the component directly, you cannot click it?

ferrannp:
Thanks!

PR #672, opened 2018-11-22
feat: smooth icon change in `IconButton`

jaulz:
### Motivation

Whenever the icon of `IconButton` changes, the change looks quite hard. This change will introduce the same effect as in `FAB`s.

### Test plan

n/a

### DIFF
--- a/src/components/IconButton.js
+++ b/src/components/IconButton.js
@@ -5,7 +5,7 @@ import { View, StyleSheet } from 'react-native';
 import color from 'color';
 
 import TouchableRipple from './TouchableRipple';
-import Icon from './Icon';
+import CrossFadeIcon from './CrossFadeIcon';
 import { withTheme } from '../core/theming';
 import type { IconSource } from './Icon';
 import type { Theme, $RemoveChildren } from '../types';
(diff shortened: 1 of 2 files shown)


ferrannp:
@jaulz can you give an example of use case for this? Some gifs maybe?

jaulz:
@ferrannp I am not able to create a gif right now but my use case is that I want to change the icon of an `Appbar.Action` button depending on whether items are selected or not. I could of course render a different component but that transition looks hard and not really appealing. If we use `CrossFadeIcon` instead the change would look nice and smooth and it would not change the API.

ferrannp:
Oh I see. Will check. I know this is the use case for the Contextual Action Bar on Android. Would be nice to have this on Paper.

Trancever:
@jaulz Can you upload a gif showing transition? I still see it quite sharp, just the transition is a bit delayed.

jaulz:
@ferrannp yep, that's basically my use case
@Trancever I don't have a gif right now but the animation is the same as in `FAB`s (which are the reason why the `CrossFadeIcon` was introduced I assume)

Trancever:
@jaulz So I figured out that there is something wrong with my device and animation is not shown.. but it works fine on other devices :)

PR #674, opened 2018-11-22
feat: ability to change button text case

raajnadar:
### Motivation

Ability to change case of text in button. The button with mode text look a bit ugly when the text is long. This PR will give more control to the developer.

### Test plan

Pass `upperCase={false}` as props to the button element


raajnadar:
This can be merged right??

PR #676, opened 2018-11-24
Add Keyboard awareness to BottomNavigation

surajj2223:
### Motivation

The Bottom Navigation does not hide  when keyboard gets open in case for android with windowSoftInputMode set to adjustResize.

### DIFF
--- a/src/components/BottomNavigation.js
+++ b/src/components/BottomNavigation.js
@@ -9,6 +9,7 @@ import {
   SafeAreaView,
   StyleSheet,
   Platform,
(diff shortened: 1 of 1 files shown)


satya164:
Unfortunately, this approach makes it look janky as the bar is visible for a few milliseconds after the keyboard shows up. I created #680 which hides the bar with an animation so it doesn't look as bad.

surajj2223:
Thanks @satya164 . Much appreciated ! Hope to see that merged soon.

PR #680, opened 2018-11-26
feat: hide bottom navigation when keyboard is shown. closes #676

satya164:
ezgif-1-584caa2b0500

### DIFF
--- a/src/components/BottomNavigation.js
+++ b/src/components/BottomNavigation.js
@@ -9,6 +9,7 @@ import {
   SafeAreaView,
   StyleSheet,
   Platform,
+  Keyboard,
 } from 'react-native';
 import { polyfill } from 'react-lifecycles-compat';
 import color from 'color';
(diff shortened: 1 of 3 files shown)


src/components/BottomNavigation.js
```diff
@@ -220,6 +225,10 @@ type State = {
    * List of loaded tabs, tabs will be loaded when navigated to.
    */
   loaded: number[],
+  /**
+   * Trak whether the keyboard is visible to show and hide the tab bar.
```
brentvatne: Trak -> Track

src/components/BottomNavigation.js
```diff
@@ -583,210 +629,232 @@ class BottomNavigation<T: *> extends React.Component<Props<T>, State> {
...
+            {
+              transform: [
+                {
+                  translateY: this.state.visible.interpolate({
+                    inputRange: [0, 1],
+                    outputRange: [this.state.layout.height, 0],
+                  }),
+                },
+              ],
+              position: this.state.keyboard ? 'absolute' : null,
```
brentvatne: only question I have here is whether it would look better to have the keyboard set to false immediately on hide rather than waiting for the animation to complete

brentvatne:
we might want to put this behind a flag in case for some reason people don't want to hide the tab bar when the keyboard is visible

src/components/BottomNavigation.js
satya164: So the reason I change the position to `absolute` is so that it overlays the content below it during the animation, which looks better than empty space behind the tab bar. So it must stay absolute while the animation happens and change back after it finishes.

DenisDov:
Still a problem

brentvatne:
@DenisDov - that is actually the exact point of this pr, to make that happen. you might want to create a new issue to describe what you want in more detail.

PR #684, opened 2018-11-27
feat: allow custom content in DataTableCells

jaulz:
### Motivation

Currently, only text and numbers can be rendered in `DataTableCell`s.

### Test plan

Test included.

### DIFF
--- a/example/src/DataTableExample.js
+++ b/example/src/DataTableExample.js
@@ -2,7 +2,13 @@
 
 import * as React from 'react';
 import { ScrollView, StyleSheet } from 'react-native';
-import { DataTable, Card, withTheme, type Theme } from 'react-native-paper';
+import {
+  DataTable,
+  Card,
+  IconButton,
+  withTheme,
+  type Theme,
+} from 'react-native-paper';
 
 type Props = {
   theme: Theme,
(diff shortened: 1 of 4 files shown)


satya164:
I thought about it, but you can just pass the element directly instead of wrapping it in `DataTable.Cell` which is more flexible. Granted you have to wrap it in `TouchableRipple` yourself or align to right by yourself if you need those, but it's fairly straightforward.

jaulz:
Hm, if we follow that argumentation wouldn't that apply to text as well and we don't need `DataTable.Cell` at all?

jaulz:
In that case we could also think about dropping the `Text` component inside the `DataTable.Cell`. The `right` styling is anyway only applied to the `View`. Then we would be more flexible for any future changes in Material design.

By the way, I thought about the naming of `right` in terms of RTL support. I'm not sure if we should come up with a different name. 
@iyadthayyil the `right` property was designed to support alignment of numbers to be able to compare numbers more easily. How would you do that with RTL?

satya164:
@jaulz yeah we could drop `View`, but it'll be inconvenient for the most common case since you always have to wrap it in `Text`, so I'm not entirely sure.

> By the way, I thought about the naming of right in terms of RTL support. I'm not sure if we should come up with a different name.

Good call. Maybe `end` is a better name since the `start` and `end` terminology used everywhere in styles?

jaulz:
If we drop the `Text`, users will also be able to have multi-line text which is not possible at the moment since we hard coded the number of lines. So it's only for the one case where the user needs the text to be in one line. Though it's very common, thinking twice about it I am a strongly in favor of dropping it.

In general I like `end` but will it left align the content in case of RTL or will it still be right aligned? In case of the first it will not make it easier to compare numbers vertically. In case of the latter it's not really the `end` anymore. We could also call it `numeric`?

PR #693, opened 2018-12-12
Add title and subtitle scaling attributes to Appbar.Content

netshade:
### Motivation

This adds the ability to indicate whether or not navigation bar title and subtitle should be affected by font scaling.  It adds two properties to `Appbar.Content`, `allowTitleFontScaling` and `allowSubtitleFontScaling`, which allow you to individually control whether or not the `title` or `subtitle` should be affected by font scaling.

### Test plan

1. Create an `Appbar` with an `Appbar.Content` whose `title` and `subtitle` are set to a string value.
1. Run the app in an iOS Simulator
1. Open the *Accessibility Inspector*  ( `/Applications/Xcode.app/Contents/Applications/Accessibility Inspector.app` ), and set its `Target` to the Simulator running your test app
1. Choose the `Settings` button ( a Gear icon ) in Accessibility Inspector, and change `Font size` to the maximum possible value
1. Observe the title and subtitle being clipped out of the navigation bar
1. Change `allowTitleFontScaling` and `allowSubtitleFontScaling` to `true`, and reload the app
1. Observe the title and subtitle not being affected by the font scaling changes

Clipped behavior:
screen shot 2018-12-12 at 2 15 56 pm

Not clipped behavior:
screen shot 2018-12-12 at 2 15 08 pm

I could see an argument for the default being to *not* allow for scaling, as that would likely correspond with the desirable behavior of not causing the text content to be clipped, but I erred on the side of respecting older behavior first and allowing you to opt in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-paper",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "description": "Material design for React Native",
   "main": "src/index.js",
   "typings": "typings/index.d.ts",
--- a/src/components/Appbar/AppbarContent.js
+++ b/src/components/Appbar/AppbarContent.js
@@ -90,23 +90,6 @@ class AppbarContent extends React.Component<Props> {
             style={[
               {
                 color: titleColor,
-                fontFamily: Platform.OS === 'ios' ? fonts.regular : fonts.medium,
-              },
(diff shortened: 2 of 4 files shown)


netshade:
Closing this because the history after the rebase was messier than I liked, new PR is #708.

PR #694, opened 2018-12-12
Allow Drawer.Item and Drawer.Section to render multi-line text

netshade:
### Motivation

`Drawer.Item` and `Drawer.Section` were biased to only show 1 line of text, and only at a specific height ( `40` pixels in height in the case of `Drawer.Section` ). This PR changes it such that, when the font size is increased or the content is larger than expected, both `Drawer.Item` and `Drawer.Section` **may** wrap the content and display it as multiline text.

The behavior change is such that:

* IF no `numberOfLines` is specified (`undefined`), the previous behavior is maintained, which is to say that only 1 line of text will be shown
* IF a `numberOfLines` is specified that is a number, the text will show up to that many lines of text
* IF a `numberOfLines` is specified that is `null`, that will be sent as the property value, which will cause the **default** value of the `Text` element to be used for the number of lines.

This is a slightly convoluted path, but it allows the prior behavior to be maintained.

### Test plan

1. Create a `Drawer` which has a `Drawer.Section` and `Drawer.Item` whose label values will be multiple lines worth of content
1. Run the app 
1. Observe that the contents of the `Drawer.Section` and `Drawer.Item` are truncated
1. Change the `numberOfLines` attribute to a number, observe that the number of lines specified are now shown
1. Change the `numberOfLines` attribute to `null`, observe that the default behavior of the `Text` element is used ( show all content ).

Clipped Content
screen shot 2018-12-12 at 2 51 28 pm

Unclipped Content
screen shot 2018-12-12 at 2 52 45 pm

The above screenshots exhibit the clipped / unclipped behavior by increasing the font size using the Accessibility Inspector ( `/Applications/Xcode.app/Contents/Applications/Accessibility Inspector.app` ).

### DIFF
--- a/src/components/Drawer/DrawerItem.js
+++ b/src/components/Drawer/DrawerItem.js
@@ -24,6 +24,10 @@ type Props = {
    */
   active?: boolean,
   /**
+   * Number of lines to display for text, defaults to 1.
+   */
+  numberOfLines?: number,
+  /**
    * Function to execute on press.
    */
   onPress?: () => mixed,
@@ -53,7 +57,7 @@ class DrawerItem extends React.Component<Props> {
   static displayName = 'Drawer.Item';
 
   render() {
-    const { icon, label, active, theme, style, onPress, ...rest } = this.props;
+    const { icon, label, active, theme, style, onPress, numberOfLines, ...rest } = this.props;
     const { colors, roundness } = theme;
     const backgroundColor = active
       ? color(colors.primary)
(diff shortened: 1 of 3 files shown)


netshade:
Sorry for the mess on commits on this one, lint and prettier fixes caused a bit more of a mess than I intended.

ferrannp:
Same as https://github.com/callstack/react-native-paper/pull/708.

netshade:
Same question as #708.

PR #704, opened 2018-12-17
Add borderless prop to Button

ferrannp:
Specially useful for Android P where `borderless = true` is currently broken.


satya164:
Can you conditionally disable it on Android P instead of adding a prop?

ferrannp:
We could but, isn't `borderless = false` something we might want to have in a button?

ferrannp:
Fixed in separated commit.

PR #718, opened 2018-12-31
Allow TouchableRipple for Modern Android Versions

Kjens93:
### Motivation

TouchableRipple uses hard-coded values to prohibit use on Android versions > 27. This means that each time a new version of android is released, the maintainers of this library must update this file to allow TouchableRipple to be used. The proposed change is to allow TouchableRipple to be used on newer versions of Android, namely API level 28.

### Test plan

1. Verify that TouchableRipple now works on Android API level 28.


satya164:
It's intentional as you can see in the commit message: https://github.com/callstack/react-native-paper/commit/23cdb3cd20c105a0a41b0c51bedc6aef11435355

PR #719, opened 2019-01-03
feat: created a counter Component and added to the example app

harisbaig100:
Created PlusMinusCounter component which call be accessible from the main library as well as the demo can be seen by running the example app.

One can use this component for their app and use the counter value provided by the client (as prop) and let the app behave asynchronously according to the value provided by the app.

You can learn more about the component by going [here](https://6577-71323749-gh.circle-artifacts.com/0/docs/plus-minus-counter.html)


Trancever:
Hi @harisbaig100 Could you point me to the Material Design Guidelines that shows this component?

harisbaig100:
> Hi @harisbaig100 Thanks for PR. Could you point me to the Material Design Guidelines that shows this component?

Material Design doesn't show this component, the inspiration was taken from an e-commerce app where you can put items into the cart by choosing number of products.
I think it would help community to use this component in their apps for various purposes.

Trancever:
Ok your point makes sense, but we try to strictly follow Material Design Guidelines and it's very unlikely we will merge this PR, since this counter is not MD component. @satya164 Would you add few words from yourself? 

What I can suggest is publishing this component (with other reusable components that you created) in npm as a separate package or you can try to submit this PR to other RN component libraries which doesn't follow any particular design language and can benefit of merging it :)

harisbaig100:
Thanks for the suggestion @Trancever, i will figure out how can i let community use this component as there is no such component available to use for React Native apps.

Issue #562, opened 2018-09-29
TextInput with RTL is buggy


yaronlevi:
Saw this now on Expo (pixel 2 xl, android pie, hebrew)
![screenshot_20180930-012528](https://user-images.githubusercontent.com/5398977/46251003-11cee680-c450-11e8-8f58-dcfec5f190c8.png)

PR #721, opened 2019-01-06
fix: textInput label wrong position in RTL

hamdiwanis:
### Motivation
Fixes issues like this #562 due to the transformed offset that imitates transform origin on label scale do not count for RTL and make the offset in the wrong direction


Trancever:
hi @hamdiwanis Thanks for the PR!

PR #723, opened 2019-01-08
feat: ActivityIndicator component

jaulz:
### Motivation

This component serves as a drop-in replacement for the component that is delivered by React Native out of the box. It's design tries to imitate the indeterminate circular progress indicator from the Material guidelines:
https://material.io/design/components/progress-indicators.html#circular-progress-indicators 

### Test plan

Open examples and see examples of `ActivityIndicator`.


example/src/ActivityIndicatorExample.js
```diff
@@ -0,0 +1,84 @@
+/* @flow */
+
+import * as React from 'react';
+import { View, StyleSheet } from 'react-native';
+import { Colors, FAB, withTheme, type Theme } from 'react-native-paper';
+import { ActivityIndicator } from '../../src';
```
Trancever: Can we import it from 'react-native-paper' instead?

src/components/ActivityIndicator.js
```diff
@@ -0,0 +1,269 @@
+/* @flow */
+
+import * as React from 'react';
+import { Animated, Easing, Platform, StyleSheet, View } from 'react-native';
+import { withTheme } from '../core/theming';
+import type { Theme } from '../../types';
+
+type Props = {|
```
Trancever: I think in future we should add `type` prop for Determinate and Indeterminate states, but it shouldn't be breaking change so we can keep that API as it is.

jaulz: You are right but at the moment this component is supposed to be a drop in replacement for the native component. Anyway in the future we could enhance it and just make `type` as `indeterminate` as default.

example/src/ActivityIndicatorExample.js
jaulz: Done

PR #724, opened 2019-01-08
fix: use fontFamily of theme for Searchbar

HassanHeydariNasab:
### Motivation

Searchbar doesn't use fontFamily described in custom theme.

### Other solutions

I wonder that why Searchbar doesn't use fontFamily of TextInput. maybe someone can explain it.


satya164:
Thanks for the PR. Looks like a mistake. Can you fix the lint errors?

Issue #687, opened 2018-11-30
IconButton appearing as "?" 


Wilson-Lim:
### Current behaviour
![image](https://user-images.githubusercontent.com/15623614/49265733-5269d100-f48e-11e8-9998-a9a823b7fbf3.png)

### Expected behaviour
Trying to use "calendar_today" in Material icons, and many other icons, some are appearing, and some are just "?" 

Is there a list of whats supported and not ?

### Code sample
<IconButton
          icon="calendar_today"
          style={styles.child}
          size={20}
          color={"#161616"}
        />

### Screenshots (if applicable)
![image](https://user-images.githubusercontent.com/15623614/49265831-bab8b280-f48e-11e8-987d-cd5553e57f73.png)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | 8.1 
| react-native          | 0.57.7
| react-native-paper    | ^2.2.8
| node                  | 
| npm or yarn           |

raajnadar:
This means the icon is not available..

satya164:
We use `MaterialIcons` from https://github.com/oblador/react-native-vector-icons. So only the icons available there are supported.

devpolish:
I'm having the same issue, I'm trying to add Navigation's icons from this source but most of the icons are not supported.

https://material.io/resources/icons/?style=baseline

"react": "16.6.0-alpha.8af6728"
"react-native": "0.57.4"
"react-native-paper": "^2.16.0"

lukewalczak:
Hey @devpolish , from version 3.0 we are using [MaterialCommunityIcons](https://oblador.github.io/react-native-vector-icons/) by default.

CLantigua2:
The site still points to https://materialdesignicons.com/ but none of those seem to actually work. Should it reflect https://oblador.github.io/react-native-vector-icons/ instead?

sebastianext:
Currently you can see the list of all supported icons right in 

https://material.io/resources/icons/?style=baseline

**Use example:**
![image](https://user-images.githubusercontent.com/6503043/67135786-345ea980-f1e3-11e9-93dc-45a080431f71.png)

**Name icon:** _alarm_add_

**Name react native**: 

```
...
<IconButton
     icon="alarm-add"
     size={20}
     onPress={() => console.log('Pressed')}
 />
...
```

Vidyack:
Link the react-native-paper module to your project after installing. It will work fine.

1. npm install react-native-paper --save
2. react-native link
[...]

PR #728, opened 2019-01-14
docs: add SearchBar to find an icon name on Icons page

lukewalczak:
### Motivation

Issue no #687  - That PR will add a SearchBar component in documents on Icons page to allow user search for specific icon.

image

### DIFF
--- a/docs/assets/styles.css
+++ b/docs/assets/styles.css
@@ -1,9 +1,10 @@
-a, a:hover {
+a,
+a:hover {
   color: #6200ee;
 }
 
 a:hover {
-  color: #873DF2;
+  color: #873df2;
 }
 
 .screenshots {
@@ -47,4 +48,10 @@ a:hover {
   .screenshots img {
     width: 240px;
   }
+}
+
+@font-face {
+  font-family: 'MaterialIcons';
+  src: url('../../node_modules/react-native-vector-icons/Fonts/MaterialIcons.ttf')
+    format('truetype');
 }
--- a/docs/pages/4.icons.mdx
+++ b/docs/pages/4.icons.mdx
@@ -1,6 +1,8 @@
----
-title: Icons
----
+export const meta = {
+  title: "Icons"
+}
+
+import IconsList from './src/components/IconsList.js'
 
 ## Configuring icons
 
@@ -23,6 +25,10 @@ Example:
   Press me
 </Button>
 ```
+
+<details><summary>See the list of supported icons</summary>
+  <IconsList />
+</details>
 
 ### 2. An image source
 
--- a/docs/index.js
+++ b/docs/index.js
@@ -21,6 +21,15 @@ if (!fs.existsSync(dist)) {
   fs.mkdirSync(dist);
 }
 
+function getType(file: string) {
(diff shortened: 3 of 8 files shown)


docs/assets/styles.css
```diff
@@ -48,3 +49,8 @@ a:hover {
     width: 240px;
   }
 }
+
+@font-face {
```
satya164: Does it work if you move it to the CSS in the icon search component? It'll be nice to have it where it's used if it works.

docs/assets/styles.css
```diff
@@ -48,3 +49,8 @@ a:hover {
...
+  font-family: 'MaterialIcons';
+  src: url('./fonts/MaterialIcons.ttf') format('truetype');
```
satya164: Could probably do `url(./node_modules/react-native-vector-icons/Fonts/MaterialIcons.ttf)` so we won't have to copy this file.

docs/pages/4.icons.mdx
```diff
@@ -1,6 +1,8 @@
----
-title: Icons
----
+export const meta = {
+  title: "Icons"
+}
+
+import SearchBar from './src/components/SearchBar.js'
```
satya164: Maybe rename this component to `IconList` to be more descriptive

docs/pages/4.icons.mdx
```diff
@@ -24,6 +26,9 @@ Example:
 </Button>
 ```
 
+<SearchBar />
```
satya164: Would be nice to have some kind of heading/label indicating what it is. Also maybe hide the list by default and show an accordion? Something like this should work I think:

```html
<details><summary>See list of icons</summary>
  <SearchBar />
</details>
```

docs/pages/src/components/SearchBar.js
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  render() {
+    return (
+      <form>
+        <input
+          type="search"
+          value={this.state.query}
+          onChange={this.handleInputChange}
+          placeholder="Search for icon..."
```
satya164: Use an ellipsis instead of 3 dots:

```suggestion
          placeholder="Search for icon…"
```

docs/pages/src/components/SearchBar.js
```diff
@@ -0,0 +1,115 @@
...
+    value: string,
+  },
+};
+
+class SearchBar extends React.Component<{}, State> {
+  state = {
+    query: '',
+  };
+
+  handleInputChange = (e: Event) => {
```
satya164: nit: prefix the methods with `_`

docs/pages/src/components/SearchBar.js
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      return icons.map<React.Node>(iconName => (
+        <div className={iconContainer} key={iconName}>
+          <p>{iconName}</p>
+          {this.getIconCharacter(iconName)}
+        </div>
+      ));
+    }
+    return null;
```
satya164: I think if there's no search term, we should show the full list

PR #730, opened 2019-01-15
Appbar.Content multiline title

danikane:
If subtitle is present limit the number of lines for the title to 1, else let it spread over 2 lines.
Suitable for longer titles, especially when Appbar.Action's taking up space are present.

### Motivation

On smaller screens a title of decent length is truncated. It can be rendered over 2 lines at best.


juarezfranco:
Usage example:

```
        <Appbar.Content
          title={
            <>
              <Text variant="titleLarge">Playground</Text>
              <Text>Second Line</Text>
            </>
          }
          style={{ alignItems: 'flex-start' }}
        />
```

Issue #500, opened 2018-08-20
FAB ripple is wrong on Android P (everything that uses borderRadius)


ferrannp:
This is more a react-native issue that we need to keep track on. See https://github.com/facebook/react-native/issues/6480.

![screenshot_20180820-124834_2](https://user-images.githubusercontent.com/774577/44336424-b33d4280-a477-11e8-8335-3ddffadd4787.png)

yaronlevi:
Duplicate:
https://github.com/facebook/react-native/issues/6480

ferrannp:
Fixed #788 here. For 3.0 we might go with gesture handler.

PR #731, opened 2019-01-16
feat: use BaseButton from react-native-gesture-handler

lukewalczak:
### Motivation

Solving the issue no #500.

### DIFF
--- a/src/components/TouchableRipple.js
+++ b/src/components/TouchableRipple.js
@@ -2,29 +2,22 @@
 
 import * as React from 'react';
 import {
-  TouchableNativeFeedback,
   TouchableHighlight,
   TouchableWithoutFeedback,
   Platform,
-  View,
 } from 'react-native';
+import { BaseButton } from 'react-native-gesture-handler';
 import color from 'color';
 import { withTheme } from '../core/theming';
 import type { Theme } from '../types';
 
 const ANDROID_VERSION_LOLLIPOP = 21;
-const ANDROID_VERSION_NOUGAT = 27;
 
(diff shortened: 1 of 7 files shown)


.eslintrc
```diff
@@ -1,4 +1,8 @@
 {
+  "globals": {
```
satya164: We should remove this

src/components/TouchableRipple.js
```diff
@@ -145,6 +123,19 @@ class TouchableRipple extends React.Component<Props, void> {
...
         {React.Children.only(children)}
       </TouchableHighlight>
+    ) : (
+      <BaseButton
+        {...rest}
+        disabled={disabled}
+        style={[style]}
+        rippleColor={
+          background !== undefined
+            ? background
```
satya164: We should get rid of `background` prop

ferrannp:
We need to test if it works with https://github.com/kmagiera/react-native-gesture-handler/pull/400 cc @Trancever

ferrannp:
@satya164 gesture handle is out, 3.0-alpha time.

Trancever:
@ferrannp We wait for new expo release with updated gesture handler, without it we can't test it in our example app.

ferrannp:
@brentvatne new Expo release with Gesture Handler coming soon? ❤️

ferrannp:
We won't have this because the buttons are not accessible yet https://github.com/kmagiera/react-native-gesture-handler/issues/468 :(

waqas19921:
@ferrannp There is fix for that now.

RichardLindhout:
There is now also web support.

PR #735, opened 2019-01-18
feat: ProgressBar component

jaulz:
### Motivation

Currently, the progress bar differs on every platform. This component is a drop-in for the native `ProgressBar` and follows the guidelines from https://material.io/design/components/progress-indicators.html

### Test plan

See examples app.


src/components/ProgressBar.js
```diff
@@ -0,0 +1,232 @@
...
+   */
+  color?: string,
+  /**
+   * If the progress bar will show indeterminate progress.
+   */
+  indeterminate?: boolean,
+  /**
+   * Whether to show the ProgressBar (true, the default) or hide it (false).
+   */
+  animating?: boolean,
```
Trancever: Can we rename that prop to `visible`? Animating does not really makes sense for us and keeping compatible api with ProgressBarAndroid is not something we care about.

src/components/ProgressBar.js
```diff
@@ -0,0 +1,232 @@
...
+
+type Props = {|
+  /**
+   * Progress value (between 0 and 1).
+   */
+  progress?: number,
+  /**
+   * Color of the progress bar.
+   */
+  color?: string,
```
Trancever: Can we mention that `backgroundColor` is going to be calculated based on this prop?

src/components/ProgressBar.js
```diff
@@ -0,0 +1,232 @@
...
+  /**
+   * If the progress bar will show indeterminate progress.
+   */
+  indeterminate?: boolean,
+  /**
+   * Whether to show the ProgressBar (true, the default) or hide it (false).
+   */
+  animating?: boolean,
+  style?: any,
```
Trancever: Can we mention that passing `backgroundColor` here will override the default color?

typings/components/ProgressBar.d.ts
```diff
@@ -1,11 +1,13 @@
...
-
-export declare class ProgressBar extends React.Component<ProgressBarProps> {}
+import * as React from 'react';
+import { ThemeShape } from '../types';
+
+export interface ProgressBarProps {
+  progress?: number;
+  color?: string;
+  indeterminate?: boolean;
+  animating?: boolean;
```
Trancever: Do not forget to change typescript type also :D

example/src/ProgressBarExample.js
```diff
@@ -16,43 +17,47 @@ type Props = {
...
       <View style={[styles.container, { backgroundColor: background }]}>
+        <Button
+          onPress={() => this.setState({ animating: !this.state.animating })}
+        >
+          Toggle animating
+        </Button>
+        <Button onPress={() => this.setState({ progress: Math.random() })}>
+          Random progress
+        </Button>
         <Paragraph>ProgressBar primary color</Paragraph>
```
Trancever: Can we add some vertical padding between each example?

src/components/ProgressBar.js
jaulz: It's renamed now.

jaulz:
@Trancever thanks for the feedback. Could you please have another look? 😊

Trancever:
@jaulz I also pushed one commit with updated typescript types, cause they were lost during merge I think.

jaulz:
@Trancever oops, sorry, but thanks!

Trancever:
@satya164 Should we consider this change as breaking? We changed `animating` prop to `visible`.. I haven't though about this earlier :/

deyanskiq:
satya164 any update?

Trancever:
@deyanskiq This is breaking change so we are going to introduce it with 3.0 version, but I can't say exactly when it's going to happen.

deyanskiq:
I have trouble changing the height of the ProgressBar. This breaking change is related to my problem right?

Trancever:
@deyanskiq This is breaking change, cause we slightly change props to better align with existing components, but yeah, after this is merged, you should be able to change height via `style` prop. If you can't wait for it, you can copy and paste this component into your project and use it :)

deyanskiq:
@Trancever Applying height property to style set height only to the background object. Is there a way to apply height to the front object too. Screenshot here: https://prnt.sc/mxrcqg

jaulz:
@deyanskiq here you go...

deyanskiq:
Great thank you. You helped other devs as well.

jaulz:
@Trancever do you think it would make sense to have a 2.0 and a 3.0 version with the subtle change of the property name?

Trancever:
@jaulz We are going to release 3.0 just after we merge this PR with Menu component. Shouldn't take much time, probably it will happen this week.

jaulz:
@Trancever great news, thanks for the update 😊

Trancever:
@jaulz Could you change target branch to 3.0? Once it's done we will merge this :)

jaulz:
@Trancever changed 😊

Issue #716, opened 2018-12-29
List.Subheader component


ocarreterom:
Add a new component for lists subheaders so it can be use in [FlatList.ListHeaderComponent](https://facebook.github.io/react-native/docs/flatlist#listheadercomponent) and [SectionList.renderSectionHeader](https://facebook.github.io/react-native/docs/sectionlist#rendersectionheader) detached from `List.Section`.

https://github.com/callstack/react-native-paper/blob/38b729597a9bbf1721b4d9b5e8ec47a4dba2a8f5/src/components/List/ListSection.js#L74-L79

ocarreterom:
@ferrannp Ok! I will try this weekend.

ocarreterom:
There are some Flow errors when commiting.
```sh
yarn run v1.13.0
$ flow
Error ------------------------------------------------------ docs/node_modules/@webassemblyjs/ieee754/src/index.js:37:40

Cannot resolve name `Byte`.

   37| export function decodeF32(bytes: Array<Byte>): number {
                                              ^^^^


Error ---------------------------------------------- example/node_modules/metro-config/src/configTypes.flow.js.flow:22:8

Cannot resolve module `metro/src/DeltaBundler/types.flow.js`.

   22| } from 'metro/src/DeltaBundler/types.flow.js';
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


Error ------------------------------------------------------ node_modules/metro-config/src/configTypes.flow.js.flow:22:8

Cannot resolve module `metro/src/DeltaBundler/types.flow.js`.

   22| } from 'metro/src/DeltaBundler/types.flow.js';
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


Error ----------------------------------------------------- node_modules/metro-config/src/configTypes.flow.js.flow:24:40

Cannot resolve module `metro/src/JSTransformer/worker`.

   24| import type {JsTransformerConfig} from 'metro/src/JSTransformer/worker';
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^



Found 4 errors
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Trancever:
@ocarreterom Create PR so we can have a look.

satya164:
[...]

PR #739, opened 2019-01-19
feat: ListSubheader component

ocarreterom:
### Motivation
Add new component `ListSubheader` to be use in `FlatList`.
I didn't found specific specs for this component in material.io but there are some examples of lists with subheader. https://material.io/design/components/lists.html#anatomy

closes https://github.com/callstack/react-native-paper/issues/716

### Test plan
|Reference|react-native-paper|
|----------|-------------------|
|![reference]|![paper]|

[reference]: https://user-images.githubusercontent.com/11599942/51439310-8efcc080-1cb8-11e9-9347-3d8ff99856ac.png
[paper]: https://user-images.githubusercontent.com/11599942/53601703-fb6ab980-3bac-11e9-8efc-9d390a61920e.png


ferrannp:
@ocarreterom can you fix merge conflicts?

src/components/List/ListSection.js
```diff
@@ -1,15 +1,15 @@
...
 import { View, StyleSheet } from 'react-native';
-import Text from '../Typography/Text';
+import ListSubheader from './ListSubheader';
 import { withTheme } from '../../core/theming';
 import type { Theme } from '../../types';
 
 type Props = React.ElementConfig<typeof View> & {
   /**
    * Title text for the section.
+   * @deprecated
```
ferrannp: What's this about?

ocarreterom: I'm trying to avoid a breaking change. 😅
How can I inidicate that `title` prop should not be used?

ocarreterom:
Rebased. 👌

Trancever:
@satya164 Any comments on this? Imo this change is ok and it allows passing `Subheader` to `FlatList`. I am also thinking if we should keep title prop on List.Section, remove it or maybe mark as deprecated and remove while releasing 3.0 version. What do you think?

src/components/List/ListSection.js
Trancever: Let's remove this warning, and keep previous prop as it is.

src/components/List/ListSubheader.js
```diff
@@ -0,0 +1,62 @@
...
+ * class MyComponent extends React.Component {
+ *  render () {
+ *    return <List.Subheader>My List Title</List.Subheader>;
+ *  }
+ * }
+ *
+ * export default MyComponent;
+ * ```
+ */
+class ListSubheader extends React.Component<Props> {
```
Trancever: Why do we call it Subheader? Isn't it simply Header?

ocarreterom: The name comes from [material.io](https://material.io/design/components/lists.html#anatomy). 🤷‍♂️

Trancever: It's ok then :D

src/components/List/ListSection.js
ocarreterom: Done! 👌

src/components/List/ListSubheader.js
```diff
@@ -0,0 +1,62 @@
...
+  static displayName = 'List.Subheader';
+
+  render() {
+    const { style, theme, ...rest } = this.props;
+    const { colors, fonts } = theme;
+
+    const computedStyle = StyleSheet.flatten([
+      styles.container,
+      {
+        color: colors.text,
```
Trancever: color should be: `color(colors.text).alpha(0.54).rgb().string();` just like before.
Same for font, it should be: `fonts.medium`

src/components/List/ListSubheader.js
```diff
@@ -0,0 +1,62 @@
...
+ * ```
+ */
+class ListSubheader extends React.Component<Props> {
+  static displayName = 'List.Subheader';
+
+  render() {
+    const { style, theme, ...rest } = this.props;
+    const { colors, fonts } = theme;
+
+    const computedStyle = StyleSheet.flatten([
```
Trancever: Can we remove this variable and pass array with 3 objects directly to style prop?

src/components/List/ListSubheader.js
Trancever: Also, let's extract them to separate variables just like before.

src/components/List/ListSubheader.js
```diff
@@ -0,0 +1,62 @@
+/* @flow */
+
+import * as React from 'react';
+import { Text, StyleSheet } from 'react-native';
```
Trancever: Text should be imported from `react-native-paper`

src/components/List/ListSubheader.js
```diff
@@ -0,0 +1,62 @@
...
+      style,
+    ]);
+
+    return <Text numberOfLines={1} {...rest} style={computedStyle} />;
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    fontSize: 14,
```
Trancever: Use same styles as before 
`{
  marginVertical: 13,
  marginHorizontal: 16
}`

ocarreterom: Isn't it better to use padding than margin?
I think that using padding is more easy to apply custom common styles like background color.

Trancever: I am ok with paddings, let's change it but we should keep the previous values. When those issues are solved, we will be ready to merge it :)

Trancever: I am ok with paddings, let's change it but we should keep the previous values. When those issues are solved, we will be ready to merge it :)

ocarreterom:
Issues addressed 👍

PR #744, opened 2019-01-22
Added ProgressBarComponent as fallback for generic platforms

robertherber:
### Motivation

Tried using react-native-paper with react-native-dom and saw ProgressBarComponent crashes because it doesn't have a fallback defined for generic platforms (react-native-dom listens to .dom.js rather than .web.js). As far as I can see it's the only file missing fallback in the project.

### Test plan

Use it with react-native-dom or any other platform using an undefined pattern.

### DIFF
--- a/src/components/ProgressBar/ProgressBarComponent.js
+++ b/src/components/ProgressBar/ProgressBarComponent.js
@@ -1,0 +1,1 @@
+export { ProgressBar as default } from 'react-native';


Trancever:
@robertherber Merged in [bcf9178c4](https://github.com/callstack/react-native-paper/pull/749/commits/5a0ff2289e46f69f9d7077b1ca68858bcf9178c4). We decided to rename .web.js to .js.

PR #746, opened 2019-01-22
feat: add badge support in bottom navigation. closes #288

### LINKED ISSUE #288: Support badges in bottom navigation
image

satya164:


src/components/Badge.js
```diff
@@ -81,6 +81,7 @@ class Badge extends React.Component<Props, State> {
 
     return (
       <Animated.Text
+        numberOfLines={1}
```
ferrannp: Was this a bug from the badge?

src/components/BottomNavigation.js
```diff
@@ -924,4 +954,9 @@ const styles = StyleSheet.create({
     textAlign: 'center',
     backgroundColor: 'transparent',
   },
+  badgeContainer: {
+    position: 'absolute',
+    left: 0,
+    top: -2,
```
ferrannp: Does the -2 work ok everywhere?

src/components/Badge.js
satya164: Yea

src/components/BottomNavigation.js
satya164: Height of badge stays constant, so yea

src/components/BottomNavigation.js
```diff
@@ -797,6 +808,25 @@ class BottomNavigation<T: *> extends React.Component<Props<T>, State> {
...
+                                (badge != null && typeof badge !== 'boolean'
+                                  ? String(badge).length * -2
+                                  : 0) - 2,
+                            },
+                          ]}
+                        >
+                          {typeof badge === 'boolean' ? (
+                            <Badge visible={badge} size={8} />
+                          ) : (
+                            <Badge visible={badge != null} size={16}>
```
brunohkbx: Shouldn't it be !== ?
```suggestion
                            <Badge visible={badge !== null} size={16}>
```

dylandislers:
Is there a way to change the background color of the badge from the default red?

PR #747, opened 2019-01-23
fixing parenthesis locations

napolev:
On the following line the parenthesis are not properly located:
https://github.com/callstack/react-native-paper/blob/bcef4a60ba72669d51cd9418a341193c9bb0cfc8/example/App.js#L41

this is how it is right now:
drawerPosition: Platform.OS === 'android' && (I18nManager.isRTL ? 'right' : 'left'),

this is how it should be:
drawerPosition: (Platform.OS === 'android' && I18nManager.isRTL) ? 'right' : 'left',

since "drawerPosition" accepts the values: { 'left', 'right' }

Thanks.

### DIFF
--- a/example/App.js
+++ b/example/App.js
@@ -38,7 +38,7 @@ const App = createDrawerNavigator(
     ),
     // set drawerPosition to support rtl toggle on android
     drawerPosition:
-      (Platform.OS === 'android' && I18nManager.isRTL) ? 'right' : 'left',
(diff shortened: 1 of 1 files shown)


Trancever:
@napolev Thanks for PR! Could you fix eslint error? It's a problem with formatting.

Trancever:
Yes, thanks for that!

PR #750, opened 2019-01-23
fix: replace 'orderRadius' with 'borderRadius' in Snackbar component

itsjgf:
### Motivation

The `borderRadius` / `roundness` prop was not working as expected because there was a typo in its name.

### Test plan

n/a

### DIFF
--- a/src/components/Snackbar.js
+++ b/src/components/Snackbar.js
@@ -196,7 +196,7 @@ class Snackbar extends React.Component<Props, State> {
           style={[
             styles.container,
             {
-              orderRadius: roundness,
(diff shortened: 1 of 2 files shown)


Trancever:
Hi @itsjgf Thanks for PR. It looks like bug was introduced with [2b54b06bee105](https://github.com/callstack/react-native-paper/commit/2618c067f6e9ffdeed17ce7b12d2b54b06bee105)
Could you fix tests (yarn test -u) so we can proceed with merge?

itsjgf:
@Trancever done

PR #753, opened 2019-01-24
fix TypeScript typings for TextInput methods

Nkzn:
### Motivation

* paper version: v2.6.1
* typescript version: v3.2.2

The `TextInput` class has public methods `isFocused`, `clear`, `focus`, `blur`.

https://github.com/callstack/react-native-paper/blob/v2.6.1/src/components/TextInput.js#L348-L374

However, they are not declared in d.ts as below.

### Test plan

Typecheck for below function.

```typescript
import { TextInput } from "react-native-paper";

function doAnything(textInput: TextInput) {
  textInput.isFocused();
  textInput.clear();
  textInput.focus();
  textInput.blur();
}
```

### DIFF
--- a/typings/components/TextInput.d.ts
+++ b/typings/components/TextInput.d.ts
@@ -36,4 +36,9 @@ export interface RenderProps extends NativeTextInputProps {
   value?: string;
 }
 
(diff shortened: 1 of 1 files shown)


satya164:
Thank you for the PR!

PR #758, opened 2019-01-25
fix: Allow to change background for FAB Group

gawrysiak:
### Motivation

It's currently possible to change the background color for FAB and also items within FAB.Group components. Changing the background color in FAB Group causes the overlay background to change.

This PR makes it possible to change the background for the FAB Group main button.

Current behavior:
screen shot 2019-01-25 at 17 41 56

### Test plan

https://snack.expo.io/ryjNEaO7E

### DIFF
--- a/src/components/FAB/FABGroup.js
+++ b/src/components/FAB/FABGroup.js
@@ -72,6 +72,10 @@ type Props = {|
    */
   style?: any,
   /**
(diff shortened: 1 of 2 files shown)


satya164:
Can you also update the typescript declaration files too?

gawrysiak:
Ah, sorry, missed that. It's now added.

PR #769, opened 2019-01-27
fix: makes icon load errors more informative

robbiemu:
### Motivation

The error messages from try block loading Icons were conditional and did not include the actual error. makes icon load errors more informative

### Test plan

don't install icons, so that you render the squares. look for error in console

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -10,6 +10,9 @@ try {
   // Optionally require vector-icons
   MaterialIcons = require('react-native-vector-icons/MaterialIcons').default;
 } catch (e) {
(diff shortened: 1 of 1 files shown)


src/components/Icon.js
```diff
@@ -10,6 +10,9 @@ try {
   // Optionally require vector-icons
   MaterialIcons = require('react-native-vector-icons/MaterialIcons').default;
 } catch (e) {
+  // eslint-disable-next-line no-console
+  console.error(e)
```
AndrewJack: this line causes a red error screen in debug mode when using https://github.com/satya164/babel-plugin-optional-require

simulator screen shot - iphone 8 plus - 2019-02-06 at 16 38 24

satya164: Ouch. @AndrewJack can you send a PR moving it to the icon component?

AndrewJack: Raised a PR here #818

Issue #768, opened 2019-01-27
Fonts are not correct on React Native Web


satya164:
Try on Safari.

We'll need to do the same as this: #590

jaulz:
This could be solved by https://github.com/callstack/react-native-paper/pull/786

ferrannp:
#786 was merged.

PR #778, opened 2019-01-29
fix: improve fonts on web

jaulz:
### Motivation

`Roboto Medium` was not properly displayed in the example app since only `Roboto` font was loaded. This PR loads all necessary fonts locally (which also helps when you work in places without internet) and also adds fallback fonts.

### Test plan

Open examples app on web.

### DIFF
--- a/example/App.web.js
+++ b/example/App.web.js
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { render } from 'react-dom';
 import App from './src/index';
+import './assets/styles/fonts.css';
 
 render(<App />, document.getElementById('root'));
--- a/example/assets/styles/fonts.css
+++ b/example/assets/styles/fonts.css
@@ -1,0 +1,47 @@
+@font-face {
+    font-family: 'Roboto-Thin';
+    font-style: normal;
+    font-display: swap;
+    font-weight: 100;
+    src:
+        local('Roboto Thin '),
+        local('Roboto-Thin'),
+        url('~typeface-roboto/files/roboto-latin-100.woff2') format('woff2'), /* Super Modern Browsers */
+        url('~typeface-roboto/files/roboto-latin-100.woff') format('woff'); /* Modern Browsers */
+}
+  
+@font-face {
+    font-family: 'Roboto-Light';
+    font-style: normal;
+    font-display: swap;
+    font-weight: 300;
+    src:
+        local('Roboto Light '),
+        local('Roboto-Light'),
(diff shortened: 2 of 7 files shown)


jaulz:
Should solve this: https://github.com/callstack/react-native-paper/issues/768

satya164:
> Should solve this: #768

I think we should make it work with google fonts. So I'll leave that open.

jaulz:
@satya164 I would actually even suggest to include the Material font in this library. Especially for offline scenarios, e.g. develop while commuting but also PWAs that should work without any internet connection, it does not make sense to use Google Fonts since they won't be cached. If we anyway would like to use Google Fonts, we need to to define Roboto for all (`regular`, `medium`, `light` and `thin`) because there is no `RobotoMedium` as far as I know. If we do that, we could set the `fontWeight` explicitely based on `family` here:
https://github.com/callstack/react-native-paper/blob/77936b8d17823c781389e581abbbcbcac3093c33/src/components/Typography/StyledText.js#L31

satya164:
I think it totally makes sense to document how to include these fonts on web. We can add it in the guide for web in our docs.

> it does not make sense to use Google Fonts since they won't be cached

That depends on the use case. There are advantages of google fonts:
1. Lots of webpages use google fonts. It's very likely that the font is already in the browser cache, so it'll load instantly (since Roboto is a common font)
2. Google will serve the best-supported fonts for user's browser and can optimize it better

If we include the font in the library, there are some problems:
1. The setup is more complex now. User has to configure `css-loader`, `style-loader`, `mini-css-extract-plugin` before using Paper on web.
2. The user is not required to use Roboto with paper and they can change the fonts to a different one. In that case, they'll need a way to opt-out of the Roboto fonts.

If the user has a PWA and wants to cache the fonts, they can include the fonts themselves. It'll be easier than having ways to opt-out, as long as we document it properly of course.

Regarding how we support Roboto, we'll need to change the format of `theme.fonts` which will be a breaking change for 3.0. We'll need to do it regardless of web support to be able to San Francisco on iOS without needing to bundle the fonts separately (see the linked issue in #768). We can use the same mechanism to support Google Fonts or anything else for that matter.

jaulz:
Yep, Google Fonts has definitely its advantages and you are absolutely right. Though, since Roboto is an essential part of the Material design system I meant to include the Roboto font files in this library so the user does not need to include another third-party module (which are in fact rare and most of them seem to be abandoned or rarely updated). I will send a PR in a minute with various changes so we can discuss it in there.

Issue #767, opened 2019-01-27
Card doesn't have shadow on React Native Web


satya164:

jaulz:
The thing is that `elevation` is an `Animated.Value` which cannot be used in `shadow` function:
https://github.com/callstack/react-native-paper/blob/eefcff37b56c12213c50d48b89d2fccc92de5d70/src/components/Card/Card.js#L93

https://github.com/callstack/react-native-paper/blob/eefcff37b56c12213c50d48b89d2fccc92de5d70/src/components/Surface.js#L68

PR #779, opened 2019-01-29
fix: elevation on web

jaulz:
### Motivation

This will fix https://github.com/callstack/react-native-paper/issues/767. However, there is one downside as there seems to be an issue with `Animated` on web and shadows. For some reason it does not animate the shadow properly and we have to forcefully update the component:
https://github.com/callstack/react-native-paper/blob/fix/elevation-on-web/src/components/Surface.js#L88

### Test plan

Open examples app.


satya164:
Thanks for looking into it. I just pushed an update to the shadow helper to handle animated values: https://github.com/callstack/react-native-paper/commit/ec872a5c6188d8fdded269ac2e98f4ba5881e679

I'm not sure about doing `setState` for the press as that might not feel smooth/instant depending on what's inside the card and how long the re-render is. Maybe we should use a different approach to animate shadows on web, e.g. using the `boxShadow` and `transition` properties directly instead of using the animated API. Probably we can make a component which handles this and Button, Card etc. will use it instead of animating the value manually.

jaulz:
Sounds valid. Then I will close this PR in the meantime and we can check on shadows on web in a different PR.

Issue #764, opened 2019-01-27
Toggle button icon rotates when switching on React Native Web


satya164:
![kapture 2019-01-27 at 14 40 21](https://user-images.githubusercontent.com/1174278/51801689-8456a400-2241-11e9-8cdb-d9825a87ccb7.gif)

jaulz:
See https://github.com/callstack/react-native-paper/pull/780

PR #780, opened 2019-01-29
fix: animated icon change

jaulz:
### Motivation

This PR introduces a new property `animated` on `IconButton` which determines whether an icon change is animated or not. For example, for `ToggleButton` we do not want to have an animation but for `Appbar.Action` we want it (i.e. for contextual appbar). This will fix https://github.com/callstack/react-native-paper/issues/764.

### Test plan

See examples app.


satya164:
@all-contributors please add jaulz for code and docs

Issue #782, opened 2019-01-29
Shadow doesn't animate on Web (Button, Card etc.)


satya164:

idl99:
Yes, i've experienced the same, using the Button. When the button mode is contained, the ripple doesn't appear. Tried changing rippleColor but to no effect. The ripple doesn't overlay on top of contained button.

Trancever:
I believe this work properly in v3.x.x

PR #785, opened 2019-01-30
fix: consistent elevation behavior across platforms

jaulz:
### Motivation

This PR introduces a consistent behavior across platforms for elevations. It seems that on web we need to animate width and height in order to trigger the animation correctly. However, when I did that with the existing interpolations the shadow was pretty dark. So I changed the height to 0 as well and in my opinion it still looks good on native. 

### Test plan

Open example app on web and press "Pressable Chameleon" card. Before this PR, the card elevation does not appear. After this PR, the card elevation does appear and is animated.

#### Web:
Unpressed:
unpressed

Pressed:
pressed

#### iOS:
Unpressed:
Simulator Screen Shot - iPhone 8 - 2019-03-14 at 14 02 56

Pressed:
Simulator Screen Shot - iPhone 8 - 2019-03-14 at 14 03 05


jaulz:
@satya164 could you please check if the look on native is fine for you? If you confirm it should fix https://github.com/callstack/react-native-paper/issues/782 😊 If not, we could think about setting height to 0 on web and keep the existing interpolation on native.

satya164:
cc @Trancever

Trancever:
Hey @jaulz It looks good to me on native, do you have any reference how it should look like on web? I couldn't find any good example on material.io..

jaulz:
@Trancever I don't think there is any specification for web but instead it should be the same on native and web but without this PR there is no shadow at all.

Trancever:
@satya164 Any comments on this? For me this shadow on web is a bit too sharp, but if there is nothing we can do, I am ok with merging it.

ferrannp:
Is this PR not affecting android and iOS? Are they still working the same?

jaulz:
@ferrannp I just pushed another commit which should make it absolutely equal on web and native. Please have a look at the screenshots I just took and added to the original text. And indeed @Trancever was right that the shadow on web was too sharp and the issue was that the opacity and color were not respected as long as they were plain numbers and no `Animated.Value`s.

src/components/Card/Card.js
```diff
@@ -93,6 +93,15 @@ class Card extends React.Component<Props, State> {
     elevation: new Animated.Value(this.props.elevation),
   };
 
+  componentDidUpdate(prevProps) {
+    if (this.props.elevation && this.props.elevation !== prevProps.elevation) {
```
satya164: How about moving this to `Surface` and adding an `animated: boolean` prop to control it? It can reduce code in other places like `Button`

satya164: We should also remove this elevation prop in 3.0 in favor of `elevation` in `style`. cc @Trancever

src/styles/shadow.js
```diff
@@ -3,42 +3,62 @@
...
+  const shadowOpacity = new Animated.Value(0.24);
+
+  // We cannot used strings in Animated.Value and hence we need to interpolate something
+  const shadowColor = animatedElevation.interpolate({
+    inputRange,
+    outputRange: inputRange.map(() => Colors.black),
+  });
+
+  // Width and height are required to be numbers on web (i.e. not undefined)
+  const width = animatedElevation.interpolate({
```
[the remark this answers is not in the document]
satya164: Also a comment saying why we use animated value instead of a number here will be useful

src/components/Card/Card.js
jaulz: I just removed this piece since I think it actually does not belong to the original purpose of the PR but anyway a good idea.

src/styles/shadow.js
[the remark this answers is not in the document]
jaulz: Just tidied up and the code should be easier to read... also I created a return type to ensure that we always return animated values.

Trancever:
@satya164 Can we merge this?

Trancever:
@jaulz I am not sure why but when I am on this branch I can't see Ripple effect, both on web and mobile. Could you check it?

jaulz:
@Trancever I think it's because the ripple effect for cards was only added to 3.0 in https://github.com/callstack/react-native-paper/pull/983 and this PR is for master

Trancever:
@jaulz I am talking about whole app, every component :D Even `TouchableRipple` behaves like `TouchableHighlight` on Android when I am on this branch. I am not sure if something is wrong with this PR or it's just my device... Will check that later on different Android.

jaulz:
@Trancever a rebase fixed it 😊

jaulz:
Can we merge this anytime soon?

Trancever:
@jaulz I'll try to check once again this PR and merge it today

Trancever:
@jaulz There are some issues with native animations. Example app crashed when I tried to use dialog or menu.

jaulz:
Strange. Can you tell me how to reproduce it? I will have a look next week once I am back from vacation.

Trancever:
Sure, Just checkout this branch, run example app on iOS or Android and try to open dialog or menu. Thanks for taking care of it!

jaulz:
@Trancever fix is in... could you please check again? Still there is a warning:
```
Failed prop type: Invalid prop `shadowOffset.width` of type `object` supplied to `SafeAreaView`, expected `number`.
```

I'm not exactly sure why because the docs clearly state the it should be an object: https://facebook.github.io/react-native/docs/shadow-props#shadowoffset

Trancever:
@jaulz Thanks for your patience :D We're shipping it finally!

Trancever:
@jaulz Would you like to try to cherry-pick it to current master?

jaulz:
@Trancever you mean I should migrate it to Typescript? Shall I create another PR for that?

Trancever:
@jaulz Yeah, another PR would be cool. I think the best option is to cherry-pick the merge commit and then resolve the conflicts/migrate to typescript.

Issue #590, opened 2018-10-10
Feature Request: Use San Francisco font instead of Helvetica Neue for iOS


dylancompanjen:
Since iOS9, San Francisco became the default font. The library is using Helvetica Neue though (While San Francisco is available in React Native.
https://material.io/design/platform-guidance/cross-platform-adaptation.html#cross-platform-guidelines

ferrannp:
We should do this but that would be a breaking change. We need to think where and how to adopt it.

andrictham:
@ferrannp 

Right now, if you want to set San Francisco as a font on iOS, you can override the fonts with `System` in your custom theme, like so:

```
import { DefaultTheme, Provider as PaperProvider } from 'react-native-paper';

const ModifiedDefaultTheme = {
  ...DefaultTheme,
  fonts: {
    thin: 'System',
    light: 'System',
    regular: 'System',
    medium: 'System',
  }
}
```

Then do

```
<PaperProvider theme={ModifiedDefaultTheme}>
  <App />
</PaperProvider>
```

The downside of this is, all the fonts will be in the default weight. The reason is, iOS sets font weight using the `fontWeight` style property, but Android doesn’t – it takes in a different `fontFamily` per font weight.

The way theming works will have to be more cross-platform friendly. Is there a way we could somehow use [react-native-typography](https://github.com/hectahertz/react-native-typography) to achieve this?

satya164:
I think we need to change our theme to have these like:

```js
  fonts: {
    regular: { fontFamily: 'System', fontWeight: 'normal' },
    ...
  }
```

However, it'll be a breaking change. For now, the workaround is to bundle those fonts yourself.

andrictham:
@satya164 How would that workaround work, exactly?

brunotalhate:
Following @satya164 comment, the workaround may return a style object with fontFamily and/or fontWeight instead of the fontFamily attribute. Looks like the change lies in change the fontFamily to a font config.

Using the `Button` component as example, the actual style:
```
const fontFamily = theme.fonts.medium
...
<Text
  numberOfLines={1}
  style={[
    styles.label,
    compact && styles.compactLabel,
    textStyle,
    { fontFamily },
  ]}
>
...
```

May become:
```
const fontStyle = theme.fonts.medium
...
<Text
  numberOfLines={1}
  style={[
    styles.label,
    compact && styles.compactLabel,
    textStyle,
    fontStyle,
  ]}
>
```

[...]

PR #786, opened 2019-01-30
feat: configure fonts more granuar

jaulz:
### Motivation

This PR allows to configure fonts more granular (i.e. currently, it let's you specify `fontWeight` but it can be extended). As mentioned by @satya164 in https://github.com/callstack/react-native-paper/issues/590 this will most likely land in 3.0.

### Test plan

See examples app.


jaulz:
By the way, does it make sense to include `letter-spacing` as well in Typography? See https://material.io/design/typography/the-type-system.html#type-scale and https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/ which both specify such.

satya164:
@jaulz yea why not

RichardLindhout:
Excited for this feature ;)

jaulz:
I just had a look at the letter spacing but since the spacing depends on the font and the size it makes it difficult to configure. The only way I could think of at the moment would be to define a letter spacing map in the theme, e.g.:
```
  ios: {
    regular: {
      fontFamily: 'System',
      fontWeight: '400',
      letterSpacing: {
        // See https://github.com/hectahertz/react-native-typography/blob/master/src/helpers/sanFranciscoSpacing.js
        6: 0.246,
        7: 0.223,
        8: 0.208,
        9: 0.171,
        10: 0.12,
        11: 0.06,
        12: 0,
      },
    },
  },
```

@satya164 what do you think?

satya164:
My idea was that for fonts was that we'd be able to do `style={fonts.regular}`. I think if there is a letter spacing map inside, it'll get more complicated. The solution I can think of is that our `Text` component (or another component) accepts a prop to customize the letter spacing and adjusts it automatically based on font size. But I think let's leave it for now, we can come back to it later.

jaulz:
Yep, that's what I thought as well. Though, in the future we could think about a `letterSpacing` function that takes the `fontFamily` and the `fontSize` as input parameters and calculates the letter spacing for predefined fonts - maybe even via the theme. Let's park it for now...

geminiyellow:
good job, that is really what i want. 
i need to set a big font for old men.

Trancever:
Hey @jaulz, Could you change target branch to 3.0?

jaulz:
@Trancever done 😊 unfortunately, I don't have time to check for the conflicts right now.

Trancever:
@jaulz No worries, take your time. Maybe someone from callstack will fix it in the meantime.

jaulz:
@Trancever done

jaulz:
@NoemiRozpara snapshots were still wrong but I'm sure you noticed it 😊 thanks for merging!

PR #787, opened 2019-01-30
feat: fonts on web

jaulz:
### Motivation

This PR adds fonts and stylesheets so users can simply include them from `react-native-paper` instead of relying on any third-party module. Since Roboto is an essential part of the Material design system I think it makes sense to provide them as well. 

### Test plan

See examples app on web.


ferrannp:
Can we add just instructions on how to do this instead of including into Paper?

jaulz:
Instructions will change over time in case required third-party modules (e.g. the Robot fonts) change. With this approach we can make sure that the user will always stick to the latest specification in a non-breaking manner.

ferrannp:
I discussed this PR with the internal and they not want to add this into Paper. I suggest you to start a discussion either in an issue or Discord if you really think we need it but I don't want to keep PRs open that I know we are not gonna merge. Thanks for the work here!

PR #788, opened 2019-01-30
fix: Android P ripple effect using overflow hidden and useForeground

ferrannp:
This https://github.com/facebook/react-native/issues/6480#issuecomment-454183413 works.

### DIFF
--- a/src/components/TouchableRipple.js
+++ b/src/components/TouchableRipple.js
@@ -114,9 +114,6 @@ class TouchableRipple extends React.Component<Props, void> {
       Platform.OS === 'android' &&
       Platform.Version >= ANDROID_VERSION_PIE &&
       borderless;
-    const overflowStyle = {
-      overflow: useForeground ? 'hidden' : 'visible',
-    };
 
     if (TouchableRipple.supported) {
       return (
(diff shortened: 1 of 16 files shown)


src/components/TouchableRipple.js
```diff
@@ -113,11 +108,27 @@ class TouchableRipple extends React.Component<Props, void> {
         .rgb()
         .string();
 
+    // A workaround for ripple on Android P is to use useForeground + overflow: 'hidden'
+    // https://github.com/facebook/react-native/issues/6480
+    const useForeground =
+      Platform.OS === 'android' &&
```
satya164: You should extract `Platform.OS === 'android' && Platform.Version >= ANDROID_VERSION_PIE` to a `shouldUseForeground` variable and use it in both

src/components/TouchableRipple.js
```diff
@@ -113,11 +108,27 @@ class TouchableRipple extends React.Component<Props, void> {
...
+      Platform.Version >= ANDROID_VERSION_PIE &&
+      borderless;
+    const overflowStyle = {
+      overflow:
+        Platform.OS === 'android' &&
```
satya164: Can replace this condition with `useForeground`

RichardLindhout:
@ferrannp Nice work! 👍

PR #796, opened 2019-02-01
feat: card title

jaulz:
### Motivation

This component implements `Card.Title` as described here:
https://material.io/design/components/cards.html#specs

### Test plan

From the example app:
image

image

image


src/components/Card/CardTitle.js
```diff
@@ -0,0 +1,144 @@
...
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    paddingLeft: 16,
+  },
+
+  avatar: {
+    justifyContent: 'center',
+    marginRight: 16,
+    height: 40,
```
Trancever: Shouldn't we also add `width: 40` here?

src/components/Card/CardContent.js
```diff
@@ -57,23 +58,27 @@ class CardContent extends React.Component<Props> {
...
+      if (prev === cover || prev === title) {
         contentStyle = styles.only;
       } else {
         contentStyle = styles.last;
       }
-    } else if (prev === cover) {
+    } else if (prev === cover || prev === title) {
       contentStyle = styles.first;
-    } else if (next === cover) {
+    } else if (next === cover || prev === title) {
```
Trancever: Is this correct?

Trancever:
Btw, This card with Title under Cover doesn't look good on screenshot, there is not enough space between them, what do you think @jaulz ?

src/components/Card/CardContent.js
jaulz: No, it's fixed now, thanks 😊

src/components/Card/CardTitle.js
jaulz: Yep, added it.

jaulz:
> Btw, This card with Title under Cover doesn't look good on screenshot, there is not enough space between them, what do you think @jaulz ?

Just updated the screenshots after the new commits 😊 You are right, some space was missing.

Trancever:
Thanks for great work!

PR #805, opened 2019-02-04
fix ToggleButton onPress

usrbowe:
### Motivation

When using `ToggleButton` like this:
```js
<ToggleButton
    icon="update"
    status={jobSort === "recent" ? "checked" : "unchecked"}
    disabled={jobSort === "recent"}
    value={"recent"}
    onPress={value => {
        console.log(value);
    }}
/>
```
the output after pressing on button is `SyntheticEvent`.  Expected output should be the value either `checked` or `unchecked`.

### Test plan

When onPress prop is available, expect return value to be `checked || unchecked`.

### DIFF
--- a/src/components/ToggleButton/ToggleButton.js
+++ b/src/components/ToggleButton/ToggleButton.js
@@ -126,9 +126,9 @@ class ToggleButton extends React.Component<Props> {
             <IconButton
               borderless={false}
               icon={icon}
-              onPress={() => {
+              onPress={(e) => {
                 if (this.props.onPress) {
-                  this.props.onPress(status);
+                  this.props.onPress(e);
                 }
 
                 if (context) {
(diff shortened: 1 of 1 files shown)


Trancever:
@usrbowe What is your use case? I don't think it is good pattern for controlled button.

usrbowe:
@Trancever For now I don't use it under a `ToggleButton.Group`. Have two ToggleButtons which are linked to one state. So only one of the two can be active.
But I think my use case doesn't have to do anything with the code change I propose. The fix is passing properly back the status for the callback function.

satya164:
Oops xD
I think we should fix `onPress` not being called correctly, but shouldn't pass `status` as an argument to keep it same as radio button: https://github.com/callstack/react-native-paper/blob/fb6762ee10ed72b51b0f871e535ed119c3ebe212/src/components/RadioButtonIOS.js#L28

PR #806, opened 2019-02-04
fix: fix path to ToggleButtonGroup component.

Trancever:
Path to ToggleButtonGroup is wrong and that's why it's not displayed correctly in docs.


satya164:
@all-contributors add @Trancever for code please

PR #810, opened 2019-02-04
Remove whitespace around icon selection

usrbowe:
### Motivation

On [docs/react-native-paper/icons](https://callstack.github.io/react-native-paper/icons.html), when I click on icon and copy this selection to clipboard it contains whitespace after.

Example:
```js
1  account-balance
2
```
The number represents lines in editor.

To prevent whitespace we can select the inner text node, which will result in trimmed string like this:
```js
1 account-balance
```


satya164:
Thanks :D

PR #812, opened 2019-02-05
fix: pass down accessible from Card props to Touchable for Appium tests

t-d-o-g:
### Motivation

The testID prop is not accessible on children of the TouchableWithoutFeedback component without explicitly setting accessible={false}. So accessible needs to be passed to touchable to find elements by testID value for Appium testing.

### Test plan

TextInput in Card component will be made accessible by setting accessible={false\} on Card
```
<Card style={styles.container} accessible={false}>
  <TextInput 
    style={styles.inputContainerStyle}
    testID='text-input'
    mode='outlined'
  />
</Card>
```

Test will validate that TextInput field is accessible
```
it('test that text input field is accessible', () => {
  $('~text-input').addValue('foo')
});
```

### DIFF
--- a/src/components/Card/Card.js
+++ b/src/components/Card/Card.js
@@ -40,6 +40,10 @@ type Props = React.ElementConfig<typeof Surface> & {|
    * Pass down testID from card props to touchable
    */
   testID?: string,
+  /**
+   * Pass down accessible from card props to touchable
+   */
+  accessible?: boolean,
 |};
 
 type State = {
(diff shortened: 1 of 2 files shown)


BoazWarshawsky:
Hi, I am also having this issue with `TouchableHighlight` which they are children of `SafeAreaView`
Any idea how to solve it? adding `accessible={false}` into `SafeAreaView` did not solve the problem

Issue #657, opened 2018-11-15
TextInput outlined mode: label has unknown shadow


mehulmpt:
### Current behaviour
Please see attached:

![screenshot 2018-11-15 at 7 01 25 pm](https://user-images.githubusercontent.com/10388889/48556136-13257700-e909-11e8-9c63-6ea064be8d79.png)

### Expected behaviour

There should not be any weird shadow.

### Code sample

Here's what I'm using to render this input field:

```jsx
<TextInput
	label='Name'
	testID="input"
	mode="outlined"
	theme={{ colors: { placeholder: 'grey', background: '#f5f6f5', text: 'grey', primary: '#5d5d5d' }}}
	style={styles.input}
	value={ this.state.name === undefined ? name : this.state.name }
	onChangeText={name => this.setState({ name })}
/>
```

and here's the parent config:

```js
const theme = {
	...DefaultTheme,
	colors: {
		...DefaultTheme.colors,
		text: '#565656',
		disabled: '#ffffff',
		primary: '#3f51b5'
	},
	fonts: {
		regular: "AirbnbCereal"+fontSuffix+"-Book",
		medium: "AirbnbCereal"+fontSuffix+"-Medium",
		bold: "AirbnbCereal"+fontSuffix+"-Bold"
	}
}
```

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | android
| react-native          | 0.57.3
| react-native-paper    | 2.2.4
| node                  |  8.9
| npm or yarn           | yarn

raajnadar:
Hey, create a reproduce able [snack here](https://snack.expo.io), It will be easier to debug the issue

mehulmpt:
No its still same @ferrannp

imerkle:
Same

ayaanmd21:
same here

ayaanmd21:
please fix it

raajnadar:
No one can help you without an issue reproducing snack or github repo.

imerkle:
Its fixed now atleast for me

ayaanmd21:
how

ayaanmd21:
> No one can help you without an issue reproducing snack or github repo. Comments like "same", "me too", "+1" won't help to find the issue.

in snack its working fine

raajnadar:
Generate this issue in a test repository.

ayaanmd21:
there are two label which label text color is transparent that is creating shawdow

TheRealNate:
[...]

PR #815, opened 2019-02-06
Updated the color of the Label to match the background color. 

Romain-Spielmann:
Fixes the "shadow" problem in the issue #657

The label on the outlined mode had a "shadow" like shown in the issue, caused by the "transparent" color of the font. I've set the color to the same as the background color of the label

### Motivation

I had the same problem as the issue in my application

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -510,6 +510,7 @@ class TextInput extends React.Component<Props, State> {
             style={[
               styles.outlinedLabelBackground,
               {
+                color: backgroundColor,
                 backgroundColor,
                 fontFamily,
                 fontSize: MINIMIZED_LABEL_FONT_SIZE,


Trancever:
@Romain-Spielmann Hey, thanks for PR! Do you have a small repo that reproduces issue so we could really see if this PR solves it? Do you have any idea why our approach (setting backgroundColor to transparent) is not working correctly?

ferrannp:
Can you provide a repo where I can reproduce the problem that is suppose to happen without this change?

syedquadri1982:
> @romain-spielmann Hey, thanks for PR! Do you have a small repo that reproduces issue so we could really see if this PR solves it, or maybe do you have any idea why our approach (setting backgroundColor to transparent) is not working correctly?

setting backgroundColor to transparent outputs the UI incorrectly, the border line appears on top of the label, has anyone tried to fix this yet?

syedquadri1982:
image

PR #817, opened 2019-02-06
fix: optimize Avatar.Icon look

jaulz:
### Motivation

The icons in `Avatar.Icon` look quite small if the size is high.

### Test plan

Before:
image

After:
image


Trancever:
Hey @jaulz We think that the size after change is much to big. Something around `size * 0.60` looks better and it's more similar to default android icon's look.

jaulz:
@Trancever sure, here we go:
image

PR #818, opened 2019-02-06
Remove console.error from Icon.js

AndrewJack:
### Motivation
the `console.error` looks like it was included in #769 by accident.
it causes a red error screen to popup in DEV mode.
simulator screen shot - iphone 8 plus - 2019-02-06 at 16 38 24

### Test plan
Use react-native-paper with https://github.com/satya164/babel-plugin-optional-require and check if red error screen is shown.

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -10,9 +10,6 @@ try {
   // Optionally require vector-icons
   MaterialIcons = require('react-native-vector-icons/MaterialIcons').default;
 } catch (e) {
-  // eslint-disable-next-line no-console
-  console.error(e);
-
   if (global.__expo && global.__expo.Icon && global.__expo.Icon.MaterialIcons) {
     // Snack doesn't properly bundle vector icons from subpath
     // Use icons from the __expo global if available


satya164:
I think we should move it to the component instead of removing it, so you only see the error if you try to use material icons instead of removing it entirely.

AndrewJack:
A console.warn already exists for that doesn't it? line 21?

satya164:
@AndrewJack yes, but it doesn't show the actual error message. If some error occurred other than module not found (eg. some error from Babel), it's not obvious without the error message.

AndrewJack:
Ah ok, I'm a bit busy right now, butI may have time to get my head around this tomorrow. If you haven't fixed it already 😁.

AndrewJack:
Thanks @satya164 💯

robbiemu:
this shouldn't have been closed should it? stay said he _would get around to it_, not _let's just use this_. this merge breaks error reporting in the case of a configuration error

AndrewJack:
@robbiemu This PR was closed, not merged.

@satya164 fixed it in this commit https://github.com/callstack/react-native-paper/commit/bd512f3bbc4e87da4424b50c098234bc2a4178fe

PR #829, opened 2019-02-09
docs: Add new showcase project

gawrysiak:
Here's a Paper based app built for my customer. It was an intense process to create it along with the rapid growth of Paper (starting from ver 1), doing many redesigns and customizations, but it somehow made it. The app doesn't have public registration yet, but hopefully that's okay to show anyway. Thanks to all the people involved in creating components and fixes :+1:

Screenshot resolution was adjusted to match Showman.


ferrannp:
Awesome @gawrysiak ! Would love to see more from people in showcase 😄

Issue #830, opened 2019-02-09
Remove SafeAreaView around Snackbar [suggestion]


RichardLindhout:
It should be better to remove the SafeAreaView around the snackbar because now it's difficult to show it on top of a FAB button. Or e.g. on top of something instead of bottom.

Maybe we could have two components Snackbar and SnackbarContent or something.

Trancever:
Hi @RichardLindhout Can you provide Snack that reproduces issue?

RichardLindhout:
Your pull request could fix the issue (I would only need to add margin to my snackbar of the fab height)
But I don't know. Personally I like freedom and create the safeareviews myself instead of every component. Like three SafeAreaView (one in my screen, one in my fab, one in my snackbar)

RichardLindhout:
+ sometimes I would like the snackbar component just to stay there without duration or something like that and handle it manually.


Material-UI references:
https://material-ui.com/api/snackbar-content/
https://material-ui.com/api/snackbar/

PR #831, opened 2019-02-10
fix: Add SafeAreaView to FAB Group

gawrysiak:
FAB Group currently doesn't support SafeAreaView. It results in an inconsistent behavior between different phones when using a bottom navigation or such.

Since SafeAreaView is already there for components like Snackbar or BottomNavigation, it might be a good idea to add it here as well.

It's also not so obvious to handle it per project cause of the dark overlay that breaks when SafeAreaView applied outside the component. For now I've ended up using react-native-iphone-x-helper package for my local projects to apply different padding values when notch is there, so hopefully the PR provides a better solution.

Before:
unadjustednonraw_thumb_b4a

After:
unadjustednonraw_thumb_b4b

### DIFF
--- a/src/components/FAB/FABGroup.js
+++ b/src/components/FAB/FABGroup.js
@@ -7,6 +7,7 @@ import {
   Animated,
   TouchableWithoutFeedback,
   StatusBar,
+  SafeAreaView,
 } from 'react-native';
 import { polyfill } from 'react-lifecycles-compat';
 import color from 'color';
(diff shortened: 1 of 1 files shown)


Trancever:
Hey @gawrysiak Thanks for PR. Should this fix this issue https://github.com/callstack/react-native-paper/issues/830 ?

gawrysiak:
hi @Trancever, I don't think that's related. The issue seems to be about standard FAB + Snackbar. Although it might be worth looking into how does it all play together

RichardLindhout:
I think we should *not* add SafeAreaView around all components but reference it in the docs it's available so we don't have safeareaviews all over the place.

gawrysiak:
@RichardLindhout normally I'm using a SafeAreaView inside my app files, so I don't really need that for most of the components, but this case is slightly different. The component inside is using `StyleSheet.absoluteFillObject`, so my SafeAreaView is ignored. It's not really possible to apply it from outside

RichardLindhout:
Ah got it ;-)

Issue #823, opened 2019-02-07
questions about TextInput Types


buncismamen:
is `onBlur` in react-native-paper's TextField

intended to be different than `onBlur`  in React-Native's TextField ?

because I need to pass function from `Formik` that supposed to be parameter for React-Native's TextField `onBlur`

for now it gives this error 
```
Type `(e: any) => void` is not assignable to type `() => any`
```
if this is because not updated Types I think I could send a PR

### Current behaviour
```
export interface TextInputProps extends NativeTextInputProps {
...
   onBlur?: () => any;
...
}
```
### Expected behaviour

onBlur type should have same type as react-native TextField

```
export interface TextInputProps extends ViewProps, TextInputIOSProps, TextInputAndroidProps, AccessibilityProps {
...
    onBlur?: (e: NativeSyntheticEvent<TextInputFocusEventData>) => void;
...
}
```
### Proposed behaviour
if onBlur from react-native-paper supposed to be different than onBlur from react-native
to make it compatible I propose this change to the typing 
```
export interface TextInputProps extends NativeTextInputProps {
...
   onBlur?: (e?: any) => any;
...
}
```
### Your Environment

| software              | version
| --------------------- | -------
| react-native          | "0.58.3"
| react-native-paper    | ^2.10.1"

buncismamen:
solved in #833

PR #833, opened 2019-02-10
update TextInput Typing

buncismamen:
update the types
so it have compatibility with React-Native's TextInput Typing since react-native-paper is extending it

its related to my questions in here #823

### DIFF
--- a/typings/components/TextInput.d.ts
+++ b/typings/components/TextInput.d.ts
@@ -12,8 +12,8 @@ export interface TextInputProps extends NativeTextInputProps {
   underlineColor?: string;
   multiline?: boolean;
   numberOfLines?: number;
-  onFocus?: () => any;
-  onBlur?: () => any;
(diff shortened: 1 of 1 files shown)


typings/components/TextInput.d.ts
```diff
@@ -12,8 +12,8 @@ export interface TextInputProps extends NativeTextInputProps {
   underlineColor?: string;
   multiline?: boolean;
   numberOfLines?: number;
-  onFocus?: () => any;
-  onBlur?: () => any;
+  onFocus?: (e?: any) => any;
```
satya164: Can we change the types to the actual types instead of any?

buncismamen: @satya164 I've changed it to be the exact type from react-native's

satya164: @buncismamen will be awesome

PR #835, opened 2019-02-11
fix: limit Dialog.ScrollArea height

jaulz:
### Motivation

If you don't set `maxHeight` on the `DialogScrollArea` itself but instead on `Dialog` (which is mostly what you want) the content of the dialog takes too much space and the actions won't be visible anymore:
image

### Test plan

Open the example for the dialog with long text:
image

Tests on iDevice and web.

### DIFF
--- a/example/src/Examples/Dialogs/DialogWithLongText.js
+++ b/example/src/Examples/Dialogs/DialogWithLongText.js
@@ -15,7 +15,7 @@ const DialogWithLongText = ({
     <Dialog
       onDismiss={close}
       visible={visible}
-      style={{ maxHeight: 0.9 * Dimensions.get('window').height }}
+      style={{ maxHeight: 0.6 * Dimensions.get('window').height }}
     >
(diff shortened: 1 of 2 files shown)


example/src/Examples/Dialogs/DialogWithLongText.js
```diff
@@ -12,9 +12,13 @@ const DialogWithLongText = ({
   close: Function,
 }) => (
   <Portal>
-    <Dialog onDismiss={close} visible={visible}>
+    <Dialog
+      onDismiss={close}
+      visible={visible}
+      style={{ maxHeight: 0.9 * Dimensions.get('window').height }}
```
Trancever: Can we change it to something smaller, e.g height * 0.5? That big modal does not look good :D

jaulz: It's implemented now.

Trancever:
Thanks!

PR #836, opened 2019-02-11
fix: automatically apply contentContainerStyle in Dialog.ScrollArea

jaulz:
### Motivation

`Dialog.ScrollArea` is supposed to be a container for any scrollable content. Hence, it should be safe to assume that its child has a `contentContainerStyle` prop (see https://facebook.github.io/react-native/docs/scrollview#contentcontainerstyle). This change automatically applies the padding to the content instead of the wrapper. 

~~Since this is a breaking change I would suggest to introduce it with 3.0.~~ I changed the API so the behaviour is not breaking and still compatible to the current version.

### Test plan

Open dialog examples with scroll areas:
- "Show dialog with long text" shows how to use React nodes as children
- "Show dialog with radio buttons" shows to use a callback as children

### DIFF
--- a/example/src/Examples/Dialogs/DialogWithLongText.js
+++ b/example/src/Examples/Dialogs/DialogWithLongText.js
@@ -14,8 +14,8 @@ const DialogWithLongText = ({
   <Portal>
     <Dialog onDismiss={close} visible={visible}>
       <Dialog.Title>Alert</Dialog.Title>
-      <Dialog.ScrollArea style={{ maxHeight: 220, paddingHorizontal: 0 }}>
-        <ScrollView contentContainerStyle={{ paddingHorizontal: 24 }}>
+      <Dialog.ScrollArea style={{ maxHeight: 220 }}>
+        <ScrollView>
           <Paragraph>
             Material is the metaphor
             {'\n'}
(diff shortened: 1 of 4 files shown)


jaulz:
@satya164 @Trancever I slightly changed the API so it behaves more like the `image` callback in `Banner` and allows more flexible scenarios. What do you think?

jaulz:
Hey @Trancever, could you take a look at it? Please let me know if you feel that this should not be merged at all. Thanks!

Trancever:
Hi @jaulz . To be honest, I don't think it should be merged. There is no really difference between applying padding on a wrapper or content and also keeping api as simple as possible is our goal.

jaulz:
@Trancever thanks for the feedback! Though, in my opinion we should also try to minimize the effort for developers to have a user interface that follows the Material design guidelines. Any design related styles should be part of the library and not something the user needs to implement manually. 

We could also think about a new component `Dialog.ScrollView` that has the style of a `Dialog.ScrollArea` but uses a `ScrollView` instead. However, if you then want to use a `FlatList` or `SectionList` inside the dialog you would face the same problem again. Since lists also share `ScrollView` props I think that the approach suggested in this PR offers both the flexibility to always apply correct styles (e.g. also to apply `indicatorStyle` automatically) and keep any future migration effort at minimum (e.g. if Material guidelines change in the future). This PR is non-breaking so the developer can still pass React elements if he wants, keeping the API simple as today.

Trancever:
Hey @jaulz We decided not to merge this, cause we don't see any real value this PR adds. What can be accomplished thanks to that PR can be also accomplished using current approach.

If you want, you can create an issue, so we can discuss is further. I hope more people from our team will be able to answer and discuss.

PR #842, opened 2019-02-15
fix: update types

jaulz:
### Motivation

Some types were inaccurate so I implemented the following fixes:
- updated `ProgressBar` according to https://facebook.github.io/react-native/docs/progressbarandroid
- exported complete `Avatar` component
- remove unnecessary `StyledProps` in `Card`

### Test plan

n/a

### DIFF
--- a/typings/components/Avatar.d.ts
+++ b/typings/components/Avatar.d.ts
@@ -30,3 +30,9 @@ export interface TextProps {
 }
 
 export declare class Text extends React.Component<TextProps> {}
+
+export declare class Avatar {
+  static Icon: Icon;
+  static Image: Image;
+  static Text: Text;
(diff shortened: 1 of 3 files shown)


typings/components/Avatar.d.ts
```diff
@@ -30,3 +30,9 @@ export interface TextProps {
 }
 
 export declare class Text extends React.Component<TextProps> {}
+
+export declare class Avatar {
+  static Icon: Icon;
+  static Image: Image;
+  static Text: Text;
+}
```
satya164: Not related to your changes, but 
image

jaulz: Good catch, thanks!

PR #853, opened 2019-02-19
Fix react-native-web example

dalcib:
1. Add the package `react-dom` that was missing.
2. Update `react` to `16.7.0` to fix this error:
`index.js:28 Uncaught TypeError: (0 , _unstableNativeDependencies.injectEventPluginsByName) is not a function`
3. Update `react-native-web` to `10.0.0`


example/package.json
```diff
@@ -18,11 +18,12 @@
     "create-react-context": "^0.2.3",
     "expo": "^31.0.3",
     "file-loader": "^3.0.1",
-    "react": "16.5.0",
+    "react": "16.7.0",
```
satya164: Pretty sure upgrading react version will break the react natve app

dalcib:
react 16.5.1 also fix the problem, but it is better to wait for a new version of Expo.

PR #855, opened 2019-02-19
fix: When we switch the visibility props to quickly the Snackbar stay hidden even if the visibility props is set to true

zer0nim:
### Motivation

When we switch the visibility props to quickly the Snackbar stay hidden even if the **visibility** props is set to `true`.

The problem is related to the state **hidden** which prevent the Snackbar from rendering when not visible.
The **show** function set **hidden** to `false` directly, but **hide** function set **hidden** to `true` only on the animation callback.

This is the desired workflow but if we call **show** just after **hide** and before the end of the **hide** animation, the snackbar will be hidden directly after the animation finish.

To handle that I implemented a queue to wait **show**/**hide** animation finished before calling them again. 

### Test plan

Toggle visibility props quickly

### DIFF
--- a/src/components/Snackbar.js
+++ b/src/components/Snackbar.js
@@ -46,6 +46,8 @@ type Props = {|
 type State = {
   opacity: Animated.Value,
   hidden: boolean,
+  inProgress: boolean,
+  queue: Array<boolean>,
 };
 
 const DURATION_SHORT = 4000;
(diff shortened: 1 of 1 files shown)


Trancever:
Hi @zer0nim . I created https://github.com/callstack/react-native-paper/pull/860 which also fixes this issue, but without this queue solution that we think is not good for this use case.

PR #861, opened 2019-02-21
Make font-size smaller in search bar

RichardLindhout:
https://www.slashgear.com/gmail-app-android-ios-material-design-update-released-21566652/
Looks like it's te way Google want it

### DIFF
--- a/src/components/Searchbar.js
+++ b/src/components/Searchbar.js
@@ -188,7 +188,7 @@ const styles = StyleSheet.create({
   },
   input: {
     flex: 1,
-    fontSize: 18,
+    fontSize: 16,
     paddingLeft: 8,
     alignSelf: 'stretch',
     textAlign: I18nManager.isRTL ? 'right' : 'left',


src/components/Searchbar.js
```diff
@@ -188,7 +188,7 @@ const styles = StyleSheet.create({
   },
   input: {
     flex: 1,
-    fontSize: 18,
+    fontSize: 16,
```
ferrannp: Tbh I use `Searchbar` and I prefer `18`. We are passing `{...rest}` to `TextInput` but with style won't work here. I would prefer to have a prop like `inputStyle` and merge it with the other styles from the `TextInput` inside the `Searchbar` in order to change `fontSize` if desired.

ferrannp:
Also CI needs to pass.

src/components/Searchbar.js
RichardLindhout: That's a better idea indeed, I will push a pull request when I have time.

PR #864, opened 2019-02-22
Update Searchbar.test.js.snap

RichardLindhout:


ferrannp:
This should be in #861 PR not in a separated one.

PR #866, opened 2019-02-23
editable and disabled is different

### LINKED ISSUE #865: TextInput editable not works
### Current behaviour

TextInput editable not works

### Expected behaviour
TextInput editable and disabled is different behavior 

### Code sample

```js
import { TextInput } from 'react-native-paper';

  state = {
    text: ''
  };

  render(){
    return (
      <TextInput
        label='Email'
        value={this.state.text}
        editable={false}
      />
    );
  }
}
```

### What have you tried

see https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput.js#L647 , you can see here not disabled is equal to editable

### Your Environment

| ios or android        |
| react-native          |
| react-native-paper    |
| node                  |
| npm or yarn           |
| expo sdk              |

sm2017:
### Motivation

Fixed #865

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -383,6 +383,7 @@ class TextInput extends React.Component<Props, State> {
     const {
       mode,
       disabled,
+      editable,
       label,
       error,
       selectionColor,
@@ -644,7 +645,7 @@ class TextInput extends React.Component<Props, State> {
           onChangeText: this._handleChangeText,
           placeholder: label ? this.state.placeholder : this.props.placeholder,
           placeholderTextColor: placeholderColor,
-          editable: !disabled,
+          editable,
           selectionColor:
             typeof selectionColor === 'undefined'
               ? activeColor


src/components/TextInput.js
```diff
@@ -644,7 +645,7 @@ class TextInput extends React.Component<Props, State> {
           onChangeText: this._handleChangeText,
           placeholder: label ? this.state.placeholder : this.props.placeholder,
           placeholderTextColor: placeholderColor,
-          editable: !disabled,
+          editable,
```
ferrannp: What's wrong with using our prop `disabled` here?

Trancever:
@ferrannp It makes sense, cause we apply different styles based on disabled prop. Users want to have TextInput that looks like currently not disabled TextInput but also they want to make it not clickable at the same time. With current approach it's not possible.

@sm2017 I am ok with that change but have a suggestion. Can we keep previous behaviour when editable prop is not specified and only apply that prop when user explicitly passed it? Thanks to that, it won't be breaking.

sm2017:
> Users want to have TextInput that looks like currently not disabled TextInput but also they want to make it not clickable at the same time. With current approach it's not possible.

Yes, exactly!

@Trancever , why you think that it's breaking now?

Trancever:
@sm2017 Current users relay on this behaviour:

They are passing `disabled={true}` which means that TextInput is not clickable etc. With proposed changes, passing `disabled={true}` will change only styles, but TextInput will be still clickable which is breaking for them. 
So my suggestion is to set `editable={typeof editable !== 'undefined' ? editable : !disabled}`, so users who don't pass editable prop at all will have previous behaviour.

sm2017:
It's OK too

Trancever:
@sm2017 Could you change it as I suggested so we can proceed with a merge?

sm2017:
I'm so busy, take me more time

Trancever:
@sm2017 No worries, take your time :)

raajnadar:
I think instead of merging this PR we can wait till v3 with a breaking change. Because there is also a conflict with textinput component. It is refactored in another PR.

sm2017:
@raajnadar ok

Trancever:
@sm2017 It should be merged into 3.0 branch. Once it's changed, we will proceed with merge

sm2017:
@Trancever https://github.com/callstack/react-native-paper/pull/986

sm2017:
@Trancever https://github.com/callstack/react-native-paper/pull/986

sm2017:
@Trancever https://github.com/callstack/react-native-paper/pull/986 merged at 3.0 branch

PR #872, opened 2019-02-25
Add style overwrite for title and description for ListItem Component

ctravail:
### Motivation

I wanted to be able to change the color of the title/description independently of each other.

An alternative would be to use two different theme colors for title and description rather than having both use 'text', but I would like to be able to control the opacity so my solution allows you to be able to change other aspects besides color

My use case: Want title to be color #041F41 without opacity, but description to black with opacity  Currently, changing the 'text' theme to #041F41 creates a hard to read description.

### Test plan

```jsx
import { List } from "react-native-paper";

const listItem = () => (<List.Item
    title="Title"
    description="Description"
    style = {{
      title: {
        color:'#041F41'
      },
      description: {
        color:'#000000'
      },
    }}
  />)

### DIFF
--- a/src/components/List/ListItem.js
+++ b/src/components/List/ListItem.js
@@ -95,7 +95,7 @@ class ListItem extends React.Component<Props> {
           <View style={[styles.item, styles.content]} pointerEvents="none">
             <Text
               numberOfLines={1}
-              style={[styles.title, { color: titleColor }, style.title]}
+              style={[styles.title, { color: titleColor }, style.title || {}]}
             >
               {title}
             </Text>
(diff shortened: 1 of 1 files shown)


ctravail:
Working on this with @Darkmyr

ctravail:
Team, I will need some assistance in updating your tests. I'd be happy to jump on a call for this, but in the meantime I'm doing research around jest.

satya164:
I think `titleStyle` and `descriptionStyle` props will be better.

Also regarding snapshot tests, run `yarn test -u`

Trancever:
Closing in favour of #911

Issue #286, opened 2018-03-27
Create optional setting to disable all animations entirely


ferrannp:
Useful for example when battery saver on Android is on. We can show how to listen for:

```xml
<action android:name="android.os.action.POWER_SAVE_MODE_CHANGED"/> 
```

onitzschke:
Hi, i'm using the experimental `LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);` and the animation of FAB icon is stuck at the end. Any idea?

satya164:
@onitzschke `LayoutAnimation` is very invasive and can affect other components in unexpected ways. Unfortunately, it's not something we can help with. You should consider removing `LayoutAnimation`.

onitzschke:
Yes, that's the problem. I'll try to abstract the components. Maybe that helps. Thanks.

onitzschke:
For all with the same issue:
I use a custom animation with less duration and it works great now.
```
const CustomAnimation = {
  duration: 100,
  create: {
    type: LayoutAnimation.Types.easeInEaseOut,
    property: LayoutAnimation.Properties.opacity
  },
  update: {
    type: LayoutAnimation.Types.easeInEaseOut,
    duration: 100
  }
};
```

`              LayoutAnimation.configureNext(CustomAnimation);
`

Subv:
For anyone stumbling upon this issue looking for a way to disable all animations, it seems there is now a theme property to do that: https://github.com/callstack/react-native-paper/pull/1594
https://github.com/callstack/react-native-paper/blob/d2f61fbee9bd9bd11daa206313d8069876fbccd5/src/styles/DefaultTheme.tsx#L24

PR #880, opened 2019-03-01
added 'transitionDuration' prop to Modal and Dialog components

AlexeyVlasenko:
### Motivation

Modal and Dialog components have too long fadeIn/fadeOiut animations (for me).

### Test plan

1. You can add transitionDuration to Modal or Dialog component.
2. transitionDuration prop accept number (in this case, both fadeIn/fadeOut animation will use this timing)
3. transitionDuration prop accept object: {enter: number, exit: number}

### DIFF
--- a/src/components/Dialog/Dialog.js
+++ b/src/components/Dialog/Dialog.js
@@ -24,6 +24,10 @@ type Props = {|
    */
   visible: boolean,
   /**
+   * Duration of the Dialog transition.
+   */
+  transitionDuration: number | {enter: number, exit: number},
+  /**
    * Content of the `Dialog`.
    */
   children: React.Node,
@@ -102,6 +106,7 @@ class Dialog extends React.Component<Props, void> {
       dismissable,
       onDismiss,
       visible,
+      transitionDuration,
       style,
       theme,
     } = this.props;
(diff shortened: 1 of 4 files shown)


satya164:
I think we want a property in theme which controls the animation speed: https://github.com/callstack/react-native-paper/issues/286

ferrannp:
I agree we need something like this. The current dialog animation is very slow compared to what it should be.

Trancever:
@satya164 How do you want to have it stored inside theme? We are waiting for your decision :D

satya164:
@Trancever something like `animation.scale: 2.0`. similar to the one in Android developer tools. It'll allow people to speed up animations, or entirely disable them (like we wanted to have). it can also help in debugging animations by slowing them down.

Trancever:
Closing in favour of #942

msageryd:
@Trancever This seems like a good solution, but I'm confused.

- I can see this fix in the source at GH
- The fix is *not* in my installed npm package (2.16.0)
- I cannot see any commit to Modal.tsx which would have added the fix.

Did something go wrong when you converted to Typescript? The Modal.js file in the installed package seems to be an old version without the code for applying scale to the animation durations.

Modal.tsx in the GH master repo has code for applying theme.animation.scale to the animation duration:
https://github.com/callstack/react-native-paper/blob/8767b76db43af475301ad1fa1ba15f3e8d769120/src/components/Modal.tsx#L126

msageryd:
Oh, sorry. I see now that this fix is in version 3. 
I also see that the Master branch seems to be version 3. Does this mean that v3 will be published to npm soon?

PR #883, opened 2019-03-02
Improve OutlinedInput on the web

RichardLindhout:
### Motivation

Autocomplete was rendered above the input + it was too big so it showed above the outline.
schermafbeelding 2019-03-02 om 16 47 42

I have changed the height of the input so the auto fill does not overlap the input.
On the web I make use of the fieldset and label the remove border on the outline.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#Example
This approach is also used by Material-UI.
https://material-ui.com/

### Test plan

WEB:
schermafbeelding 2019-03-02 om 17 07 24

ANDROID:
b9bb3bf1-be37-404d-b9be-4756ddb17686

### DIFF
--- a/example/public/index.html
+++ b/example/public/index.html
@@ -10,7 +10,9 @@
     html, body, #root {
       height: 100%;
     }
-
+    input:focus,textarea:focus{
+      outline:none;
+    }
     #root {
       display: flex;
       flex-direction: column;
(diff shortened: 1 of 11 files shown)


RichardLindhout:
@satya164 this looks even better on web, what do you think about this or something like this implementation.

PR #854, opened 2019-02-19
feat: button inner content can be aligned with alignContent prop

ThingsIDoForLove:
### Motivation
We normally require feature where button content can be  aligned left, right or centered

PR #885, opened 2019-03-02
fix: add default touchable style to always fill parent container with custom styles

RichardLindhout:
https://github.com/callstack/react-native-paper/pull/854


RichardLindhout:
@satya164 this looks even better on web, what do you think about this or something like this implementation.

PR #886, opened 2019-03-02
fix: add default touchable style to always fill parent container with…

RichardLindhout:
Motivation:
Only small part of a custom button was clickable (in web and native apps)

Before:
schermafbeelding 2019-03-02 om 18 38 09

After:
schermafbeelding 2019-03-02 om 18 41 45

It's a different app so colors are a bit different ;)

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -246,7 +246,7 @@ class Button extends React.Component<Props, State> {
           accessibilityStates={disabled ? ['disabled'] : undefined}
           disabled={disabled}
           rippleColor={rippleColor}
-          style={touchableStyle}
+          style={[styles.touchable, touchableStyle]}
         >
           <View style={[styles.content, contentStyle]}>
             {icon && loading !== true ? (
@@ -289,6 +289,10 @@ const styles = StyleSheet.create({
   button: {
     minWidth: 64,
     borderStyle: 'solid',
+  },
+  touchable: {
+    width: '100%',
+    flex: 1,
   },
   compact: {
     minWidth: 'auto',
--- a/src/components/__tests__/__snapshots__/Banner.test.js.snap
+++ b/src/components/__tests__/__snapshots__/Banner.test.js.snap
@@ -205,9 +205,15 @@ exports[`renders visible banner, with action buttons and with image 1`] = `
               Object {
                 "overflow": "hidden",
               },
-              Object {
-                "borderRadius": 4,
-              },
+              Array [
+                Object {
+                  "flex": 1,
+                  "width": "100%",
+                },
+                Object {
+                  "borderRadius": 4,
+                },
+              ],
             ]
           }
         >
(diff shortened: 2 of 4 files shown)


src/components/Button.js
```diff
@@ -290,6 +290,10 @@ const styles = StyleSheet.create({
     minWidth: 64,
     borderStyle: 'solid',
   },
+  touchable: {
+    width: '100%',
+    flex: 1,
```
satya164: Have you tried it in the RN app? I previously had issues with `flex: 1` here. I think just `width: '100%'` should be enough?

RichardLindhout: I did at least on Android. I thought flex: 1 was needed to fill the height. Will test if I have time.

Trancever:
@RichardLindhout Hey, could you tell me steps to reproduce your issue?

RichardLindhout:
```
<Button
                    mode={'contained'}
                    onPress={this._login}
                    type="submit"
                    loading={loading}
                    uppercase={false}
                    style={styles.button}
                    contentStyle={styles.buttonContentStyle}
                    touchableStyle={styles.buttonTouchableStyle}
                  >
                    Inloggen
                  </Button>
```


```
  button: {
    height: 56,
    alignItems: 'center',
    justifyContent: 'center',
    // width: '100%',
  },
  buttonTouchableStyle: {
    height: 56,
    width: '100%',
  },

  buttonContentStyle: {
    height: 56,
    width: '100%',
  },
```

src/components/Button.js
RichardLindhout: Flex: 1 is not needed indeed

RichardLindhout: Flex:1 is needed else you need to set the height on 
contentStyle={styles.buttonContentStyle} and
touchableStyle={styles.buttonTouchableStyle}

RichardLindhout: I think we should even add more flex:1 for the content style also.

Trancever:
Can you try with following styles and let me know about results?

```
  buttonContentStyle: {
    height: 56,
    width: '100%',
  },
```

Don't pass touchableStyle cause there is no such prop in Button component, also I think you don't need to pass style.

RichardLindhout:
```
<Button
  mode={'contained'}
  style={styles.button}
  contentStyle={styles.buttonContentStyle}>
  Inloggen
</Button>
```

```
  button: {
    height: 56,
    alignItems: 'center',
    justifyContent: 'center',
    width: '100%',
  },

  buttonContentStyle: {
    height: 56,
    width: '100%',
  },
```

Results in:
schermafbeelding 2019-03-07 om 13 35 39

RichardLindhout:
I'm pretty sure the touchable needs the width: '100%' somehow to fix this.
Also if we use flex: 1 also one the inner component we maybe only need to set height on the root button.

Trancever:
Height and width should be only set through contentStyle prop. Please remove style prop completely, cause it does not have any impact and try again please.

Trancever:
ezgif com-video-to-gif 3
It works for me when I passed 
```
height: 56,
width: '100%'
```
to contentStyle prop.

Trancever:
@RichardLindhout Do you still have any issue with that?

RichardLindhout:
You're right. This pull request is not needed!

PR #887, opened 2019-03-02
Improve outlined input on the web

RichardLindhout:
### Motivation

Autocomplete was rendered above the input + it was too big so it showed above the outline.
schermafbeelding 2019-03-02 om 16 47 42

### Test plan

WEB:
schermafbeelding 2019-03-02 om 19 26 13

ANDROID:
8f1024f1-eaff-4f5f-ba4a-d6cc0cad215f

PS: I did another pull request which was more changes but did have support for backgroundless labels. This creates even better effect for the label on the web because it uses fieldset en legend to remove the border: https://github.com/callstack/react-native-paper/pull/883

### DIFF
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -655,8 +655,11 @@ class TextInput extends React.Component<Props, State> {
           multiline,
           style: [
             styles.input,
-            mode === 'outlined' && {
+            mode === 'outlined' ? {
               borderRadius: theme.roundness,
+            } : {
+              borderTopLeftRadius: theme.roundness,
+              borderTopRightRadius: theme.roundness,        
             },
             mode === 'outlined'
               ? styles.inputOutlined
--- a/src/components/TextInput.js
+++ b/src/components/TextInput.js
@@ -528,6 +528,7 @@ class TextInput extends React.Component<Props, State> {
             style={[
               styles.outlinedLabelBackground,
               {
+                height: hasActiveOutline ? 2 : 1,
                 backgroundColor,
                 fontFamily,
                 fontSize: MINIMIZED_LABEL_FONT_SIZE,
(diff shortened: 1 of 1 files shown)


src/components/TextInput.js
```diff
@@ -709,6 +712,7 @@ const styles = StyleSheet.create({
     left: 8,
     paddingHorizontal: 4,
     color: 'transparent',
+    borderRadius: 5,
```
satya164: This is probably a mistake?

src/components/TextInput.js
```diff
@@ -655,6 +655,9 @@ class TextInput extends React.Component<Props, State> {
           multiline,
           style: [
             styles.input,
+            mode === 'outlined' && {
+              borderRadius: theme.roundness,
```
satya164: Maybe we should do it in all modes?

src/components/TextInput.js
RichardLindhout: When the roundness is very big it creates a better look with cropping away the border.

src/components/TextInput.js
RichardLindhout: I think that's a good idea if the flat textinputs are also rounded by the theme roundess

RichardLindhout: Only the top are rounded in the flat ine so I think we should not do it.

src/components/TextInput.js
RichardLindhout: schermafbeelding 2019-03-06 om 13 35 48

Trancever:
@satya164 What do we do with this?

RichardLindhout:
It's still needed, but needs a rewrite since the components are changed. Also on native this problem is there too when you fill in with 1password for example.

Trancever:
@RichardLindhout I am closing this, because it's very outdated. If the issue is still here, please submit a new PR or create an RFC issue so we can discuss it before implementing.

PR #890, opened 2019-03-05
fix: make onDismiss optional

jaulz:
### Motivation

There are use cases when you don't need `onDimiss`.

### Test plan

n/a


jaulz:
This is by the way anyway already covered by the code:
https://github.com/callstack/react-native-paper/blob/master/src/components/Modal.js#L137

Trancever:
Thx @jaulz !

jaulz:
@Trancever thanks to you as always 😊

PR #904, opened 2019-03-07
Fix AppbarHeader elevation on Android

okitoxd:
The Appbar.Header had no elevation in Android. This is because you did not write the elevation property when you passed it to style prop.
Tip: Add the elevation property in the shadow function return

### DIFF
--- a/src/components/Appbar/AppbarHeader.js
+++ b/src/components/Appbar/AppbarHeader.js
@@ -111,7 +111,7 @@ class AppbarHeader extends React.Component<Props> {
 
     return (
       <Wrapper
-        style={[{ backgroundColor, zIndex }, elevation && shadow(elevation)]}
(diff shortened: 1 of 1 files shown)


src/components/Appbar/AppbarHeader.js
```diff
@@ -111,7 +111,7 @@ class AppbarHeader extends React.Component<Props> {
 
     return (
       <Wrapper
-        style={[{ backgroundColor, zIndex }, elevation && shadow(elevation)]}
+        style={[{ backgroundColor, zIndex }, elevation && {shadow(elevation), elevation}]}
```
Trancever: Did you test it? It's not valid javascript syntax.

ferrannp:
@satya164 looks like we removed elevation for Android.

PR #908, opened 2019-03-08
Add style override to wrapper of Modal component

Darkmyr:
### Motivation

The motivation for this PR is to allow style overriding to the wrapper for the Modal. This allows us to do things like justify content to the bottom if we so choose. We could also allow just an override to JustifyContent, but this solution is further extendable should users of the library have other style considerations they want to execute.

### Test plan
```jsx
import { Modal, Portal, Text, Button, Surface, withTheme, Divider } from 'react-native-paper';

class ModalExampleComponent extends React.Component {
  state = {
    visible: true,
  };

  _showModal = () => this.setState({ visible: true });
  _hideModal = () => this.setState({ visible: false });

  render() {
    const { visible } = this.state;
    return (
      <Portal>
        <Modal wrapperStyle={{ justifyContent:'flex-end'}} visible={visible} onDismiss={this._hideModal}>
          <View>
          <Surface style={{borderRadius:5, margin:10}}>
              <Text style={{textAlign:'center'}}>Select a Photo</Text>
              <Divider />
              <Button>Take Photo...</Button>
              <Divider />
              <Button>Choose from Library...</Button>
            </Surface>
            <Surface style={{borderRadius:5, margin:10}}>
              <Button onPress={this._hideModal}>Cancel</Button>
            </Surface>
          </View>
        </Modal>
      </Portal>
    );
  }
}
```

image

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -8,6 +8,7 @@ import {
   StyleSheet,
   TouchableWithoutFeedback,
   BackHandler,
+  ViewStyle,
 } from 'react-native';
 import { polyfill } from 'react-lifecycles-compat';
 import Surface from './Surface';
@@ -43,7 +44,7 @@ type Props = {|
    * @optional
    * Overrides Styles for the Wrapper element.
    */
-  wrapperStyle?: StyleSheet.Styles,
+  wrapperStyle?: StyleSheet.NamedStyles<ViewStyle>,
 |};
 
 type State = {
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -43,7 +43,7 @@ type Props = {|
    * @optional
    * Overrides Styles for the Wrapper element.
    */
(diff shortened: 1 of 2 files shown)


satya164:
Why not `{ flex: 1, justifyContent:'flex-end' }` on the `View` inside?

Darkmyr:
Hey team, so I'm new to flow and will need help with the flow imports for type `StyleProp<ViewStyle>` or some equivalent of that. I've tried to get those 2 things imported from `react-native`but I can't seem to figure it out.

Darkmyr:
@satya164 if you run the component, you'll see that doesn't work

Darkmyr:
Hey @satya164,
Just following up, anything else needed?

Trancever:
Pass {{ justifyContent:'flex-end', flex: 1 }} to View component

PR #922, opened 2019-03-14
chore: Remove typescript tests generation

pbitkowski:
- remove typescript tests generation script,
- remove check of typescript generated tests,
- remove ts check from CI

### Motivation

We've got script that extracted example usage from every component docs. Then it was checked with typescript. This thing was created to skip handling TS tests. I think this is not the best solution, in fact it's blocking our PRs right now. 

If we want to keep TS tests, then I propose to keep them in the repo not automagically generate them on CI. 

@satya164 @ferrannp @Trancever @Esemesek WDYT?


Trancever:
I am ok with removing them, but if anybody wants to fix this script, it's fine too.

Esemesek:
We have already briefly discussed this with @pbitkowski and I think we should leave this script as a part of library tools, which we can use in the future to generate tests for new components. We should fix the script and generate tests for all components and commit those tests to the repository.

pbitkowski:
I agree. The problem is generating tests on the fly, not the ts tests in general. So it would be good to do it. Okay, I'll try to make it happen then, @Esemesek :)

Trancever:
I am closing this in favour of #936

Issue #781, opened 2019-01-29
Appbar doesn't have shadow on Web


satya164:

pnkapadia6:
@satya164 Can I pick this up?

satya164:
Go ahead

pnkapadia6:
@satya164 https://github.com/callstack/react-native-paper/pull/923

edmundojr:
`Appbar.Header` doesn't have shadow at all, even on react-native.

Taym95:
@edmundojr work in [progress](https://github.com/callstack/react-native-paper/pull/785) will be fixed!

Alain00:
Any progess on this? the shadow on the Appbar.Header is't working for me too

PR #923, opened 2019-03-14
fix: appbar shadow on web

pnkapadia6:
Issue: https://github.com/callstack/react-native-paper/issues/781

### Motivation

Shadow wasn't visible for the Appbar component on web.
Without increasing the z-index, the shadow wasn't visible. Another way could be adding space between the sibling component (the one rendered below), but the space may or may not be required. So z-index will fix the issue.

Screen Shot 2019-03-15 at 12 02 46 AM


ferrannp:
Hey! Thanks for your PR. We are going to fix this as part of #785 so I'll close this one.

PR #932, opened 2019-03-17
feat FastImage support in CardCover component

espipj:
### Motivation

I discovered `react-native-fast-image` library some days ago and I believe that it improves the performance when loading big amount of images, I wanted to use it in a Card.Cover component yesterday but I wasn't able as it is hard coupled with the Image component. This changes I believe will not strongly affect the use of Card.Cover as it is now but will add the feature for those who want to use FastImages.

#### Features

-   [x] Aggressively cache images.
-   [x] Add authorization headers.
-   [x] Prioritize images.
-   [x] Preload images.
-   [x] GIF support.
-   [x] Border radius.

[FastImage repository on GitHub](https://github.com/DylanVann/react-native-fast-image)

### Test plan

Documentation has been added to show the usage, the only change is adding the prop fastImage when you feel that you want to use it:
 ```js
 import { Card } from 'react-native-paper';
 
 const MyComponent = () => (
   <Card>
     <Card.Cover fastImage source={{ uri: 'https://picsum.photos/700' }} />
   </Card>
 );

 export default MyComponent;
 ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "create-react-context": "^0.2.3",
     "hoist-non-react-statics": "^3.1.0",
     "react-lifecycles-compat": "^3.0.4",
+    "react-native-fast-image": "^5.2.0",
     "react-native-safe-area-view": "^0.12.0"
   },
   "devDependencies": {
--- a/src/components/Card/CardCover.js
+++ b/src/components/Card/CardCover.js
@@ -21,6 +21,10 @@ type Props = React.ElementConfig<typeof Image> & {|
    * @optional
    */
   theme: Theme,
+  /**
+   * @optional
+   */
+  fastImage: boolean,
 |};
 
 /**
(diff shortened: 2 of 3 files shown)


Trancever:
Hi @espipj . Thanks for PR, but I am afraid it can't be merged. `Paper` is UI component library and components that we provide should be only responsible for rendering and not trying to optimize Image caching etc. 

As a workaround I suggest to just copy paste CardCover component into your project and adjust it to your needs.

smkhalsa:
@Trancever, while I agree that adding FastImage as a dependency and using it by default is not the right way to go, I feel the underlying issue deserves more discussion.

`react-native-paper` doesn't currently provide a way to override the default Image component within built-in components like Card.Cover and Avatar.Image. This is an issue for certain use cases such as offline caching. Your suggestion above of re-writing the component is not a very maintainable approach.

Any other ideas?

YajanaRao:
Hi,
I am facing issue with the performance while loading the image as card cover. Progressive loading would help here.

PR #933, opened 2019-03-18
RNW setup: only transpile node_modules

slorber:
I think the newly added rule should not apply to files outside of node modules, to avoid messing up with another js loader that may already be present

This is for example required for my Gatsby plugin setup to work, because Gatsby already has its own babel loader for app sources:
https://github.com/slorber/gatsby-plugin-react-native-web/blob/master/examples/todos/gatsby-node.js
https://gatsby-rnw-todos.netlify.com/

### DIFF
--- a/docs/pages/4.react-native-web.md
+++ b/docs/pages/4.react-native-web.md
@@ -30,6 +30,7 @@ yarn add --dev react-app-rewired
 module.exports = function override(config, env) {
   config.module.rules.push({
     test: /\.js$/,
+    include: /node_modules/,
     exclude: /node_modules[/\\](?!react-native-paper|react-native-vector-icons|react-native-safe-area-view)/,
     use: {
       loader: "babel-loader",
@@ -175,6 +176,7 @@ Now, add the following in the `module.rules` array in your webpack config:
 ```js
 {
   test: /\.js$/,
+  include: /node_modules/,
   exclude: /node_modules[/\\](?!react-native-paper|react-native-vector-icons|react-native-safe-area-view)/,
   use: {
     loader: 'babel-loader',


ferrannp:
I think we can update docs to say you need this for Gatsby but it cannot be the main example.

slorber:
@ferrannp just wondering can you expand on your thinking?

For me:
- You are pushing a new rule to an array, so you basically expect the user to already handle his own babel config for his own sources (at least it's what I assumed)
- That "extra rule" should only be responsible for node modules transpilation (which is not common practice in web envs and unlikely to already be in existing user setup)
- That "extra rule" should not conflict with existing user rule for transpiling regular user sources (in ./src)

So all of this depends on weither or not you consider the new rule you push an "extra rule" that completes an existing rule, or if you expect to use a single babel rule and merge it with your RNW rule to form only one rule in the end (but in this case I'd argue the use of push is confusing)

satya164:
Hey @slorber

Thanks for the PR. I think there's a bit of misunderstanding about the docs. In your PR, you have 2 parts:

1. The first one is for CRA, where we're pushing an extra new rule. And as you said, it should only include the files in `node_modules`
2. The second one is for a vanilla configuration, where we assume that you're configuring everything from scratch and don't have another babel loader config.

But I agree it's very not clear regarding that. I think what we should do is, mention that *if* the user is already compiling their application code with Babel and have an existing `babel-loader` config, or using a tool such as CRA, Gatsby, Next.js etc., then they need to add this extra `include`. But if they don't have an existing babel loader config, then they don't need the `include`.

slorber:
Hi,

In my PR maybe I should not modify the 2nd part and only modify the first part with push then?

Or maybe both parts should use "push". Because maybe your own transpilation config should not "pollute" the user one? I mean he may not be using flow in his own codebase so asking him to add the flow preset for his whole babel loader may be unnecessary (but probably wouldn't harm much)

jaysbytes:
It's inactive for a while so I'm closing this

PR #936, opened 2019-03-18
chore: remove TS test generation from CI, fix ts gen script, add tests

pbitkowski:
- remove TS generation script from CI checks (leave only testing, not generation - script will remain in repo as utility),
- generate all TS tests, fix them and commit,
- fix broken example usages.

I propose to remove TS generation script from CI checks. After fixing script that extract TS tests from example usage I commited all tests and fixed them (also fixed couple of broken examples). Squash and merge please, we need to move forward with this TS madness. 

### Motivation

Tests shouldn't be generated at runtime. 

### Test plan

run `yarn typescript`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,7 +53,6 @@ android/keystores/debug.keystore
 # Build
 #
 dist/
-__ts-tests__/
 
 # Release
 #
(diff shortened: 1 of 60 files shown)


Trancever:
@Esemesek Do you want to take a look at this?

PR #942, opened 2019-03-21
feat: add scale to control modal animation speed

Trancever:


ferrannp:
@Trancever would this control only Modal/Dialog? More in the future? I want this PR ASAP :D but how the user knows that the theme prop `scale` is for the dialog animation speed?

Trancever:
@ferrannp For now it's just for modal but we want to add support for other components too. Feel free to send PRs :D

PR #943, opened 2019-03-21
fix: Searchbar TS

gawrysiak:
Missing TS typings for Searchbar methods

Screen Shot 2019-03-21 at 15 12 50

### DIFF
--- a/typings/components/Searchbar.d.ts
+++ b/typings/components/Searchbar.d.ts
@@ -9,4 +9,9 @@ export interface SearchbarProps extends TextInputProps {
   onChangeText?: (query: string) => void;
 }
 
-export declare class Searchbar extends React.Component<SearchbarProps> {}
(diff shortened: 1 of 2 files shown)


typings/components/Searchbar.d.ts
```diff
@@ -9,4 +9,9 @@ export interface SearchbarProps extends TextInputProps {
   onChangeText?: (query: string) => void;
 }
 
-export declare class Searchbar extends React.Component<SearchbarProps> {}
+export declare class Searchbar extends React.Component<SearchbarProps> {
+  isFocused: () => boolean;
+  clear: () => void;
```
Trancever: Why is type for clear and focus different? Can't w unify it?

gawrysiak: yeah, you're right, it was copied from TextInput, so sent a fix for TextInput as well now

PR #944, opened 2019-03-21
List item ellipsize position

Darkmyr:
### Motivation

Ability to change the ellipsize position in the Title/Description of list items

### Test plan
add the following to your list item properties and see that you can change the ellipses position in the following text elements. 
```
titleEllipsizeMode='middle'
descriptionEllipsizeMode='middle'
```
image


ferrannp:
Thanks for your PR!

PR #946, opened 2019-03-22
Use native driver in Modal

auzwang:
### Motivation
The animation performance for showing and hiding modals can be improved using the native driver.

https://facebook.github.io/react-native/blog/2017/02/14/using-native-driver-for-animated

Supply an optional `useNativeDriver` prop to the Modal component.

### Fixes
Suppress a warning in `ActivityIndicator.test.js`.

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -39,10 +39,6 @@ type Props = {|
    * @optional
    */
   theme: Theme,
-  /**
-   * Animate using native driver.
-   */
-  useNativeDriver?: boolean,
 |};
 
 type State = {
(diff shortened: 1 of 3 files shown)


src/components/Modal.js
```diff
@@ -39,6 +39,10 @@ type Props = {|
    * @optional
    */
   theme: Theme,
+  /**
+   * Animate using native driver.
+   */
+  useNativeDriver?: boolean,
```
ferrannp: Can you remove the prop and just make `useNativeDriver` true by default?

auzwang: Sure!

PR #954, opened 2019-03-27
feat: add prop textStyle to Chip

brunohkbx:
### Motivation
It would be nice to be able to customize the style of chip's content .

example


brunohkbx:
Hey @satya164. Could you please take a look? :smiley:

Issue #596, opened 2018-10-12
TextInput support Leading Icon


ButuzGOL:
Any plans for 
TextInput support Left Icon
or could you suggest as workaround to freeze label on the top in mode="outlined"

----
from @ferrannp: 

![screenshot 2019-01-31 at 13 13 41](https://user-images.githubusercontent.com/774577/52053584-0bfe2480-255a-11e9-81dd-171a6b0dd5f8.png)

webdevsyd:
Agree to this hoping to get a support from this.

Manoj002:
+1

ferrannp:
Depending on https://github.com/callstack/react-native-paper/pull/587.

viv3kk:
any update on this??

jbinda:
Guys - good news :) Work is in progress ! It will be cover with PR posted to this issue: #903

khanhubgit:
<TextInput
   ....
   left={<TextInput.Icon name="account"/>}
/>

ignaciolopez12:
> <TextInput
> ....
> left={<TextInput.Icon name="account"/>}
> />

I only recognize the "account" icon, another does not recognize me ... how can I solve it?

khanhubgit:
https://callstack.github.io/react-native-paper/icons.html

![Uploading 95650198...]

khanhubgit:
> > <TextInput
> 
> > ....
> 
> > left={<TextInput.Icon name="account"/>}
> 
> > />
> 
> 
> 
> I only recognize the "account" icon, another does not recognize me ... how can I solve it?


https://callstack.github.io/react-native-paper/icons.html

![Uploading 95650213...]

khanhubgit:
https://callstack.github.io/react-native-paper/icons.html

![D3DC08DC-4BEC-4A0C-ADE7-72D4AC21236C](https://user-images.githubusercontent.com/36913458/95650266-04976880-0afc-11eb-832a-ee796d75a088.png)

khanhubgit:
> https://callstack.github.io/react-native-paper/icons.html
> 
> ![D3DC08DC-4BEC-4A0C-ADE7-72D4AC21236C](https://user-images.githubusercontent.com/36913458/95650266-04976880-0afc-11eb-832a-ee796d75a088.png)

PR #961, opened 2019-03-29
Leading Icon try 1

murtraja:
### Motivation
#587 #596 
I would like to work on Leading and Trailing Icon implementation. This was my first try at the feat. I would love some mentor to help me land this patch

### Test plan
leading_icon

### DIFF
--- a/src/components/TextInput/TextInputFlat.js
+++ b/src/components/TextInput/TextInputFlat.js
@@ -270,6 +270,14 @@ class TextInputFlat extends React.Component<ChildTextInputProps, {}> {
             ],
           }: RenderProps)
         )}
+        {trailingIcon ? (
+            <TouchableOpacity 
+              style={[styles.icon, { marginLeft: 0, marginRight: 12 }]} 
+              onPressIn={trailingIconOnPressIn} 
+              onPressOut={trailingIconOnPressOut}>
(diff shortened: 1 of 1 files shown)


jaysbytes:
@murtraja Thank you for the PR
Unfortunately, I cannot merge it because there are missing parts:
- lint-and-flow task is failing on CI
- you used `TouchableOpacity` and paper uses `TouchableRipple`
- missing RTL support
- props naming convention do not follow other components naming - we would rather have `leftIcon` and `rightIcon`, also it would be good to have a possibility to specify icon as render prop.
-  `[styles.icon, { marginLeft: 0, marginRight: 12 }]` Why not put all styles in styles.icon instead of defining them inline?
I'm closing this one due to too much code needs to be refactored. It was try 1 so don't worry, you can try one more time :)

murtraja:
@jaysbytes, thank you for your valuable feedback. I have understood what more needs to be done here. But I am not sure what should I do here

> missing RTL support

should the leftIcon and rightIcon be swapped in case of RTL? Is there some piece of code that I can take a look to understand this better? Also, what constitutes RTL support? Is there a doc somewhere which I can read?

jaysbytes:
https://material.io/design/components/text-fields.html#anatomy
Find `Icons` section there and there is a sentence
`Leading and trailing icons change their position based on LTR or RTL contexts.` 
So as I understand in RTL context you should swap the icons

PR #962, opened 2019-03-29
fix: fix borderRadius support for the Chip component

### LINKED ISSUE #955: Chip | Support border-radius + default to theme roundness
## Describe the feature
The chip component currently ignores the [roundness](https://callstack.github.io/react-native-paper/theming.html) setting of the react-native-paper theme. It would be great if it would set its borderRadius based on the roundness.

## Motivation
Help achieve a consistent [theme](https://material.io/design/components/chips.html#theming).

## Related Issues
Currently, when a borderRadius is explicitly added, the selection animation does not follow. The white background disappeared during the gif export, but it is possible to see the animation roundness issue in the gif below.

Mar-28-2019 14-34-46

agarant:
### Motivation
Fixes #955 

### Test plan
- Tested in the example app by setting the borderRadius on the chip.
Mar-29-2019 10-56-38

### DIFF
--- a/src/components/Chip.js
+++ b/src/components/Chip.js
@@ -152,6 +152,7 @@ class Chip extends React.Component<Props, State> {
         : dark
           ? '#383838'
           : '#ebebeb',
+      borderRadius = 16,
     } = StyleSheet.flatten(style) || {};
 
     const borderColor =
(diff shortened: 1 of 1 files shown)


Trancever:
Thanks for PR!

Issue #549, opened 2018-09-17
TextInput with transparent background


luissmg:
### Current behaviour

When you set the `TextInput` component to have a `transparent` background color, the spaces still persist.

### Expected behaviour

I check MD and they don't mention anything like this.
I was thinking that when the `TextInput` has a `transparent` background, the horizontal spaces on the input should be removed.
I don't actually know if it is supposed to have a transparent background on MD's input. I just had a case where I needed it and the spaces were there.

### Screenshots (if applicable)

The background is from my app. Input has no background color.

![screenshot_1537178841](https://user-images.githubusercontent.com/20660551/45617375-136cd780-ba6a-11e8-9ba6-4705ebab5f1e.png)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | both
| react-native          | 0.55.4
| react-native-paper    | 2.0.1 (latest)
| node                  | 8.11.3
| npm or yarn           | 6.4.1

jsp3536:
I would second that this would be a nice addition

raajnadar:
+1 the textinput looks ugly when the background is transparent

luissmg:
I wanted some feedback from the team. What do you think of this @satya164?

luissmg:
@ferrannp What kin of investigation do you need?

ferrannp:
I think when background is transparent, `TextInput` should be like in `v1`. Which is as your screenshot but the bottom underline to be aligned with text. An option to remove the bottom underline should be there too (or make it transparent too).

gawrysiak:
I wouldn't really go for changing the paddings based on backgroundColor setting. It's somehow breaking and also hard to understand why one attribute affects another. This way if padding set to 0, should it change the background to transparent?

One of the ways is to first tackle #791 and then create a new TextInput mode, called maybe "legacy" or "v1" (or TextInput.Legacy). It doesn't really go by the recent guidelines, but it's a highly requested feature and still works visually, even with projects using the new MD. With #791 finished, it won't create any more mess inside the TextInput component.

The other way I'd really prefer here is to go back to the drawing board and work on a decent customizations system to be released for 3.0. Such simple case should be possible by just setting the background & padding. Should also allow fontSize, so it can cover probably the number 1 requested feature 😄

hussainahmad:
Is there any workaround to get this feature,  Much needed.

Trancever:
Fixed in #1018

abhiburk:
[...]

PR #964, opened 2019-03-30
TextInputLegacy Try 1

murtraja:
### Motivation
https://github.com/callstack/react-native-paper/issues/549

### Test plan
gifeditor_20190330_112121

This is work in progress. Need help with animating color, fontSize, etc. whose support was removed

### DIFF
--- a/src/components/TextInput/TextInput.js
+++ b/src/components/TextInput/TextInput.js
@@ -6,6 +6,7 @@ import { polyfill } from 'react-lifecycles-compat';
 
 import TextInputOutlined from './TextInputOutlined';
 import TextInputFlat from './TextInputFlat';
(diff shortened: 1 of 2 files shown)


raajnadar:
You made this PR very complex and hard to maintain!

murtraja:
@raajnadar thank you for your inputs. Can you be a bit more specific though? What can I do to make it less complex and more maintanable?

All I did is that I copied the TextInput code from v1 branch and added it as a new Legacy mode as described [here](https://github.com/callstack/react-native-paper/issues/549#issuecomment-462543506)

ferrannp:
I wrote in #967 how we will approach this issue. Closing it in favour of such.

PR #967, opened 2019-03-31
feat: add legacy textinput

raajnadar:
### Motivation

closes https://github.com/callstack/react-native-paper/issues/549

### Test plan

Added a new mode type **standard**.

Textinput looks perfect in example app

### DIFF
--- a/src/components/TextInput/TextInput.js
+++ b/src/components/TextInput/TextInput.js
@@ -1,19 +1,20 @@
 /* @flow */
 
 import * as React from 'react';
-import { Animated, TextInput as NativeTextInput } from 'react-native';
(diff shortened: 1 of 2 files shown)


ferrannp:
We are planning to use a prop to make flat `TextInput` to look like in v1.

raajnadar:
Even I used the prop method you should review my changes.

https://github.com/callstack/react-native-paper/pull/967/files#diff-c5c6fd9ea546b7ce3b786d110089550dR26

I think you should review once again and test the example app.

ferrannp:
@raajnadar ok we can take it from here. We are thinking though to use another prop instead of adding such `standard` mode.

PR #972, opened 2019-04-03
fix(appbar): does not pass any additional props to Searchbar component

### LINKED ISSUE #605: Searchbar crashes on Android when wrapped in Appbar
### Current behaviour

`Searchbar` component crashes on Android with 
`Error while updating property 'color' in shadow node of type: AndroidTextInput` 
when it's wrapped in `Appbar` or `Appbar.Header`.  I guess this happens because `Appbar` changes the theme based on the `dark` property. This works fine on iOS.

### Expected behaviour

Should not crash.

### Code sample
```js
class SearchContainer extends React.Component {
    render() {
        return (
            <Appbar.Header>
                <Searchbar
                    value=""
                />
            </Appbar.Header>
        )
    }
}
```

It seems to happen because of [this](https://github.com/callstack/react-native-paper/blob/master/src/components/Appbar/Appbar.js#L136) and [this](https://github.com/callstack/react-native-paper/blob/master/src/components/Searchbar.js#L165). `...rest` contains `color` prop, which crashes the `TextInput`.

### Your Environment

| ios or android        | android 28
| react-native          | 0.57.1
| react-native-paper    | 2.1.3
| node                  | 10.11.0
| npm or yarn           | 1.10.1

brunohkbx:
Fix #605

### DIFF
--- a/src/components/Appbar/Appbar.js
+++ b/src/components/Appbar/Appbar.js
@@ -133,7 +133,15 @@ class Appbar extends React.Component<Props> {
         {React.Children.toArray(children)
           .filter(child => child != null && typeof child !== 'boolean')
           .map((child, i) => {
-            if (!React.isValidElement(child)) {
+            if (
+              !React.isValidElement(child) ||
+              ![
+                'AppbarContent',
+                'AppbarAction',
+                'AppbarBackAction',
+                'AppbarHeader',
+              ].includes(child.type.name)
+            ) {
               return child;
             }
 
--- a/src/components/__tests__/Appbar/Appbar.test.js
+++ b/src/components/__tests__/Appbar/Appbar.test.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Appbar from '../../Appbar/Appbar';
+import Searchbar from '../../Searchbar';
+
+describe('Appbar', () => {
+  it('does not pass any additional props to Searchbar', () => {
+    const tree = renderer
+      .create(
+        <Appbar>
+          <Searchbar placeholder="Search" />
+        </Appbar>
+      )
+      .toJSON();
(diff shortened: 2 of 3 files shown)


brunohkbx:
Hey @satya164 @Trancever. My initial approach was to remove the [`...rest`](https://github.com/callstack/react-native-paper/blob/master/src/components/Searchbar.js#L179) props passed to TextInput, but there are some meaningful cases like passing `autoFocus` to TextInput.
Thus, I just have checked if the children are a searchbar.
What do you guys think?

jaysbytes:
Hi @brunohkbx! Thank you for your interest in this issue. I've checked your solution and I think we can do it better.  Please check my comment.
Let me know if you have time for changing your PR or I will handle this

src/components/Appbar/Appbar.js
```diff
@@ -133,7 +134,7 @@ class Appbar extends React.Component<Props> {
         {React.Children.toArray(children)
           .filter(child => child != null && typeof child !== 'boolean')
           .map((child, i) => {
-            if (!React.isValidElement(child)) {
+            if (!React.isValidElement(child) || child.type === Searchbar) {
```
jaysbytes: I'm afraid this kind of issue could appear in future not only for `Searchbar` component. My proposal is to use the following condition
```js
!React.isValidElement(child) ||
![
  'AppbarContent',
  'AppbarAction',
  'AppbarBackAction',
  'AppbarHeader',
].includes(child.type.name)
```
because these color and style changes are related to these four components.

brunohkbx:
Thanks for reviewing the changes @jaysbytes. I've already pushed your suggestions. :grin:

jaysbytes:
Now the fix looks legit. WDYT @Trancever, can we merge?

PR #982, opened 2019-04-10
fix: component RadioGroup animation

brunohkbx:
**What existing problem does the pull request solve?**
`RadioButtonAndroid` wasn't triggering the animation on `componentDidUpdate` when it was wrapped by a `RadioButtonGroup` because it was getting the props in the context.  

Before:
ezgif com-video-to-gif

Fixed
ezgif com-video-to-gif (1)

### DIFF
--- a/src/components/RadioButton.js
+++ b/src/components/RadioButton.js
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 import { Platform } from 'react-native';
-import RadioButtonGroup from './RadioButtonGroup';
+import RadioButtonGroup, { RadioButtonContext } from './RadioButtonGroup';
 import RadioButtonAndroid from './RadioButtonAndroid';
 import RadioButtonIOS from './RadioButtonIOS';
 import { withTheme } from '../core/theming';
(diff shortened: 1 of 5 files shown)


brunohkbx:
@satya164 @ferrannp what do you guys think? :grin:

src/components/RadioButton.js
```diff
@@ -103,11 +103,32 @@ class RadioButton extends React.Component<Props> {
...
+  isChecked = context =>
+    context.value === this.props.value ? 'checked' : 'unchecked';
+
   render() {
-    return Platform.OS === 'ios' ? (
-      <RadioButtonIOS {...this.props} />
-    ) : (
-      <RadioButtonAndroid {...this.props} />
+    const Button = Platform.select({
+      android: RadioButtonAndroid,
```
Trancever: Could you use default instead android here? We should use RadioButtonAndroid also on different platforms like web for example.

brunohkbx: Good catch, thanks.

Trancever:
@brunohkbx Awesome job! Thanks

PR #983, opened 2019-04-10
fix: use TouchableRipple for Card

jaulz:
### Motivation

Replaces `TouchableWithoutFeedback` with `TouchableRipple` in Cards.

### Test plan

- Open example app
- Open `Card` examples
- Press "Pressable Chameleon" card

### DIFF
--- a/__ts-tests__/TouchableRipple.test.tsx
+++ b/__ts-tests__/TouchableRipple.test.tsx
@@ -1,0 +1,15 @@
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Text, TouchableRipple } from '..';
+
+const MyComponent = () => (
+  <TouchableRipple
+    onPress={() => console.log('Pressed')}
+    rippleColor="rgba(0, 0, 0, .32)"
+  >
+    <Text>Press me</Text>
+  </TouchableRipple>
+);
+
+export default MyComponent;
--- a/src/components/Card/Card.js
+++ b/src/components/Card/Card.js
@@ -1,17 +1,13 @@
 /* @flow */
 
 import * as React from 'react';
-import {
-  Animated,
-  View,
-  TouchableWithoutFeedback,
-  StyleSheet,
-} from 'react-native';
+import { Animated, View, StyleSheet } from 'react-native';
 import CardContent from './CardContent';
 import CardActions from './CardActions';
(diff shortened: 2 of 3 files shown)


ferrannp:
@jaulz the `Card` has `TouchableWithoutFeedback` because the effect is when you click is to increase elevation (as Material guide states). Pretty sure if you want something different, the best would be to wrap `Card` in your own Touchable.

jaulz:
@ferrannp hm, do you have a link where this is described? I could only find the following links that indicate that a ripple should be the desired effect but I could be wrong.

- The [official components](https://material-components.github.io/material-components-web-catalog/#/component/card?type=basic) have a ripple.

- [Here](https://material.io/develop/web/components/cards/) (under Primary Action) it says:

> If a majority of the card (or even the entire card) should be actionable, you can add the mdc-card__primary-action class to the region to give it MDC Ripple styles. 

- [Here](https://material.io/design/components/cards.html#card-collections) it says:
> By default, cards in a collection are coplanar, sharing the same resting elevation unless they are picked up or dragged.

- [Here](https://material.io/design/components/cards.html#behavior) it says:
> On desktop and mobile, cards can have a resting elevation of 0dp. They elevate to 8dp on hover.

ferrannp:
@jaulz being checking here https://material.io/design/environment/elevation.html#depicting-elevation and you might be right it is ripple + elevation. But for sure if we want this, we need to consider it for 3.0.

jaulz:
@ferrannp then it's maybe the perfect point in time 😊 could we consider this for 3.0?

If I read it correctly, the elevation is only for the case when the card is in motion (i.e. is dragged). At least it's explicitly mentioned in this table: https://material.io/design/environment/elevation.html#default-elevations

ferrannp:
We can. @Trancever told me we tried it in the past and didn't work for some reason?

jaulz:
@Trancever do you remember what exactly wasn't working? Maybe I missed something while testing?

lukewalczak:
Unfortunately, the ripple effect is broken on different android versions. Once you click the outside the `Card.Cover` e.g. `Card.Title` ripple effect source position is wrong. Probably we'll have to close that PR.

jaysbytes:
It's inactive for a while so I'm closing for now

Issue #686, opened 2018-11-29
Feature request: Customisable numberOfLines for List.Item


abn-volk:
### Current behaviour

`List.Item`'s title and description texts/components are [wrapped inside a `Text` with fixed `numberOfLines`](https://github.com/callstack/react-native-paper/blob/0ad98f3bf2c397c12e258927b5bdd504ebf9f275/src/components/List/ListItem.js#L97). As a result, long text elements are always cut off (even when using custom text components with `numberOfLines` set).

### Expected behaviour

It'd be better if there is a way to customise the `numberOfLines` for item title/description, or at least a way to disable the default behaviour. I know that I can easily replace `List.Item` with a `View`, but I'd like to use `List.Item` for its standard padding/margin and ripple.

### Code sample
Snack: https://snack.expo.io/@jibi/react-native-paper-list.item-numberoflines

### Screenshots (if applicable)
![Imgur](https://i.imgur.com/J4MQOJ2.png)

### What have you tried

* Use a `Subheading` for title with `numberOfLines`
* Read the source code -> Write a feature request

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android
| react-native          | 0.57 (expo 31)
| react-native-paper    | 2.3.1
| node                  | 10.2.1
| npm or yarn           | yarn 1.12.3

raajnadar:
The feature that you are asking is not under material design guidelines. Can tell me what are you trying to do? Maybe I can give you a better alternative solution.

abn-volk:
I wanted to show some list items with long text, but the text is always cut off. I've switched to a View and it's working fine now. 

Oh, and I didn't realise the 3 line limitation is to comply with the MD guideline. I'll close this then.

tonyxiao:
@raajnadar sometimes guidelines do not correspond to the reality of the data we're trying to present to users.

tonyxiao:
What alternative would you recommend for a list item that takes a few lines?

zerobytes:
I must disagree with @raajnadar and say that the implementation that took place on list item does not corresponde to the guidelines, itself.

Material guideline is very clear on giving 3 independent lines or 2 lines and the last can grow wrapping down.
![image](https://user-images.githubusercontent.com/439445/54005847-c4128480-4139-11e9-8db4-4c377e543ed8.png)

raajnadar:
Ok Ok, I am opening this issue again..!

satya164:
[...]

PR #996, opened 2019-04-16
refactor: make ListItem accepts a custom description

brunohkbx:
### Motivation
Currently, there's no way to customize the description of `List.Item`. In order to make it able to be customized, `List.Item` now accepts a custom node as description.

With these changes we can implement a list item like Gmail app:
customized-list-description

```
<List.Item
  left={() => (
    <Image
      source={require('../../assets/images/email-icon.png')}
      style={styles.image}
    />
  )}
  right={props => <List.Icon {...props} icon="star-border" />}
  title="List Item 1"
  description={({
    ellipsizeMode,
    color: descriptionColor,
    fontSize,
  }) => (
    <View style={[styles.container, styles.column]}>
      <Text
        numberOfLines={2}
        ellipsizeMode={ellipsizeMode}
        style={{ color: descriptionColor, fontSize }}
      >
        React Native Paper is a high-quality, standard-compliant
        Material Design library that has you covered in all major
        use-cases.
      </Text>
      <View style={[styles.container, styles.row, { paddingTop: 8 }]}>
        <Chip icon="picture-as-pdf" onPress={() => {}}>
          DOCS.pdf
        </Chip>
      </View>
    </View>
  )}
/>
```

### Test plan

1. Download this branch
2. Run the example app
3. Open `List.Section` to see the new example.

### DIFF
--- a/example/src/Examples/ListSectionExample.js
+++ b/example/src/Examples/ListSectionExample.js
@@ -1,8 +1,8 @@
 /* @flow */
 
 import * as React from 'react';
-import { ScrollView, StyleSheet, Image } from 'react-native';
-import { List, Divider, withTheme, type Theme } from 'react-native-paper';
+import { ScrollView, StyleSheet, Image, View, Text } from 'react-native';
+import { Chip, List, Divider, withTheme, type Theme } from 'react-native-paper';
 
 type Props = {
   theme: Theme,
@@ -80,6 +80,42 @@ class ListSectionExample extends React.Component<Props> {
             description="Describes item 2. Example of a very very long description."
           />
         </List.Section>
+        <Divider />
+        <List.Section>
+          <List.Subheader>Custom description</List.Subheader>
+          <List.Item
+            left={() => (
+              <Image
+                source={require('../../assets/images/email-icon.png')}
+                style={styles.image}
+              />
+            )}
+            right={props => <List.Icon {...props} icon="star-border" />}
+            title="List Item 1"
+            description={({
+              ellipsizeMode,
+              color: descriptionColor,
+              fontSize,
+            }) => (
+              <View style={[styles.container, styles.column]}>
+                <Text
+                  numberOfLines={2}
+                  ellipsizeMode={ellipsizeMode}
+                  style={{ color: descriptionColor, fontSize }}
+                >
+                  React Native Paper is a high-quality, standard-compliant
+                  Material Design library that has you covered in all major
+                  use-cases.
+                </Text>
+                <View style={[styles.container, styles.row, { paddingTop: 8 }]}>
+                  <Chip icon="picture-as-pdf" onPress={() => {}}>
+                    DOCS.pdf
+                  </Chip>
+                </View>
+              </View>
+            )}
+          />
+        </List.Section>
       </ScrollView>
     );
   }
@@ -93,6 +129,12 @@ const styles = StyleSheet.create({
     height: 40,
     width: 40,
     margin: 8,
+  },
+  row: {
+    flexDirection: 'row',
+  },
+  column: {
+    flexDirection: 'column',
   },
 });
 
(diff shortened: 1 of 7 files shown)


brunohkbx:
Hey @ferrannp @Trancever. I'm not sure if all the flow declarations are correct. Could you guys please review it? :grimacing:

brunohkbx:
Thanks @Trancever. Done 5ae1695

Trancever:
@brunohkbx Awesome, one more thing,  can we improve `description` prop description? We should add info that it can be string or render prop and what arguments will be passed to that render prop. Please check other components that follow this pattern for reference.

brunohkbx:
Good catch, thanks. Added in cc56e4f89422fc884fd3323a0f4e706bfe3caff4

example/src/Examples/ListSectionExample.js
```diff
@@ -80,6 +80,42 @@ class ListSectionExample extends React.Component<Props> {
...
+              <Image
+                source={require('../../assets/images/email-icon.png')}
+                style={styles.image}
+              />
+            )}
+            right={props => <List.Icon {...props} icon="star-border" />}
+            title="List Item 1"
+            description={({
+              ellipsizeMode,
+              color: descriptionColor,
```
satya164: imo the color and fontSize should be in style, not as separate props.

Trancever: @satya164 hmm, other components expose them as separate props, I wanted it to be consistent

satya164: hmm I think in most places where it's this way, it's usually for icons etc., but here looks like the render callback is passing a bunch of props (do we need to pass `ellipsizeMode` and `fontSize` for example?)

brunohkbx: @satya164 I found this components doing the same: [ListItem](https://github.com/callstack/react-native-paper/blob/master/src/components/List/ListItem.js#L27), [ListAccordion](https://github.com/callstack/react-native-paper/blob/master/src/components/List/ListAccordion.js#L28), [Banner](https://github.com/callstack/react-native-paper/blob/master/src/components/Banner.js#L26)
I think we should pass `ellipsizeMode` and `fontSize` because could have cases when the user only wants to add more content to the description and keep the `Text` as the original. e.g. #686

BenWildeman:
This should have never been merged as it's a breaking change but aside from that. the implementation here isn't really the "react way". What do I mean by that? Well you shouldn't have to pass in a callback in order to use a react component.

Because of this breaking change, it's no longer possible to pass in a text component to it which was perfectly valid beforehand. Here it checks that the description is explicitly a string, in order to put it into the text component, but then this PR #1199 has reversed it and is now explicitly checking for a function.

This PR did not remove the React.Node type which meant that this didn't through any errors within TS

The ideal way to fix this is to do the following which would have satisfied all needs, without creating a breaking change:
```js 

// renderDescription function

const isDescReactElement = React.isValidElement(description); // would be false for anything that's not a react element, such as a string;

if (isDescReactElement) {
    const desc = React.cloneElement(description, {
        ellipsizeMode: descriptionEllipsizeMode,
        color: descriptionColor,
        fontSize: styles.description.fontSize,
    });

    return desc;
}

return (
    <Text	
         ellipsizeMode={descriptionEllipsizeMode}	
         numberOfLines={2}	
         style={[	
             styles.description,	
             {	
                 color: descriptionColor,	
              },	
              descriptionStyle,	
         ]}	
    >	
        {description}	
    </Text>
);
```

brunohkbx:
@BenWildeman I didn't realize that it would break the current implementation :disappointed:. I apologize for the inconvenience.
Thanks for showing your implementation, it's good to know but I'm not sure if it would be ok to pass a component as a prop because this entire project is using render props to return elements.
Maybe @satya164, @Trancever or @wojteg1337 can check your fix.

BenWildeman:
it's fine with this, it's just that the original types are incorrect really. there shouldn't have been React.Node if it was only supposed to accept a string. you fixed it to be inline with the rest of the codebase. it's just annoying that a minor bump included a breaking change that broke our app in places that we didn't know about. luckily we do regression testing so it was caught

PR #1001, opened 2019-04-17
fix(appbar): make allowed children to accept additional props

brunohkbx:
Comparing a string to `child.type.name` doesn't work when the component is wrapped by `withTheme`. So we need to check against its type.
 
### Motivation
#972 

Before:
WhatsApp Image 2019-04-17 at 10 48 54 (1)

Fixed:
WhatsApp Image 2019-04-17 at 10 48 54

### DIFF
--- a/src/components/Appbar/Appbar.js
+++ b/src/components/Appbar/Appbar.js
@@ -135,12 +135,9 @@ class Appbar extends React.Component<Props> {
           .map((child, i) => {
             if (
               !React.isValidElement(child) ||
-              ![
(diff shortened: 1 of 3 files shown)


brunohkbx:
@jaysbytes, I should have tested in the example app our last changes. What do you think?

awinograd:
I just upgraded from 2.15.2 to 2.16.0 to get the fix for Searchbar nested inside Appbar. I now see this bug (text is black), but only in development. In production, the type check works properly and the colors are as expected (white).

PR #1002, opened 2019-04-17
feat: paddinghorizontal props for legacy input

raajnadar:
### Motivation

closes https://github.com/callstack/react-native-paper/issues/549

### Test plan

cc @ferrannp


Trancever:
@raajnadar We merged #1018 instead, cause we think it better align with api of existing components.

raajnadar:
Yup my approach was not that good!

PR #1004, opened 2019-04-17
refactor: declare typography colors into theme file

brunohkbx:
### Motivation
This PR inserts a new node in the theme file containing the default style of typography in order to avoid the need of declaring variables with the title and description style in each component.

DefaultTheme.js
```
typography: {
  primary: 'rgba(0, 0, 0, 0.87)',
  secondary: 'rgba(0, 0, 0, 0.54)',
},
```

DarkTheme.js
```
typography: {
  primary: 'rgba(255, 255, 255, 0.87)',
  secondary: 'rgba(255, 255, 255, 0.54)',
},
```

### DIFF
--- a/src/components/Appbar/AppbarAction.js
+++ b/src/components/Appbar/AppbarAction.js
@@ -1,10 +1,10 @@
 /* @flow */
 
 import * as React from 'react';
-import color from 'color';
-import { black } from '../../styles/colors';
 import IconButton from '../IconButton';
 import type { IconSource } from '../Icon';
+import type { Theme } from '../../types';
+import { withTheme } from '../../core/theming';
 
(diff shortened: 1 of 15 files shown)


ocarreterom:
A reference of theme system can be viewed in [material-components-web](https://github.com/material-components/material-components-web/blob/master/packages/mdc-theme/_variables.scss).
We need a dark and light version of text colors to contrast with primary and accent (now secondary) colors.

brunohkbx:
@ocarreterom the color variation will be chosen according to the selected theme(light or dark). You can download this branch and change the theme to dark in the example app to verify the behavior.

ferrannp:
We decided this makes maintaining the library more complicated. For the future it would be better if you open an RFC before sending changes. Thank you!

Issue #877, opened 2019-02-28
Textinput height


a1lu:
### Current behaviour

If i change (make it smaller) the height of the TextInput component the placeholder text is not centered.

### Expected behaviour
A smaller textinput with the placeholder centered

### Code sample

<TextInput style={{width: '100%', height: 44}} mode='outlined' label='Test' />

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | android 7
| react-native          | 0.58.5
| react-native-paper    | 2.11
| node                  |
| npm or yarn           |
| expo sdk              |

viv3kk:
Facing the same issue..

jcarloTA:
Same problem, when I add custom styles for height, these are reflected only in the parent but the component seems to have default styles for the child, a suggestion would be to add something like childStyles so that we can customize the child component as well

`style: [
            styles.input,
            childStyles, // <---just here!
            mode === 'outlined'
              ? styles.inputOutlined
              : this.props.label
                ? styles.inputFlatWithLabel
                : styles.inputFlatWithoutLabel,
            {
              color: inputTextColor,
              fontFamily,
              textAlignVertical: multiline ? 'top' : 'center',
            },
  ],`

elizamaria:
I also tried something for this to work, first for the flat input. Is it ok to add a new props style for the input child component that @jcarloTA said? I also make us of the input height in order to fix the label position when the height is smaller.

jbinda:
fixed in #1020

jmoliugp:
It doesn't seem to have been fixed.

I tried it out using this component with `2.16.0` version:

```
import * as React from 'react';

import { TextInput } from 'react-native-paper';

export default class MyComponent extends React.Component {
  state = {
    text: ''
  };

  render(){
    return (
      <TextInput
        label='Email'
        value={this.state.text}
        onChangeText={text => this.setState({ text })}
        style={ { marginVertical: 50, height: 150, textAlign: 'center', backgroundColor: 'grey' } }
        dense={true}
      />
    );
  }
}
```

jbinda:
Hi @jmoliugp

Issue was fixed in 3.0 branch which is current master.

Regards

Noorulhuda682:
{this.state.emailMsg !== '' &&
                  <HelperText type="error" visible={true} >
                         {this.state.emailMsg}
                      </HelperText> 
                  }    
               <TextInput
                  label='Password'
                  value={this.state.password}
[...]

PR #1009, opened 2019-04-19
#877 fix: placeholder position text input flat

elizamaria:
### Motivation

Fixes placeholder position when the input height in smaller - this is **only for the flat TextInput**. If the approach is okay I will continue with the outlined mode as well.

### Test plan

Before: with height 44

Screenshot 2019-04-19 at 18 42 27

After: height: 44

Screenshot 2019-04-19 at 18 43 14

### DIFF
--- a/example/src/Examples/TextInputExample.js
+++ b/example/src/Examples/TextInputExample.js
@@ -57,6 +57,10 @@ class TextInputExample extends React.Component<Props, State> {
             onChangeText={text => this.setState({ text })}
           />
           <TextInput
+            style={styles.inputWithHeightStyle}
+            placeholder="Type something"
+          />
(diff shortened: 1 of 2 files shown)


itzsaga:
Looking at PRs to achieve this and found yours @elizamaria. Awesome work 😃. I would assume that this needs to be done for `/src/components/TextInput/TextInputOutlined.js` as well and figured I'd point that out so hopefully when it's reviewed that isn't an issue for you. Cheers.

jbinda:
Hi, thanks for your suggestion and PR.

I had been working on font size feature and I need to deal also with the placement of label and placeholder. 

I close your PR because it is fixed with : #1020

Issue #1015, opened 2019-04-23
Card.Title expects type `React.ReactNode` but is rendered inside a `StyledText` -- which means in effect it can only accept a Text node


chadlavi-casebook:
## Description

Per the type definitions[[1](https://github.com/callstack/react-native-paper/blob/901745bc840bdef85446fc7d58160642c8f4b7de/typings/components/Card.d.ts#L19)], `Card.Title`'s `title` prop should accept any react node. However, because it is wrapped in a `StyledText` via the `<Title>` component[[2](https://github.com/callstack/react-native-paper/blob/master/src/components/Typography/Title.js)], it can only accept _text_ nodes. This is quite limiting for devs who want to customize their Card titles, especially since the `left` prop does not have style controls. 

For example, in my use case: I want to put an Icon before the text of the title, at the same size as the title text, and with only a gap of 16px between the icon and the text. I have to either put the icon in the `left` then give the title a negative left margin (because the `left` has a fixed width of 40), or I have to use an icon set that renders as text glyphs so that I can include it in the `title` itself; passing anything that is/contains a `<View>` to `title` results in the error `Invariant Violation: Nesting of <View> within <Text> is not currently supported`.

The `title` prop of `Card.Title` should not be wrapped in a `<Text>`-based component, so that `<View>`-based children can be passed to it. Either that or the type definition for `CardTitleProps` should be updated to admit its limitation to only `<Text>` components.

for example, here's the desired state:
<img width="689" alt="Screen Shot 2019-04-23 at 12 32 28 PM" src="https://user-images.githubusercontent.com/45078010/56599138-f67e2100-65c3-11e9-94ab-85b1909ee930.png">

and here's how that looks if I _don't_ do the hacky negative left margin on the `title`:
<img width="689" alt="Screen Shot 2019-04-23 at 12 32 51 PM" src="https://user-images.githubusercontent.com/45078010/56599165-07c72d80-65c4-11e9-86ad-cac62b511295.png">

jaulz:
Use left or right prop.

chadlavi-casebook:
> Use left or right prop.

I mentioned left explicitly in my issue. Left has a fixed width of 40, and cannot be styled. It's not possible to do this without a hacky workaround where you apply a negative left margin to the title.

jaulz:
[...]

PR #1016, opened 2019-04-23
add props `leftStyle` and `rightStyle` to `Card.Title` so devs can style the `left` and `right` independently

chadlavi:
### Motivation
This PR adds style APIs for the `left` and `right` of the `Card.Title` component. See also issue #1015. 

### Test plan
This can be tested by creating a card whose `Card.Title` has `left`, `right`, `leftStyle` and `rightStyle` props. The node passed to `left` should be seen do have the style passed to `leftStyle`, and the node passed to `right` should be seen to have the style passed to `rightStyle`.

### DIFF
--- a/src/components/Card/CardTitle.js
+++ b/src/components/Card/CardTitle.js
@@ -9,7 +9,7 @@ import type { Theme } from '../../types';
 
 type Props = React.ElementConfig<typeof View> & {|
   /**
-   * Text for the title.
+   * Text for the title. Note that this will only accept a `<Text>`-based node.
    */
   title: React.Node,
   /**
--- a/src/components/Card/CardTitle.js
+++ b/src/components/Card/CardTitle.js
@@ -6,6 +6,7 @@ import { withTheme } from '../../core/theming';
 import Caption from './../Typography/Caption';
 import Title from './../Typography/Title';
 import type { Theme } from '../../types';
+import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
 
 type Props = React.ElementConfig<typeof View> & {|
   /**
@@ -29,17 +30,17 @@ type Props = React.ElementConfig<typeof View> & {|
    */
   left?: (props: { size: number }) => React.Node,
   /**
-   * Style for the left.
+   * Style for the left element wrapper.
    */
-  leftStyle?: any,
+  leftStyle?: ViewStyleProp,
   /**
    * Callback which returns a React element to display on the right side.
    */
   right?: (props: { size: number }) => React.Node,
   /**
-   * Style for the right.
+   * Style for the right element wrapper.
    */
-  rightStyle?: any,
+  rightStyle?: ViewStyleProp,
   /**
    * @internal
    */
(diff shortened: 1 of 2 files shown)


src/components/Card/CardTitle.js
```diff
@@ -9,7 +9,7 @@ import type { Theme } from '../../types';
 
 type Props = React.ElementConfig<typeof View> & {|
   /**
-   * Text for the title.
+   * Text for the title. Note that this will only accept a `<Text>`-based node.
```
jaulz: Please remove this comment again and change the type to `string`

typings/components/Card.d.ts
```diff
@@ -22,7 +22,10 @@ export interface CardTitleProps extends ViewProps {
   subtitleStyle?: StyleProp<TextStyle>;
   style?: StyleProp<ViewStyle>;
   left?: (props: { size:number }) => React.ReactNode;
+  leftStyle?: StyleProp<ViewStyle>;
   right?: (props: { size:number }) => React.ReactNode;
+  rightStyle?: StyleProp<ViewStyle>;
+
```
jaulz: Can you change the `title` type here as well to `string`?

chadlavi: It doesn't actually have to be a string, you can pass a `<Text>`. Changing this to a string would be a breaking change and would remove functionality. 

They said, `React.Node` isn't specific enough, but I don't know how to make it a more specific type that only accepts nodes that are based on react native `<Text>`

jaulz: Why would you pass a `<Text>` if `<Title>` is already a text?

chadlavi: In my case, our component library has pre-approval styled text components, and it's much simpler to pass that component to the title than to pass a string and a style

jaulz: Makes sense 👍  Can you please adjust the comment then to something like this and add it to `subtitle` as well to keep it consistent?
> Text for the title. Note that this will only accept a string or a `<Text>`-based node.

chadlavi: Sure, done.

jaulz:
Looks fine to me. @Trancever what do you think?

chadlavi:
Any updates on this? Looks like @Trancever hasn't reviewed yet.

src/components/Card/CardTitle.js
```diff
@@ -28,10 +28,18 @@ type Props = React.ElementConfig<typeof View> & {|
    * Callback which returns a React element to display on the left side.
    */
   left?: (props: { size: number }) => React.Node,
+  /**
+   * Style for the left.
```
jaysbytes: I think `Style for the left element wrapper would more precise.`

src/components/Card/CardTitle.js
```diff
@@ -28,10 +28,18 @@ type Props = React.ElementConfig<typeof View> & {|
...
+   */
+  leftStyle?: any,
```
jaysbytes: Can you type is as `ViewStyleProp`?
You can import it like this (see [ListAccordion](https://github.com/callstack/react-native-paper/blob/master/src/components/List/ListAccordion.js) as a refference)
```
import type {
  ViewStyleProp
} from 'react-native/Libraries/StyleSheet/StyleSheet';
```

Please apply these two comments for `rightStyle` also

PR #1019, opened 2019-04-24
Elevation applied after mode changes

jordy-wijman:
### Motivation
The wrong elevation is being applied to an outlined button when it is first being rendered as a contained button.

Reproduce steps:
1. Render a button using `mode={this.propsIsContained ? 'contained' : 'outline'}` and set `propsIsContained = 'true'`.
2. Change the property in the component to false during runtime `propsIsContained = false`
3. Observe that elevation is applied on the outlined button while it shouldn't.

**WRONG**
image

**GOOD**
image

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -221,7 +221,7 @@ class Button extends React.Component<Props, State> {
     };
     const touchableStyle = { borderRadius: roundness };
     const textStyle = { color: textColor, fontFamily };
-    const elevation = disabled ? 0 : this.state.elevation;
+    const elevation = disabled || mode !== 'contained' ? 0 : this.state.elevation;
 
     return (
       <Surface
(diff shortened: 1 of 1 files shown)


src/components/Button.js
```diff
@@ -221,7 +221,11 @@ class Button extends React.Component<Props, State> {
     };
     const touchableStyle = { borderRadius: roundness };
     const textStyle = { color: textColor, fontFamily };
-    const elevation = disabled ? 0 : this.state.elevation;
+    const elevation = disabled
+      ? 0
+      : mode !== 'contained'
+        ? new Animated.Value(0)
```
ferrannp: Does this need to be Animated? I think 0 will suffice.

jordy-wijman: I added this after I saw that the tests failed. I’m not familiar with the testing library so not sure how I can fix those.

Trancever: @jordy-wijman Try to run `yarn test -u` after you change it to `0` and add changed snapshots.

PR #1035, opened 2019-04-29
fix: set forceInset in Appbar.Header by default

lucasmotta:
Follows a similar fix made on the BottomNavigation where we explicitly tell `SafeAreaView` that the Header is placed on the top, so it does not have to calculate using `onLayout`.

### Motivation

The SafeAreaView component uses `onLayout` to calculate the insets. It would sometimes cause some visual flickering as the layout is being calculated. They also mention about this on their [documentation](https://github.com/react-native-community/react-native-safe-area-view#forceinset).

### Test plan

- [x] Unit tests (snapshots)
- [x] Tested on simulator and real device, the problem is gone

### DIFF
--- a/src/components/Appbar/AppbarHeader.js
+++ b/src/components/Appbar/AppbarHeader.js
@@ -1,7 +1,8 @@
 /* @flow */
 
 import * as React from 'react';
-import { View, SafeAreaView, StyleSheet } from 'react-native';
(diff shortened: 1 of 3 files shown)


ferrannp:
@lucasmotta can you add some gifs or visual on what this change is doing?

PR #1037, opened 2019-04-30
fix: elevation being applied after mode changes on a button while it shouldn't

jordy-wijman:
### Motivation
The wrong elevation is being applied to an outlined button when it is first being rendered as a contained button.

Reproduce steps:
1. Render a button using `mode={this.propsIsContained ? 'contained' : 'outline'}` and set `propsIsContained = 'true'`.
2. Change the property in the component to false during runtime `propsIsContained = false`
3. Observe that elevation is applied on the outlined button while it shouldn't.

**WRONG**
image

**GOOD**
image

FYI: duplicate of #1019 didn't know how to update that one.


ferrannp:
Thanks!

PR #1043, opened 2019-05-02
Fix placeholder alignment in inputFlatWithoutLabel

### LINKED ISSUE #870: TextInput with multiline and without label has incorrect placeholder alignment on iOS
### Current behaviour

`TextInput` placeholder has incorrect alignment on iOS when used with `multiline={true}` and without `label`.

### Expected behaviour

Expect `TextInput` with `multiline={true}` and `placeholder` to look the same, whether with or without `label`.

Issue happens only on iOS, not on Android (see screenshots below).

### Code sample

```jsx
<TextInput
  placeholder={"Placeholder"}
  multiline={true}
  numberOfLines={1}
/>
```

iOS  | Android 🤖
------------ | -------------
ios | android

Note: none of the inputs are focused, problem is on the first screenshot, second input.

### What have you tried

Have tried to set `textAlignVertical` to `center` but that's an Android-only property. Also tried to have the label as an empty string, which fix the alignment but hides the placeholder – which I need.

### Your Environment

| ios or android        | iOS
| react-native          | 0.57.8
| react-native-paper    | 2.10.1
| node                  | 11.9.0
| npm or yarn           | yarn
| expo sdk              | no

Taym95:
Try to fix #870

### DIFF
--- a/src/components/TextInput/TextInputFlat.js
+++ b/src/components/TextInput/TextInputFlat.js
@@ -280,6 +280,6 @@ const styles = StyleSheet.create({
     paddingBottom: 6,
   },
   inputFlatWithoutLabel: {
-    paddingVertical: 15,
+    paddingTop: 20,
   },
 });


ferrannp:
Hey @Taym95 we are fixing several TextInput issues all together in 3.0. I think we should check if this is still an issue in 3.0.

Taym95:
Yes sure!

Issue #455, opened 2018-07-16
Add customizable icon size for BottomNavigation


peacechen:
### Current behaviour

Icons are fixed at size 24 in BottomNavigation.

### Expected behaviour

Add a prop to support custom icon sizes.

### Code sample

I'm using `react-navigation-material-bottom-tabs` which in turn uses this library.  If I specify icon sizes greater than 24, the labels render in the center of the icon instead of below (see screenshot below).

```
let iconSize = sizes.h2;
let iconContainerStyle = [s.aic, {width: iconSize, height: iconSize}];

this.tabNav = createMaterialBottomTabNavigator(
  {
    Docs: { screen: Docs,
      navigationOptions: {
        tabBarLabel: 'Docs',
        tabBarIcon: ({ tintColor }) => (<View style={iconContainerStyle}><Icon style={[{color: tintColor}]} size={iconSize} name={'document'}/></View>),
        gesturesEnabled: false,
      }
    },
    Add: { screen: Add/>,
      navigationOptions: {
        tabBarLabel: 'Add',
        tabBarIcon: ({ tintColor }) => (<View style={iconContainerStyle}><Icon style={[{color: tintColor}]} size={iconSize} name={'add'}/></View>),
        gesturesEnabled: false,
      }
    },
    Search: { screen: Add/>,
      navigationOptions: {
        tabBarLabel: 'Search',
        tabBarIcon: ({ tintColor }) => (<View style={iconContainerStyle}><Icon style={[{color: tintColor}]} size={iconSize} name={'search'}/></View>),
        gesturesEnabled: false,
      }
    },
    Profile: { screen: Profile,
      navigationOptions: {
        tabBarLabel: 'Profile',
        tabBarIcon: ({ tintColor }) => (<View style={iconContainerStyle}><Icon style={[{color: tintColor}]} size={iconSize} name={'user'}/></View>),
        gesturesEnabled: false,
      }
    },
  },
  {
    initialRouteName: 'Manifest',
    stateName: props.name,
    activeTintColor: lightColors.primaryButtonFontColor,
    inactiveTintColor: lightColors.secondaryColor,
    barStyle: [navBackground, s.h4],
  }
);
```

### Screenshots (if applicable)

<img src="https://user-images.githubusercontent.com/6295083/42716526-8669bb2c-86c1-11e8-9149-b2be31558053.png">

### What have you tried

Different icon sizes.  Only renders properly if icon size is 24, the fixed size set in BottomNavigation.js.
[...]

PR #1061, opened 2019-05-13
feat: custom icon container and font sizes for BottomNavigation

JBaczuk:
### Motivation

In the BottomNavigation component, icons sizes are fixed to 24, and label font to 12. This makes it difficult to change the label font and size of the icons without clipping as mentioned in https://github.com/callstack/react-native-paper/issues/455 and https://github.com/react-navigation/material-bottom-tabs/issues/18

### Test plan

You can now add optional props `iconContainerSize` and `labelFontSize` to the component to set these.

### DIFF
--- a/src/components/BottomNavigation.js
+++ b/src/components/BottomNavigation.js
@@ -193,6 +193,14 @@ type Props<T> = {
   barStyle?: any,
   style?: any,
   /**
+   * Custom icon container size (square dimensions)
+   */
+  iconContainerSize?: Number,
+  /**
+   * Custom label font size
+   */
+  labelFontSize?: Number,
+  /**
    * @optional
    */
   theme: Theme,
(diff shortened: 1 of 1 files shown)


x5engine:
why the heck isn't this merged yet or approved? this is very important! I am struggling to change hte icon size which is very absurd to rely on that fixed size

x5engine:
this seriously very important

x5engine:
actually you don't even need this!
just do this

```
  handleIcon = ({route, focused= false, color }) => {
    return <Image style={{width:25, height: 25}} source={route.icon}/>
  }

  render() {
    return (
      <BottomNavigation
        shifting={false}
        navigationState={this.state}
        onIndexChange={this._handleIndexChange}
        renderIcon={this.handleIcon}
        renderScene={this._renderScene}
      />
    );
  }
```

PR #1063, opened 2019-05-14
refactor: restructure DrawerSection component

### LINKED ISSUE #1062: Applying backgroundColor to Drawer.Section leaks out to Divider component
## Description
I was trying to add color to the background of `Drawer.Section` but it is leaking out below the `Divider`.

Its highlighted in the picture.
bug

## Reproducible Demo
```js
<Drawer.Section style={{ backgroundColor: '#e5e0e0' }}>
  <View style={{ height: 100 }} />
</Drawer.Section>
```

brunohkbx:
Apply marginTop to Divider instead of marginVertical
Apply style and marginBottom to View wrapper

### Motivation
Fix #1062 

fix

### Test plan
1. Download this branch
2. Apply style with backgroundColor to a `Drawer.Section` in the example app.

### DIFF
--- a/src/components/Drawer/DrawerSection.js
+++ b/src/components/Drawer/DrawerSection.js
@@ -2,11 +2,12 @@
 
 import color from 'color';
 import * as React from 'react';
-import { View } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import Text from '../Typography/Text';
 import Divider from '../Divider';
 import { withTheme } from '../../core/theming';
 import type { Theme } from '../../types';
+import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet';
 
 type Props = React.ElementConfig<typeof View> & {
   /**
@@ -17,6 +18,7 @@ type Props = React.ElementConfig<typeof View> & {
    * Content of the `Drawer.Section`.
    */
   children: React.Node,
+  style?: ViewStyleProp,
   /**
    * @optional
    */
@@ -61,7 +63,7 @@ class DrawerSection extends React.Component<Props> {
   static displayName = 'Drawer.Section';
 
   render() {
-    const { children, title, theme, ...rest } = this.props;
+    const { children, title, theme, style, ...rest } = this.props;
     const { colors, fonts } = theme;
     const titleColor = color(colors.text)
       .alpha(0.54)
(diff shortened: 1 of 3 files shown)


src/components/Drawer/DrawerSection.js
```diff
@@ -70,19 +70,21 @@ class DrawerSection extends React.Component<Props> {
...
-              numberOfLines={1}
-              style={{ color: titleColor, fontFamily, marginLeft: 16 }}
-            >
-              {title}
-            </Text>
-          </View>
-        )}
-        {children}
-        <Divider style={{ marginVertical: 4 }} />
+      <View style={{ marginBottom: 4 }}>
```
Trancever: Do we really need additional `View` here? Can't we pass that style to the `View` on line 74?

Trancever: If margin won't do the work there, w can use padding probably. Also we should destructure `style` from props and pass it like that `style={[style, { paddingBottom: 4 }]}`. Default value for desctructured style should be empty object.

brunohkbx: @Trancever it worked with margin :smile: 
Please, take a look ba8db427e022f37b579475a6a1501f20c392cd7b

src/components/Drawer/DrawerSection.js
```diff
@@ -82,7 +84,7 @@ class DrawerSection extends React.Component<Props> {
           </View>
         )}
         {children}
-        <Divider style={{ marginVertical: 4 }} />
+        <Divider style={{ marginTop: 4 }} />
```
satya164: Move style objects to `StyleSheet.create`

src/components/Drawer/DrawerSection.js
```diff
@@ -70,7 +72,7 @@ class DrawerSection extends React.Component<Props> {
     const fontFamily = fonts.medium;
 
     return (
-      <View {...rest}>
+      <View style={[style, { marginBottom: 4 }]} {...rest}>
```
satya164: User passed `style` should be at the end
```suggestion
      <View style={[{ marginBottom: 4 }, style]} {...rest}>
```

brunohkbx:
Thanks @satya164. Done: 34bace7

src/components/Drawer/DrawerSection.js
Trancever: Good catch Satya :D

PR #1074, opened 2019-05-17
Fix typo in accessibilityRole in DialogTitle

sstur:
### Motivation

This solves the following error when using this library on web:
Screen Shot 2019-05-17 at 15 02 21

See [documentation here](https://github.com/necolas/react-native-web/blob/master/docs/guides/accessibility.md#accessibilityrole).

### Test plan

Manually opened in a web browser and confirmed the error is gone.

### DIFF
--- a/src/components/Dialog/DialogTitle.js
+++ b/src/components/Dialog/DialogTitle.js
@@ -59,7 +59,7 @@ class DialogTitle extends React.Component<Props> {
     return (
       <Title
         accessibilityTraits="header"
-        accessibilityRole="header"
+        accessibilityRole="heading"
         style={[styles.text, { color: theme.colors.text }, style]}
         {...rest}
       >


sstur:
Turns out there's a Flow error (below). I'm now wondering if it's react-native-web that has the header/heading term mixed up.

Screen Shot 2019-05-17 at 16 01 34


/cc @necolas

necolas:
https://github.com/necolas/react-native-web/commit/39d2e18ccfda6e6fd27f01f61198c27d2281107b

sstur:
Thanks @necolas! I've created [PR 1350](https://github.com/necolas/react-native-web/pull/1350) to update the docs to match those of react-native and hopefully avoid the confusion in future.

Sorry for the noise on react-native-paper.

PR #1078, opened 2019-05-20
fix: move paddingHorizontal out from content style

jose96043:
### Motivation

The ability of overriding style properties makes things more flexible. In order to achieve this within DataTableRow,  we have moved paddingHorizontal out from content style which can't be overridden, and put it in container style which can be overridden with the style class property.

### DIFF
--- a/src/components/DataTable/DataTableRow.js
+++ b/src/components/DataTable/DataTableRow.js
@@ -40,7 +40,7 @@ class DataTableRow extends React.Component<Props> {
         onPress={onPress}
         style={[styles.container, { borderBottomColor }, style]}
       >
-        <View style={styles.content}>{this.props.children}</View>
+        <View style={[styles.content, style]}>{this.props.children}</View>
       </TouchableRipple>
     );
   }
--- a/src/components/DataTable/DataTableRow.js
+++ b/src/components/DataTable/DataTableRow.js
@@ -40,7 +40,7 @@ class DataTableRow extends React.Component<Props> {
         onPress={onPress}
         style={[styles.container, { borderBottomColor }, style]}
       >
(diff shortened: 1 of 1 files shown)


src/components/DataTable/DataTableRow.js
```diff
@@ -40,7 +40,7 @@ class DataTableRow extends React.Component<Props> {
         onPress={onPress}
         style={[styles.container, { borderBottomColor }, style]}
       >
-        <View style={styles.content}>{this.props.children}</View>
+        <View style={[styles.content, style]}>{this.props.children}</View>
```
satya164: I don't think we want to duplicate style like this, you apply some padding and now it'll get applied 2 times

jose96043: sorry good point can we have a content field exposed as part of class then?

satya164: I don't remember why the inner view is needed at all. Can we remove it? cc @jaulz

jaulz: I think I put it there because a `TouchableRipple` can only have a single child.

jose96043: @jaulz Is there a reason why there needs to be a horizontalPadding of 16?

satya164: How about moving the padding to the `TouchableRipple` styles so it can be overriden by user's passed style?

jaulz: Yep, that should be fine. The padding is part of the [specs](https://material.io/design/components/data-tables.html#specs).

jose96043: @satya164  I have updated the commit with your suggestion. I tested it and it works as expected. Please take a look. Thanks!

PR #1079, opened 2019-05-20
feat: Allow Override of Button Label Text Styles

scandinaro:
### Motivation

I wanted to be able to increase the font size of the text displayed within the Button component

### Test plan

- Add an optional `textStyles` object parameter to your react-native-paper Button component as displayed below:

```javascript
<Button
  {...props}
  mode="contained"
  textStyles={{ fontSize: 25 }}
/>
```

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -65,6 +65,11 @@ type Props = React.ElementConfig<typeof Surface> & {|
    * Use this prop to apply custom height and width.
    */
   contentStyle?: any,
+  /**
+   * Style of button's text component.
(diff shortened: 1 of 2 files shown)


scandinaro:
@ferrannp @satya164 anything else I can do to move this along?

ferrannp:
@scandinaro the reason we do not have such prop is that you can pass your own Text component as children. That should be sufficient for your case.

scandinaro:
@ferrannp It looked like the code drops the children into a text field so wouldn't doing that as you suggested put a text component within another text component?

PR #1083, opened 2019-05-23
fix: CrossFadeIcon with function-based source

mars-lan:
Expand isValidIcon check to include function-based icon source so that CrossFadeIcon will work properly in this case.


mars-lan:
Actually this causes some unintended problems as icons generated from render function are not easily comparable.

PR #1084, opened 2019-05-23
fix: wrap overflowing menu in scrollView

### LINKED ISSUE #1075: Menu goes outside of screen when many items are present
## Environment
Pretty sure it happens everywhere, I checked on iOS and Android, not web.

## Description
When using the menu component with many items, it overflows out of the screen, towards the top. See the demo for an example. I noticed that it also expands towards the wrong side (top instead of bottom), this only happens when lots of items are present. It would be great if it would just expand towards the bottom with a scrollable view inside of it.

## Reproducible Demo
Here is a snack that demo's the issue, I just took the example in the docs, and copy pasted one of the items many times. You will see that the component will overflow out of the screen.
https://snack.expo.io/Syejcq62E

iyadthayyil:
### Motivation
fixes #1075

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -303,7 +303,6 @@ class Menu extends React.Component<Props, State> {
     );
   };
 
-
   _hide = () => {
     this._removeListeners();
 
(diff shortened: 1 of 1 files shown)


jaysbytes:
Thanks for the PR @iyadthayyil! Sorry for such a long delay

PR #1088, opened 2019-05-24
Feat/tooltip

brunohkbx:
### Motivation
Add component Tooltip

## Description
In this section, I'm going to highlight some key points of the implementation that I'm not sure if it's ok.

1. I had to pass a prop `onLongPress` on [L242](https://github.com/callstack/react-native-paper/pull/1088/files#diff-533ac842d2af86cd70bf4e8d88baf6a6R242) to prevent the children's touchable to respond to an `onPress` event right after the end of `onLongPress`

2. I don't know why but I had to use a setTimeout with 500 ms to properly compute the Y position of the children on [L126](https://github.com/callstack/react-native-paper/pull/1088/files#diff-533ac842d2af86cd70bf4e8d88baf6a6R126). It will get random values for Y without this setTimeout or with a smaller value.
I've found the same problem on this implementation: https://github.com/react-native-training/react-native-elements/blob/master/src/tooltip/Tooltip.js#L145 and it's somehow related to https://github.com/facebook/react-native/issues/19497

3. The tooltip starts with `opacity: 0` and then it changes to `0.9` when its measure gets calculated at [L132](https://github.com/callstack/react-native-paper/pull/1088/files#diff-533ac842d2af86cd70bf4e8d88baf6a6R132). Should I have used animation to change this value? 

## TODOs
- [x] ~Fix tooltip coords when orientation change~ - 8ea66a8
onLayout doesn't trigger on children component when the orientation changes, so I had to remove it.
- [x] ~Fix flow~ - b7bd00dba65c7d70759332b5c048695f6e51970d
- [x] ~Test on IOS~ - It works!! :smiley: 
- [x] ~Add normal and tsx tests~ - b50fadd
- [x] ~Add docs~ - 8875998
- [ ] Implement tooltip placement prop ?

## Material Design specs
1. Don't crop Tooltip on the left side
left

2. Don't crop Tooltip on the right side
right

3. Don't crop Tooltip on the bottom
bottom

4. Landscape mode
landscape

### Test plan
1. Download this branch
2. Run example app
3. Open `Appbar` or `Toggle Button` section and hold press on any icon with Tooltip.

## References
* [Material Design Tooltip specs](https://material.io/design/components/tooltips.html)
* Material-UI - [Tooltip API](https://material-ui.com/api/tooltip/)

## Screenshots
tooltip1

tooltip1


andresmtz98:
Any update on this?

brunohkbx:
@andresmtz98 I'm waiting for @Trancever's review.

Trancever:
@brunohkbx I resolved all the conflicts + added support for typescript locally but I can't push it, because you created PR from your fork. I am pretty sure you have a write access, so please use main repo instead of fork next time :D

Could you open new PR from `brunohkbx-feat/tooltip` branch from main repo so the contribution will be counted as yours?

brunohkbx:
I didn't realize that I was working from the forked repository. :disappointed: 
Thanks for the rebase @Trancever.
Done #1185

Issue #926, opened 2019-03-15
use MaterialCommunityIcons instead of MaterialIcons


alikazemkhanloo:
## Describe the feature
Use MaterialCommunityIcons instead of MaterialIcons by default in all the components that take the `icon` prop.

## Motivation
I found out MaterialIcons lacks a lot of icons like `logout` of `thumb-up-outline`. But MaterialCommunityIcons looks like MaterialIcons and also have lots of other useful icons. Is there a reason you chose MaterialIcons? at least we can let the devs to select their preferences.

Zefau:
I wrote a component for myself to use all the sets from `react-native-vector-icons`. Can probably be optimized, but currently looks like:

```
import React from 'react'
import { View, StyleSheet } from 'react-native'

import MaterialIcons from 'react-native-vector-icons/MaterialIcons'
import MaterialCommunityIcons from 'react-native-vector-icons/MaterialCommunityIcons'
import FontAwesome from 'react-native-vector-icons/FontAwesome'

let Icons = {
	MaterialIcons,
	MaterialCommunityIcons,
	FontAwesome,
}


export default class Icon extends React.Component
{
	static defaultProps = {
		size: 40,
	}
	
	render()
	{
		let { icon, size, style, backgroundColor, textColor, color, onPress } = this.props
		
		let iconSet = 'MaterialIcons'
		if (icon.indexOf('|') > -1) [ iconSet, icon ] = icon.split('|') // this.props.icon may contain icon set by defining "iconSet|icon"
		
		backgroundColor = backgroundColor || 'transparent'
		textColor = color || (textColor || '#525252')
		
		let TagName = Icons[iconSet]
		return(

<View
	style={[
		{
			width: size,
			height: size,
			borderRadius: size / 2,
			backgroundColor,
		},
		styles.container
	]}>
	
	<TagName name={icon} size={size * 0.6} color={textColor} style={style} onPress={onPress} />
</View>

		)
	}
}
```

Can be used by `<Icon icon="android" />` (which uses MaterialIcons as default) or `<Icon icon="MaterialCommunityIcons|android-head" />` (which uses MaterialCommunityIcons).
@satya164 Maybe you'll find is useful to use

connercms:
you can render any icon you want from react-native-vector-icons by passing it as a prop

example 

```
<IconButton 
icon={({ size, color}) => (
    <MaterialCommunityIcons 
    name={'logout'} 
    size={size} 
    color={color}/>
)} 
[...]

PR #1091, opened 2019-05-26
Feat: use Material Design Community Icons

jaysbytes:
This PR introduces the replacement of Material Icons to Material Community Icons. 
Some of the icons have different names, that's why there are so many changes.

### Motivation
#926 
### Test plan

Unfortunately, it needs to be tested visually. 
1.  Open example app and go through all examples to verify if there is no `?` in place of an icon
2. Open all snacks from the documentation page.

I changed all the icons one-by-one so there shouldn't be anything missing, but it can be double checked.

### DIFF
--- a/example/src/DrawerItems.js
+++ b/example/src/DrawerItems.js
@@ -29,7 +29,7 @@ const DrawerItemsData = [
   { label: 'Inbox', icon: 'inbox', key: 0 },
   { label: 'Starred', icon: 'star', key: 1 },
   { label: 'Sent mail', icon: 'send', key: 2 },
-  { label: 'Colored label', icon: 'color-lens', key: 3 },
+  { label: 'Colored label', icon: 'palette', key: 3 },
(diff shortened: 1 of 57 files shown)


ferrannp:
@jaysbytes do you think is possible to have a list between old icons vs new icons ? Like `favorite vs heart` ?

jaysbytes:
We can prepare a list of icons used in our app based on diff in PR. But I'm not sure if it is possible to whole icons set. But users can use this page https://materialdesignicons.com/ to search by icon names. Icons are usually indexed for names form both packages. For some icons there are no equivalents. i.e  there is no `add-a-photo` in new icons set. There is only `camera` but without this `+` sign

Trancever:
I am ok with merging it when we add info to docs where to look for the icons + we have to make sure that our icon search in docs works properly with new icons.

Trancever:
Closing this in favour of #1210

Issue #1098, opened 2019-06-02
Dynamically choose Icon set from react-native-vector-icons through a prop


yoeven:
Wouldn't it better to choose an icon by exposing a prop rather than having a render method to render another icon component from the same react-native-vector-icons library. I would suggest a prop for the Icon component that allows us to choose an icon set. This has actually been done by [react-native-elements](https://react-native-training.github.io/react-native-elements/docs/icon.html), it would be great to have this simple feature in this library along with all the other components.

jbinda:
I agree that its the most flexible way to select Icons.

I will take a closer look on that when we merge #1091

jbinda:
Feature will be handle with #1210

PR #1099, opened 2019-06-02
feat: added ability to choose icon set

yoeven:
Any component that takes a prop of type IconSource will be able to use key "type" in the source object to change the Icon set from the react-native-vector-icons library.

Example:
```
<Button icon={{ source: 'folder', direction: 'rtl', type: 'Feather' }}>
                  Press me
</Button>
```

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -16,6 +16,7 @@ let AntDesign,
   Octicons,
   Zocial,
   SimpleLineIcons;
+
 const IconSets = {
   AntDesign,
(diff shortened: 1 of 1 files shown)


jbinda:
reffers to #1098

jbinda:
I have also rebase to current callstack:master branch, adjust logic and fix flow/lint issues

jbinda:
After internal discussion we decided to leave implement this feature after figure out better way to handle iconsets

PR #1109, opened 2019-06-05
Fix example erros in docs

netochaves:
### Motivation
Fix erros on modal and fab groups examples. #1105 

### Test plan

Check the docs, and run the snack examples.


Taym95:
Looks good, since the error was sating that you forget the provider.

netochaves:
@Taym95 Yes, all components that need a portal wrapper also need a provider wrapper.

PR #1118, opened 2019-06-09
Customize main container style in Snackbar

zmnv:
### Motivation

In your case with:
```
  wrapper: {
    position: 'absolute',
    bottom: 0,
    width: '100%',
  },
```

I can't use more than one snackbar at the same time. They overlap each other.
I need to change 'position' to 'relative' inside my components for example.

Keep 'absolute' by default;

### Test plan

```jsx
<Fragment>
    <Snackbar />
    <Snackbar />
</Fragment>
```

### DIFF
--- a/src/components/Snackbar.js
+++ b/src/components/Snackbar.js
@@ -38,7 +38,14 @@ type Props = {|
    * Text content of the Snackbar.
    */
   children: React.Node,
+  /**
+   * Snackbar Surface style
+   */
   style?: ViewStyleProp,
+  /**
+   * Snackbar SafeAreaView style
+   */
+  wrapperStyle?: ViewStyleProp,
   /**
    * @optional
    */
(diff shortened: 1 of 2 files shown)


ferrannp:
@zmnv thanks for sending a PR. However, in this library we try to match the spec of Material Design as close as possible. Showing multiple `SnackBar`(s) is not recommended. See https://material.io/design/components/snackbars.html#behavior (Consecutive snackbars). 

This is the reason why the team has decided no to merge this PR.

Thank you again!

zmnv:
Okay. How about to change Button style in Snackbar? 

Your letter spacing is UGLY and has old material design version.

zmnv:
ALSO HOW ABOUT ABSOLUTE POSITION? 

I can't use Material requirements with absolute.... It just overlap BottomNavigation. I don't wanna handle that case with offsets and other things...

Okay.... OKAY.....

ferrannp:
Everything you said is not related to this PR which was passing a style prop to the `SafeAreaView`.

- If you think the component has bugs, please fill an issue
- If you think the component needs to be written different or need extra functionality according to Material Design, you can open an issue and write a RFC

Thank you.

PR #1124, opened 2019-06-10
docs: Adding <Text> to avoid errors

rodriigovieira:
## The change

Add a `<Text>` in the documentation of the `Dialog.ScrollArea` to avoid errors while copying without making changes.

When copying the code and trying it in the editor, an error is returned, demanding that the "This is a scrollable area" text be wrapped in a `<Text>` component.

Although in most cases the scrollable content won't be a `<Text>` component, I think it would nice to do something to avoid the error. As an alternative, a `<View>` with a few `<Text>` inside could be added.

## Motivation

By fixing this, it'll avoid errors by beginners or by someone who's just copying the code and trying to check things out, avoiding the unnecessary trouble.

### DIFF
--- a/src/components/Dialog/DialogScrollArea.js
+++ b/src/components/Dialog/DialogScrollArea.js
@@ -19,8 +19,8 @@ type Props = React.ElementConfig<typeof View> & {
  * ## Usage
  * ```js
  * import * as React from 'react';
- * import { ScrollView, Text } from 'react-native';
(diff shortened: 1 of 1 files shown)


ferrannp:
@rodriigovieira can you import `Text` from `react-native-paper` ? :)

rodriigovieira:
I just did. Did I do it correctly?

ferrannp:
Yup, thank you!

PR #1132, opened 2019-06-14
fix: pass flattened styles to adjustPadding

jaulz:
### Motivation

If plain styles are passed, they cannot be used to read properties of it.

### Test plan

n/a

### DIFF
--- a/src/components/TextInput/TextInputFlat.js
+++ b/src/components/TextInput/TextInputFlat.js
@@ -155,7 +155,7 @@ class TextInputFlat extends React.Component<ChildTextInputProps, {}> {
       label,
       scale: fontScale,
       isAndroid: Platform.OS === 'android',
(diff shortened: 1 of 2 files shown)


Trancever:
@jaulz What is this PR fixing?

jaulz:
The issue is that if you pass a style to `StyleSheet.create` it will generate an id. Hence, you cannot simply pass styles to `adjustPaddingFlat`:
https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput/TextInputFlat.js#L147-L166

and read them afterwards like this:
https://github.com/callstack/react-native-paper/blob/cbfa18194c7548eb00cd842b4c5da69f6198e263/src/components/TextInput/helpers.js#L147

On web, this caused an issue which positioned the text input content incorrectly (to close to the label):
image

jbinda:
Thanks @jaulz ! I missed broken padding in web.

PR #1143, opened 2019-06-21
feat: add contentStyle prop to banner

jaulz:
### Motivation

Add `contentStyle` prop to banner.

### Test plan

n/a


satya164:
Couldn't you pass width in the style prop? If it doesn't work, we should make it work. cc @Trancever

jaulz:
@satya164 the goal is to implement banners like mentioned [here](https://material.io/design/components/banners.html#placement) that have a specific `maxWidth` in the inner container. I don't see any possibility via `style` prop only or do you have an idea? 
image

satya164:
Thanks for clarifying. It'll be better to have this info when sending a PR :)

Maybe we should do this by default? It seems reasonable.

jaulz:
Sorry for that 😊 We could default it and instead add a `maxWidth` property. What do you think?

satya164:
I'll leave it upto @Trancever

wojteg1337:
@jaulz I'm up to `style` prop passed to the component instead `maxWidth`.
Additionally, you will have to solve conflicts

@Trancever make decision 🥉

Trancever:
@jaulz Can we have:
-  maxWidth specified by default (is there any information in MD Guidelines about value?)
- style prop that can override it

jaulz:
@Trancever I would suggest to use 960 since that's the minimum width for a tablet in landscape mode from the [breakpoint system](https://material.io/design/layout/responsive-layout-grid.html#breakpoints). There is no value specified and I assume it usually depends on the content which is below. Would you like to have the `maxWidth` as part of the `style` prop or as part of a new `contentStyle` prop?

Trancever:
@jaulz Go ahead with `contentStyle` and 960 :)

jaulz:
@Trancever please check 😊

jaulz:
@Trancever thanks!

PR #1147, opened 2019-06-23
fix: make appbarcontentprops extend touchableripple props

yogurtandjam:
### Motivation

In a typescript project, I am attempting to add an onPress property as it is listed in the react-native-paper api https://callstack.github.io/react-native-paper/appbar-content.html but it is not listed in the AppBar.Content 
```
export interface AppbarContentProps {
  color?: string;
  title: React.ReactNode;
  titleStyle?: StyleProp<TextStyle>;
  subtitle?: React.ReactNode;
  subtitleStyle?: StyleProp<TextStyle>;
  style?: StyleProp<ViewStyle>;
  theme?: ThemeShape;
}
```

### Test plan

Import Appbar component from react-native-paper and add an onPress property to Appbar.content


ferrannp:
Can you check if this is still an issue in the latest alpha? We don't have Appbar.d.ts anymore.

PR #1157, opened 2019-06-27
fix: FAB - always display activity indicator if loading property is set to true

YajanaRao:
Displaying activity indicator if loading property is set to true. activity indicator was not displaying if label is not set. 
Removed condition to check for label is loading property is set true.

### Motivation
Pull request to address callstack#1130

### Test plan

Testing different combination like
Icon and Loading 
Label and Loading
No Label and Icon


ferrannp:
@YajanaRao can you provide some gifs or screenshots about this change? Is there any example in the Material guide? With the team we are discussing if this makes really sense.

YajanaRao:
Here is the screenshot of the feature implemented:
image

It is required in changing the icon which is dependent on some data load.

PR #1165, opened 2019-06-30
feat: added prop to override fontSize

netbull:
Possibility to set the fontSize when "string" is passed as child.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -57,6 +57,10 @@ type Props = React.ComponentProps<typeof Surface> & {
    */
   accessibilityLabel?: string;
   /**
+   * Font size if the children are string
+   */
+  fontSize?: number;
+  /**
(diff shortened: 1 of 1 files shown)


jaysbytes:
Thank you for the PR but unfortunately, this prop is too specific. In the future, someone would like to have bolded text, different font size, different font family etc. Adding `labelStyle` prop would be a better approach. See related issue #1273 .

netbull:
@jaysbytes Thanks! That's a valid point. :)

pan-pawel:
@netbull I added PR with labelStyle for button. I will notify you if we add it to paper

pan-pawel:
@netbull we added labelStyle to master, so with next release you will be able to set custom font for button

netbull:
@pan-pawel that's great! thanks :)

PR #1166, opened 2019-06-30
fix: disable menu interaction while hiding

iyadthayyil:
### Motivation
This fix disables interaction with menu while it is hiding.
Similar to #1116 

### Test plan
N/A


Trancever:
@iyadthayyil Thanks!

PR #1167, opened 2019-07-01
feat: added possibility for transparent input

netbull:
when used with the prop `transparent={true}` the input's background is removed and the label is aligned with the border

### Motivation
When the input is transparent the label is not aligned with the bottom border and there is no way so far to align it.

### Test plan
All original tests should pass, because there are no breaking changes

### DIFF
--- a/src/components/TextInput/TextInput.js
+++ b/src/components/TextInput/TextInput.js
@@ -57,6 +57,10 @@ export type TextInputProps = {|
    */
   multiline?: boolean,
   /**
+   * Whether the input should be transparent.
+   */
+  transparent?: boolean,
+  /**
    * The number of lines to show in the input (Android only).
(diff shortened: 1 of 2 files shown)


satya164:
I believe there is `padding: 'none'` prop

netbull:
@satya164 can you give a quick example please.. I was not able to find any info.. :(

satya164:
Please check master.

netbull:
@satya164 that is for v3... what about v2?

pixelize:
@netbull this would be awesome feature, hope it will be merged soon ! :)

jbinda:
Hi ! Thanks for raising the PR. As Satya said this feature is already implemented in 3.0 branch (please check master)

PR #1169, opened 2019-07-01
feat: add ToggleButton.Row component

### LINKED ISSUE #1168: Add a `ToggleButtonRow` component
We have a `ToggleButtonGroup` component which is supposed to provide a way to manage multiple toggle buttons from one place, similar to `RadioButtonGroup`. However, when grouping toggle buttons, it maybe desirable to group them in a row and remove left/right borders of adjacent buttons.

The `ToggleButtonGroup` confusingly adds some styles for borders, but doesn't add style for a row. We should add a `ToggleButtonRow` component which does this and remove all style related code from `ToggleButtonGroup`. The `ToggleButtonRow` component could include `ToggleButtonGroup` by default.

Trancever:
### Motivation

Fix #1168

### DIFF
--- a/src/components/ToggleButton/ToggleButtonGroup.tsx
+++ b/src/components/ToggleButton/ToggleButtonGroup.tsx
@@ -25,7 +25,8 @@ export const ToggleButtonGroupContext = React.createContext<
 >(null as any);
 
 /**
- * Toggle group allows to control a group of toggle buttons.
+ * Toggle group allows to control a group of toggle buttons.</br>
+ * It doesn't change the appearance of the toggle buttons. If you want to group them in a row, checkout <a href="/toggle-button-row.html">`ToggleButton.Row`</a>.
  *
  * ## Usage
  * ```js
--- a/src/components/ToggleButton/ToggleButtonGroup.tsx
+++ b/src/components/ToggleButton/ToggleButtonGroup.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
-import { StyleSheet } from 'react-native';
-import ToggleButton from './ToggleButton';
+import { View, StyleProp, ViewStyle } from 'react-native';
 
 type Props = {
   /**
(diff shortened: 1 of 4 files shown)


src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -68,51 +67,10 @@ class ToggleButtonGroup extends React.Component<Props> {
...
-                  : styles.middle,
-                // @ts-ignore
-                child.props.style,
-              ],
-            });
-          }
-
-          return child;
-        })}
+        <View style={style}>{children}</View>
```
satya164: I think this wrapper view can be removed

Trancever: Ok, then we don't want to accept style prop, right?

satya164: Yea

Trancever: Fixed :)

src/components/ToggleButton/ToggleButtonRow.tsx
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * React elements containing toggle buttons.
+   */
+  children: React.ReactNode;
+  style?: StyleProp<ViewStyle>;
+};
+
+/**
+ * Toggle row renders buttons in a row and applies predefined styles.
+ *
```
satya164: nit: remove extra new line

src/components/ToggleButton/ToggleButtonRow.tsx
```diff
@@ -0,0 +1,111 @@
...
+  value: string;
+  /**
+   * React elements containing toggle buttons.
+   */
+  children: React.ReactNode;
+  style?: StyleProp<ViewStyle>;
+};
+
+/**
+ * Toggle row renders buttons in a row and applies predefined styles.
```
satya164:
```suggestion
 * Toggle button row renders a group of toggle buttons in a row.
```

PR #1171, opened 2019-07-03
Rename index.native.js to index.native.tsx

tasn:
This looks like an oversight while moving from flow to typescript.

## Extra information

There really isn't much to it, just a wrongly named file.


satya164:
I think this file needs to be migrated to TS properly

tasn:
Sorry, upon taking a second look, it looks like it's still in flow so was just not converted. I guess I'll close this PR for now.

PR #1174, opened 2019-07-03
fix(appbar): make allowed children to accept additional props

### LINKED ISSUE #1215: Appbar doesn't pass props to Appbar content
## Environment

Latest React Native Paper. Both Android and iOS

## Description

In appbar [render](https://github.com/callstack/react-native-paper/blob/1899651cdc517a6996e228a4cf0d8962ae350f89/src/components/Appbar/Appbar.tsx#L138-L169) we check if child is valid Appbar child like "Appbar Content" and then we passing props to it, otherwise we just returning child as it is.
https://github.com/callstack/react-native-paper/blob/1899651cdc517a6996e228a4cf0d8962ae350f89/src/components/Appbar/Appbar.tsx#L138-L149
I noticed that Appbar Content `child.type.name` is undefined, that's why it always return in line 148 and never passes props to it
https://github.com/callstack/react-native-paper/blob/1899651cdc517a6996e228a4cf0d8962ae350f89/src/components/Appbar/Appbar.tsx#L169

Zrzut ekranu 2019-07-15 o 19 25 16
Zrzut ekranu 2019-07-15 o 19 24 53
In those examples appbar Content didnt recieve correct color prop from Apbbar, that's why title has wrong color

Can anyone check if it's my bad or Appbar Content doesn't receive props from Appbar (for example color prop)

brunohkbx:
### Motivation
This commit(#1001) is missing from master and I couldn't find a branch to cherry-pick, so I'm opening this PR again.

Resolve #1215

### DIFF
--- a/src/components/Appbar/Appbar.tsx
+++ b/src/components/Appbar/Appbar.tsx
@@ -138,12 +138,11 @@ class Appbar extends React.Component<Props> {
             if (
               !React.isValidElement(child) ||
               ![
-                'AppbarContent',
-                'AppbarAction',
-                'AppbarBackAction',
-                'AppbarHeader',
+                AppbarContent,
+                AppbarAction,
+                AppbarBackAction,
                 // @ts-ignore
-              ].includes(child.type.name)
+              ].includes(child.type)
             ) {
               return child;
             }
--- a/src/components/__tests__/Appbar/Appbar.test.js
+++ b/src/components/__tests__/Appbar/Appbar.test.js
@@ -15,4 +15,18 @@ describe('Appbar', () => {
 
     expect(tree).toMatchSnapshot();
   });
+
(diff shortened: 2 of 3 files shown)


src/components/Appbar/Appbar.tsx
```diff
@@ -138,12 +138,11 @@ class Appbar extends React.Component<Props> {
             if (
               !React.isValidElement(child) ||
               ![
-                'AppbarContent',
-                'AppbarAction',
-                'AppbarBackAction',
-                'AppbarHeader',
```
Trancever: Why is AppbarHeader completely removed?

brunohkbx: It doesn't have a `color` prop.

brunohkbx: @Trancever It doesn't have a color prop.

pan-pawel:
Resolving #1215

PR #1175, opened 2019-07-04
feat: Export useTheme hook from react-theme-provider

### LINKED ISSUE #994: Export useTheme from react-theme-provider
This might also give us the possibility for better typings when extending theme.

jshearer:
### Motivation
Instead of wrapping every component that needs the theme with `withTheme`, in React 16.8+ and RN 0.59+ you can use `useTheme()` as a hook to get the theme with much less hassle.

This would close #994.

### Test plan

I don't believe this would break any tests.

### DIFF
--- a/src/core/theming.tsx
+++ b/src/core/theming.tsx
@@ -2,4 +2,6 @@ import { createTheming } from '@callstack/react-theme-provider';
 import DefaultTheme from '../styles/DefaultTheme';
 import { Theme } from '../types';
 
-export const { ThemeProvider, withTheme, useTheme } = createTheming<Theme>(DefaultTheme);
+export const { ThemeProvider, withTheme, useTheme } = createTheming<Theme>(
+  DefaultTheme
+);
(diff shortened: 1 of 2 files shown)


jshearer:
Thanks for pointing that out @satya164 😄Made the change you suggested. Looks good?

src/core/theming.tsx
```diff
@@ -2,4 +2,4 @@ import { createTheming } from '@callstack/react-theme-provider';
 import DefaultTheme from '../styles/DefaultTheme';
 import { Theme } from '../types';
 
-export const { ThemeProvider, withTheme } = createTheming<Theme>(DefaultTheme);
+export const { ThemeProvider, withTheme, useTheme } = createTheming<Theme>(DefaultTheme);
```
wojteg1337: Fix linting issue:
  5:76  error  Replace `DefaultTheme` with `⏎··DefaultTheme⏎`  prettier/prettier

jaulz:
@wojteg1337 I just run prettier on his branch. Could you review please?

PR #1180, opened 2019-07-05
fix: add $Omit for lower Typescript versions

jaulz:
### Motivation

Only recently Typescript introduced `Omit`. If you are still with a lower version you will encounter several type issues. This PR will add a local helper `$Omit`.

### Test plan

n/a


wojteg1337:
hello @jaulz, how it would help if we use typescript from package.json. I think no one should use global TS version

PR #1181, opened 2019-07-05
fix: add contentStyle prop to Menu

jaulz:
### Motivation

This will allow us to apply inner styles which could be useful if you want to have a Menu for a TextInput and don't want to have rounded corners at the top for example.

### Test plan

n/a

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -47,6 +47,10 @@ type Props = {
    * Content of the `Menu`.
    */
   children: React.ReactNode;
+  /**
+   * Style of menu's inner content.
+   */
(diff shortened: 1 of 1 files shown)


satya164:
`theme={{ roundness: 0 }}` ?

jaulz:
That would also clear the bottom corners.

wojteg1337:
I'm not sure, but adding new prop would break current snapshots. If all tests pass it's fine for me :)

jaulz:
@wojteg1337 actually, adding this new prop did not break any snapshots as long as it was undefined and all test passed. However, I added a new test case and also reworked the existing tests because they were not really working as they did not cover the actual menu but only the anchor.

PR #1183, opened 2019-07-05
fix: reduce delay of Menu

jaulz:
### Motivation

The menu does only appear after half a second on iOS and Android. This fix will reduce the delay to milliseconds and was tested on all devices.

### Test plan

Open Menu examples.

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -262,7 +262,7 @@ class Menu extends React.Component<Props, State> {
       !anchorLayout.width ||
       !anchorLayout.height
     ) {
-      setImmediate(this._show);
+      requestAnimationFrame(this._show);
       return;
     }
 
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -262,7 +262,7 @@ class Menu extends React.Component<Props, State> {
       !anchorLayout.width ||
       !anchorLayout.height
     ) {
-      setTimeout(this._show, ANIMATION_DURATION);
+      setImmediate(this._show);
       return;
     }
 


src/components/Menu/Menu.tsx
```diff
@@ -262,7 +262,7 @@ class Menu extends React.Component<Props, State> {
       !anchorLayout.width ||
       !anchorLayout.height
     ) {
-      setTimeout(this._show, ANIMATION_DURATION);
+      setImmediate(this._show);
```
satya164: Any reason this needs to be delayed at all?

jaulz: There's a comment above the condition saying that there are issues on native devices sometimes so I kept it as it is.

satya164: Can we replace the non-standard `setImmediate` with `requestAnimationFrame` here?

jaulz: Sure, done 😊

Akshay-Katariya:
Is this fix incorporated in latest version because I am still facing delay while opening menu :( 
It's not that snappy there's still long delay

```
"react-native-paper": "^2.16.0",
```

Trancever:
@Akshay-Katariya This is going to be included in 3.0 version.

denniske:
Is this as fast as it gets?

`"react-native-paper": "^3.10.1",`

Screen5

libragopi:
Yeah, I have the same problem. It takes about a second to load the menu. 

I use: "react-native-paper": "^4.0.1"

FreshlyBrewedCode:
ZJkxXRQIbN

not as bad as @denniske but does not feel snappy at all. It also has roughly the same delay when disapearing.

PR #1185, opened 2019-07-05
Feat/tooltip

brunohkbx:
### Motivation
Add component Tooltip

## Description
In this section, I'm going to highlight some key points of the implementation that I'm not sure if it's ok.

1. I had to pass a prop `onLongPress` on [L242](https://github.com/callstack/react-native-paper/pull/1088/files#diff-533ac842d2af86cd70bf4e8d88baf6a6R242) to prevent the children's touchable to respond to an `onPress` event right after the end of `onLongPress`

2. I don't know why but I had to use a setTimeout with 500 ms to properly compute the Y position of the children on [L126](https://github.com/callstack/react-native-paper/pull/1088/files#diff-533ac842d2af86cd70bf4e8d88baf6a6R126). It will get random values for Y without this setTimeout or with a smaller value.
I've found the same problem on this implementation: https://github.com/react-native-training/react-native-elements/blob/master/src/tooltip/Tooltip.js#L145 and it's somehow related to https://github.com/facebook/react-native/issues/19497

3. The tooltip starts with `opacity: 0` and then it changes to `0.9` when its measure gets calculated at [L132](https://github.com/callstack/react-native-paper/pull/1088/files#diff-533ac842d2af86cd70bf4e8d88baf6a6R132). Should I have used animation to change this value? 

## TODOs
- [x] ~Fix tooltip coords when orientation change~ - 8ea66a8
onLayout doesn't trigger on children component when the orientation changes, so I had to remove it.
- [x] ~Fix flow~ - b7bd00dba65c7d70759332b5c048695f6e51970d
- [x] ~Test on IOS~ - It works!! :smiley: 
- [x] ~Add normal and tsx tests~ - b50fadd
- [x] ~Add docs~ - 8875998
- [ ] Implement tooltip placement prop ?

## Material Design specs
1. Don't crop Tooltip on the left side
left

2. Don't crop Tooltip on the right side
right

3. Don't crop Tooltip on the bottom
bottom

4. Landscape mode
landscape

### Test plan
1. Download this branch
2. Run example app
3. Open `Appbar` or `Toggle Button` section and hold press on any icon with Tooltip.

## References
* [Material Design Tooltip specs](https://material.io/design/components/tooltips.html)
* Material-UI - [Tooltip API](https://material-ui.com/api/tooltip/)

## Screenshots
tooltip1

tooltip1


src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+  componentWillUnmount() {
+    Dimensions.removeEventListener(
+      'change',
+      this._getChildrenPosition.bind(this)
+    );
+    this._clearTimeouts();
+  }
+
+  _getChildrenPosition() {
```
talyssonoc: Since it doesn't just _return_ the children position but actually changes it, maybe it'd be better if the name of this method was `setChildrenPosition`

talyssonoc: Also, you can create it as an arrow function so you don't have to bind it :smile:

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+        tooltipOpacity: 0.9,
+        tooltipMeasured: true,
+      });
+  };
+
+  _showTooltip = () => this.setState({ tooltipVisible: true });
+
+  _hideTooltip = () => this.setState({ tooltipVisible: false });
+
+  _clearTimeouts = () => clearTimeout(this._longPressTimeout);
```
talyssonoc: Looks like you're not passing it standalone so you don't need to define it as an arrow function.

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+          height: layout.height,
+        },
+        tooltipOpacity: 0.9,
+        tooltipMeasured: true,
+      });
+  };
+
+  _showTooltip = () => this.setState({ tooltipVisible: true });
+
+  _hideTooltip = () => this.setState({ tooltipVisible: false });
```
talyssonoc: Same thing here about not needing to be an arrow function, and for `_showTooltip` as well.

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+    const {
+      childrenLayout: { x, width },
+      tooltipLayout: { width: tooltipWidth },
+    } = this.state;
+
+    const centerDistanceFromChildren = x + (width - tooltipWidth) / 2;
+    const { width: layoutWidth } = Dimensions.get('window');
+
+    // Does it overflow to the right? If so, subtracts tooltipWidth
+    if (centerDistanceFromChildren + tooltipWidth > layoutWidth) {
```
talyssonoc: What do you think about creating a method with the name of this logic instead of adding a comment?

Something  like:

```js
if(this.overflowRight(centerDistanceFromChildren)) {

}
```

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+      return x + width - tooltipWidth;
+    }
+    // Does it overflow to the left? If so, starts from children start position
+    else if (centerDistanceFromChildren < 0) {
```
talyssonoc:
```suggestion
    else if (this.overflowLeft(centerDistanceFromChildren)) {
```

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+          onTouchStart={this._handleTouchStart}
+          onTouchEndCapture={this._handleTouchEndCapture}
+          onTouchCancel={this._handleTouchCancel}
+        >
+          {React.cloneElement(childElement, {
+            onLongPress: () => {}, // Prevent touchable to trigger onPress after onLongPress
+            ref: (el: React.ReactNode) => {
+              this._children = el;
+            },
+            ...rest,
```
talyssonoc: Do you intend to allow the user to override `onLongPress` and `ref`? If no, you should do the spread _before_ `onLongPress` and `ref`.

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+          )}
+        </Portal>
+        <View
+          onTouchStart={this._handleTouchStart}
+          onTouchEndCapture={this._handleTouchEndCapture}
+          onTouchCancel={this._handleTouchCancel}
+        >
+          {React.cloneElement(childElement, {
+            onLongPress: () => {}, // Prevent touchable to trigger onPress after onLongPress
+            ref: (el: React.ReactNode) => {
```
talyssonoc: What do you think about creating a method as an arrow function instead of doing it directly in render?

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,279 @@
...
+                styles.tooltip,
+                {
+                  opacity: tooltipOpacity,
+                  left: this._getTooltipXPosition(),
+                  top: this._getTooltipYPosition(),
+                },
+              ]}
+              onLayout={this._handleTooltipLayout}
+            >
+              <Text style={{ color: Colors.white }}>{title}</Text>
```
talyssonoc: Should this be custom styled as well?

src/components/Tooltip.tsx
brunohkbx: > Also, you can create it as an arrow function so you don't have to bind it

I was trying to figure out why I had to use bind here. Good to know about that btw.
Thanks!

src/components/Tooltip.tsx
brunohkbx: Sounds good!

src/components/Tooltip.tsx
brunohkbx: I think only `onLongPress` could be overridden by some component which has its own implementation such a Button.
What do you think @Trancever?

src/components/Tooltip.tsx
brunohkbx: I don't know.
Should we get this value from theme @Trancever?

satya164:
Can you rebase this against master? This PR has a lot of unrelated changes.

Easiest way will be to:
- Create a backup branch from this branch: `git checkout -b brunohkbx-feat/tooltip-backup`
- Delete the branch `brunohkbx-feat/tooltip`: `git branch -D brunohkbx-feat/tooltip`
- Create branch `brunohkbx-feat/tooltip` from master: `git checkout master && git checkout -b brunohkbx-feat/tooltip`
- Cherry-pick all the commits you have to this branch
- Force push

brunohkbx:
Done @satya164 :+1:

PR #1193, opened 2019-07-08
fix: add missing types in example project

wojteg1337:
### Motivation
When installing dependencies bob throws error:
Errors found when building definition files.

### Test plan
Install dependencies and generate def files with no errors.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -32,7 +32,7 @@
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.3.3",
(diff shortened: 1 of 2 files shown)


example/package.json
```diff
@@ -32,7 +32,7 @@
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.3.3",
-    "@types/react-navigation": "^2.13.10",
+    "@types/react-navigation": "^3.0.7",
```
Trancever: types v3.0.7 for package v2.18.0. Are those types correct?

satya164:
Are you sure? CI passes on master

PR #1192, opened 2019-07-08
feat: pass fab as children to FABGroup

brunohkbx:
### Motivation
Resolve #1145

Issue #1145, opened 2019-06-21
[Feature] Pass FABs as children to FAB.Group


hamidfzm:
## Describe the feature
Currently `FAB.Group` uses `actions` property to pass FAB buttons. Like this:

```
<FAB.Group
    open={this.state.open}
    icon={this.state.open ? "keyboard_arrow_down" : "add"}
    actions={[
    { icon: 'add', onPress: () => console.log('Pressed add') },
    { icon: 'star', label: 'Star', onPress: () => console.log('Pressed star')},
    { icon: 'email', label: 'Email', onPress: () => console.log('Pressed email') },
    { icon: 'notifications', label: 'Remind', onPress: () => console.log('Pressed notifications') },
    ]}
    onStateChange={this.handleStateChange}
/>
```

We can change this to:
```
<FAB.Group
    open={this.state.open}
    icon={this.state.open ? "keyboard_arrow_down" : "add"}
    onStateChange={this.handleStateChange}
>
  <FAB
    style={styles.fab}
    small
    icon="add"
    onPress={() => console.log('Pressed')}
  />
  <FAB
    style={styles.fab}
    small
    icon="add"
    onPress={() => console.log('Pressed')}
  />
</FAB.Group>
```
## Motivation
- Beautiful Code OFC
- More control over rendering fab elements without creating new arrays each time (using conditionals and PureComponents and ...)

satya164:
Yea, it seems reasonable

> More control over rendering fab elements without creating new arrays each time (using conditionals and PureComponents and ...)

I think it won't make a difference in perf. `FAB.Group` doesn't do anything expensive.

joeframbach:
I worked around this by writing this small component.

```
import React from 'react';
import {FAB} from 'react-native-paper';

type FabGroupProps = Omit<React.ComponentProps<typeof FAB.Group>, 'actions'>;
type Props = FabGroupProps & {
  children: React.ReactElement | React.ReactElement[];
};

export function FabGroup({children, ...fabGroupProps}: Props) {
  return (
    <FAB.Group
      actions={React.Children.map(children, child => child.props)}
      {...(fabGroupProps as FabGroupProps)}
    />
  );
}
```

PR #1194, opened 2019-07-08
feat: add labelStyle prop to FAB.Groups's item

ElChurros:
### Motivation

I'd like to be able to pass a custom style to the item's label in a FAB.Group.

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -25,6 +25,7 @@ type Props = {
    * - `accessibilityLabel`: accessibility label for the action, uses label by default if specified
    * - `color`: custom icon color of the action item
    * - `style`: pass additional styles for the fab item, for example, `backgroundColor`
+   * - `labelStyle`: pass additional styles for the fab item's label, for example `backgroundColor`
    * - `onPress`: callback that is called when `FAB` is pressed (required)
    */
   actions: Array<{
@@ -33,6 +34,7 @@ type Props = {
     color?: string;
     accessibilityLabel?: string;
     style?: StyleProp<ViewStyle>;
+    labelStyle?: StyleProp<ViewStyle>;
     onPress: () => void;
   }>;
   /**
@@ -262,6 +264,7 @@ class FABGroup extends React.Component<Props, State> {
                           transform: [{ scale: scales[i] }],
                           opacity: opacities[i],
                         },
+                        it.labelStyle,
                       ] as StyleProp<ViewStyle>
                     }
                     onPress={() => {
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -25,7 +25,7 @@ type Props = {
    * - `accessibilityLabel`: accessibility label for the action, uses label by default if specified
    * - `color`: custom icon color of the action item
    * - `style`: pass additional styles for the fab item, for example, `backgroundColor`
-   * - `labelStyle`: pass additional styles for the fab item's label, for example `backgroundColor`
+   * - `labelViewStyle`: pass additional styles for the view containing the fab item's label, for example `backgroundColor`
    * - `onPress`: callback that is called when `FAB` is pressed (required)
    */
   actions: Array<{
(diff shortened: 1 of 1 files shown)


src/components/FAB/FABGroup.tsx
```diff
@@ -262,6 +264,7 @@ class FABGroup extends React.Component<Props, State> {
                           transform: [{ scale: scales[i] }],
                           opacity: opacities[i],
                         },
+                        it.labelStyle,
```
Trancever: This is not a label. Label is rendered on line 283 as a `Text` component and labelStyle should be applied there.

ElChurros: I guess I was not clear enough. My goal is to style the view containing the label. I'll make the modifications and ping you.

src/components/FAB/FABGroup.tsx
```diff
@@ -33,6 +34,7 @@ type Props = {
     color?: string;
     accessibilityLabel?: string;
     style?: StyleProp<ViewStyle>;
+    labelViewStyle?: StyleProp<ViewStyle>;
```
Trancever: Hmm, Can we rename that prop to `style` and  pass it to `Card` + rename current `style` to `labelStyle` and pass it to the Text on line 283? When this is done, we can merge this PR :)

ElChurros: I don't think this will work since the current `style` prop is a `StyleProp<ViewStyle>` and not a `StyleProp<TextStyle>`. Hence it cannot be applied to the Text on line 283.
My proposal would be the following:
 - Add a `StyleProp<TextStyle>` called `labelStyle` passed to Text on line 283
 - Leave the current `style` prop as it is and keep applying it to the FAB.
 - rename the `labelViewStyle` I created to `cardStyle` since it actually applies to the card containing the label.  + For clarity sake, mention in the doc that it style the card containing the label.

Trancever: Sorry, I wasn't clear. We are ok with breaking current behaviour, cause we are about to release new 3.0 version. Let's change `style` to be `StyleProp<ViewStyle>` and apply it to `Card` and also apply `labelStyle` of type `StyleProp<TextStyle>` to `Text`.

ElChurros: I don't think this would be good as this removes the option of styling the FAB itself. I believe there should be 3 custom styles that one can pass to the action of a `FAB.Group`:
 - Custom style for the `FAB`itself which should be `style` in my opinion, as a `FAB.Group`'s item is first of all a `FAB`
 - Custom style for the label text which could be named `labelStyle`
 - Custom style for the card containing the label, for which I can't find a better name than `cardStyle`

Your proposal would actually remove the option of styling the main item of the `FAB.Group`: the `FAB`.
Please tell me if I missed something @Trancever !

johnwils:
I'll be using this library in the future and require a custom background color for each text label. It sounds like this PR is close to providing this capability. If I can help speed this along, please let me know, happy to contribute if needed.

jaysbytes:
Hi everyone. In 3.0 (master) version, we are going to remove actions prop and add children prop to specify FABs. See related issue #1145 and PR #1192. This PR if still needed can me merge to `2.0`. @ElChurros change a branch is this change is still needed.
In 3.0 we will need to take a different approach and add props to FAB.Group I think

Trancever:
@johnwils Would you like to try to finish this PR? What we want is: 
- `fabStyle` prop that is passed to main fab button
- `style` prop for every action that is passed to `Card`
- `labelStyle` prop for every action that is passed to the action's `Text`

johnwils:
@Trancever when you say:

`fabStyle` prop that is passed to the man fab button

Are you referring to adding a `fabStyle` prop to the `<TouchableRipple>`?

The other two I understand.

Also I'm having trouble with the cloned github repository. Running `yarn install` or `npm install` gives me this error:

```
Error: Failed to build definition files.
```
Not sure what I am doing wrong. But if I can test to make sure it is al working, I'll create a PR.

Trancever:
I am closing this due to the inactivity. If it is still relevant, feel free to open a new PR.

PR #1195, opened 2019-07-09
feat: optional key to FAB:Group's actions if the actions order changes

ElChurros:
### Motivation

I'd like to be able to dynamically choose the actions present in the FAB.Group, based on the context of the surrounding application. When I do so, the first batch of actions is rendered properly. However, if the context of the application around changes and the actions provided to the FAB.Group change accordingly, the label are correctly updated, but not the icons. 

In a FAB.Group, the index of the action in the array is currently used for the React key. That is perfectly fine as long as the actions in the FAB.Group remain constant.
In situations where the actions depend on the context of the application around, you can end up with label-icon mismatches in the FAB.Group.
My proposal is to let the user provide a custom key for the actions in the FAB.Group in case the actions are not constant throughout the application lifetime.

### Test plan

While tested and validated in the example app and the app I am working on, I have not pushed any modification to the example app, as this use case is hard to showcase in an example app.

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -26,6 +26,7 @@ type Props = {
    * - `color`: custom icon color of the action item
    * - `style`: pass additional styles for the fab item, for example, `backgroundColor`
    * - `onPress`: callback that is called when `FAB` is pressed (required)
+   * - `key`: key used by React in case the actions order is not constant
    */
   actions: Array<{
     icon: IconSource;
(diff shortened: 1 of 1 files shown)


satya164:
> In situations where the actions depend on the context of the application around, you can end up with label-icon mismatches in the FAB.Group.

This doesn't sound right. Why would the icon-label mismatch? Can you post a repro of what's happening?

ElChurros:
@satya164 Here is the repro:
https://github.com/ElChurros/FABGroupKeyIssueRepro

Trancever:
We are closing this because we are not able to reproduce it.

PR #1199, opened 2019-07-09
fix(list.item): properly render prop description

### LINKED ISSUE #1160: List.Item: description is not a function (regarding #996)
## Intro
Recently, I updated my [app](https://github.com/HenryQuan/WoWs-Info-Origin) and there were massive amount of crashes due to pull 996. It is mainly my fault but I still want to report it here.

~~~
"react-native": "^0.59.9",
"react-native-paper": "^2.16.0"
~~~

## About #996 
~~~javascript
inside a list...
<List.Item title={item.nickname} onPress={() => this.pushToPlayer(item)} 
    description={item.account_id} key={String(item.account_id)}
    right={() => <IconButton color={Colors.grey500} icon='close' 
    onPress={() => this.removeFriend(item)}/> }/>}
~~~
This was the code I used before. `item.account_id` was an integer but nevertheless, it was fine because description was simply passed to Text like `<Text>{description}</Text>`. 
Currently, it is checking if [description is a string](https://github.com/callstack/react-native-paper/blob/12f3cf109683d365e7c672d7df626b5900ee1e02/src/components/List/ListItem.tsx#L119). I think it might be better to check if description is a `function` instead of a string because people like me might pass an integer in and break everything. However, it is still mainly my fault because I didn't pass a string to it and I haven't seen other having the same issue.

brunohkbx:
### Motivation
Resolve #1160

### DIFF
--- a/src/components/List/ListItem.tsx
+++ b/src/components/List/ListItem.tsx
@@ -116,7 +116,13 @@ class ListItem extends React.Component<Props> {
   ) {
     const { descriptionEllipsizeMode, descriptionStyle } = this.props;
 
-    return typeof description === 'string' ? (
+    return typeof description === 'function' ? (
+      description({
+        ellipsizeMode: descriptionEllipsizeMode,
+        color: descriptionColor,
+        fontSize: styles.description.fontSize,
+      })
+    ) : (
       <Text
         numberOfLines={2}
         ellipsizeMode={descriptionEllipsizeMode}
@@ -128,14 +134,6 @@ class ListItem extends React.Component<Props> {
       >
         {description}
       </Text>
-    ) : (
-      description &&
-        typeof description === 'function' &&
-        description({
-          ellipsizeMode: descriptionEllipsizeMode,
-          color: descriptionColor,
-          fontSize: styles.description.fontSize,
-        })
     );
   }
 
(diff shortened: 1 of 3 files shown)


src/components/List/ListItem.tsx
```diff
@@ -116,27 +116,25 @@ class ListItem extends React.Component<Props> {
...
-        typeof description === 'function' &&
-        description({
+    return typeof description === 'function'
+      ? description({
           ellipsizeMode: descriptionEllipsizeMode,
           color: descriptionColor,
           fontSize: styles.description.fontSize,
         })
-    );
+      : description && (
```
Trancever: It's gonna crash when description is empty string. Can we check if description is falsy at the beginning of the function and return null in that case?

brunohkbx: Thanks, @Trancever.
Please check it again.

src/components/List/ListItem.tsx
```diff
@@ -116,26 +116,27 @@ class ListItem extends React.Component<Props> {
...
-          styles.description,
-          { color: descriptionColor },
-          descriptionStyle,
-        ]}
-      >
-        {description}
-      </Text>
-    ) : (
+    return (
       description &&
```
Trancever: It can still crash when description is empty string, because it will evaluate to false and empty string will be rendered not in `Text` component.

brunohkbx: Now I got it :sweat_smile: 
@Trancever Please check it again

Trancever:
@brunohkbx Thanks!

PR #1203, opened 2019-07-12
fix: exclude example folder from tsconfig.json

jaulz:
### Motivation

Currently, it's not feasible to install the library via `yarn add https://github.com/callstack/react-native-paper` because the following error occurs during installation:
```
ℹ Generating type definitions with tsc
✖ Errors found when building definition files.
example/src/index.native.tsx(1,25): error TS2307: Cannot find module 'expo'.
example/src/index.native.tsx(2,30): error TS2307: Cannot find module 'expo-keep-awake'.
example/src/index.native.tsx(17,39): error TS2307: Cannot find module 'react-navigation'.
example/src/RootNavigator.native.tsx(2,38): error TS2307: Cannot find module 'react-navigation'.
```

The reason is that no dependencies for the example folder exist during installation. You can easily reproduce this by removing `example/node_modules` and afterwards run `yarn prepare`.

### Test plan

Run `yarn prepare` without running `yarn bootstrap` before (i.e. make sure that no node_modules in example folder exist).


satya164:
We also want to typecheck the example folder. That's one of the main ways we ensure types are correct.

jaulz:
Wouldn't a separate tsconfig.json make sense in that case?

satya164:
> Wouldn't a separate tsconfig.json make sense in that case?

Doesn't it require running separate tsc command to check example separately?

jaulz:
@satya164 yep, I added it to the `typescript` command so it is still executed before commits but not during preparation phase.

satya164:
I don't really want to add a separate config and a separate tsc command. Seems moving the `@types/x` deps to root `package.json` will be simpler and do the job

jaulz:
It might do the job but in my opinion it is even more confusing if you install dependencies of the example project in the top level project. Those are two different projects each having its own `package.json` so it would also make sense if they each have its own `tsconfig.json`, wouldn't it?

satya164:
We already do it for eslint, prettier etc. The root project contains all the required configs and development dependencies for them. It should be the same for typescript.

Duplicating configs is worse because then we have to keep both configs updated if I change something.

jaulz:
eslint and prettier are not runtime related so it's not really comparable I would say. However, I just found the `extends` property so we can keep the `tsconfig.json` at minimum in the example folder.

jaulz:
@satya164 what do you think? 😊

jaysbytes:
As @satya164  said, it is not needed and actually, it will introduce more maintenance work than improvement. Closing for now.

lrojas94:
It's a shame this didn't go forward. Installing from Git would be nice for features that can make it into the offciial/final release but while this happens I can use them in my projects. Can we evaluate this again?

imhashir:
Agreed with @lrojas94 
I wanted to use #1797 which is pending for a very long time, but now I can't use that fork because of this issue. 
Would be awesome if someone can approve that PR or this one.

Issue #1089, opened 2019-05-24
Menu position wrong when anchored to upper right corner view


HolySamosa:
## Environment
Paper v2.15.2

## Description
My project is presently using React Native Navigation and I would like to display a `Menu` when a button is clicked in the native top navigation bar / app bar. Since the top bar is native, it's button can't serve as an anchor for the `Menu`, so I planned to instead anchor to an invisible `View` immediately below the native toolbar button.

Unfortunately, when the `Menu` renders it does not appear below the anchor view in the top-right corner.  I'm assuming that the `Menu` _should_ display immediately below this anchor, so this would be a bug?  Is there a better way to accomplish the effect I am after?

Thanks!

## Reproducible Demo

I've created a snack which demonstrates the issue:
https://snack.expo.io/@holysamosa/c16159

## Screenshot

Here's an example:
(I've given the anchor view a red background for visibility)

![image](https://user-images.githubusercontent.com/1156546/58351295-60455100-7e36-11e9-86a0-4875b6d4c0e4.png)

iyadthayyil:
I don't know why this issue is caused.

As a quick workaround wrap the menu in another view with position styles.

```javascript
const styles = StyleSheet.create({
  container: {
    flex: 1,
    flexDirection: 'row',
    flexWrap: 'wrap',
  },
  row: {
    width: '100%',
    backgroundColor: 'green',
  },
  mainMenuAnchor: {
    width: 24,
    height: 5,
    backgroundColor: 'red',
  },
  menuWrapper: {
    alignSelf: 'flex-end',
    margin: 8,
  }
});
```

```javascript
<Provider>
        <View style={[styles.container]}>
          <View style={[styles.row]}>
          <View style={styles.menuWrapper}>
            <Menu
              visible={this.state.visible}
              onDismiss={this._closeMenu}
              anchor={
              <View style={[styles.mainMenuAnchor]} />
              }
            >
              <Menu.Item onPress={() => {}} title="Item 1" />
              <Menu.Item onPress={() => {}} title="Item 2" />
            </Menu>
            </View>
          </View>
        </View>
        <Button style={[styles.button]} onPress={this._openMenu}>Show menu</Button>
</Provider>
```

HolySamosa:
@iyadthayyil Thanks for the work around!

ferrannp:
Wdyt @Trancever ?

HolySamosa:
[...]

PR #1208, opened 2019-07-14
feat: add support to anchor menu on x,y coordinates

iyadthayyil:
### Motivation
This PR allows to pass x,y coordinates as a anchor prop for Menu component
This would be a solution for #1089 too.
Also this will add ability to open contextual menus

### Test plan
Run the example app to see contextual menu


src/components/Menu/Menu.tsx
```diff
@@ -165,6 +165,8 @@ class Menu extends React.Component<Props, State> {
   _anchor?: View | null = null;
   _menu?: View | null = null;
 
+  _isAnchorCoord = !React.isValidElement(this.props.anchor);
```
ferrannp: Can you call this any time you call `_measureAnchorLayout` ? Maybe just memoize the result. If I change the props, this will be broken.

iyadthayyil: Done!

PR #1210, opened 2019-07-15
feat: add ability to use own icon component

### LINKED ISSUE #926: use MaterialCommunityIcons instead of MaterialIcons

Trancever:
### Motivation

Fix #926 

### Test plan

Run example app on web and native.


src/core/settings.tsx
```diff
@@ -0,0 +1,22 @@
...
+    style,
+    pointerEvents,
+    accessibilityProps,
+  }: {
+    name: string;
+    color: string;
+    size: number;
+    style: StyleProp<ViewStyle>;
+    pointerEvents: 'none';
+    accessibilityProps: any;
```
Trancever: @satya164 Is there any way to type this correctly? There is a difference between web and native.

src/core/settings.tsx
```diff
@@ -0,0 +1,22 @@
...
+    color,
+    size,
+    style,
+    pointerEvents,
+    accessibilityProps,
+  }: {
+    name: string;
+    color: string;
+    size: number;
+    style: StyleProp<ViewStyle>;
```
Trancever: With this approach style will already have `scaleX` set based on direction. Do we want to expose direction prop and let user handle it on his own?

src/components/Icon.tsx
```diff
@@ -7,27 +7,34 @@ import {
...
 } from 'react-native';
+import { Consumer as SettingsConsumer } from '../core/settings';
 
-let MaterialIcons: any;
+let MaterialCommunityIcons: any;
 
 try {
   // Optionally require vector-icons
-  MaterialIcons = require('react-native-vector-icons/MaterialIcons').default;
+  MaterialCommunityIcons = require('react-native-vector-icons/MaterialCommunityIcons')
```
Trancever: Maybe we should completely drop support for default Icon set?

Trancever:
@satya164 Can you review this? This is the last thing we need to release 3.0 version.

cassioseffrin:
@Trancever How can I test it en alpha.3 ? How can I set the MaterialIcons instead of
MaterialCommunityIcons ?

That feature will be very appreciated. Mainly because in MaterialCommunityIcons the search icon is not nice, there's not good options for it. And pass the icon as argument is too much verbose.

Trancever:
@cassioseffrin Check this https://github.com/callstack/react-native-paper/wiki/Migration-guide-from-2.0-to-3.0

cassioseffrin:
@Trancever Thks,  I'll checkout!

Issue #1159, opened 2019-06-27
Wrong Dialog size


m-ochyra:
## Environment

React Native Paper version 2.16.0 and 3.0.0-alpha.2.

## Description

Dialog component in horizontal orientation is too wide and on the iPhone X part of the component is hidden under the notch:
<img width="866" alt="Zrzut ekranu 2019-06-27 o 19 28 23" src="https://user-images.githubusercontent.com/5583015/60287699-ced46d80-9912-11e9-828e-ed38aa369a47.png">
![Zrzut ekranu 2019-06-27 o 19 28 36](https://user-images.githubusercontent.com/5583015/60287706-d431b800-9912-11e9-908b-60af698fc8c8.png)

Original Android dialog:
![Zrzut ekranu 2019-06-27 o 19 29 02](https://user-images.githubusercontent.com/5583015/60287726-e27fd400-9912-11e9-82dc-44d6a2d95541.png)

## Reproducible Demo

Example Dialog code.

devdebajyotidas:
Any update on this ?

robclouth:
For now you can set maxWidth and alignSelf on the Dialog:
```
<Dialog style={{ maxWidth: 500, alignSelf: "center" }} />
```

sepsol:
> For now you can set maxWidth and alignSelf on the Dialog:
> 
> ```
> <Dialog style={{ maxWidth: 500, alignSelf: "center" }} />
> ```

This doesn't work anymore. Any new solutions?

federicogomezcactus:
> For now you can set maxWidth and alignSelf on the Dialog:
> 
> ```
> <Dialog style={{ maxWidth: 500, alignSelf: "center" }} />
> ```

This causes the dialog to occupy the entire screen's height if I don't set a maxHeight for the Dialog.Content

hakanayaz14159:
> For now you can set maxWidth and alignSelf on the Dialog:
> 
> ```
> <Dialog style={{ maxWidth: 500, alignSelf: "center" }} />
> ```


Small addition on this answer. It helps to solve height issue.
```javascript
<Dialog style={{ maxWidth: 500, alignSelf: "center" }}>
   {...otherThings}
   <Dialog.Actions style={{flexGrow:0}}>
      <Button>Ok</Button>
   </Dialog.Actions>
</Dialog>
```

JakubTiimo:
> > For now you can set maxWidth and alignSelf on the Dialog:
> > ```
[...]

PR #1214, opened 2019-07-15
fix: wrap modal children in SafeAreaView

lukewalczak:
### Motivation

Fixes: #1159 

### Test plan

default | horizontal left | horizontal right |
--- | --- | ---
Screenshot 2019-07-15 at 15 43 54 | Screenshot 2019-07-15 at 15 42 44 | Screenshot 2019-07-15 at 15 43 42

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -7,8 +7,8 @@ import {
   StyleSheet,
   TouchableWithoutFeedback,
   ViewStyle,
-  SafeAreaView,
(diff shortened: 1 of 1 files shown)


satya164:
Can you use `react-native-safe-area-view`? The built-in `SafeAreaView` doesn't have ideal behaviour when a screen is transitioning. For example, it'll keep changing height when you push a new screen which slides from bottom in React Navigation.

PR #1218, opened 2019-07-16
fix: expose ref in text input

jaulz:
### Motivation

Currently, it is not possible to pass a `ref` prop to the `TextInput`.

### Test plan

Use `TextInput` with `ref` prop.

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -395,7 +395,7 @@ class TextInput extends React.Component<TextInputProps, State> {
   }
 
   render() {
-    const { mode, padding, ...rest } = this.props;
+    const { mode, padding, ref: outerRef, ...rest } = this.props;
 
     return mode === 'outlined' ? (
       <TextInputOutlined
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -15,7 +15,7 @@ import { Theme } from '../../types';
 const BLUR_ANIMATION_DURATION = 180;
 const FOCUS_ANIMATION_DURATION = 150;
 
-export type TextInputProps = React.ComponentProps<typeof NativeTextInput> & {
+export type TextInputProps = React.ComponentPropsWithRef<typeof NativeTextInput> & {
   /**
    * Mode of the TextInput.
    * - `flat` - flat input with an underline.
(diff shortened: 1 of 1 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -395,7 +397,8 @@ class TextInput extends React.Component<TextInputProps, State> {
   }
 
   render() {
-    const { mode, padding, ...rest } = this.props;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { mode, padding, ref: outerRef, ...rest } = this.props;
```
ferrannp: Where do you use `outerRef` ?

jaulz: Nowhere but we need to remove it from `rest` since we cannot pass it on to the inner components.

jaulz: The `ref` will be passed to `TextInput` and React will take care of it so you can access the methods of it, e.g. https://github.com/callstack/react-native-paper/blob/29f3ffc139909d80ab0d2872f83de2e6b5fb2636/src/components/TextInput/TextInput.tsx#L388-L390

`_root` again will be set by these piece of codes:
https://github.com/callstack/react-native-paper/blob/29f3ffc139909d80ab0d2872f83de2e6b5fb2636/src/components/TextInput/TextInput.tsx#L407-L409

https://github.com/callstack/react-native-paper/blob/29f3ffc139909d80ab0d2872f83de2e6b5fb2636/src/components/TextInput/TextInput.tsx#L419-L421

I didn't want to break anything with the `innerRef` so that was the simplest solution I think.

Trancever:
@jaulz Can you describe what do you want to achieve? Why do you need ref to the React-Native's `TextInput`.

jaulz:
@Trancever at the moment you will see the following error in case you use TypeScript:
```
Property 'ref' does not exist on type 'IntrinsicAttributes & Pick<TextInputProps, "label" | "style" | "hitSlop" | "onLayout" | "pointerEvents" | "removeClippedSubviews" | "testID" | "nativeID" | "collapsable" | ... 99 more ... | "dense"> & { ...; } & { ...; }'.ts(2322)
```

You can of course still pass a `ref` property and it works but it will cause a warning. 

This PR will expose the `ref` property using `React.ComponentPropsWithRef` here: https://github.com/callstack/react-native-paper/pull/1218/files#diff-3a9942f56421e77a4b6910cb9cac8381R18

Since this means that `ref` is part of the `props` we need to explicitly exclude it here: https://github.com/callstack/react-native-paper/pull/1218/files#diff-3a9942f56421e77a4b6910cb9cac8381R401

Trancever:
Ahh so it's only to fix typescript right? I think we can accept that

jaulz:
Sorry for the confusion 😇

Issue #1086, opened 2019-05-23
Change dark theme default values following new Material guidelines


ferrannp:
See https://material.io/design/color/dark-theme.html

- Toolbar: add `primary` prop
- BottomNavigation: backgroundColor should use `surface` from theme

PR #1219, opened 2019-07-16
Feat: new dark theme colors

pan-pawel:
Resolving #1086 
### Motivation
There is [new guidline](https://material.io/design/color/dark-theme.html) for dark theme.
main list

In dark theme we should use surface color for appbars and bottom bars instead of primary color
appbar normal
I added a `primary` prop to appbar. If you pass true there appbar will still use primary color from the theme in dark mode. In light theme appbar still use primary as before
abbar primary
I also added change to bar style (time, battery etc. colors) to be consistent with appbar color
appbarcustomcolor
Bottom bars also should use surface color
buttom navigation
Snackbars should use onSurface color witch is black on light and white on dark theme
snackbar light
snackbar dark
Surfaces in dark mode should beside shadow use white overlay with correct transparency level to show elevation level
Those elevation transparency values are hardcoded in `overlay` function that is used to calculate color for surface
elevation
surfaces
searchbar
context menu
alert
data table
cards
buttons
I also changed colors in chip example to be more visible in dark mode
chips


src/components/Snackbar.tsx
```diff
@@ -233,11 +232,18 @@ class Snackbar extends React.Component<Props, State> {
...
               style,
             ] as StyleProp<ViewStyle>
           }
         >
-          <Text style={[styles.content, { marginRight: action ? 0 : 16 }]}>
+          <Text
+            style={[
+              styles.content,
+              { marginRight: action ? 0 : 16 },
+              { color: colors.surface },
```
wojteg1337: use one object to pass styles

example/src/Examples/AppbarExample.tsx
```diff
@@ -31,7 +33,10 @@ class AppbarExample extends React.Component<Props> {
 
     return {
       header: (
-        <Appbar.Header>
+        <Appbar.Header
+          primary={params.showTopPrimary}
+          style={params.showCustomColor && { backgroundColor: '#ffff00' }}
```
wojteg1337: we keep convention `foo ? style.a : null` in paper

src/components/Surface.tsx
```diff
@@ -56,19 +62,21 @@ class Surface extends React.Component<Props> {
...
+    const { dark: isDarkTheme, colors } = theme;
     return (
       <Animated.View
         {...rest}
         style={[
-          { backgroundColor: theme.colors.surface },
-          shadow(elevation),
+          { backgroundColor: colors.surface },
+          elevation && shadow(elevation),
+          isDarkTheme && {
```
wojteg1337: change it to `{ backgroundColor: isDarkTheme ? overlay(elevation) : colors.surface}` so you do not override same style

PR #1221, opened 2019-07-16
Add component group annotation for Typography

### LINKED ISSUE #757: put all  Typography components in a group in documentation
### Current behaviour

Typography components are in seperate headings in docs
### Expected behaviour

They should be grouped together

byCedric:
### Motivation

This PR fixes #757, the [full motivation and other approaches are listed here](https://github.com/callstack/react-native-paper/issues/757#issuecomment-511245429).

### Test plan

It's a docs change only, all other tests are passing. But just to be complete 😄 

Screenshot 2019-07-16 at 22 48 38

### DIFF
--- a/docs/component-docs.config.js
+++ b/docs/component-docs.config.js
@@ -62,9 +62,11 @@ function getPages() {
     })
     .reduce((acc, file) => {
       const content = fs.readFileSync(file).toString();
+      const groupMatch = /\/\/ @component-group (\w+)/gm.exec(content);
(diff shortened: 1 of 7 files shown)


byCedric:
Ah, I see I broke the alphabetical order. Give me a sec to fix that 😄

byCedric:
There we go! This should do it. The regex and group-lookup might be a bit sloppy, I'm not a master with regex. But at least it's working right 😉

ferrannp:
@satya164 wdyt ? Looks good to me.

PR #1230, opened 2019-07-19
fix: upgrade color missing changes

pan-pawel:
There was missing usage of old color version function "light"


jaulz:
Oops, sorry bu good catch! Actually, I thought I used find & replace all...

PR #1241, opened 2019-07-24
refactor: change outline textinput label behaviour

jbinda:
PR makes few things:
1. Refactor outline input label - moves components to separate folder and extract one of the component (LabelBackground)
2. Fix spacing around label in outline input when control has focus or value
3. Adjust label background when backgroundColor is passed in styles to the outline input
4. change input padding from 12 to 14 according to material guidelines [inputs](https://material.io/design/components/text-fields.html#spec)

image

### Test plan
**BEFORE**
Note: more white space on right side of the label and gray background outside the outline border

Screenshot 2019-07-24 at 15 57 41

Screenshot 2019-07-24 at 15 57 47

**AFTER**
Note: Equal margin of 4px on both sides of the minimized label, correct spacing and border cutoff view sized properly

image

image

### DIFF
--- a/src/components/TextInput/Label/LabelBackground.tsx
+++ b/src/components/TextInput/Label/LabelBackground.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react';
+import { StyleSheet } from 'react-native';
+
+import AnimatedText from '../../Typography/AnimatedText';
+
+import { LabelBackgroundProps } from '../types';
+
+const LabelBackground = ({
+  parentState,
+  labelProps: {
+    placeholderStyle,
+    topPosition,
+    hasActiveOutline,
+    label,
+    backgroundColor,
+  },
+  labelStyle,
+}: LabelBackgroundProps) =>
+  label ? (
+    <AnimatedText
+      style={[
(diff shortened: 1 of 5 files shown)


src/components/TextInput/Label/LabelBackground.tsx
```diff
@@ -0,0 +1,99 @@
...
+                ...labelStyle.transform,
+                {
+                  scaleY: parentState.labeled.interpolate({
+                    inputRange: [0, 1],
+                    outputRange: [0.2, 1],
+                  }),
+                },
+              ],
+            },
+            styles.outlinedLabel || {},
```
Trancever: Do we need `|| {}` here ?

Trancever: Can we move it just after labelStyle on line 56 and merge style objects on line 57 and 70 ?

jbinda: Yeah, it's not neccessary, thanks

jbinda: Fixed

PR #1285, opened 2019-08-17
Removed unnecessary parenthesis

mcrowder65:
### Motivation
When copying and pasting this code sample, I got an error saying that text needs to be wrapped in a <Text> component, but it was an extra `)` that I missed because I copy and pasted the solution.


Trancever:
@mcrowder65 This closing parenthesis is needed to close the one opened on line 42.

PR #1294, opened 2019-08-21
fix(ProgressBar): only start/stop animation if visible prop changes

jaulz:
### Motivation

Currently, the `ProgressBar` restarts in between if any other prop than `visible` changes. This fix only starts/stops the animation if the `visible` prop has change.d

### Test plan

Open `ProgressBar` component in example app.


Trancever:
Thanks @jaulz !

PR #1300, opened 2019-08-23
Feat/props to drawer items

sritharanpalani:
### Motivation
Update To the DrawerItem component Label, Icon, and TouchableRipple

- Add a style to Label

- RippleColor props to TouchableRipple 

- IconSize to Left Icon

### Test plan

Add Drawer.Item component with addtional props 

`<Drawer.Item
      label="First Item"
      labelStyle={{fontSize: 20}}
      iconSize={29}
      rippleColor={'transperant'}
 />`

### DIFF
--- a/example/src/DrawerItems.tsx
+++ b/example/src/DrawerItems.tsx
@@ -6,7 +6,6 @@ import {
   Switch,
   TouchableRipple,
   Text,
-  Colors,
   Theme,
 } from 'react-native-paper';
 
@@ -49,13 +48,14 @@ class DrawerItems extends React.Component<Props, State> {
             <Drawer.Item
               {...props}
               key={props.key}
-              theme={
-                props.key === 3
-                  ? { colors: { primary: Colors.tealA200 } }
-                  : undefined
-              }
               active={this.state.drawerItemIndex === index}
               onPress={() => this._setDrawerItem(index)}
+              iconSize={index === 4 ? 27 : 24}
+              labelStyle={{
+                fontWeight:
+                  this.state.drawerItemIndex === index ? 'bold' : '400',
+              }}
+              rippleColor="transparent"
             />
           ))}
         </Drawer.Section>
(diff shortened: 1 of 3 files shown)


jaysbytes:
Is this change required by the community? We should first discuss if this change is needed. Everyone wants to change something in the lib, which fits some specific need, but we need to discuss if it is good from the library perspective and follows material design guidielines

sritharanpalani:
@jaysbytes I have added an example for this change! 
these additional props have not affected the guidelines provided by Material by default (every additional props are optional ) 
Screen Shot 2019-09-09 at 12 54 08 PM

sritharanpalani:
@ferrannp  hey you need a any clarification from my side!

sritharanpalani:
Hello Team
We use react-native-paper extensively and would love to have this change merged.  This change is completely backward compatible. Please let me know if I can answer any questions or make any changes to accelerate this PR. Thanks so much.

Trancever:
@sritharanpalani We are having a hard time deciding if those additional props are needed. What exactly do you want to achieve with a `rippleColor` prop? In example I see you set it to 'transparent', do you want to remove the ripple effect completely?

sritharanpalani:
@Trancever The main purpose is to customize rippleColor / remove it. by having this additional prop we can achieve more design level aspects effects.

Thanks @Trancever for the effort

Trancever:
@sritharanpalani We think that the proposed changes allows for too much customizability. We want to keep the API of the components as easy and approachable as possible and proposed changes allows making the component not MD guidelines compliant e.g. no ripple effect at all on press.

What we suggest is to create your own `Drawer.Item` component that will be compliant with your design system.  You shouldn't be afraid of the code duplication and increased bundle size, because not used paper components are not included into the final bundle thanks to the babel plugin that we provide.

sritharanpalani:
thanks @Trancever

Issue #1302, opened 2019-08-23
Cannot find module 'react-native-paper/babel' when generating production build for android


devsrv:
**Without Expo** & Using - "react-native": "0.60.4", "react-native-paper": "^3.0.0-alpha.3",

_babel.config.js--_
```javascript
module.exports = {
  presets: ['module:metro-react-native-babel-preset'],
  env: {
    production: {
      plugins: ['react-native-paper/babel'],
    },
  },
};
```

when trying to generate APK on windows system getting this error. tried on other projects using RN paper "^3.0.0-alpha.3", got same error

![rn paper issus](https://user-images.githubusercontent.com/26733312/63611154-e40b0700-c5f8-11e9-9ae6-6000f82c53f9.png)

diegorodriguesvieira:
Any solution? :(

Trancever:
@satya164 can you check what's happening?

jaysbytes:
The issue might be related to the fact,  that after migration to typescript `.js` files aren't copied to `lib` folder.  `src/babel/index.js` is a `js` file

yonahforst:
I'm using master at commit 403bfd3 and I'm getting the following error:
`Cannot find module './lib/module/babel/index.js'`

zyxel648:
I'm manually copying `babel.js` to the project's `node_modules\react-native-paper` folder, just for a temporary solution.

jaysbytes:
@yonahforst It's propably becouse content of the package on npm is different  than the content of the repo. You propably don't have `lib` directory at all. Could you please check that?

yonahforst:
@jaysbytes yeah that makes sense. Indeed i dont have the `lib` dir. I'll wait for the next alpha version

jaysbytes:
Try to build paper inside node modules using `yarn prepare`. I'm not sure when we are planning to release

slavikdenis:
@jaysbytes unfortunately it did not worked.

```
denis@DESKTOP-GS9VKQF MINGW64 /f/repos/inqool/spu/app/node_modules/react-native-paper (master)
$ yarn prepare
yarn run v1.17.3
$ bob build && node ./scripts/generate-mappings.js
ℹ Building target module
ℹ Cleaning up previous build at lib\module
ℹ Compiling 102 files in src with babel
✓ Wrote files to lib\module
ℹ Building target typescript
ℹ Cleaning up previous build at lib\typescript
ℹ Generating type definitions with tsc
[...]

PR #1305, opened 2019-08-26
chore: fix babel plugin import

jaysbytes:
`babel.js` was removed from the `files` field in package.json. It caused fails in release builds (at least Android #1302 ).

### Test plan

it can be tested with the next alpha release.


jafri:
@satya164 can this get merged as it currently blocks production builds

PR #1314, opened 2019-08-28
feat: added keyboardAware prop to Modal and Dialog components

GollyJer:
This PR adds a keyboard awareness prop to `Modal` and `Dialog` components and a `KeyboardAwareDialog` to the example app. It also removes the `SafeAreaView` wrapper from `Modal`. Details below.

### Motivation
When adding a `TextInput` inside a `Modal` or `Dialog` components the input is not keyboard aware and often ends up underneath the keyboard. [Here is a snack example](https://snack.expo.io/@gollyjer/react-native-paper-dialog-with-textinput) of the current behavior.

It would be ideal if a user could simply wrap these themselves in a KeyboardAwareView, but the way `Modal` is structured it can't work.  Once a child of the `KeyboardAwareView`, or any view for that matter, is absolutely positioned it is out of flexbox flow.

From [w3c.org](https://www.w3.org/TR/css-flexbox-1/#abspos-items)
image

All of that is to say... the `SafeAreaView` (a fix for this should be separate from this PR) in `Modal` and wrapping `KeyboardAwareView` around `Dialog` both have no effect because the `Surface` component has to be wrapped in an absolutely positioned view.

### Test plan
I tested on both iOS and Android with real devices.
- Run the test app.
- Open Dialog > Show Keyboard Aware Dialog
- Tap on the input.
- Keyboard should come up and the view should shrink to fit.

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -3,12 +3,13 @@ import {
   Animated,
   BackHandler,
   Easing,
+  KeyboardAvoidingView,
   StyleProp,
   StyleSheet,
   TouchableWithoutFeedback,
   ViewStyle,
+  View,
 } from 'react-native';
-import SafeAreaView from 'react-native-safe-area-view';
 import Surface from './Surface';
 import { withTheme } from '../core/theming';
 import { Theme } from '../types';
@@ -26,6 +27,10 @@ type Props = {
    * Determines Whether the modal is visible.
    */
   visible: boolean;
+  /**
+   * Determines whether the view resizes when the keyboard is visible.
+   */
+  keyboardAware?: boolean;
   /**
    * Content of the `Modal`.
    */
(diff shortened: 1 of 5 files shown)


GollyJer:
Documentation looks good.

src/components/Modal.tsx
```diff
@@ -176,26 +182,42 @@ class Modal extends React.Component<Props, State> {
...
+        style={styles.container}
       >
         <TouchableWithoutFeedback
           onPress={dismissable ? this._hideModal : undefined}
         >
-          <Animated.View
+          <View
             style={[
               styles.backdrop,
-              { backgroundColor: colors.backdrop, opacity: this.state.opacity },
```
Trancever: Why did you remove the animation?

src/components/Modal.tsx
```diff
@@ -176,26 +182,42 @@ class Modal extends React.Component<Props, State> {
...
+              },
             ]}
           />
         </TouchableWithoutFeedback>
-        <SafeAreaView style={styles.wrapper}>
+        <SurfaceWrapper
+          style={styles.wrapper}
+          pointerEvents={'box-none'}
+          behavior="padding"
+        >
```
Trancever: Can you post a screenshot how the modal looks on iphone with notch?

Also there is no way to change KeyboardAvoidingView props. We need a better API to handle this case also.

PR #1316, opened 2019-08-29
refactor: use textTransform to upperCase Button

doomsower:
### Motivation

When testing with `@testing-library/react-native` I want to find buttons by text. However, `Button` component uses `toUpperCase` explicitly, so I have to upprecase all my search strings. This is annoying, especially since I use react-i18next. With this change, I can use the same string in my tests (you can see this from snapshots)

### Test plan

I tested it in example app visually on iOS and Android simulators and web version. Everything looks properly uppercased.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -268,15 +268,12 @@ class Button extends React.Component<Props, State> {
               style={[
                 styles.label,
                 compact && styles.compactLabel,
+                uppercase && styles.uppercaseLabel,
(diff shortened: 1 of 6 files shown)


jaysbytes:
I'm totally for that, but maybe there was a reason to use uppercase. Any ideas @satya164 @Trancever ?

satya164:
`textTransform` wasn't available in React Native before. If it works on Platforms now, we can use it.

PR #1319, opened 2019-08-31
fix: add elevation to wrapper style

### LINKED ISSUE #1225: Elevation on the AppBar.Header
## Describe the feature

It would be nice if we could set elevation on the AppBar Header component.

## Motivation

Just to add depth ¯\_(ツ)_/¯

pan-pawel:
Resolve #1225 
### Motivation

Without property elevation in wrapper style there is no shadow on android.
Screenshot 2019-08-31 at 17 58 37
Screenshot 2019-08-31 at 17 54 49

With elevation added to wrapper style
```
<Wrapper
        style={
          [
            { backgroundColor, zIndex, elevation },
            shadow(elevation),
          ] as StyleProp<ViewStyle>
        }
      >
```

Screenshot 2019-08-31 at 17 52 09
Screenshot 2019-08-31 at 17 52 52

Additional
since we have default elevation value 
```
const {
  [..]
  elevation = 4,
  [..]
} = StyleSheet.flatten(style) || {};
```
we don't need to check if it's present
```
elevation &&  shadow(elevation),
```
so
```
shadow(elevation),
```
isn't enough?

### DIFF
--- a/src/components/Appbar/AppbarHeader.tsx
+++ b/src/components/Appbar/AppbarHeader.tsx
@@ -129,8 +129,8 @@ class AppbarHeader extends React.Component<Props> {
       <Wrapper
         style={
           [
-            { backgroundColor, zIndex },
-            elevation && shadow(elevation),
+            { backgroundColor, zIndex, elevation },
+            shadow(elevation),
           ] as StyleProp<ViewStyle>
         }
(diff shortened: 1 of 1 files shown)


franziz:
Great!

PR #1323, opened 2019-09-02
fix: tests and package.json scripts so they run cross-platform

GollyJer:
This PR does 3 things.

1. Adds a Windows style `fixtures` for the babel plugin tests.
2. Adds an OS based `lint` command in `package.json`.
This kind of sucks. I didn't want to add another package to the repo but I had to use [run-script-os](https://www.npmjs.com/package/run-script-os) to apply platform specific commands.
I was hoping to find a simple cross platform command. You can see [my question here on stackoverflow](https://stackoverflow.com/questions/57750884/cross-platform-eslint-command-to-run-on-current-directory-recursively). (please upvote to get it more traction).

3. Adds a `.prettierignore` file to simplify the `code.js` and `output.js` test files.
Basically, now `.eslintignore` and `.prettierignore` are the same. Generally if you want to ignore with one you want to ignore with the other. In addition I added a pattern to both to ignore the `output.js` files in tests. 
Now `/* eslint-disable prettier/prettier */` isn't needed at the top of every file.

### Motivation
I was unable to commit anything when using a Windows machine do to cross-platform issues.

### Test plan

I have tested on `Windows` and `Ubuntu`.  
**Need someone to test on `macOS`**.

- Create a branch.
- Edit a file and try and commit.
- If the tests all run then we're good to go! 👍

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,4 @@
-# NOTE: Should match .prettierignore
+# NOTE: Should match .eslintignore
 
 node_modules/
 flow-typed/
--- a/package.json
+++ b/package.json
@@ -31,7 +31,9 @@
   ],
   "scripts": {
     "typescript": "tsc --noEmit",
-    "lint": "eslint --ext '.js,.ts,.tsx' .",
+    "lint": "run-script-os",
+    "lint:win32": "eslint **/*.js **/*.ts **/*.tsx",
+    "lint:default": "eslint . --ext .js,.ts,.tsx",
     "test": "jest",
     "prepare": "bob build && node ./scripts/generate-mappings.js",
     "release": "release-it",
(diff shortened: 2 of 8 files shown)


Trancever:
@GollyJer We are ok with merging the fixtures in tests but we are not sure about the changes regarding eslint. Isn't eslint supposed to work cross-platform by default? Maybe we can tweak parameters we pass to make it working on both systems?

GollyJer:
@Trancever Can you review this pull request instead?
https://github.com/callstack/react-native-paper/pull/1477

I was able to find a linting command that works cross-platform without the need for a new npm package.
Also, I removed the eslint stuff as it was just a convenience and not important for this PR.

Please close this request after you've read this. 😃

PR #1356, opened 2019-09-28
Feat: dark theme elevation overlay

pan-pawel:
1. Added possibility for the user to set a custom surface colour on dark theme mode.
2. Added **primary**  prop to bottomNavigation (same as for appbar) It forces component to use primary colour instead of surface one (in new dark theme guideline huge surfaces like appbar or bottom navigation in opposite to light theme use surface instead of primary colour)
3. Added UseSurfaceOverlay boolean property to DarkTheme (true by default) 
In dark theme we use shadow & white overlay with % opacity based on elevation to display elevation differences. (the bigger elevation the lighter surface).
if user set UseSurfaceOverlay to false paper will use only shadow (same as in light theme)

### DIFF
--- a/src/styles/DarkTheme.tsx
+++ b/src/styles/DarkTheme.tsx
@@ -6,6 +6,7 @@ import { Theme } from '../types';
 const DarkTheme: Theme = {
   ...DefaultTheme,
   dark: true,
+  useSurfaceOverlay: true,
   colors: {
     ...DefaultTheme.colors,
     primary: '#BB86FC',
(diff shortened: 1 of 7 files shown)


src/styles/DarkTheme.tsx
```diff
@@ -6,6 +6,7 @@ import { Theme } from '../types';
 const DarkTheme: Theme = {
   ...DefaultTheme,
   dark: true,
+  useSurfaceOverlay: true,
```
Trancever: Can you update this to the format which Satya proposed?

ferrannp:
@pan-pawel working in different PR for this.

PR #1359, opened 2019-10-02
Fix RadioButton onPress

### LINKED ISSUE #1355: OnPress event on radio button does not work
## Environment

react-native  0.60.4
react-native-paper 2.16.0

## Description

When I pass the function to the RadioButton prop OnPress, the event doesn't fire no matter how many times I press the button.

## Reproducible Demo

The example itself passed by the documentation is enough to reproduce the error.

```javascript
import { RadioButton } from 'react-native-paper';

  state = {
    checked: 'first',
  };

  render() {
    const { checked } = this.state;

    return (
      <View>
        <RadioButton
          value="first"
          status={checked === 'first' ? 'checked' : 'unchecked'}
          onPress={() => { this.setState({ checked: 'first' }); }}
        />
        <RadioButton
          value="second"
          status={checked === 'second' ? 'checked' : 'unchecked'}
          onPress={() => { this.setState({ checked: 'second' }); }}
        />
      </View>
    );
  }
}
```

Taym95:
Fix #1355

### DIFF
--- a/src/components/RadioButton.tsx
+++ b/src/components/RadioButton.tsx
@@ -112,7 +112,7 @@ class RadioButton extends React.Component<Props> {
 
   handlePress = (context: RadioButtonContextType) => {
     const { onPress } = this.props;
-    const { onValueChange } = context;
+    const onValueChange = context ? context.onValueChange : () => {};
 
     onPress ? onPress() : onValueChange(this.props.value);
   };
@@ -132,7 +132,7 @@ class RadioButton extends React.Component<Props> {
           <Button
             {...this.props}
             status={this.props.status || (context && this.isChecked(context))}
-            onPress={() => context && this.handlePress(context)}
+            onPress={() => this.handlePress(context)}
           />
         )}
       </RadioButtonContext.Consumer>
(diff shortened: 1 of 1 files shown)


src/components/RadioButton.tsx
```diff
@@ -114,7 +114,7 @@ class RadioButton extends React.Component<Props> {
     const { onPress } = this.props;
     const { onValueChange } = context;
 
-    onPress || onValueChange(this.props.value);
+    onPress ? onPress() : onValueChange(this.props.value);
```
Trancever: Please change line 135 to this: 
```
            onPress={
              this.props.onPress ||
              (() => {
                context && this.handlePress(context);
              })
            }
```

This change is also needed to make it fully working

Taym95: Yes sure I will, I have a question what if context is null this `const { onValueChange } = context || {};` will not throw an error?

Trancever: It will, but context will be undefined when RadioButton is not wrapped with RadioButton.Group. You can double check that but I am pretty sure that's how context should work.

Taym95: This one `const { onValueChange } = context || {};` will trigger this typescript error `Initializer provides no value for this binding element and the binding element has no default value.` because Typescript notices that `onValueChange ` is not present in `{}`.

Trancever: It's fine, let's do this

PR #1361, opened 2019-10-03
fix: omit ref in render method of TextInput

jaulz:
### Motivation

Currently, the following warning is printed in console when using a `TextInput` on web:
````
TextInput: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.
````

This is because the types say there is a `ref` prop but in fact the `ref` prop cannot be accessed at runtime.

### Test plan

Open `TextInput` page in example project for web.


redbluenat:
thanks for solving this issue :) when do you plan to release it :) ?

Annihil:
Nice, when is 3.0.0-alpha.6 going to be released? 🙂

redbluenat:
any news?

PR #1370, opened 2019-10-08
Ability to add icon to the right side of text in button

raajnadar:
### Motivation

Fix https://github.com/callstack/react-native-paper/issues/452

### Test plan

Check the example app -> button section.

### DIFF
--- a/example/src/Examples/ButtonExample.tsx
+++ b/example/src/Examples/ButtonExample.tsx
@@ -35,7 +35,10 @@ class ButtonExample extends React.Component<Props, State> {
             <Button disabled onPress={() => {}} style={styles.button}>
               Disabled
             </Button>
-            <Button icon="camera" onPress={() => {}} style={styles.button}>
+            <Button iconStart="camera" onPress={() => {}} style={styles.button}>
+              Icon
+            </Button>
+            <Button iconEnd="camera" onPress={() => {}} style={styles.button}>
               Icon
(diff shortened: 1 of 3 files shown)


lukewalczak:
Hello @raajnadar! Actually Material Design isn't directly saying that you can't place the icon after the label, but on the all presented examples icon is always before the label.

That PR also creates a potential risk of adding two icons at the same time which is against Material Design.

image

raajnadar:
Maybe the right side icon is not as per the material guidelines..! What if we need a text as well as icon for pagination button?? Right arrow before the text will not look good. If we already have a solution for the pagination issue please let me know.

After creating the PR I found some flaws, but I was waiting for the code review.

A prop for the position of the icon will be good I guess!!

`iconAlign: "right | left"`

lukewalczak:
Honestly speaking, I think in the case of pagination, one of two indicators is enough - a label or an icon. I understand that you can have a design where label along with an icon is a requirement and currently you have a problem, however It's kinda against MD if there is no example for that.

Wdyt @satya164 @Trancever ?

dimbslmh:
image

kopax:
Hello, I am having the same issue, it would be good to allow that, it seems like it's permited with MD guidlines

angelxmoreno:
So why was this not merge?

paour:
Any news on this important PR?

PR #1379, opened 2019-10-10
fix: make ListIcon color optional

jaulz:
### Motivation

If I want to use images instead of icons in the `ListIcon` a color prop causes the image to have a tint (i.e. all the non-alpha pixels will be converted to the tint color) and thus it appears completely black on web. However, excluding the color would solve the issue:
```
                      left={props => {
                        const { color, ...rest } = props

                        return <List.Icon {...rest} icon={{
                          uri: providerType.icon || ''
                        }} />
                      }}
```

Currently, the color prop is required but this PR makes it optional.

### Test plan

This is just a prop type change.

### DIFF
--- a/src/components/Icon.tsx
+++ b/src/components/Icon.tsx
@@ -7,22 +7,28 @@ import {
 } from 'react-native';
 import { Consumer as SettingsConsumer } from '../core/settings';
 import { accessibilityProps } from './MaterialCommunityIcon';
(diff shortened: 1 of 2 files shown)


Trancever:
@jaulz Can you make CI pass? There is some issue with typescript types.

jaulz:
@Trancever sure, here you go.

PR #1399, opened 2019-10-20
docs: add pandadeals app in the showcase

miazga:
Simple change extending the showcase with an app called PandaDeals available on both mobile platforms.


MattKoboski:
Screenshot:
image

Issue #1404, opened 2019-10-21
[web] Focused TextInput label has large left padding


jaulz:
## Environment

react-native-paper `3.0.0-alpha.8`

## Description

The label of the focused `TextInput` has a large left padding on web:
![image](https://user-images.githubusercontent.com/5358638/67232018-bdbee780-f440-11e9-9e15-c21c149b7187.png)

Seems to happen since `3.0.0-alpha.6` and I assume it's related to: https://github.com/callstack/react-native-paper/commit/708fdf3463c676507f64286bbf790f02c5426cf2

Is this intentionally?

## Reproducible Demo

Just open the web example and check the focused `TextInput`.

PR #1405, opened 2019-10-22
fix: correct FlatTextInput padding on web

jaulz:
### Motivation

Fix https://github.com/callstack/react-native-paper/issues/1404

### Test plan

Open `FlatTextInput` example on web.


Trancever:
@jaulz Thanks for fixing this!

Issue #1346, opened 2019-09-23
Passing TextInput props dynamically hides text input field Value


gayathri-perumal:
## Description
The following object is my form data.

`const formData = [`
    `{`
      `label: 'Username',`
      `name: 'username',`
    `},`
    `{`
     ` label: 'Email',`
      `name: 'email',`
    `},`
    `{`
      `label: 'Password',`
      `name: 'password',`
   ` },`
  `];`

And I am passing this values into my text input like this,
`{`
      `formData.map( data => (`
            ` <TextInput`
                     `label={data.label}`
                      `placeholder={data.placeholder}`
                `/>`
           `)`
   `)`
`}`

In IOS it is working fine,
But In Android When enter value in username and moving to next text input, the username value suddenly goes hiddden.

## Reproducible Demo
It should show the entered value.

PR #1434, opened 2019-10-30
fix: fix text input disappearing value on blur when in uncontrolled mode

mmarkusik:
### Motivation

This pull request fixes the issue mentioned in [Passing TextInput props dynamically hides text input field Value #1346](https://github.com/callstack/react-native-paper/issues/1346). When TextInput is in uncontrolled mode it's text value disappears when blurred. 

### Test plan

- Go to `TextInputExample` component
- Remove `value` and `onChangeText` props from any text input
- Run the example 
- Type something to this input, then focus another one. The value of the blurred input should not disappear


Brendan-csel:
I gave this a test and it works for me - and solves my problem using TextInput uncontrolled.

Issue #1380, opened 2019-10-10
TextInput Label wrong position without focus and given value


Brainyoo:
## Environment

RN-Version: 0.61.2
React-Verison 16.9.0
RN-Paper-Version: 3.0.0-alpha.5

## Description

If I use the TextInput-Component and set the label with an existing value, the label gets padding on the left side. The longer the label, the longer the padding.
If I click in the input-field, the label corrects his position.
It happens on Android and iOS.

**(without focus)**
![Bildschirmfoto 2019-10-10 um 12 54 48](https://user-images.githubusercontent.com/54897979/66563057-3f6b6700-eb5d-11e9-927b-838b8aaf28f0.png)
![Bildschirmfoto 2019-10-10 um 12 55 01](https://user-images.githubusercontent.com/54897979/66563063-42665780-eb5d-11e9-8b01-f94aac9339e3.png)
![Bildschirmfoto 2019-10-10 um 12 55 07](https://user-images.githubusercontent.com/54897979/66563068-44c8b180-eb5d-11e9-9fd7-e41c00afd428.png)
![Bildschirmfoto 2019-10-10 um 12 55 24](https://user-images.githubusercontent.com/54897979/66563070-45614800-eb5d-11e9-8b31-3120fcbac60e.png)

**(with focus)**
![Bildschirmfoto 2019-10-10 um 12 56 49](https://user-images.githubusercontent.com/54897979/66563147-717cc900-eb5d-11e9-98bc-650a44c84cbf.png)

Brainyoo:
Im closing this issue because it seems to happen only with my Dialog. In other cases it works

Brainyoo:
Ok I found out how to cause this issue.
If you already defined the value for your TextInput, it causes the padding problem.

`const MainScreen = () => {`

   ` const [text, setText] = useState('bla');
    return <TextInput label={"translate('dialog.editWord')"} value={text} onChangeText={setText} padding={'none'} />;
};`

RichardLindhout:
Not true, I have the same bug. Its not your dialog. It does the padding on the flat input which should only be done on the outlined one.

hugopretorius914:
I too have this bug, and it's on both Android and iOS.

jaulz:
Please try `3.0.0` which was released today. Maybe it's related to https://github.com/callstack/react-native-paper/pull/1405

Shywim:
Same bug on 3.0.0 (not alpha).

[...]

Issue #1354, opened 2019-09-27
TextField label has margin initialValues


douglaszaltron:
## Environment

react-nativo > 0.61.1
react-native-paper `3.0.0-alpha.5`

## Description

In input text fields, when fields are filled in by redux or the label has a different margin than the text. after you change the value back to normal

## Reproducible Demo

![image](https://user-images.githubusercontent.com/4502670/65796245-7a73a080-e142-11e9-83f1-ca94a0f0520d.png)

mishojan:
PLEASE FIX THIS :((((((

satya164:
@mishojan please send a pull request

RichardLindhout:
Will try to fix this, this week probably Tuesday.

RichardLindhout:
Ok little bit later than Tuesday but I  found the cause :-)

RichardLindhout:
The code here is never ran when you have an initial value
```
https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput/TextInput.tsx#L210
```
There is an initial value available in the state so it's never changed there
https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput/TextInput.tsx#L191

RichardLindhout:
Pull request: https://github.com/callstack/react-native-paper/pull/1440

RichardLindhout:
FIx for this merged: https://github.com/callstack/react-native-paper/pull/1441

PR #1440, opened 2019-10-31
fix: minimize label on initial render with a value

RichardLindhout:
### Motivation

Fixes 
https://github.com/callstack/react-native-paper/issues/1380
https://github.com/callstack/react-native-paper/issues/1354

The code here is never ran when you have an initial value

https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput/TextInput.tsx#L210

There is an initial value available in the state so it's never changed there
https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput/TextInput.tsx#L191

### Test plan
Tested on Android, iOS and web

### Extra

We should maybe rewrite this using hooks, that should prevent these bugs from happening since we could use 'useEffect'

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -184,11 +184,20 @@ class TextInput extends React.Component<TextInputProps, State> {
   }
 
   state = {
-    labeled: new Animated.Value(this.props.value || this.props.error ? 0 : 1),
+    labeled: new Animated.Value(
+      (this.props.value !== undefined
+      ? this.props.value
+      : this.props.defaultValue)
+        ? 0
+        : 1
+    ),
     error: new Animated.Value(this.props.error ? 1 : 0),
     focused: false,
-    placeholder: this.props.error ? this.props.placeholder : '',
-    value: this.props.value || this.props.defaultValue,
+    placeholder: '',
+    value:
+      this.props.value !== undefined
+        ? this.props.value
+        : this.props.defaultValue,
     labelLayout: {
       measured: false,
       width: 0,
@@ -202,13 +211,14 @@ class TextInput extends React.Component<TextInputProps, State> {
     if (
       prevState.focused !== this.state.focused ||
       prevState.value !== this.state.value ||
-      prevProps.error !== this.props.error ||
-      this.props.defaultValue
+      // workaround for animated regression for react native > 0.61
+      // https://github.com/callstack/react-native-paper/pull/1440
+      prevState.labelLayout !== this.state.labelLayout
     ) {
       // The label should be minimized if the text input is focused, or has text
       // In minimized mode, the label moves up and becomes small
-      if (this.state.value || this.state.focused || this.props.error) {
-        this.minmizeLabel();
+      if (this.state.value || this.state.focused) {
+        this.minimizeLabel();
       } else {
         this.restoreLabel();
       }
@@ -216,13 +226,12 @@ class TextInput extends React.Component<TextInputProps, State> {
 
     if (
       prevState.focused !== this.state.focused ||
-      prevProps.label !== this.props.label ||
-      prevProps.error !== this.props.error
+      prevProps.label !== this.props.label
     ) {
-      // Show placeholder text only if the input is focused, or has error, or there's no label
+      // Show placeholder text only if the input is focused, or there's no label
       // We don't show placeholder if there's a label because the label acts as placeholder
       // When focused, the label moves up, so we can show a placeholder
-      if (this.state.focused || this.props.error || !this.props.label) {
+      if (this.state.focused || !this.props.label) {
         this.showPlaceholder();
       } else {
         this.hidePlaceholder();
@@ -305,7 +314,7 @@ class TextInput extends React.Component<TextInputProps, State> {
       }),
     }).start();
 
-  private minmizeLabel = () =>
+  private minimizeLabel = () =>
     Animated.timing(this.state.labeled, {
       toValue: 0,
       duration: BLUR_ANIMATION_DURATION,
--- a/src/components/TextInput/Label/InputLabel.tsx
+++ b/src/components/TextInput/Label/InputLabel.tsx
@@ -25,6 +25,7 @@ const InputLabel = (props: InputLabelProps) => {
     topPosition,
     paddingOffset,
     placeholderColor,
+    errorColor,
   } = props.labelProps;
 
   const labelTranslationX = {
@@ -123,7 +124,7 @@ const InputLabel = (props: InputLabelProps) => {
           labelStyle,
           paddingOffset,
           {
-            color: placeholderColor,
+            color: error && errorColor ? errorColor : placeholderColor,
             opacity: placeholderOpacity,
           },
         ]}
--- a/src/components/TextInput/TextInputFlat.tsx
+++ b/src/components/TextInput/TextInputFlat.tsx
@@ -88,7 +88,11 @@ class TextInputFlat extends React.Component<ChildTextInputProps, {}> {
       paddingHorizontal: number;
     };
 
-    let inputTextColor, activeColor, underlineColorCustom, placeholderColor;
+    let inputTextColor,
+      activeColor,
+      underlineColorCustom,
+      placeholderColor,
+      errorColor;
 
     if (disabled) {
       inputTextColor = activeColor = color(colors.text)
@@ -101,6 +105,7 @@ class TextInputFlat extends React.Component<ChildTextInputProps, {}> {
       inputTextColor = colors.text;
       activeColor = error ? colors.error : colors.primary;
       placeholderColor = colors.placeholder;
+      errorColor = colors.error;
       underlineColorCustom = underlineColor || colors.disabled;
     }
 
@@ -202,6 +207,7 @@ class TextInputFlat extends React.Component<ChildTextInputProps, {}> {
       hasActiveOutline,
       activeColor,
       placeholderColor,
+      errorColor,
     };
 
     const minHeight =
--- a/src/components/TextInput/TextInputOutlined.tsx
+++ b/src/components/TextInput/TextInputOutlined.tsx
@@ -84,6 +84,7 @@ class TextInputOutlined extends React.Component<ChildTextInputProps, {}> {
       activeColor,
       outlineColor,
       placeholderColor,
+      errorColor,
       containerStyle;
 
     if (disabled) {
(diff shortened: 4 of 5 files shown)


satya164:
> We should maybe rewrite this using hooks

Unfortunately it'll break compatibility with older versions of RN. Maybe sometime for next major release.

RichardLindhout:
Yeah feels a little bit wrong to minimize on every render. Maybe we should check if the labelLayout has changed and value is set and then to the minimize

src/components/TextInput/TextInput.tsx
```diff
@@ -202,7 +202,8 @@ class TextInput extends React.Component<TextInputProps, State> {
     if (
       prevState.focused !== this.state.focused ||
       prevState.value !== this.state.value ||
-      this.props.defaultValue
+      this.props.defaultValue ||
+      this.props.value
```
satya164: Shouldn't the `this.state.value` check be enough for this? Since we synchronise `this.props.value` with `this.state.value`

RichardLindhout:
```
  (prevState.labelLayout !== this.state.labelLayout && this.props.value) ||
      (prevState.labelLayout !== this.state.labelLayout &&
        this.props.defaultValue) ||
```

src/components/TextInput/TextInput.tsx
RichardLindhout: Yes but it's done in the constructor too, so it does not change. Thats why minimize is never run (labelLayout should be calculated first)

RichardLindhout: I changed this check to be more performant.

satya164: I'm not following. Why wouldn't it change if it's done in constructor? Maybe we want to add `prevState.labelLayout !== this.state.labelLayout` here instead?

RichardLindhout: Flow chart
first render -> 
props.value= 'dd'
state.value= 'dd'
```javascript
componentDidUpdate() {
if (prevState.value !== this.state.value) {
// this will never run since values are the same
// but label needs to minimize
}
}
```

satya164: But we're not checking `props.value` against `state.value`, so they being same shouldn't matter. Isn't the reason this is not run on first render is because `componentDidUpdate` doesn't run on first render?

Also on first render it should be initialised with the label as minimized if value is provided. I think the bug is here: `new Animated.Value(this.props.value ? 0 : 1),` here we don't consider `defaultValue`

src/components/TextInput/TextInput.tsx
```diff
@@ -202,7 +202,8 @@ class TextInput extends React.Component<TextInputProps, State> {
     if (
       prevState.focused !== this.state.focused ||
       prevState.value !== this.state.value ||
-      this.props.defaultValue
+      (prevState.labelLayout !== this.state.labelLayout && this.props.value) ||
```
satya164: Do we actually need to check for `this.props.value` and `this.props.defaultValue`

Also since this is an "or" clause, does it actually prevent the inner code from being run? since `state.value` is going to change every render anyway, triggering the condition.

src/components/TextInput/TextInput.tsx
RichardLindhout: It does not work on first mount, I think because the labellayout is not yet calculated.

src/components/TextInput/TextInput.tsx
RichardLindhout: This is only to prevent the Animated call when a parent component is changed, then the component will be rendered again.

E.g. when you have 1000 textinputs, and in 1 you type they will all 1000 be calling an animated minimizeLabel

src/components/TextInput/TextInput.tsx
satya164: The `minimizeLabel` and `restoreLabel` methods don't use the layout values, so it shouldn't matter.

satya164: I think changing we need to change `new Animated.Value(this.props.value ? 0 : 1)` to `new Animated.Value(this.props.value || this.props.defaultValue ? 0 : 1)` first since the state value is initialized from that, and the label animation state is not consistent with it.

Actually it shouldn't be `this.props.value || this.props.defaultValue`, it should be `this.props.value !== undefined ? this.props.value : this.props.defaultValue`

RichardLindhout: Schermafbeelding 2019-10-31 om 22 44 03
Schermafbeelding 2019-10-31 om 22 43 39

src/components/TextInput/TextInput.tsx
satya164: I think that's because the conditions are wrong here: `this.props.defaultValue` and `this.props.value` shouldn't be in this condition at all. The main issue is this: https://github.com/callstack/react-native-paper/pull/1440#discussion_r341378499

src/components/TextInput/TextInput.tsx
RichardLindhout: They do in my opinion, but not directly. See above.

satya164: @RichardLindhout the interpolation is inside render. It'll always run every render regardless of re-running the animation.

Can you try changing the initial state of label animated value like I mentioned? It seems wrong, and explains the inconsistent initial animated value of label.

RichardLindhout:
```
new Animated.Value(this.props.value ? 0 : 1)
```

This bug happens with value="something" so it could not fix anything to put the defaultValue there?

RichardLindhout: > I think we need to change `new Animated.Value(this.props.value ? 0 : 1)` to `new Animated.Value(this.props.value || this.props.defaultValue ? 0 : 1)` first since the state value is initialized from that, and the label animation state is not consistent with it.
> 
> Actually it shouldn't be `this.props.value || this.props.defaultValue`, it should be `this.props.value !== undefined ? this.props.value : this.props.defaultValue`

That's not gonna fix this issue. The bug is we need to run the interpolated with an *updated* baseLabelTranslateX.
Schermafbeelding 2019-10-31 om 22 55 25

Schermafbeelding 2019-10-31 om 22 54 54

RichardLindhout: The label width and height are not yet known on first render and that's why the transform-origin fix is never applied.

src/components/TextInput/TextInput.tsx
RichardLindhout: I think you have a valid point with default value, but the problem is with value. This fixed both unneeded minimizeLabel calls with defaultValue + fixes the transform origin on label when the textinput is rendered for the first time with a *value* prop.

src/components/TextInput/TextInput.tsx
satya164: @RichardLindhout yes, but then when we have the height and width the component will re-render and the new interpolation should be applied.

satya164: Maybe here's a bug with animated which results in the update not being applied. Can you provide a small repro so I can debug this?

RichardLindhout: You're probably right, maybe it's an animated bug in iOS.

RichardLindhout: Happens only on iOS btw..

RichardLindhout: Uhmm, not sure about that.

satya164: Should this repro it?

```js
<TextInput
  label="Some label"
  placeholder="Type something"
  value="Some test"
/>
```

I can't seem to repro it on iOS simulator

src/components/TextInput/TextInput.tsx
RichardLindhout: But will interpolate run when nothing has changed parentState.labeled and start is not run.

Then it will never run with the new baseLabelTranslateX?

satya164: Interpolation runs every render. If anything inside the interpolation changes, it will apply the update.

src/components/TextInput/TextInput.tsx
RichardLindhout:
```
import React from 'react'
import { View } from 'react-native'
import { TextInput } from 'react-native-paper'

export default function Bug() {
  return (
    <View style={{ flex: 1 }}>
      <View style={{ height: 100 }} />
      <TextInput value={'value'} label={'A very nice label'} />
    </View>
  )
}

```

RichardLindhout: IMG_45781DB11ACE-1

RichardLindhout: With this code
Schermafbeelding 2019-10-31 om 23 25 25
Result
IMG_E2B4E2352DAD-1

satya164: Hmm, that looks fine for me on iOS simulator 🤔

satya164: Kapture 2019-10-31 at 23 30 04

RichardLindhout: Please refresh

RichardLindhout: It's only on first view, not after 1 fast refresh

RichardLindhout: Do you see it?

satya164: I both navigated to the view and reloaded the page. It's not a fast refresh.

What React Native version are you on? I'm guessing it happens on new RN versions since I don't see it in our example app which has RN 0.59, and both the linked issues are on RN 0.61.

Which makes me think that there is some regression in new RN version. And if the change works around the issue, we can add it with a note that it's an workaround for bug in RN 0.61 where updated values don't apply for interpolation.

For the check, `prevState.labelLayout !== this.state.labelLayout` should be enough.

src/components/TextInput/TextInput.tsx
RichardLindhout: Perhaps only when the animated value is changed somehow, thats the only thing that could cause this bug.

src/components/TextInput/TextInput.tsx
RichardLindhout: I'm on 0.61.3

satya164: Ok, so it's a regression in RN. Let's change the condition to add `prevState.labelLayout !== this.state.labelLayout` and add a comment that this is to workaround regression in RN 0.61

RichardLindhout: Did that, thanks for the all your help with this pull request.

RichardLindhout:
Yes, but first a beer too celebrate

RichardLindhout:
Done? I don't fully understand rebasing. I don't get why there are still conflicts since I pulled the upstream master.

satya164:
@RichardLindhout you can try doing a merge instead.

RichardLindhout:
I don't know, I did something wrong with merging.
@satya164 New pull request here: https://github.com/callstack/react-native-paper/pull/1441

PR #1465, opened 2019-11-11
Removed letterSpacing

sm2017:
### Motivation

Languages like arabic and persian has ugly view with low letter spacing


Trancever:
@sm2017 This breaks the Material Design Guidelines. You can pass style prop to override letterSpacing.

PR #1490, opened 2019-11-26
fix: support backdrop dismiss

RichardLindhout:
### Motivation
Dismiss with the backdrop does not work since the safeareaview is in front of it.

### Test plan

See here a discussion in another library where this fix is used
https://github.com/react-native-community/react-native-modal/issues/11#issuecomment-318853831


Trancever:
@RichardLindhout We are planning to release new version early next week. This is gonna be included.

PR #1538, opened 2019-12-20
feat: support loading indicator to Searchbar

brunohkbx:
### Motivation
It would be nice to be able to display a loading indicator when searching

ezgif com-video-to-gif

### Test plan
1. Download this branch
2. Open the Searchbar example
3. See the example using the loading indicator.

### DIFF
--- a/example/src/Examples/SearchbarExample.tsx
+++ b/example/src/Examples/SearchbarExample.tsx
@@ -17,6 +17,7 @@ type State = {
   firstQuery: string;
   secondQuery: string;
   thirdQuery: string;
+  fourthQuery: string;
 };
 
 class SearchExample extends React.Component<Props, State> {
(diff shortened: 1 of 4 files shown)


Trancever:
@brunohkbx Why do you hide close icon in "loading" state? Did you find it in the guidelines? Can you point me to the reference?

brunohkbx:
@Trancever I hid the close icon to show up the loading indicator. 
There isn't any mention of that on MD guidelines and I don't think there's any about the button loading indicator either.
I've just thought it would be a nice addition.

Trancever:
@brunohkbx We are not sure if replacing close icon with loading icon is the correct solution. The UX is not that great and MD Guideline does not mention it anywhere.

satya164:
Yea, user should be able to delete the text when loading results.

Trancever:
Closing this because the user can't delete text when the loading icon is shown. Feel free to submit a new PR if you come up with a solution to the mentioned problem.

PR #1542, opened 2019-12-23
feat: add outlineColor prop in TextInputOutlined & add titleStyle prop in MenuItem & add wrapInSafeArea prop in Modal

ahce:
#### Modal
Determines whether the modal is wrapped in a SafeAreaView
#### MenuItem
Allows change title style.
#### TextInput
Allows to controls the outline color.
Example with outlineColor='transparent'

#### Normal input
image
#### Focused input
image

### Test plan

1. Download this branch
2. Open the TextInput or MenuItem or Modal example
3. Change the prop outlineColor (TextInput) or titleStyle (MenuItem) or wrapInSafeArea (Modal)


src/components/Modal.tsx
```diff
@@ -178,8 +183,15 @@ class Modal extends React.Component<Props, State> {
...
-    const { children, dismissable, theme, contentContainerStyle } = this.props;
+    const {
+      children,
+      dismissable,
+      wrapInSafeArea = true,
+      theme,
+      contentContainerStyle,
+    } = this.props;
     const { colors } = theme;
+    const Wrapper = wrapInSafeArea ? SafeAreaView : View;
```
Trancever: Why do you want to remove SafeAreaView? Do you have any issues with it?

src/components/TextInput/TextInput.tsx
```diff
@@ -55,6 +55,10 @@ export type TextInputProps = React.ComponentPropsWithRef<
    * Underline color of the input.
    */
   underlineColor?: string;
+  /**
+   * Outline color of the input.
+   */
+  outlineColor?: string;
```
Trancever: We need to rethink the strategy of customizing a TextInput. We won't accept this change now. Please give us some more time and we will come up with a good plan for both Flat and Outlined TextInput.

src/components/Modal.tsx
ahce: Yeah problems with the background color, also when I want to stretch an image in the whole modal

src/components/TextInput/TextInput.tsx
ahce: I don't see high impact on this change.
Is there any chance that you will approve it or should I reverse it?

Trancever: It should be done in a separate PR anyway, so yes, please revert this change.

src/components/Modal.tsx
Trancever: Could you provide a [Snack](https://snack.expo.io/) that reproduces your issue? Thanks in advance!

Issue #1549, opened 2020-01-04
Add support for Rating component 


asif-ir:
## Motivation
I was using this library but it lacked a simple rating component which is usually present in every material based UI libraries. I had to implement a simple component for personal use. Instead of needing standalone rating libraries which raise the bundle sizes too much, we can provide a small component for this purpose under this module itself for general use, just like other component libraries.

asif-ir:
I raised the #1550 PR implementing this.

Trancever:
@asif-ir We try to follow MD Guidelines very strictly and only provide components that are described in the guidelines. Since I can't find any information about "Rating" component in the guidelines, I am closing this.
If you want to reuse this in other projects etc. consider making a library. It won't add any additional overhead since it will be only one file with only peer dependencies.

PR #1550, opened 2020-01-04
feat: add support for rating component (#1549)

asif-ir:
Implements: https://github.com/callstack/react-native-paper/issues/1549


asif-ir:
The PR has been manually tested as well as jest unit test coverage was ensured to be 100% for the new component files. 

The idea is to use the rating component for general purpose without any need of a 3rd party module. 

Example code:
```jsx
<Rating value={4} max={5} />
```
The above will render this:

image

Trancever:
I am closing this, because "Rating" doesn't seem to be a MD component @1549

PR #1567, opened 2020-01-11
feat: added prefixes to text inputs

lededje:
### Motivation
To allow users to specify prefixes to text inputs as specified in the [material ui documentation](https://material.io/components/text-fields/)

### Test plan

If the proposed approach is good, I'll add snapshot tests for various combinations of input type, size etc.


jayu:
Thank you for the PR, but we are planning to add a prefix prop that would accept text, icon, etc. We need a more flexible solution and leading and tailing prefixes. I'm closing it for now, but feel free to contribute, maybe preceded with a discussion about API (in issue)

PR #1571, opened 2020-01-14
fix: Added optional prop theme in typescript format for TextInput

espipj:
### Motivation

The current typescript interface make the theme prop to be mandatory but there's a comment over it stating that it is in fact optional. This fix will add more consistancy.


Trancever:
@espipj TextInput is wrapped with `withTheme` HOC that always provides this prop, so there is no reason to make the prop optional.

PR #1573, opened 2020-01-14
feat: added testID to FAB and FAB.Group

digorath:
### Motivation
Added a testID to enable testing with detox. You can now add a testID as it is done for other components to the FAB and the FAB.Group.
Additionally the you can hand in testIDs to any action provided to a FAB.Group.

### Test plan
Check, if the way that the testIDs of the FAB children of the FAB.Group are generated in a reasonable way (especially in respect to naming conventions)

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -78,6 +78,10 @@ type Props = {
    * @optional
    */
   theme: Theme;
+  /**
+   * Pass down testID from Group props to FAB and child FABs with indices.
+   */
+  testID?: string;
 };
 
 type State = {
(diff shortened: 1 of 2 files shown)


src/components/FAB/FABGroup.tsx
```diff
@@ -306,6 +311,7 @@ class FABGroup extends React.Component<Props, State> {
                   accessibilityTraits="button"
                   accessibilityComponentType="button"
                   accessibilityRole="button"
+                  testID={`${testID}-${i}`}
```
Trancever: Each action should accept a testID instead of doing this.

digorath: changed it.

Issue #1113, opened 2019-06-07
Enabling scrolling in datatable


rodriigovieira:
## Describe the feature

It would be nice to be able to pass a prop to Datatable choosing whether it should be scrollable or not. Something such as `scrollable`.

If it is currently possible, I'm sorry, I looked for it but I didn't find it.

tiendn:
I think you can customized your own code with a `ScrollView` wrap here. 
```js
import * as React from 'react';
import { DataTable } from 'react-native-paper';

export default class MyComponent extends React.Component {
render() {
    return (
      <DataTable>
        <DataTable.Header>
          <DataTable.Title>Dessert</DataTable.Title>
          <DataTable.Title numeric>Calories</DataTable.Title>
          <DataTable.Title numeric>Fat</DataTable.Title>
        </DataTable.Header>

        <ScrollView> /* Add here */
          <DataTable.Row>
            <DataTable.Cell>Frozen yogurt</DataTable.Cell>
            <DataTable.Cell numeric>159</DataTable.Cell>
            <DataTable.Cell numeric>6.0</DataTable.Cell>
          </DataTable.Row>
          <DataTable.Row>
            <DataTable.Cell>Ice cream sandwich</DataTable.Cell>
            <DataTable.Cell numeric>237</DataTable.Cell>
            <DataTable.Cell numeric>8.0</DataTable.Cell>
          </DataTable.Row>
        </ScrollView>

        <DataTable.Pagination
          page={1}
          numberOfPages={3}
          onPageChange={(page) => { console.log(page); }}
          label="1-2 of 6"
        />
      </DataTable>
    );
  }
}
```

rodriigovieira:
That's correct, it indeed works. The line became scrollable.

What I am trying to make, though, is for the table to work in smaller screens, by not displaying the three dots (...).

When I wrap the `Datatable.Row` in a `ScrollView` with the `horizontal` property, all elements get together, without any spacing between them. I tried to style it in several different ways, but I wasn't able to make it work so far.

Also, it'll probably be necessary to wrap the `Datatable.Header` in a `ScrollView` as well, so it can have the same width as it's children.

It would be nice if there were an out-of-the-box feature that could do this.

tiendn:
You meant, `horizontal scroll`?
![giphy](https://user-images.githubusercontent.com/15717476/59287981-f2ee4880-8c9c-11e9-8562-1fc76dfe7b91.gif)

HenokMengistu:
[...]

PR #1574, opened 2020-01-14
feat: added Column to DataTable

marchenk0va:
### Motivation

Column was added to make table scrollable. DataTable contains independent components, and we can not apply such prop on them.

### Test plan

1. Install example and run on expo
2. Check DataTable.Column


jayu:
Hi @marchenk0va ! Please edit example usage to reflect the use case for this component :) I cannot verify whether it is working or not. 
One of the possible use cases is in this comment https://github.com/callstack/react-native-paper/issues/1113#issuecomment-500909928

jayu:
Hey @marchenk0va will you manage to update the example that I mentioned in the previous comment?

PR #1576, opened 2020-01-15
feat: Multiple line for card title and subtitle

TVGSoft:
The title property of a card would wrap lines
The subtitle property of a card would wrap lines

### Motivation

58108975-fb75c680-7ba9-11e9-9c36-851042827d08

### DIFF
--- a/src/components/Card/CardTitle.tsx
+++ b/src/components/Card/CardTitle.tsx
@@ -22,10 +22,6 @@ type Props = React.ComponentProps<typeof View> & {
    */
   titleStyle?: StyleProp<TextStyle>;
   /**
-   * Number of line for the title.
-   */
-  titleNumberOfLines?: number;
-  /**
    * Text for the subtitle. Note that this will only accept a string or `<Text>`-based node.
    */
   subtitle?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


jaulz:
I'm not sure... I can only see examples where the subtitle has more than one line but never the title itself: https://material.io/components/cards/#specs

Thus I would suggest to limit it to the subtitle. What do you think @Trancever ?

TVGSoft:
Hi @jaulz 

Thank you for your suggestion. I understand the specs defined by material.io.
In some cases on a small screen, it was cut off. 
So by default, it still limits one line, however, still allows developers can customize it.

Trancever:
Yeah, we should only accept this prop for subtitle.

TVGSoft:
Hi @Trancever 

I already removed props for the title and only accept the subtitle

leonardoballand:
please @jaulz or @wojteg1337 can you review this PR? my team needs this fix. Thanks

jaulz:
You can always use https://github.com/ds300/patch-package in such cases as long as it's not released.

IngenieroLata:
Just updated the version and I'm requiring the card title to be a single line with ellipsis. Any change to get the `numberOfLines` for the title back?

It would be awesome some extra prop similar to `subtitleNumberOfLines` like `titleNumberOfLines `

jaulz:
@IngenieroLata could you please raise a PR for that?

IngenieroLata:
@jaulz sure here it is https://github.com/callstack/react-native-paper/pull/1809

PR #1586, opened 2020-01-18
removed textinput upderline

SimaRajput:
removed textinput upderline


SimaRajput:
removed textinput underline

Trancever:
@SimaRajput  Sorry, but this breaks the MD Guidelines and we can't accept it.

PR #1599, opened 2020-01-22
fix: disable Modal background when not dismissable

jaulz:
### Motivation

Currently, even if not `dismissable` the Modal background is clickable on web (i.e. pointer is cursor). This change disables the background view so it does not get a focus index and the cursor is the default one.

### Test plan

Open Modal with `dismissable={false}` on web and ensure that no pointer cursor is used on the background.


jaulz:
@Trancever any opinion on this one? 😊

PR #1600, opened 2020-01-22
feat(searchbar): add search and clear accessibility labels

jamsea:
### Motivation

We need a way to add a generated accessibility label to the search and clear buttons. The copy changes based on a user's locale settings.

### Test plan

As long as the typescript and eslint checks pass this should be good to go.

### DIFF
--- a/src/components/Searchbar.tsx
+++ b/src/components/Searchbar.tsx
@@ -96,6 +96,10 @@ type Props = React.ComponentProps<typeof TextInput> & {
  * ```
  */
 class Searchbar extends React.Component<Props> {
+  static defaultProps = {
+    searchAccessibilityLabel: 'search',
+    clearAccessibilityLabel: 'clear',
+  };
   private handleClearPress = () => {
     this.clear();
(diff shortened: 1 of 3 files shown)


src/components/Searchbar.tsx
```diff
@@ -17,6 +17,14 @@ import { IconSource } from './Icon';
 import MaterialCommunityIcon from './MaterialCommunityIcon';
 
 type Props = React.ComponentProps<typeof TextInput> & {
+  /**
+   * Accessibility label for the button. This is read by the screen reader when the user taps the button.
+   */
+  clearAccessibilityLabel?: string;
```
Trancever: This is nice, but we should also specify a default value for these labels: "search" and "clear". Default values should be defined in `static defaultProps`.

Trancever: This is nice, but we should also specify a default value for these labels: "search" and "clear". Default values should be defined in static defaultProps.

jamsea:
@Trancever how's this look?

PR #1627, opened 2020-01-31
chore: add QR code to expo preview messages

satya164:


Trancever:
Wow, magic :D Nicely done @satya164

PR #1641, opened 2020-02-05
Clear snackbar timeout when content has changed

TrueTiem:
I don't know if this is intentional but, currently snackbar doesn't reset it's timeout when we change the snackbar's content while it is visible. This causes the second content to show for a short time.

### Motivation

I'm using single snackbar for a page to keep the code much more cleaner. But if user activates snackbar and does something fast, second content shown for a short time.

### Test plan

Use ne snackbar
Create two buttons that sets the snackbar's content to 2 different texts
Press first button
Press second button just before snackbar's timeout ends

### DIFF
--- a/src/components/Snackbar.tsx
+++ b/src/components/Snackbar.tsx
@@ -142,8 +142,21 @@ class Snackbar extends React.Component<Props, State> {
   componentDidUpdate(prevProps: Props) {
     if (prevProps.visible !== this.props.visible) {
       this.toggle();
+    } else if (this.props.visible && prevProps.children !== this.props.children) {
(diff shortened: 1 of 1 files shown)


satya164:
Children not being equal doesn't mean content has changed. Children will change every time the component containing the Snack bar rerenders.

TrueTiem:
Well, yes. and no,, I didn't think about that since im working with only strings.

Trancever:
@TrueTiem As a workaround, you can pass an Infinity as a `duration` and control Snackbar visibility from parent component by changing `visible` prop

PR #1674, opened 2020-02-16
feat: RadioButton.Item add labelStyle prop & support themes

firede:
### Motivation

`RadioButton.Item`:

- fix: Label element using `./Typography/Text` instead of `react-native/Text` (to support themes).
- feat: add `labelStyle` prop.
- docs: fix link.

### Test plan

Screenshots from examples.

RadioButton Item issue@2x


Trancever:
@firede Thanks!

PR #1681, opened 2020-02-18
feat: add `elevation` to Theme type

adilusman51:
feat: add `elevation` to Theme type
elevation could be used to apply custom elevation to following components
- Appbar
- Appbar.Header
- Surface
- Card
- FAB
- FAB Group
- Search Bar

feat: Add `elevation` prop to following components
- Appbar
- Appbar.Header
- FAB
- FAB Group
- Search Bar

### DIFF
--- a/src/components/Appbar/Appbar.tsx
+++ b/src/components/Appbar/Appbar.tsx
@@ -26,6 +26,10 @@ type Props = Partial<React.ComponentProps<typeof View>> & {
    */
   children: React.ReactNode;
   /**
+   * Resting elevation of the app bar which controls the drop shadow.
+   */
+  elevation?: number;
+  /**
    * @optional
    */
   theme: Theme;
@@ -86,14 +90,25 @@ class Appbar extends React.Component<Props> {
   static Header = AppbarHeader;
 
   render() {
-    const { children, dark, style, theme, ...rest } = this.props;
-
-    const { colors, dark: isDarkTheme, mode } = theme;
     const {
-      backgroundColor: customBackground,
-      elevation = 4,
-      ...restStyle
-    }: ViewStyle = StyleSheet.flatten(style) || {};
+      children,
+      dark,
+      style,
+      theme,
+      elevation: propsElevation,
+      ...rest
+    } = this.props;
+
+    const {
+      colors,
+      dark: isDarkTheme,
+      mode,
+      elevation: themeElevation = 4,
+    } = theme;
+    const { backgroundColor: customBackground, ...restStyle }: ViewStyle =
+      StyleSheet.flatten(style) || {};
+
+    const elevation = propsElevation || themeElevation;
 
     let isDark: boolean;
 
@@ -139,7 +154,12 @@ class Appbar extends React.Component<Props> {
     return (
       <Surface
         //@ts-ignore
-        style={[{ backgroundColor }, styles.appbar, { elevation }, restStyle]}
+        style={[
+          { backgroundColor },
+          styles.appbar,
+          elevation && { elevation },
+          restStyle,
+        ]}
         {...rest}
       >
         {shouldAddLeftSpacing ? <View style={styles.spacing} /> : null}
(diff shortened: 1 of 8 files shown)


satya164:
Thanks for the PR.

The `elevation` style varies depending on the component. It doesn't make much sense to have a single `elevation` property in the theme. Also `elevation` is a style property. there's no need to have a separate `elevation` prop for every component

Please open an issue to discuss what you want to change before doing the work so that unnecessary work can be avoided.

adilusman51:
Hi, @satya164 
Thanks for taking some time out to review this pull request.
Motive behind adding a elevation type in the theme is to apply consistent elevation on the component through out the App, so that it is easy to maintain consistency in design.

There are two methods for achieving this in current implementation. First method, we can override individual styles to the components, which leads to code duplication in case we have to use same component at multiple places. Second method, wrap all the components with custom components (preferably using typescript ) hard coding the style props, this has multiple side effects, namely, unnecessarily performance hit due to wrapping components and comprising on the functionality of the original component as we have to provide typed props for children component, again duplication of effort.

Flutter, through material library, also provide configurable global elevation property for design consistency.

For individual component which require different elevation than the theme elevation, they can customise it either by prop or style. Prop is convenient way of adding elevation.

satya164:
> Motive behind adding a elevation type in the theme is to apply consistent elevation on the component through out the App, so that it is easy to maintain consistency in design.

Consistency is about providing a consistent user experience, not using the same value for everything. The app's behaviour should be consistent with rest of the OS.

Elevation is about defining hierarchy of elements, FABs are above the appbar, so have higher elevation, dialogs are above everything, so have even higher elevation etc. Think of using different font sizes to denote headings, paragraphs etc. Making them the same font size removes the concept of hierarchy and provides a worse UX. Same for elevation.

> Flutter, through material library, also provide configurable global elevation property for design consistency.

I can't find where it does that and I have never used Flutter, but if it does it then it's not according to Material Design guidelines.

> they can customise it either by elevation prop or elevation in ViewStyle

There's no reason to duplicate it. Same could be done for `backgroundColor` etc.

adilusman51:
Hi @satya164,
Here is the link for the Material class library for Flutter which is maintained by Google.

[Material class](https://api.flutter.dev/flutter/material/Material-class.html)
https://api.flutter.dev/flutter/material/Material-class.html

satya164:
> Here is the link for the Material class library for Flutter which is maintained by Google.

The material class just seems like a base widget rather than global style for all components. The equivalent of global styles (theme) doesn't have anything regarding global elevation https://api.flutter.dev/flutter/material/ThemeData-class.html

Maintained by Google doesn't mean anything if it's not in the guidelines. There are different teams which constantly implement MD guidelines incorrectly.

Trancever:
@adilusman51 Hi!

We are not gonna merge this because it breaks the MD Guidelines.

If you really need these changes I would suggest forking the library.

adilusman51:
Hi @satya164 
Flutter Material library even provides more functionality of configuring material Shape and text styles for UI consistency.

Motivation here is to provide more functionality in paper library like Flutter Material

adilusman51:
Hi @Trancever 
Thanks for providing your feedback.

My suggestion is that you review the Material Design capabilities that Flutter is providing. So that those functionalities can be provided in Paper library

satya164:
Paper's goal is not to replicate flutter's functionality. The goal is to implement material design guidelines for React Native.

Regardless, the original PR regarding global elevation is against the guidelines and flutter doesn't have such functionality (though it's not relevant what flutter has). If you need specific features, you should open separate issues for them.

adilusman51:
Thanks @satya164  and @Trancever for healthy discussion.

I hope in future, I can contribute to Paper. I am exploring the idea of more customisable options in theme for configuring design elements from central Theme, rather than wrapping Paper components.

Flutter is providing similar functionality with MaterialApp Widget which provides finer control on component design.

I hope you guys also look into the Flutter approach and community can collaborate to enhance Paper.

[MaterailApp Widget provides this functionality through ThemeData attributes](https://api.flutter.dev/flutter/material/MaterialApp-class.html)

[ThemeData](https://api.flutter.dev/flutter/material/ThemeData-class.html)

PR #1685, opened 2020-02-19
feature(searchbar): adds numberOfLines and multiline props

albert-park:
### Motivation
Adding the `numberOfLines` and `multiline` props to `Searchbar` adds flexibility in styling the placeholder text. 

### Test plan
With passing eslint and Typescript tests, this should be good to go.


src/components/Searchbar.tsx
```diff
@@ -131,15 +147,19 @@ class Searchbar extends React.Component<Props> {
...
       inputStyle,
+      onIconPress,
+      multiline = false,
+      numberOfLines = 1,
+      placeholder,
+      searchAccessibilityLabel,
+      style,
+      theme,
+      value,
       ...rest
```
nogueirahy: the prop ```rest``` do it

PR #1687, opened 2020-02-20
fix: memoize MaterialCommunityIcon

jaulz:
### Motivation

While trying to identify performance bottle necks I found that the Icon is rendered too often because the styles are changing:
https://github.com/callstack/react-native-paper/blob/2a29dc6d6bca0cb8f529b192d2af3a1794e61a2a/src/components/MaterialCommunityIcon.tsx#L93 
This change will memoize the component to prevent unnecessary rerenders.

### Test plan

Nothing should change and icons should behave as usually.

### DIFF
--- a/src/components/Icon.tsx
+++ b/src/components/Icon.tsx
@@ -104,13 +104,8 @@ const Icon = ({ source, color, size, theme, ...rest }: Props) => {
   } else if (typeof s === 'string') {
     return (
       <SettingsConsumer>
-        {({ icon }) => {
-          return icon({
-            name: s,
-            color: iconColor,
-            size,
-            direction,
-          });
+        {({ icon: Icon }) => {
+          return <Icon name={s} color={iconColor} size={size} direction={direction} />;
         }}
       </SettingsConsumer>
     );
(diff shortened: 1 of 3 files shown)


Trancever:
@satya164 I remember you had some insights why shouldn't we merge this one. Do you remember what was that?

src/components/Icon.tsx
```diff
@@ -104,13 +104,15 @@ const Icon = ({ source, color, size, theme, ...rest }: Props) => {
...
-        {({ icon }) => {
-          return icon({
-            name: s,
-            color: iconColor,
-            size,
-            direction,
-          });
+        {({ icon: Icon }) => {
+          return (
+            <Icon
```
satya164, line 109: Since we accept a render callback and not a function component, this is a breaking change. It'll also cause remount of all icons if the reference to the function changed, then it'll have worse perf than re-rendering.

jaulz:
It did not cause any performance issues but I noticed it popping up a lot while using https://github.com/welldone-software/why-did-you-render and I just thought it might be a nice addition 😊  Feel free to close it then.

PR #1691, opened 2020-02-21
fix: add flexShrink to inner Card container

jaulz:
### Motivation

I noticed that the Card somehow behaves strange while resizing  and keeps having the same size. It seems that `flexShrink` was missing. This style change should not have any side-effects.

### Test plan

n/a


jaulz:
@Trancever thanks! Shall I merge it?

Trancever:
@jaulz I just forgot to merge it :D

PR #1701, opened 2020-02-27
docs: clarify color prop for FAB

LuckyLuke19:
Explicitly mention that the `color` prop of `FAB` refers to its *text* color (and not its *background* color).

### DIFF
--- a/src/components/FAB/FAB.tsx
+++ b/src/components/FAB/FAB.tsx
@@ -32,7 +32,7 @@ type Props = $RemoveChildren<typeof Surface> & {
    */
   small?: boolean;
   /**
-   * Custom text color for the `FAB`.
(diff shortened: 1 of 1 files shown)


src/components/FAB/FAB.tsx
```diff
@@ -32,7 +32,7 @@ type Props = $RemoveChildren<typeof Surface> & {
    */
   small?: boolean;
   /**
-   * Custom color for the `FAB`.
+   * Custom text color for the `FAB`.
```
satya164: Thanks for the PR. Can you tweak the description so it's obvious that it changes the icon and label color?

```suggestion
   * Custom color for the icon and label of the `FAB`.
```

Issue #1703, opened 2020-02-28
Add prop testID to Button component


daniloab:
## Describe the feature

It will add the prop testID to works for tests on react native.

## Motivation

Creating a test for my application migrating from older components to new that uses react-native-paper, my tests starting to broken. Trying to understand why I realize that is because the testID was setting on wrong component (View) and not in the `TouchableHighlight`

Example
<img width="285" alt="Screen Shot 2020-02-28 at 16 16 13" src="https://user-images.githubusercontent.com/19939822/75579963-aa9ed280-5a45-11ea-919d-069d9db32eb2.png">

PR #1704, opened 2020-02-28
feat(button): add test id prop

daniloab:
### Motivation
fix callstack/react-native-paper#1703

### Test plan
Render the default testID and the custom testID.


src/components/Button.tsx
```diff
@@ -166,6 +171,7 @@ class Button extends React.Component<Props, State> {
       theme,
       contentStyle,
       labelStyle,
+      testID = 'button',
```
sibelius: not sure we need a default value here

daniloab: I believe should be a better name than only "button"

PR #1712, opened 2020-03-04
Added prop rippleDisabled in IconButton

EduVencovsky:
### Motivation

I want to be able to use `IconButton` but without the ripple effect ( without feedback ) because I already added another way to give the feedback.

### Test plan

1. Use IconButton with any normal prop and see ripple effect
2. Now pass boolean prop rippleDisabled to remove ripple effect


Trancever:
@EduVencovsky That's something specific only to your project. You should use patch-package or fork the repo for such a use case.

Issue #1725, opened 2020-03-10
Set icon color on a <Button />


vpontis:
I'd like to be able to control the icon's color on ` Button` so that it matches the `textColor` that I am manually setting via `labelStyle`. 

I have a PR that I will submit that adds `Button.iconColor`. Another alternative is to allow specifying `Button.iconProps`.

Issue #1718, opened 2020-03-06
Button icon property should allow startIcons and endIcons


mschipperheyn:
## Describe the feature

icons can either occur at the end of a button or at the start. I would like to see a startIcon and endIcon property.

## Motivation

Not everything is strict Material Design and it's a very common pattern.

justindra:
For usage-wise, it used to be in Material Design guidelines for MobileSteppers. But it's no longer there I guess.

Material-ui is still supporting it.
https://material-ui.com/components/steppers/#mobile-stepper

kabus202:
yep good feature

mschipperheyn:
There is actually a little trick you can do get the end icon effect. Apple a `row-reverse` to the `contentStyle` and add the icon as you would normally. Make sure you pass the icon as an svg component, not an icon name. Place a `View` around it and compensate the now wrong margin on it.

Issue #1659, opened 2020-02-10
How can i change Button's icon size 


yoojeen126:
<Button icon="facebook" mode="contained" onPress={() => console.log('Pressed')} color={theme.colors.facebook}>
        Facebook Login
</Button>

I saw icon size default 16, and i know labelStyle to change text. 
So what's prop to change icon size more big?

quantjamespecutus:
agree this would be useful

Abban-Fahim:
Yes it is neccessary

PR #1726, opened 2020-03-10
feat: add iconColor to Button

vpontis:
### Motivation

Allow changing the color of an icon on a `<Button />`

Resolves https://github.com/callstack/react-native-paper/issues/1725

### Test plan

Create a button with and without `iconColor`

PR #1727, opened 2020-03-10
feat: added iconStyle prop to allow overrides of icon styling

justindra:
### Motivation

There has been a need to customize the icons on a button that may not necessarily be following the Material Guidelines as it is for specific screens and use-cases for a particular user experience.

Some current open issues

* #1725 
* #1718 
* #1659 

### Test plan

* Provide the prop into a Button and see that it changes the styling of the icon
```tsx
function CustomButton() {
  return (
    <Button iconStyle={{
      backgroundColor: '#f0f'
    }} icon="calendar">Icon should have a background</Button>
  )
}
```

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -78,6 +78,10 @@ type Props = React.ComponentProps<typeof Surface> & {
    */
   labelStyle?: StyleProp<TextStyle>;
   /**
+   * Style for the icon.
+   */
+  iconStyle?: StyleProp<ViewStyle>;
+  /**
    * @optional
    */
   theme: Theme;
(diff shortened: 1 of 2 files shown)


kamui545:
Hey, thanks for the PR.

I've got a question, does `iconStyle.color` work to change the color of the icon?
Might need something like `<Icon source={icon} size={16} color={iconStyle.color || textColor} />`

Related PR: https://github.com/callstack/react-native-paper/pull/1726

justindra:
@kamui545  good spotting and doneskies!

Trancever:
@justindra Unfortunately I have too close the PR. MD Guidelines explicitly shows that the icon in the button should have the same color as a text. 

For such use cases, you should use patch-package or fork the repo.

PR #1741, opened 2020-03-18
Fix: Update Menu to use FlatList instead of ScrollView

mlabrum:
Hi React Native Paper team,

I've replaced ScrollView with FlatList which renders items as the user scrolls to improve performance, compared to ScrollView which renders all at once.

### Motivation
I'm attempting to use a few items within the Menu Component ( Timezone settings ) and the large list of items slows down the initial render.

### Test plan
Nothing visual has changed

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -12,8 +12,8 @@ import {
   TouchableWithoutFeedback,
   View,
   ViewStyle,
-  ScrollView,
   findNodeHandle,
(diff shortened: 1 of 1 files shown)


jayu:
Hi @mlabrum How many items's slows down `Menu` in your case?  I believe that the menu in most cases should no have more than 10 items because no one want's to scroll small containers with a lot of values.
Something around `10` items should not affect performance and it's a **default batch size** for FlatList,

PR #1744, opened 2020-03-20
feat: add labelStyle prop for FABGroup labels

PiwwowPants:
### Motivation

The current implementation of the FABGroup makes it so the FABs in the group can be styled but the labels cannot. This change makes it so the FAB labels can be styled independently of the FAB buttons next to them. The main goal of this change is to be able to style the text in the label but extends to allowing you to style the container of the label as well, though that need may be more niche.

The prop added covers styling both the container for the label and the text inside of the label itself. I can imagine a lot of cases where the FAB button and the label next to it should not be in sync as far as styling (the button may have a colored background but the label may be white for readability).

### Test plan

Testing can just be done by passing the `labelStyle` prop to individual actions in the FABGroup. The prop takes `ViewStyle` and `TextStyle` properties which should not overlap but would be the biggest point of failure on this change if they did since you could somehow end up with conflicts in styles. 

image

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -7,6 +7,7 @@ import {
   TouchableWithoutFeedback,
   View,
   ViewStyle,
+  TextStyle,
 } from 'react-native';
 import color from 'color';
 import FAB from './FAB';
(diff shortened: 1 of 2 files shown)


src/components/FAB/FABGroup.tsx
```diff
@@ -33,6 +35,7 @@ type Props = {
     color?: string;
     accessibilityLabel?: string;
     style?: StyleProp<ViewStyle>;
+    labelStyle?: StyleProp<ViewStyle | TextStyle>;
```
Trancever, line 38: It's not intuitive to pass the same style prop to two components.  I would prefer to have two separate props for that

PR #1760, opened 2020-03-27
fix: expose ref on DrawerItem

jaulz:
### Motivation

Currently, it is not possible to pass ref to DrawerItem.

### Test plan

Use DrawerItem, pass ref to it and no typing errors should be there.


jaulz:
@Trancever sorry I just noticed that the `ref` property was also missing here. Could you also merge this one?

jaulz:
Thanks! 🙂

PR #1765, opened 2020-03-27
Fix: fix visible props not being applied to Fab actions

kopax:
### Motivation

When using `<Portal>` and `<Fab.Group />` with `visible` props, only the Fab main button get hidden before the page transition.

### Test plan

Using react-navigation 5, use `isFocus` on a `<FabGroup />`.

- Without it: it will keep the FAB icon
- With it: it will hide the FAB icon at the same time as the Fab group icon

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -287,34 +287,36 @@ class FABGroup extends React.Component<Props, State> {
                     <Text style={{ color: labelColor }}>{it.label}</Text>
                   </Card>
                 )}
-                <FAB
-                  small
-                  icon={it.icon}
-                  color={it.color}
-                  style={
-                    [
-                      {
-                        transform: [{ scale: scales[i] }],
-                        opacity: opacities[i],
-                        backgroundColor: theme.colors.surface,
-                      },
-                      it.style,
-                    ] as StyleProp<ViewStyle>
-                  }
-                  onPress={() => {
-                    it.onPress();
-                    this.close();
-                  }}
-                  accessibilityLabel={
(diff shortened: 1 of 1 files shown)


Trancever:
@kopax Could you provide a [Snack](https://snack.expo.io/) with an example of the issue?

kopax:
@Trancever, well for this you can trust me, it is required. 

I tested it in my app. Not really at the moment, I did a few reproductions and contribution this  week (expo, react-native,paper) and I can't do more right now as I also have some work and pressure from my boss. 

But here is the story:

- I talk to @satya164 on discord and he gave me this article:  https://reactnavigation.org/blog/2020/01/29/using-react-navigation-5-with-react-native-paper/#fab-and-portal
- I was displaying a fab and the fab was still on the next page after a transition, he told me to use `useIsFocused`. it worked fine with it, except for the actions within the fab group which remain on the page transition.
- I had to copy Fab.Group and this fixed it. 

If you read the article, then you'll get the point and example, it is the same with Fab.Group because Fab.Group is in the portal and it stay until the end of the animation. **It's required if you wish to use with Fab.Group with react navigation**

Trancever:
@kopax I know the article, cause I am the author of it xd I believe if you do something like this:

```
<FAB.Group open={isFocused && open} visible={isFocused} />
```

it's gonna hide both the fab and all the actions.

kopax:
Okay :) I couldn't click and close at the same time with the provided snack, but that's not a big deal.  I have replaced the fab group in my app for something else, but another page will use the fab group and I will get into it soon enough so I will test your option. If required, I will reopen this. Thanks for the article and the hard work.

PR #1767, opened 2020-03-29
Feat: Added loading props to show activity indicator on drawer items

kopax:
- Added a loading indicator to the component (tested on web/ios/native)

### Motivation

I am having a logout link in my Drawer and it is asynchronous task, this feature was missing so I did copy the `DrawerItem.ts` in my source code and added this.

I tought this would be appreciated by others, feel free to accept or refuse the feature.

To me, this is especially useful for the web, because the sidebar is docked on the left, while on mobile, the menu is closed directly after the click.

### Test plan

Do `<DrawerItem loading={true} />` , it will display:

image

### DIFF
--- a/src/components/Drawer/DrawerItem.tsx
+++ b/src/components/Drawer/DrawerItem.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { View, StyleSheet, StyleProp, ViewStyle } from 'react-native';
 import Text from '../Typography/Text';
 import Icon, { IconSource } from '../Icon';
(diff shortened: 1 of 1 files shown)


kopax:
I am sorry for the typescript error, I am not used (at all) to typescript, perhaps I can get a hand on that one?

Trancever:
This feature is not compliant with MD Guidelines. I suggest using patch-package or fork the repo for such use cases

kopax:
Ok. I will do. Would be nice to have another official version of `react-native-paper` that extends MD guidelines.

Issue #1678, opened 2020-02-17
useNativeDriver is required in Animation.Timing calls in newer versions of React Native


ryanmagoon:
## Describe the feature

in Animated for react native v62+, `useNativeDriver` is a required value. I saw this in particular for `onPressIn` and `onPressOut` for the `Button` component.

![image](https://user-images.githubusercontent.com/5327290/74630552-d1e1cf80-5128-11ea-929e-f53d7e42820e.png)

## Motivation

getting rid of yellow box errors on day one when 62 comes out fully

ahmetbicer:
Same problem here. Would be great if we get a fix.

hakimhassani97:
hi @ryanmagoon ,
a faced the same problem as you  
i'm on  
`"react-native": "0.62.0",`
`"react-native-paper": "^3.6.0",`

for me the problem was on **Button** with `mode='contained'`
so my work around was to :
- remove the `mode` prop
- use a custom style
`style={{backgroundColor:Theme.primary, elevation:7, border:0}}`
- set text color, which changes the **ripple effect's color**
`color={Theme.light}`
and now the **Button** animations work without warnings
😃

abgaryanharutyun:
I have some problem too.

Waiting for update. Thanks

azuxx:
Same problem too and proposed workaround does not work, I have this error:

![Screenshot 2020-04-07 at 10 04 56](https://user-images.githubusercontent.com/16132913/78645225-7970b800-78b7-11ea-8cc1-890dc12886e7.png)

Trancever:
Fixed in v3.8

Issue #959, opened 2019-03-29
TextInput throwing a Animated error


brunocrpontes:
## Environment

Expo v32.0.0
running project in a iOS v12.1.4
react-native-paper v2.15.0

## Description
When i try update error props in a TextInput it returning this error:
 
![IMG_0166](https://user-images.githubusercontent.com/13150168/55214946-e2b6e900-51d6-11e9-875d-900cbcdc58c9.PNG)

## Reproducible Demo

https://snack.expo.io/@brunocardosor/restless-apples

## Update

I tested in a Android device,  everything runs fine, this happening only in iOS

satya164:
cc @Trancever I believe we need to remove `useNativeDriver` from all animations

PR #1769, opened 2020-03-31
Fix: button animation

logustra:
### Motivation
https://github.com/callstack/react-native-paper/issues/1678

### Test plan
Nothing visual has changed

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import {
-  Platform,
   Animated,
   View,
   ViewStyle,
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {
+  Platform,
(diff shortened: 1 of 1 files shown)


Trancever:
@logustra Is it possible to animate elevation using native driver now?

vicary:
From https://github.com/callstack/react-native-paper/issues/959#issuecomment-477961362, we should be cautious to put a simple `true`. Something like what ad516cca4326bfa55a2859d85817de11c54c36e8 does should be more sensible, because the referenced issue only wants to get rid of a warning where `undefined` should be explicitly `false`.

logustra:
@Trancever 
I have checked it with expo 32 and got an error on iOS and I fixed it like you did on https://github.com/callstack/react-native-paper/commit/ad516cca4326bfa55a2859d85817de11c54c36e8

Trancever:
@logustra I believe it's safe to pass true for iOS cause we are dropping support for RN < 0.60 anyway.

Issue #1705, opened 2020-02-29
Having loading on Searchbar and TextInput Components


kopax:
## Describe the feature

I am using `Searchbar` as follow:

```jsx
        <Searchbar
          placeholder={`${t('homeView.search')}...`}
          value={cityOrPostalCode}
          onChangeText={this.onChangeText}
          loading={1}
          onSubmitEditing={this.submit}
        />
```
 
## Motivation

So far, there is no way to switch the left icon, the right icon, or to add a component within the text to display an asynchronous task.

A use case would be a `<Searchbar />` used for autocomplete, it will fetch some data.

So far I must create my own indicator, which most likely does not respect material ui guidelines

Thanks again for your time,

Dimitri

belgamo:
What do you think about adding a boolean prop such as `inverseIcons`, as a way to switch icons?

kopax:
Why not just icon left and icon right ? :/?

belgamo:
I thought you just wanted changing the position of the icons. There's already exists a prop to do that but the name is not very semantic... Maybe we can restructuring the `icon` and `clearIcon` props so that they can be treated as `leftIcon` and `rightIcon`.

See: https://callstack.github.io/react-native-paper/searchbar.html#icon

kopax:
This features would be so much appreciated, yes that would be awesome. 

I am just thinking, would it be nice to toggle the clear icon with a loading icon while searching something asynchronously (which is almost always the case). I have this proposal in mind:

- Keep `icon` for previous version with a deprectation warning
- Create `leftIcon` that will serv the purpose of `icon`
- Keep `clearIcon` for clearing the search bar
- Add `rightIcon` for example for adding a spinner displayed on the left to the clear icon in the search bar


Allow user to rearrange those 3 icons on it's while

brunohkbx:
@Trancever what do you think about this issue? Allowing the user to change the `rightIcon` shouldn't be a problem at all, right?  Instead of forcing this behavior as I did on #1538

brunohkbx:
@Trancever :point_up:

brunohkbx:
It's still open

kopax:
It's still open

zholmes1:
[...]

PR #1774, opened 2020-04-02
Customize Searchbar icons 

belgamo:
### Motivation
Provide a way to custom the `icon` and `clearIcon`. This feature changes the way how we handle the icons, so it will be treated as `leftIcon` and `rightIcon`. However, `icon` and `clearIcon` still work, but they are deprecated.

Based on #1705


belgamo:
#2001

Trancever:
Could you provide me an example of how Material Design searchbar handles the loading state? I've checked the guidelines and it looks like they show loading indicator under the searchbar, in the autocompletion zone.

Before we merge this PR we want to make sure this is something that MD really recommends. If they don't, then users will have to build it themselves.

belgamo:
> @belgamo Could you provide me an example of how Material Design searchbar handles the loading state? I've checked the guidelines and it looks like they show loading indicator under the searchbar, in the autocompletion zone.
> 
> Before we merge this PR we want to make sure this is something that MD really recommends. If they don't, then users will have to build it themselves.

I don't have an example, but I think the icons should be customizable. The user can use any icon as a "clear icon", so, it's better to have it as "right icon" or "end icon", for example. The button with loading state that the `react-native-paper` have implemented is not documented in guidelines as well.

brunohkbx:
@belgamo can you please rebase this one?

Trancever:
@belgamo I am not sure if Paper needs this change. It will not only add more complexity to the component but probably not many users will benefit cause it's not how this component should be used. According to MD Guidelines, search indicators should be placed under the searchbar and that's what Google's apps do as well.

On the other hand, the implementation of the component is not very complex so if you really need this change in your project I would suggest using patch-package.

Anyway, thanks for your contribution, it's still very valuable even tho we don't merge the PR 👍 🙏

PR #1778, opened 2020-04-03
fix(radiobuttonitem): added label theming support (#1587)

### LINKED ISSUE #1587: RadioButton.Item text is not styleable. RadioButtonContextProvider is not exported.
## Describe the feature
1. In RadioButtonItem text component is not styleable, and theme values are not applied on that Text.
Exact line (ack/react-native-paper/blob/master/src/components/RadioButton/RadioButtonItem.tsx#L79). Text from 'react-native-paper' should be used, not from 'react-native'.

2. RadioButtonContext is not exported so it's hard to use RadioButton.Android on iOS.

## Motivation

1. RadioButtons looks bad with dark theme - background is dark and text is black.
2. RadioButton.Android cannot be used on iOS.

halfmatthalfcat:
# Motivation

Add preliminary theming support for the RadioButtonItem label field to support Dark mode.

Closes #1587 

### Test plan

* Enable dark mode and view the label text (should match primary dark color)

### DIFF
--- a/src/components/RadioButton/RadioButtonItem.tsx
+++ b/src/components/RadioButton/RadioButtonItem.tsx
@@ -6,6 +6,8 @@ import {
   ViewStyle,
   TextStyle,
 } from 'react-native';
+import { Theme } from '../../types';
+import { withTheme } from '../../core/theming';
 import { RadioButtonContext, RadioButtonContextType } from './RadioButtonGroup';
 import { handlePress } from './utils';
 import TouchableRipple from '../TouchableRipple';
@@ -37,6 +39,10 @@ export type Props = {
    * Style that is passed to Label element.
    */
   labelStyle?: StyleProp<TextStyle>;
+  /**
+   * @optional
+   */
+  theme: Theme;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


src/components/RadioButton/RadioButtonItem.tsx
```diff
@@ -87,7 +101,11 @@ class RadioButtonItem extends React.Component<Props> {
               }
             >
               <View style={[styles.container, style]} pointerEvents="none">
-                <Text style={[styles.label, labelStyle]}>{label}</Text>
+                <Text
+                  style={[styles.label, labelStyle, { color: colors.primary }]}
```
dimbslmh, line 105: @halfmatthalfcat `style={[styles.label, labelStyle, { color: colors.primary }]} `
Prevents us from changing the label color.

Should be: 
`style={[styles.label, { color: colors.primary }, labelStyle]}`

Trancever: You can change label color using theme prop

dimbslmh: @Trancever Ah yes.. for some reason (?), the theme prop `{ colors }` is used to color the label.

`theme={{colors: { primary: colors.text } }}` or `theme={colors.primary = colors.text}`

Thank you.

Trancever: This is our primary way for customizing components

Issue #1597, opened 2020-01-21
Set dark or light theme according to device preferences by default


satya164:
React Native 0.62 will have an `Appearance` module which gives us the preferred color scheme of the device. We can detect and switch to dark theme automatically by default if it's preferred.

https://github.com/facebook/react-native/blob/a27e31c059971b1d554ad6c7c81706f08eafac87/RNTester/js/examples/Appearance/AppearanceExample.js#L14

It'll be a breaking change due to React Native version requirement, so we might want to wait a bit and drop support for older versions of React Native when we release it.

troZee:
Please don't close it :p

troZee:
I close this issue, because it has been implemented. 

cc @Trancever

PR #1780, opened 2020-04-03
feat: handle theme change based on preferences

troZee:
### Motivation
Closes https://github.com/callstack/react-native-paper/issues/1597

Android 10 and iOS 13 added support for dark theme. User (like me ;p) can set dark theme as default theme. Application should adapt to this theme automatically, hence I would like to have automatically theme changing in my app, so I don't need to care about user preferences. This is only a draft, because RNP needs to be bumped to 0.62. Example will be add later, once expo adds support to 0.62. 

### Test plan
* Check, it theme changes based on user preferences (it should be done automatically)
* You cannot test it due above reasons. You can use my private fork to test it https://github.com/troZee/paperExample


Trancever:
@troZee There are CI checks failing but besides that the PR looks really good 🔥

troZee:
> @troZee There are CI checks failing but besides that the PR looks really good 🔥

Should I bump RN version to 0.62 in this PR ? Or should I wait for RNP with rn 0.62 ?

troZee:
Closed in favor of https://github.com/callstack/react-native-paper/pull/2016

PR #1787, opened 2020-04-06
fix: pass useNativeDriver to all animations

Trancever:
### Motivation

Running animations without explicitly passing `useNativeDriver` property on React-Native > 0.60 is causing warnings.

### Test plan

Run the example app and check if animations work as expected.

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -151,6 +151,7 @@ class Banner extends React.Component<Props, State> {
     Animated.timing(this.state.position, {
       duration: 250 * scale,
       toValue: 1,
+      useNativeDriver: false,
     }).start();
   };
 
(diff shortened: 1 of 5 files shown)


src/components/Banner.tsx
```diff
@@ -151,6 +151,7 @@ class Banner extends React.Component<Props, State> {
     Animated.timing(this.state.position, {
       duration: 250 * scale,
       toValue: 1,
+      useNativeDriver: false,
```
jayu, line 154: If this doesn't work with the native drivers, you may try storing height as an animated value. This is only a suggestion, but it feels that banner position is animated in JS

Trancever: Yes it is, we were not able to achieve the desired behaviour with animating translation only.

kabus202:
Can you also include CheckboxAndroid.tsx?

Trancever:
@kabus202 Yeah, will do it. Thanks for pointing it!

PR #1809, opened 2020-04-10
fix: card title with 1 line only

IngenieroLata:
### Motivation
Card titles should be always **1** line. Per https://github.com/callstack/react-native-paper/pull/1576 discussion.

card

### DIFF
--- a/src/components/Card/CardTitle.tsx
+++ b/src/components/Card/CardTitle.tsx
@@ -22,6 +22,10 @@ type Props = React.ComponentProps<typeof View> & {
    */
   titleStyle?: StyleProp<TextStyle>;
   /**
+   * Number of line for the title.
+   */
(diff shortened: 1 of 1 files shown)


Trancever:
It shouldn't be configurable. We should always pass 1 to Title. https://github.com/callstack/react-native-paper/pull/1576 introduced the bug.

IngenieroLata:
Ok, I'll default it to 1 then.

jaulz:
@IngenieroLata I agree to @Trancever and it should not be configurable.

IngenieroLata:
Right!  roger that

IngenieroLata:
@jaulz Done! and just updated Motivation for future revisions

Trancever:
Thanks!

PR #1817, opened 2020-04-12
Feat: Avatar.Image can receive any props to the image

kopax:
### Motivation

I wanted to pass `defaultSource` and `resizeMode` but I was not able to.

This will fix the issue

### DIFF
--- a/src/components/Avatar/AvatarImage.tsx
+++ b/src/components/Avatar/AvatarImage.tsx
@@ -6,6 +6,7 @@ import {
   ViewStyle,
   StyleProp,
   ImageSourcePropType,
+  ImageProps,
 } from 'react-native';
 import { withTheme } from '../../core/theming';
 import { Theme } from '../../types';
@@ -18,9 +19,9 @@ type Props = {
    */
   source: ImageSourcePropType;
   /**
-   * Pass any props to the `Image`
+   * Pass any image props to the `Image`
    */
-  imageProps?: object;
+  imageProps?: ImageProps;
   /**
    * Size of the avatar.
    */
--- a/src/components/Avatar/AvatarImage.tsx
+++ b/src/components/Avatar/AvatarImage.tsx
@@ -18,6 +18,10 @@ type Props = {
    */
   source: ImageSourcePropType;
   /**
+   * Pass any props to the `Image`
+   */
+  imageProps?: object;
+  /**
    * Size of the avatar.
    */
   size?: number;
@@ -55,7 +59,7 @@ class AvatarImage extends React.Component<Props> {
   };
 
   render() {
-    const { size = defaultSize, source, style, theme } = this.props;
+    const { size = defaultSize, source, style, theme, imageProps } = this.props;
     const { colors } = theme;
 
     const { backgroundColor = colors.primary } =
@@ -76,6 +80,7 @@ class AvatarImage extends React.Component<Props> {
         <Image
           source={source}
           style={{ width: size, height: size, borderRadius: size / 2 }}
+          {...imageProps}
         />
       </View>
     );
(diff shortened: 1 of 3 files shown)


kamui545:
Glad I saw this, I was about to send a PR.
Same use case, I want to set `defaultSource`

src/components/Avatar/AvatarImage.tsx
```diff
@@ -17,6 +17,10 @@ type Props = {
    * Image to display for the `Avatar`.
    */
   source: ImageSourcePropType;
+  /**
+   * Pass any props to the `Image`
+   */
+  imageProps?: object;
```
andreialecu, line 23: Not great to have `object` here. should probably be the `ImageProps` type from react native.

kopax: I am not using typescript but based on : https://github.com/wcandillon/react-native-expo-image-cache/issues/40#issuecomment-444098997 Does it means, replacing object with:

```js
import { ImageProps } from 'react-native';
```
or
```js
import { ImageStyle } from 'react-native';
```

My IDE (intellij) does not allow to import TS source from non typescript project, I just want make sure I am not adding a bug.

kamui545: yup, ImageProps

src/components/Avatar/AvatarImage.tsx
```diff
@@ -17,6 +18,10 @@ type Props = {
    * Image to display for the `Avatar`.
    */
   source: ImageSourcePropType;
+  /**
+   * Pass any image props to the `Image`
+   */
+  imageProps?: ImageProps;
```
Trancever, line 24: I would rather see it implemented in the following way:

```
type Props = React.ComponentProps<typeof Image> & {
    // props
}
```

And then we can get `rest` from destructuring and pass it to the `Image` component.

We follow this pattern in other components so would be awesome if we could also use it here.

Trancever:
@kopax Let us know if you have a capacity to work on this or someone else should finish it, thanks!

kopax:
The problem is that I am not using TypeScript myself so even a tiny change is just speculation.

It's better if someone with TS experience takes care of it.

Trancever:
@kamui545 Do you want to fix the remaining issues so we can proceed and merge the PR?

kamui545:
Sure, however I am confused about one of your comment regarding the implementation.

> I would rather see it implemented in the following way:
> 
> ```
> type Props = React.ComponentProps<typeof Image> & {
>     // props
> }
> ```
> 
> And then we can get `rest` props from destructuring and pass it to the `Image` component.
> 
> We follow this pattern in other components so would be awesome if we could also use it here.

`rest` props are already destructured and used on the `View` component that is wrapping the `Image` component.

Are you suggesting we should use `rest` on the `Image` component instead? However, that would not be backward compatible and a breaking change?

Trancever:
@kamui545 Why do you think `rest` is already used? I am looking at the diff and I can't really see it :/

kamui545:
> @kamui545 Why do you think `rest` is already used? I am looking at the diff and I can't really see it :/

PR is too old, please check master: https://github.com/callstack/react-native-paper/blob/53c8e7a255ede4b15f23e4953518525697dd8dea/src/components/Avatar/AvatarImage.tsx#L58

kopax:
If this is solved it can be safely closed. I am still on react-native-paper 3.10.1

kamui545:
> If this is solved it can be safely closed. I am still on react-native-paper 3.10.1

It is not... `rest` is applied to the container, not the image

Trancever:
@kamui545 Please create another PR that is up to date with master.

kamui545:
> @kamui545 Please create another PR that is up to date with master.

Done https://github.com/callstack/react-native-paper/pull/2168

PR #1818, opened 2020-04-12
fix Portal.tsx example snack

joshglenn:
Made changes to example snack so that it works again

Example snack was giving the error:

```
Looks like you forgot to wrap your root component with `Provider` component from `react-native-paper`.
```

Solution: Wrapped the root component with Provider and added a couple of react-native components in the normal flow for contrast.


Trancever:
@joshglenn Could you fix lint issues? This code snippet you updated must be written in the same way as prettier would do it :D You can run lint locally to check if it passes.

PR #1822, opened 2020-04-13
feat: add wrapInSafeAreaView prop in Modal

ahce:
### Motivation

This prop is necessary when you need to see content in full screen on iPhone X/11

### Test plan
1. Download this branch
2. Open Modal example
3. Change the prop wrapInSafeArea (iPhone X/11)

### Demo
1. Clone this [repo](https://github.com/ahce/rn-paper-modal-fix)
2. Run: "yarn install"
3. Run: "yarn ios"
4. To apply fix files, run: "yarn patch" ("yarn unpatch" to undo)

### Actual behaviour
image

### Behaviour with  wrapInSafeArea set to false
image


kopax:
I would suggest to do the ternary in the constructor and set the `Wrapper` in the state to prevent recreating the Wrapper (and rerendering what's below) on each render. This will drastically improve your commit.

ahce:
> I would suggest to do the ternary in the constructor and set the `Wrapper` in the state to prevent recreating the Wrapper (and rerendering what's below) on each render. This will drastically improve your commit.

Wrapper is recreating when wrapInSafeArea is changed.
 I don't understand the suggestion, can you provide an example.

kopax:
> > I would suggest to do the ternary in the constructor and set the `Wrapper` in the state to prevent recreating the Wrapper (and rerendering what's below) on each render. This will drastically improve your commit.
> 
> Wrapper is recreating when wrapInSafeArea is changed.
> I don't understand the suggestion, can you provide an example.

    const Wrapper = wrapInSafeArea ? SafeAreaView : View;

If any other `props` than `wrapInSafeArea` change, this line will change, causing the whole view to rerender. 

That's why you avoid to do such kind of computation within the `render` method, if you're component is allowing props change necessary for it's rendering other than this one.

ahce:
> > Wrapper is recreating when wrapInSafeArea is changed.
> > I don't understand the suggestion, can you provide an example.
> 
> ```
> const Wrapper = wrapInSafeArea ? SafeAreaView : View;
> ```
> 
> If any other `props` than `wrapInSafeArea` change, this line will change, causing the whole view to rerender.
> 
> That's why you avoid to do such kind of computation within the `render` method, if you're component is allowing props change necessary for it's rendering other than this one.

I still don't understand.

These codes are the same:

 ```
import { View } from 'react-native'

render() {
    return <View />
}
 ```
 ```
import { SafeAreaView, View } from 'react-native'

render() {
    const Wrapper = wrapInSafeArea ? View : SafeAreaView;

    console.log(Wrapper === View) // result: true

    return <View />
}
 ```
It is a variable assignment. 
It is not recreate, if wrapInSafeArea doesn't change.

src/components/Modal.tsx
```diff
@@ -26,6 +27,10 @@ type Props = {
    * Determines Whether the modal is visible.
    */
   visible: boolean;
+  /**
+   * Determines Whether the modal is wrapped in a SafeAreaView.
+   */
+  wrapInSafeArea?: boolean;
```
Trancever, line 33: Can we rename that prop to `wrapInSafeAreaView`? If this is done I am ok with merging the PR.

ahce: > Can we rename that prop to `wrapInSafeAreaView`? If this is done I am ok with merging the PR.

Done!

ahce-meli:
@Trancever any chance to merge this

ahce:
@Trancever any chance to merge this?

pradel:
Would love to see this one merged, we have to use patch-package to be able to use it with a simple View instead of SafeAreaView which is creating issues with iPhones with a notch.

PR #1826, opened 2020-04-14
fix: useNativeDriver warning from CheckboxAndroid component

xiqi:
### Motivation

Fix useNativeDriver warning from CheckboxAndroid component. All other places are already fixed by #1796 but this one might have been forgotten.

### Test plan

Check whether useNativeDriver warning is going away from Checkbox component on Android.


welingtonfidelis:
Thanks =)

bachstatter:
Do you have a plan for when this will be released?

Issue #1728, opened 2020-03-11
Provide Checkbox.Item as a companion to RadioButton.Item


evantrimboli:
## Describe the feature

There should be a checkbox component that encompasses the checkbox itself and a label.

## Motivation

Checkboxes often have labels, one would expect tapping the label to change the checked state. It would also be consistent with the existing `RadioButton.Item`.

Bound3R:
and would also be super useful there was a `Checkbox.Group` like the `RadioButton.Group`.

evantrimboli:
For posterity, this was resolved in the linked PR.

https://github.com/callstack/react-native-paper/pull/1844

PR #1844, opened 2020-04-19
feat: create Checkbox.Item (#1728)

StheffanyHadlich:
### Motivation

This PR solves issue #1728  creating a `CheckboxItem` that contains the checkbox itself and a label. It is supposed to be consistent with `RadioButtonItem`. 

In the same issue was suggested an `CheckboxGroup` wich I added to this PR. I'm not sure if the group is supposed to provide any prop to its items, once the `Checkbox` component does not have a `value`, only a `status`. Currently the `CheckboxGroup` only serves to render its children.

Please let me know about any necessary changes


src/components/Checkbox/Checkbox.tsx
```diff
@@ -81,6 +82,9 @@ type Props = {
  * ```
  */
 class Checkbox extends React.Component<Props> {
+  // @component ./CheckboxGroup.tsx
+  static Group = CheckboxGroup;
```
Trancever, line 86: This should be removed

src/components/Checkbox/CheckboxGroup.tsx
```diff
@@ -0,0 +1,47 @@
...
+ *           <Checkbox status="unchecked" />
+ *         </View>
+ *       </Checkbox.Group>
+ *     )
+ *   }
+ * }
+ *```
+ */
+
+class CheckboxGroup extends React.Component<Props> {
```
Trancever, line 39: This should be removed. There is no need for this component, because Checkboxes shouldn't be grouped

src/components/Checkbox/CheckboxItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+ * import * as React from 'react';
+ * import { View } from 'react-native';
+ * import { Checkbox, Text } from 'react-native-paper';
+ *
+ * export default class MyComponent extends React.Component {
+ *
+ *   render() {
+ *     return(
+ *       <Checkbox.Group>
+ *           <Checkbox.Item label="First item" status="checked" />
```
Trancever, line 70: Just render View instead of Checkbox.Group

src/components/__tests__/Checkbox/CheckboxGroup.test.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Checkbox from '../../Checkbox/Checkbox';
+
+describe('CheckboxGroup', () => {
+  it('renders properly', () => {
+    const tree = renderer
+      .create(
+        <Checkbox.Group>
```
Trancever, line 9: We also don't need Group here

Trancever:
It looks good, however we don't need Checkbox.Group component so it would be awesome if you could remove it.

StheffanyHadlich:
> It looks good, however we don't need Checkbox.Group component so it would be awesome if you could remove it.

No problem, I'm sending the corrections now

src/components/Checkbox/CheckboxGroup.tsx
Bound3R: Hi, but checkboxes are usually represented as a group, I think that this feature should not be discarded, perhaps review it later comparing it with the motivation for which the RadioButton.Group was created :wink:

Trancever: @Bound3R The case with RadioButtons is different, cause only one radio button in the group can be selected and that's why we provide the RadioButton.Group component that handles all logic.

For Checkboxes, it doesn't make sense. What should that group component do?

Trancever:
@StheffanyHadlich I think the only missing part is to expose CheckboxItem as Checkbox.Item. This should be done in a same way as for Checkbox.Android and Checkbox.IOS.

PR #1855, opened 2020-04-22
Fix: offer a way to disable the back drop

### LINKED ISSUE #1853: Disable menu overlay and prevent focus being lost on searchinput
## Describe the feature

I would like to disable the backdrop transparent overlay when opening the menu

The feature would allow to passe style to the `absoluteFill` `<TouchableWithoutFeedback />` component, or use a props to disable totally it's render.

See here: https://github.com/callstack/react-native-paper/blob/4f592d203b3ac6fc1f2ec93c8c461358d83cfe99/src/components/Menu/Menu.tsx#L539

## Motivation

While working on an autocomplete, I am using both `<Searchbar />` and `< Menu />`

This is the code:

```jsx
/* eslint-disable */
import React, {
  useState,
  useCallback,
  useEffect,
  useRef,
} from 'react';
import {
  StyleSheet,
  View,
} from 'react-native';
import {
  Searchbar,
  withTheme,
  Menu,
} from 'react-native-paper';
import { throttle, debounce } from 'throttle-debounce';
import { connect } from 'react-redux';
import { compose } from 'redux';
import { createStructuredSelector } from 'reselect';
import withTranslations from '../../core/translations/withTranslations';
import { selectConsumptionTowns } from '../../store/consumptionZones/selectors';
import {
  changeCityOrPostalCode as changeCityOrPostalCodeAction,
} from '../../views/consumers/HomeView/actions';
import { selectCityOrPostalCode } from '../../views/consumers/HomeView/selectors';

function ConsumptionTownAutoComplete({
  theme: { colors },
  t,
  setLocale,
  locale,
  initialValue,
  onChange,
  choices = [],
  debounceDelay = 500,
  throttleDelay = 500,
  maxDebounceLength = 4,
  maxChoices = 5,
  style,
  ...rest
}) {
  const [value, setValue] = useState(initialValue);
  const [searches, setSearches] = useState([]);
  const searchbar = useRef();
  const [notFound, setNotFound] = useState(false);
  const [visible, setVisible] = useState(false);

  const autocompleteSearch = useCallback((query) => {
    const newSearches = choices.filter((choice) => choice.postalCode.includes(query) || choice.name.includes(query));
    setSearches(newSearches);
    setVisible(newSearches.length > 0);
    setNotFound(newSearches.length === 0);
    console.log(searchbar.current.focus);
    setTimeout(searchbar.current.focus);
  }, [choices]);

  const searchDebounced = useCallback(debounce(debounceDelay, autocompleteSearch), [debounceDelay, autocompleteSearch]);
  const searchThrottled = useCallback(throttle(throttleDelay, autocompleteSearch), [throttleDelay, autocompleteSearch]);

  useEffect(() => {
    if (value.length === 0) {
      setNotFound(false);
      setVisible(false);
    } else if (value.length <= maxDebounceLength) {
      searchDebounced(value);
    } else {
      searchThrottled(value);
    }
  }, [value]);

    container: {
      minWidth: 300,
      width: '50%',
      maxWidth: 500,
      position: 'relative',
    },
    or: {
      textTransform: 'uppercase',
      textAlign: 'center',
      color: colors.white,
    },
    searchBar: {
      zIndex: 2,
      ...(notFound ? {
        backgroundColor: colors.error,
      } : {}),
    },
    menuContent: {
      zIndex: 1,
      // position: 'absolute',
      top: 50,
      minWidth: 300,
      width: '50%',
      maxWidth: 500,
    },
  });

  return (
      <Menu
        contentStyle={styles.menuContent}
        visible={visible}
        onDismiss={() => setVisible(false)}
        anchor={
          <Searchbar
            ref={searchbar}
            placeholder={`${t('homeView.search')}...`}
            value={value}
            onChangeText={setValue}
            onSubmitEditing={(event) => console.log(value)}
            style={[styles.searchBar, style]}
            {...rest}
          />
        }
      >
        {value.length > 0 && searches.map((search, i) => i <= maxChoices && (
          <Menu.Item
            key={search.id}
            onPress={() => { setValue(search.name); setVisible(false); }}
            title={`${search.name} (${search.postalCode})`}
          />
        ))}
      </Menu>
    </View>
  );
}

const mapStateToProps = createStructuredSelector({
  initialValue: selectCityOrPostalCode,
  choices: selectConsumptionTowns,
});

const mapDispatchToProps = (dispatch) => ({
  onChange: (cityOrPostalCode) => dispatch(changeCityOrPostalCodeAction(cityOrPostalCode)),
});

export default compose(
  withTheme,
  withTranslations,
  connect(mapStateToProps, mapDispatchToProps),
)(ConsumptionTownAutoComplete);
```

Despite the fact that I use `searchbar.current.focus()`, I keep losing the focus while opening the menu.

I also need to click twice in order to get the focus.

I have checked into the documentation, and it seems that the backdrop cannot be removed with a props, which cause my autocomplete to be useless.

## Related Issues

None that I know

kopax:
Fix #1853

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -28,6 +28,6 @@
     "del-cli": "^3.0.0",
     "file-loader": "^4.2.0",
     "metro-react-native-babel-preset": "^0.56.0",
-    "url-loader": "^2.2.0"
+    "url-loader": "^4.1.0"
   }
 }
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -4599,6 +4599,15 @@ loader-utils@^1.4.0:
     emojis-list "^3.0.0"
     json5 "^1.0.1"
 
+loader-utils@^2.0.0:
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/loader-utils/-/loader-utils-2.0.0.tgz#e4cace5b816d425a166b5f097e10cd12b36064b0"
+  integrity sha512-rP4F0h2RaWSvPEkD7BLDFQnvSf+nK+wr3ESUjNTyAGobqrijmW92zc+SO6d4p4B1wh7+B/Jg1mkQe5NYUEHtHQ==
+  dependencies:
+    big.js "^5.2.2"
+    emojis-list "^3.0.0"
+    json5 "^2.1.2"
+
 locate-path@^2.0.0:
   version "2.0.0"
   resolved "https://registry.yarnpkg.com/locate-path/-/locate-path-2.0.0.tgz#2b568b265eec944c6d9c0de9c3dbbbca0354cd8e"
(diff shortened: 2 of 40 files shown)


maurusrv:
Hi. I was using `react-native-paper` and I encounter this same issue when i'm trying to display my search results on a `Menu` after using the `Searchbar`. I would really like to still be able to tap on the other parts of the screen even when the Menu is visible.

Any direction you can give me to somehow finish this commit? If that's even possible?

Thanks!

kopax:
I was in the exact same situation as you. RNP is not very convenient when usage is different than the stack of components offered and the callstack motivation is to keep it that way. Flexibility isn't a concern.

maurusrv:
@kopax i guess for me i just use lists instead of `Menu`. i guess, after some tinkering, the way i understood is Menu is meant to really have a backdrop, so the primary focus is to select an item from the menu. Anyway, thanks for the reply!

PR #1856, opened 2020-04-22
refactor(MenuItem): switch from class to function component

kopax:
And moving styles inside the inner body function, in order to give more fined control to the user.

This reduce the size, fit the react > 16.9 requirements, and gives more control to the end-user without changing anything.

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -47,108 +47,99 @@ type Props = {
  * A component to show a single list item inside a Menu.
  *
  */
+function MenuItem({
+  icon,
+  title,
+  disabled,
+  onPress,
+  theme,
+  style,
+  testID,
+  titleStyle,
+  minWidth = 112,
+  maxWidth = 280,
+  iconWidth = 40,
+){
+  const styles = StyleSheet.create({
+    container: {
+      paddingHorizontal: 8,
+      minWidth,
+      maxWidth,
+      height: 48,
+      justifyContent: 'center',
+    },
+    row: {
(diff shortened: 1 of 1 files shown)


src/components/Menu/MenuItem.tsx
```diff
@@ -47,109 +47,100 @@ type Props = {
...
+function MenuItem({
+  icon,
+  title,
+  disabled,
+  onPress,
+  theme,
+  style,
+  testID,
+  titleStyle,
+  minWidth = 112,
```
Trancever, line 59: What are these values? Is it mentioned in Material Design Guidelines somewhere?

kopax: Not really, it also follow the rule, which can do the more can do the less.

The reason why RNP is refussing change is to prevent maintainance due to more code written, this does not go into that case, it is refactored, less lines of code, and no feature added/removed, it just provide a way for the user to set one thing more. Is it also out of the scope to permit that?

Trancever:
@kopax Btw, what's your use case? I don't really see any reason for this change.

kopax:
I wrote an `AutoComplete` for the web with unpredictable width for my choices. The minWidth was somehow too small for the default.

The reason I changed this is that RNP needs also to switch at some point the class component to function components. And this is just a move forward and I was on the code. If you don't want to move forward now, it is fine to close that pull request.

Trancever:
@kopax Thanks for the repsonse 👍

kopax:
You are welcome, thanks for taking care and not closing too fast, that's a good way of handling contributions.

Issue #1694, opened 2020-02-23
Export Icon component


hamidfzm:
## Describe the feature
Many react-native paper components use the Icon component but it is not exported.

## Motivation
Customizing react-native paper components using the Icon component which uses the SettingsConsumer.

PR #1863, opened 2020-04-25
feat: export icon (#1694)

StheffanyHadlich:
### Motivation

This PR solves #1694, exporting `Icon` component in order to allow users to customize react-native-paper components.

Please let me know about any necessary changes

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -28,6 +28,6 @@
     "del-cli": "^3.0.0",
     "file-loader": "^4.2.0",
     "metro-react-native-babel-preset": "^0.56.0",
-    "url-loader": "^2.2.0"
+    "url-loader": "^4.1.0"
   }
 }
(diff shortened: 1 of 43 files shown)


satya164:
Why wouldn't you use `MaterialCommunityIcons` or whatever icon you want directly?

Trancever:
Please use Vector Icon directly. paper's Icon is an internal component that we don't want to expose.

RichardLindhout:
@satya164 @Trancever This can be actually difficult sometimes e.g. I'm now building a tabs component for react-native-paper and I would want the react-native-paper abstraction since it has already things inside of it like the render function e.t.c. expo/web abstractions.

If I go with the MaterialCommunityIcons directly it would not work with expo

RichardLindhout:
Actually it's less hard now since __expo dependency has dropped. Sorry for bothering ;)

PR #1865, opened 2020-04-26
fix: add wrapperStyle to Snackbar

jaulz:
# Motivation

The PR adds a new property to the `Snackbar` in order to overwrite these styles:
https://github.com/callstack/react-native-paper/blob/54254134bd61c5bd259f47ef7ebc9f5e29e14405/src/components/Snackbar.tsx#L289-L293

### Test plan

n/a

### DIFF
--- a/src/components/Snackbar.tsx
+++ b/src/components/Snackbar.tsx
@@ -234,7 +234,10 @@ class Snackbar extends React.Component<Props, State> {
     }
 
     return (
-      <SafeAreaView pointerEvents="box-none" style={[styles.wrapper, wrapperStyle]}>
+      <SafeAreaView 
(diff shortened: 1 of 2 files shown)


Trancever:
@jaulz What's your use case?

jaulz:
@Trancever I need to calculate the height of the `Snackbar` and thus it needs to be positioned relatively.

Trancever:
@jaulz Can't you just pass `onLayout`? It's passed to the underlying `Surface` component so It should return the correct dimensions and position.

jaulz:
@Trancever unfortunately not as I use a custom container that calculates the layout and triggers some other effects so I think this is the only possibility for me...

PR #1875, opened 2020-05-01
Add origin to onStateChange.

omerman:
Added Ability to know if fab state changed via action click or backdrop click.


Trancever:
@omerman What's your use case?

omerman:
@Trancever 
I wan't to change my app state when a click was made on the backdrop,
and it is different from when a close is originated via click on an action.

Thinking about it even further.. I don't see the need for onStateChange..
why not just 

onPress -> for when pressing the main FAB..
onPress -> for each child action.
onBackdropPress -> for when pressing the backdrop.

it will give much more control, the state should be fully managed by the containing component.
shouldnt it?

Issue #1781, opened 2020-04-03
After upgrading to React Native 0.62 a warning is showing


yankovalera:
## Environment
React Native 0.62
react-native-paper 3.7.0

## Description
After upgrading my app to React Native 0.62 I'm getting a warning 

%s: Calling `getNode()` on the ref of an Animated component is no longer necessary. You can now directly use the ref instead. This method will be removed in a future release. ReactNativeFiberHostComponent

I've tracked down and seems to be a problem with react-native-safe-area-view dependency

## Reproducible Demo
- Create a new RN 0.62 app
- Add react-native-paper
- Run the app, the warning shows.

pan-pawel:
@yankovalera Thanks for your message!
Can you provide more info?
I tried to install clean `0.62 `RN with `3.7.0` react-native-paper
I used modal and bottom-navigation (they both uses safe-area-view)
And i cannot see warning.
Also, i checked that paper in 3.7.0 uses safe-area-view 
https://github.com/callstack/react-native-paper/blob/73814f8bb4b0e75bafe5ad5a088962e3d6ef0480/package.json#L50

and safe-area-view with this version contain if for `getNode()`
https://github.com/react-native-community/react-native-safe-area-view/blob/84749ae2fc6ffefa2933d2845efa43ced0bf6cfc/src/index.tsx#L108-L111

Can you check what version of safe area view you have? also can you check how safe area view index.js in your node_modules looks like? for me it's
`const node = typeof this.view.measureInWindow === 'function' ? this.view : this.view.getNode();`

![Screenshot from 2020-04-19 19-36-25](https://user-images.githubusercontent.com/3886886/79696342-6927e800-827c-11ea-90a3-d681568543e1.png)
Modal:
![Screenshot from 2020-04-19 19-36-34](https://user-images.githubusercontent.com/3886886/79696343-6a591500-827c-11ea-816f-d9a5d74fc024.png)
Bottom Navigation
![Screenshot from 2020-04-19 19-36-44](https://user-images.githubusercontent.com/3886886/79696344-6a591500-827c-11ea-9660-8fe90c31491b.png)

yankovalera:
Hi @pan-pawel, it's using react-native-safe-area-view "0.14.8". The warning shows in the screen that I'm using the createMaterialBottomTabNavigator 

Definitely index.js looks different, mine says

```this.view.getNode().measureInWindow((winX, winY, winWidth, winHeight) => ````

[...]

PR #1884, opened 2020-05-04
chore: upgrade react-native-safe-area-view

matkoson:
### Motivation

#1781

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.5",
     "color": "^3.1.2",
-    "react-native-safe-area-view": "^0.14.6"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.5",
     "color": "^3.1.2",
-    "react-native-safe-area-view": "^0.14.6"
+    "react-native-safe-area-view": "0.14.9"
```
satya164, line 50: We should use `^` instead of hardcoding to 0.14.9

benoitdion, line 50: where does 0.14.9 come from? I don't see any branch or commit at https://github.com/react-native-community/react-native-safe-area-view referencing it.

Issue #1866, opened 2020-04-26
Menu in header was not closing outside of the header


gjgit:
## Description
Used Expo project with tabs template and include `<Menu/>` inside the header options and all worked perfectly by can't able to close the menu when clicking outside the header, it works well inside the header ., take a look at the GIF below

## **My code**

`navigation.setOptions({
    header: ({ color, size }) => {
      return (
        <Appbar.Header
          statusBarHeight={15}
          elevation={0}
          style={styles.surface}
        >
          <Appbar.Content title={count} />
          <Provider>
            <Menu
              visible={count}
              onDismiss={_closeMenu}
              anchor={<Button onPress={_openMenu}>Show menu</Button>}
            >
              <Menu.Item onPress={_closeMenu} title="Item 1" />
              <Menu.Item title="Item 2" />
              <Divider />
              <Menu.Item title="Item 3" />
            </Menu>
          </Provider>
        </Appbar.Header>
      );
    },
  });`
## Reproducible Demo
![ezgif-5-3ca7a923e45f](https://user-images.githubusercontent.com/2841150/80322080-1f24a080-8840-11ea-9fdb-c3b2f9aafb6f.gif)

PR #1885, opened 2020-05-04
feat: add renderBackdrop prop to menu to allow a custom backdrop component 

MichaelHettmer:
### Motivation

This PR enables the user of the library to provide a custom backdrop background for the opened menu component to e.g. provide a darker background color as seen in the screen recording below. 

This also enables the usage of a different touchable component which is necessary because using the menu in combination with a touchable or handler from react-native-gesture-handler is not working correctly. This may be related to #1866 as well.

Currently the usage of e.g. TouchableOpacity from react-native-gesture-handler as the anchor of the menu component prevents the dismiss of the menu if the user presses the default backdrop component above the TouchableOpacity **(only on android).** The react-native-gesture-handler component simply steals the event from the backdrop, which is implemented with a TouchableWithoutFeedback from react-native.

Now with renderBackdrop it is possible to use the default TouchableWithoutFeedback from react-native on iOS and the TouchableWithoutFeedback from react-native-gesture-handler on Android.

### Test plan

See ExampleMenu screen of example application. There is an additional list item now opening the menu component with a custom backdrop.

custom_backdrop


MichaelHettmer:
@Trancever still relevant

alessiocancian:
I'm having the backdrop issue with react-native-gesture-handler, this PR would really help, why it wasn't merged? @Trancever

Issue #1891, opened 2020-05-05
Add color prop to RadioButton.Item


Remagpie:
## Describe the feature

Add `color` prop to `RadioButton.Item` for color customization.

## Motivation

Other similar components like `CheckBox.Item` or simple `RadioButton` can receive `color` prop for customizing the color of the button.
It'd be nice to have this prop in `RadioButton.Item` too.

PR #1892, opened 2020-05-05
feat: add color and uncheckedColor prop to RadioButton.Item

Remagpie:
### Motivation

#1891 

It was quite a simple task so I made a PR immediately.

### Test plan


Trancever:
Thanks!

PR #1894, opened 2020-05-05
fix: use same color and size for button text and icon

### LINKED ISSUE #1725: Set icon color on a <Button />

Trancever:
### Motivation

Fix #1725 

### Test plan

Screenshot 2020-05-05 at 17 05 18
Screenshot 2020-05-05 at 17 05 48

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -245,6 +245,10 @@ class Button extends React.Component<Props, State> {
         ? StyleSheet.flatten(style).borderRadius || roundness
         : roundness,
     };
+
+    const { color: customLabelColor, fontSize: customLabelSize } =
+      StyleSheet.flatten(labelStyle) || {};
+
     const textStyle = { color: textColor, ...font };
     const elevation =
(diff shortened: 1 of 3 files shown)


src/components/Button.tsx
```diff
@@ -279,13 +283,17 @@ class Button extends React.Component<Props, State> {
           <View style={[styles.content, contentStyle]}>
             {icon && loading !== true ? (
               <View style={styles.icon}>
```
satya164, line 285: Maybe we should also add `width` here like before?

Trancever: I was thinking about it. I decided to keep it as it is now, cause I couldn't break it :D I am gonna try it one more time to make sure it works properly.

PR #1895, opened 2020-05-05
docs: add useTheme example

M4gie:
Add an example for the useTheme hook

### Where to test

In the `Theming` page, in the `Using the theme in your own components` section ([here](https://17506-71323749-gh.circle-artifacts.com/0/docs/theming.html#using-the-theme-in-your-own-components))


Trancever:
Thanks!

PR #1905, opened 2020-05-08
Updated secureTextEntry usage in docs of TextInput

khushal87:
### Motivation

Updated the doc to refer for secureTextEntry props which allow hiding the password while typing


Trancever:
@khushal87 Hey! Thanks for your contribution. Our TextInput extends the one from React Native and we have a link to all the props in the documentation at the end of "Props" section. We don't see much value in duplicating props definition and that's why we won't accept this PR.

Issue #1872, opened 2020-04-30
Drawer.Item label color


matulkum:
## Describe the feature

Right now there is no option to fully customise the color of the label. You can set the text color trough the themes text color, but unfortunately the Drawer.item will always apply an opacity of 0.68 to the text color.

## Motivation

I want to colorise the label with an opacity of 1.

Thanks.

PR #1908, opened 2020-05-10
feat: add labelOpacity prop to Drawer.Item component

StheffanyHadlich:
### Motivation

This PR solves #1872 , adding a `labelOpacity` prop to the `Drawer.Item` component. This prop is optional and if the user doesn't pass it to the component it is just going to use its default value (`0.68`).


src/components/Drawer/DrawerItem.tsx
```diff
@@ -74,7 +79,7 @@ class DrawerItem extends React.Component<Props> {
     const contentColor = active
       ? colors.primary
       : color(colors.text)
-          .alpha(0.68)
+          .alpha(labelOpacity || 0.68)
```
wojteg1337, line 82: Could you use default value for labelOpacity instead of logical OR operator?

Trancever:
@StheffanyHadlich Hey, thanks for your contribution! Are you going to apply the change that @wojteg1337 requested? If that is done we will merge the PR.

src/components/Drawer/DrawerItem.tsx
StheffanyHadlich: For sure, I'll fix it

Issue #1834, opened 2020-04-14
Add missing screenshots of the components in the documentation.


Trancever:
Couple components are missing the screenshots in the documentation. We should provide nice-looking screenshots that will improve the developer experience.

The first part of the ticket is to find out which components are missing screenshots.
The second part is to create them.

cmargieson:
- [x]  ActivityIndicator

- [x] Appbar
- [ ] Appbar.Action 
- [ ] Appbar.BackAction 
- [ ] Appbar.Content 
- [x] Appbar.Header

- [ ] Avatar
- [x] Avatar.Icon
- [x] Avatar.Image
- [x] Avatar.Text

- [ ] Badge 

- [x] Banner

- [x] BottomNavigation

- [x] Button

- [x] Card
- [ ] Card.Actions 
- [ ] Card.Content 
- [ ] Card.Cover 
- [x] Card.Title
 
- [x] Checkbox
- [x] Checkbox.Android
- [x] Checkbox.IOS

- [x] Chip

- [x] DataTable
- [ ] DataTable.Cell
- [ ] DataTable.Header
- [ ] DataTable.Pagination
- [ ] DataTable.Row
- [ ] DataTable.Title

- [x] Dialog
- [ ] Dialog.Actions 
- [ ] Dialog.Content 
- [ ] Dialog.ScrollArea 
- [ ] Dialog.Title 

- [ ] Divider 

- [ ] Drawer 
- [ ] Drawer.Item 
- [ ] Drawer.Section 

- [x] Fab
- [x] FAB.Group

- [x] HelperText

- [x] IconButton

- [ ] TouchableRipple 

- [ ] List 
- [x] List.Accordion
- [ ] List.AccordionGroup 
- [ ] List.Icon 
- [x] List.Item
- [x] List.Section
- [x] List.Subheader

- [x] Menu
- [ ] Menu.Item 

- [ ] Modal 

- [ ] Portal 
- [ ] Portal.Host 

- [x] ProgressBar

- [x] RadioButton
- [x] RadioButton.Android
- [ ] RadioButton.Group 
- [x] RadioButton.IOS
- [ ] RadioButton.Item 

- [x] Searchbar

- [x] Snackbar

- [x] Surface

- [x] Switch

- [x] TextInput

- [x] ToggleButton
- [ ] ToggleButton.Group 
- [ ] ToggleButton.Row 

- [ ] Typography 
- [ ] Text 
- [x] Subheading
- [x] Title
[...]

PR #1909, opened 2020-05-11
Feat/add missing screenshots

matkoson:
### Motivation
#1834


Trancever:
1. `card-action-example` can be removed cause there is a `card-actions` which looks much better and we can reuse it.
2. `dialog-scroll-area` The container should be higher. We need to make the component look beautiful.
3. `modal` Similarly to `dialog-scroll-area`. We should make it higher cause it doesn't look good as it is now.

matkoson:
@Trancever 
- `card-action-example` is replaced with `card-actions` and removed
- GIFs for `Modal` and `DialogScrollArea` are replaced

PR #1934, opened 2020-05-19
change animeted driver on components

vinicius0012:
some components had problems with the animation when selected (checkbox and radio button), the solution found was to change an attribute in the animation "useNativeDrive: false" to "useNativeDrive: true".

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -152,7 +152,7 @@ class Banner extends React.Component<Props, State> {
     Animated.timing(this.state.position, {
       duration: 250 * scale,
       toValue: 1,
-      useNativeDriver: false,
+      useNativeDriver: true,
     }).start();
   };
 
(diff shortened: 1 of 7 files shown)


Trancever:
@vinicius0012 Is it possible to animate height in RN using native driver now? Could you send a link to a PR which adds support for it or link to documentation?

vinicius0012:
it is possible, using the native drive in the new version, the animations are performed by the native, having a better performance.

https://reactnative.dev/docs/animated#__docusaurus.

Trancever:
@vinicius0012 https://reactnative.dev/docs/animations#caveats They claim animating layout properties is not supported by the native driver and this is what we do for RadioButton and Banner

satya164:
I think there was some PR which added it. Just make sure to test, and it's a breaking change, so needs to go in 4.0

haxonadora:
@vinicius0012 tested it on `react-native 0.62` that `react-native-paper` is currently using. Setting `useNativeDrive` to `true` won't allow us to animate width or height

Issue #893, opened 2019-03-06
NativeViewHierarchyManager: Unable to update properties for view tag error thrown in logcat android when using ActivityIndicator component from RNPaper


hvardhan617:
### Current behaviour

When using Activity Indicator component from React native paper, the below error is continuously occurring in Logcat when run on an Android emulator.

:NativeViewHierarchyManager: Unable to update properties for view tag 33
    com.facebook.react.uimanager.IllegalViewOperationException: ViewManager for tag 33 could not be found
        at com.facebook.react.uimanager.NativeViewHierarchyManager.resolveViewManager(NativeViewHierarchyManager.java:107)
        at com.facebook.react.uimanager.NativeViewHierarchyManager.updateProperties(NativeViewHierarchyManager.java:134)
        at com.facebook.react.uimanager.UIImplementation.synchronouslyUpdateViewOnUIThread(UIImplementation.java:338)
        at com.facebook.react.animated.PropsAnimatedNode.restoreDefaultValues(PropsAnimatedNode.java:76)
        at com.facebook.react.animated.NativeAnimatedNodesManager.restoreDefaultValues(NativeAnimatedNodesManager.java:337)
        at com.facebook.react.animated.NativeAnimatedModule$18.execute(NativeAnimatedModule.java:350)
        at com.facebook.react.animated.NativeAnimatedModule$2.execute(NativeAnimatedModule.java:141)
        at com.facebook.react.uimanager.UIViewOperationQueue$UIBlockOperation.execute(UIViewOperationQueue.java:583)
        at com.facebook.react.uimanager.UIViewOperationQueue$1.run(UIViewOperationQueue.java:917)
        at com.facebook.react.uimanager.UIViewOperationQueue.flushPendingBatches(UIViewOperationQueue.java:1025)
        at com.facebook.react.uimanager.UIViewOperationQueue.access$2600(UIViewOperationQueue.java:46)
        at com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.doFrameGuarded(UIViewOperationQueue.java:1085)
        at com.facebook.react.uimanager.GuardedFrameCallback.doFrame(GuardedFrameCallback.java:29)
        at com.facebook.react.modules.core.ReactChoreographer$ReactChoreographerDispatcher.doFrame(ReactChoreographer.java:166)
        at com.facebook.react.modules.core.ChoreographerCompat$FrameCallback$1.doFrame(ChoreographerCompat.java:110)
[...]

PR #1935, opened 2020-05-19
fix: activityindictor crash on android

runningdeveloper:
### Motivation

I had an issue with the activity indicator crashing on android. Happens when there is an ActivityIndicator on the first page. The same issue as this https://github.com/callstack/react-native-paper/issues/893

I followed what the discussion said and added ```collapsable={false}``` to the Animated view. And it seemed to fix the issue.

Reproducible error on this repo https://github.com/runningdeveloper/crash-test 

### Test plan

I am using expo and pixel 3a emulator. 
I had to update the snapshots on activityindicator and button. Loading on the button still looks normal.

Let me know if I have done anything incorrectly.


Trancever:
Thanks!

PR #1937, opened 2020-05-20
Feat: Apply suggestions on reduce motion PR

matkoson:
### Motivation

- @Trancever comments on  #1642.

### Test plan

- It's tested on iOS Simulator with `Reduce Motion` turned on in a11y menu, as well as on Android device with `Window animation scale` , `Transition animation scale` and `Animator duration scale` all turned off in `Developer options`.


example/src/Examples/TextInputExample.tsx
```diff
@@ -219,7 +219,7 @@ class TextInputExample extends React.Component<Props, State> {
               >
                 Error: Numbers and special characters are not allowed
               </HelperText>
-              <HelperText visible style={styles.counterHelper}>
+              <HelperText type="info" visible style={styles.counterHelper}>
```
matkoson, line 222: Looks like this props is required for `HelperText`

package.json
```diff
@@ -60,7 +60,7 @@
     "@types/node": "^12.0.4",
     "@types/reach__router": "^1.2.4",
     "@types/react-dom": "^16.8.4",
-    "@types/react-native": "^0.57.61",
+    "@types/react-native": "^0.61.4",
```
matkoson, line 63: According to [changelog](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md) `isReduceMotionEnabled` method and `reduceMotionChanged` event were added in rn `0.60`, so I guess some sort of type upgrade will be necessary. @Trancever

src/core/Provider.tsx
```diff
@@ -17,41 +17,46 @@ export default class Provider extends React.Component<Props> {
     reduceMotionEnabled: false,
   };
 
-  componentDidMount() {
+  async componentDidMount() {
     AccessibilityInfo.addEventListener(
-      'reduceMotionEnabled',
-      this._handleReduceMotionToggled
+      'reduceMotionChanged',
```
matkoson, line 22: [Docs](https://reactnative.dev/docs/accessibilityinfo) describe event controlling reduce motion settings change as `reduceMotionChanged`, so do types.

matkoson:
I can see those a11y-related changes were [reverted](https://github.com/callstack/react-native-paper/pull/1789/files) once already, so I guess the original PR will need to be eventually merged to 4.0

src/core/Provider.tsx
```diff
@@ -1,26 +1,62 @@
...
+      const reduceMotionEnabled = await AccessibilityInfo.isReduceMotionEnabled();
+
+      this.setState({ reduceMotionEnabled });
+    } catch (err) {
+      console.warn(err);
+    }
+  };
+
   render() {
+    const theme = this.props.theme
```
Trancever, line 46: Can we destructure theme from props? It's used on lines 46 and 52.

yarn.lock
```diff
@@ -1555,9 +1555,9 @@
     "@types/react" "*"
 
 "@types/react-native@^0.57.61":
-  version "0.57.62"
-  resolved "https://registry.yarnpkg.com/@types/react-native/-/react-native-0.57.62.tgz#cf08ec552a73a44fcea359da0a7529a964b0898f"
```
Trancever, line 1559: Is this version correct? Types are upgraded to 0.61 in package.json.

src/core/Provider.tsx
Trancever: We could also destructure settings and children

src/components/Chip.tsx
```diff
@@ -205,7 +205,7 @@ class Chip extends React.Component<Props, State> {
       : selectedBackgroundColor;
 
     const accessibilityTraits: AccessibilityTrait[] = ['button'];
-    const accessibilityStates: AccessibilityState[] = [];
+    const accessibilityStates: AccessibilityStates[] = [];
```
Trancever, line 208: I think this should be named `accessibilityState` and be an object - https://reactnative.dev/docs/accessibility#accessibilitystate-android-ios

yarn.lock
matkoson: Thx, now it should be fine

src/components/Chip.tsx
matkoson: Yep, you're right it's deprecated. Heh, I implemented changes, but it's already addressed in #1938

src/components/Badge.tsx
```diff
@@ -90,12 +96,13 @@ class Badge extends React.Component<Props, State> {
...
       children,
       size = defaultSize,
-      style,
       theme,
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       visible,
       ...rest
     } = this.props;
+    /** Type for this.props.style was somehow altered after upgrading to "@types/react-native": "^0.61.4", and was throwing some strange errors, the following is a workaround */
+    const style: StyleProp<ViewStyle> = this.props.style;
```
Trancever: There is a bug in this implementation. If we don't destructure `style` prop then it will override whole style prop when we pass `{...rest}` on line 132. Can we somehow handle it in a different way? 

Please remember to update snapshot tests after fixing this.

docs/pages/2.theming.md
```diff
@@ -72,6 +72,9 @@ A theme usually contains the following properties:
 
 When creating a custom theme, you will need to provide all of these properties.
 
+Keep in mind that your custom theme will need to manually handle animation scaling for users with
```
Trancever, line 75: Can we add an animation scale to the theme properties on line 72? I think it is missing.

docs/pages/2.theming.md
```diff
@@ -72,6 +72,9 @@ A theme usually contains the following properties:
...
+disabled interface animations(`Reduce Motion` on iOS, `Remove animations` on Android) - Paper's [default theme](https://github.com/callstack/react-native-paper/blob/master/src/styles/DefaultTheme.tsx) takes care of those settings out of the box.
```
Trancever, line 76: I would also rephrase this section about animation scale handling. I would prefer to first say that if you don't use custom theme then Paper will automatically turn on/off animations based on device settings. And then I would write that if you pass a custom theme then you have to handle it manually using AccessibilityInfo API (It can be a link to RN docs). I believe this will be easier for users to follow. What do you think?

matkoson: Oks, it's simplified.

src/components/Badge.tsx
matkoson: It should be fixed already.

Issue #1920, opened 2020-05-16
react-native 0.62.x removed accessibilityStates


firede:
## Environment

- react: 16.11.0
- react-native: 0.62.2
- react-native-paper: 3.10.1
- typescript: 3.9.2

## Description

> Remove accessibilityStates: We’ve [removed](https://github.com/facebook/react-native/commit/7b35f427fd66cb0f36921b992095fe5b3c14d8b9) the deprecated accessibilityStates property in favor of the new accessibilityState prop which is a more semantically rich way for components to describe information about their state to accessibility services.
> -- https://reactnative.dev/blog/#breaking-changes

![image](https://user-images.githubusercontent.com/157338/82108724-12baa600-9763-11ea-914c-2048f27478cc.png)

## Reproducible Demo

Code:

```tsx
import React from "react"
import { View } from "react-native"
import { Text } from "react-native-paper"

interface Props {}

const Component: React.FC<Props> = () => {
  return (
    <View>
      <Text>Component</Text>
    </View>
  )
}

export default Component
```

GoMino:
This is a real issue

numToStr:
I was about a start a project and faced this issue :cry: . Hope it will be fixed soon.

hb12DevTn:
I Hope that a fix will be published soon..

pan-pawel:
Hey, thanks for messages.
To solve this issue we need to upgrade types for react-native also but It's abreaking change so we will  publish this change in 4.0 since we have users with older versions of react than 0.6.

ianmartorell:
What's the timeline for the release of 4.0? Any chance of getting the updated types some other way if that's not happening very soon?

stage88:
This is not a fix.... however I've removed all occurrences of '"accessibilityStates" | ' in ./node_modules/react-native-paper/lib/typescript, which makes the annoying errors go away.

Don't follow my advice, I know nothing.

firede:
A way to **IGNORE** errors using `patch-package`:

```patch
diff --git a/node_modules/@types/react-native/index.d.ts b/node_modules/@types/react-native/index.d.ts
index 98681f1..825c536 100644
--- a/node_modules/@types/react-native/index.d.ts
+++ b/node_modules/@types/react-native/index.d.ts
@@ -2069,6 +2069,7 @@ export interface AccessibilityProps extends AccessibilityPropsAndroid, Accessibi
[...]

PR #1938, opened 2020-05-20
[4.0] change deprecated accessiblityStates to accessibilityState

pan-pawel:
`accessiblityStates` is removed in `0.62` and that causing typescript errors. With this PR I'm changing paper to use new `accessibilityState` also we want to update `@types/react-native` because  `accessibilityState` wasn't exist in `0.57.61`|
Issue: #1920 
Unfortunately it's a breaking change for people using older version of react-native and we want to add it to 4.0


wojteg1337:
@pan-pawel 
resolve conflicts so we can merge this guy

PR #1947, opened 2020-05-25
Feat: add action to check versions of library

jbinda:
This PR updates the issue template and put it in the ISSUE_TEMPLATE folder as GitHub expected. Also it provides action to check if versions of npm packages mentioned in issue is up to date (this is replication the logic lays in [react-navigation repo](https://github.com/react-navigation/react-navigation)

### Motivation

We need to make sure if issue posted as a bug is still existing after update to recent versions of npm packages.

### Test plan

@Trancever @pan-pawel  please check the code (there is not much of it). After merge I believe we might check if everything works as expected ( we might need to adjust from repo settings to set up template also for feature )


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,20 @@
...
+  - name: Example
+    url: https://snack.expo.io/@satya164/github.com-callstack-react-native-paper:example
+    about: See the official example app.
+  - name: Contributing guide
+    url: https://github.com/callstack/react-native-paper/blob/master/CONTRIBUTING.md
+    about: See our contributing guide.
+  - name: Discord channel
+    url: https://discord.com/channels/426714625279524876/438316503113138176
+    about: Join to our official discord channel.
+  - name: Expo
```
Trancever, line 18: Why do we have "Expo" here?

jbinda: done

PR #1948, opened 2020-05-25
docs: added lib name in title

marchenk0va:
### Motivation

The lib name in title is clear and recognisable out of all open tabs. With the lib name page would be more unique.

Screenshot 2020-05-25 at 12 01 30

### Test plan

Build and run docs


jbinda:
thanks for update @marchenk0va !

@Trancever I think its ready to merge :)

PR #1953, opened 2020-05-28
feat: add iconSize prop to Button component

### LINKED ISSUE #1317: Allow styling of text in Button and modification of Icon size
## Describe the feature

At the moment the text styling in Button is [hardcoded with a combination](https://github.com/callstack/react-native-paper/blob/master/src/components/Button.tsx#L272) of `styles. label` and `textStyle` neither of which are accessible outside the component. 

Also the `size` parameter passed to `Icon` is [fixed to 16](https://github.com/callstack/react-native-paper/blob/master/src/components/Button.tsx#L256) and cannot be changed.

## Motivation

I wish to use a button as a larger styling component and in order to make the appearance as I wish I would need to use a larger font and icon

## Related Issues

N/A

elizabeth-dev:
### Motivation

This allows to set a custom icon size for the Button component. Until now, icon size was linked to font size, but there may be cases where further customization is needed. As the loading indicator used the same font size approach, and is syntactically similar to an icon, it has been linked to `iconSize` value too.

I considered adding an `iconColor` prop with a similar approach, but I didn't know if that would be useful for anyone, and i'm not even sure if that would be Material-spec compliant. This can be discussed.

Although already closed, fixes #1317 remaining demands.

### Test plan

This is a really simple semver-minor change, extensive testing should not be necessary.


elizabeth-dev:
🙄

PR #1955, opened 2020-05-29
feat: add prop accessibilityLabel to RadioButtonItem

brunohkbx:
### Motivation
It's not possible to press on a RadioButtonItem using appium or other testing tools without passing an accessibilityLabel to the touchable.


Trancever:
@brunohkbx Thanks for your contribution!

PR #1963, opened 2020-06-03
feat: ability to use number as radio btn value

raajnadar:
### Summary

Ability to use Number in value prop of radio btn. Why?

Let's say I get data from an endpoint and the returned data is in this form.

```
{
  "data": {
    "project_types": [
      {
        "id": 1,
        "name": "Web Development"
      },
      {
        "id": 2,
        "name": "Web Designing"
      },
      {
        "id": 3,
        "name": "UX/UI Designing (Prototyping)"
      }
    ]
  }
}
```

Datatype of id is number & I can safely assume it is unique so I can directly loop through the data in render.

### Test plan

```
<View>
    <RadioButton
      value={1}
      status={checked === 1 ? 'checked' : 'unchecked'}
      onPress={() => { this.setState({ checked: 1 }); }}
    />
    <RadioButton
      value={2}
       status={checked === 2 ? 'checked' : 'unchecked'}
      onPress={() => { this.setState({ checked: 2 }); }}
    />
</View>
```

### DIFF
--- a/example/src/Examples/RadioButtonGroupExample.tsx
+++ b/example/src/Examples/RadioButtonGroupExample.tsx
@@ -9,8 +9,8 @@ import {
 } from 'react-native-paper';
 
 const RadioButtonGroupExample = () => {
-  const [value, setValue] = React.useState<string>('first');
-  const [value2, setValue2] = React.useState<string>('first');
+  const [value, setValue] = React.useState<string | number>('first');
+  const [value2, setValue2] = React.useState<string | number>('first');
 
   const {
     colors: { background, primary },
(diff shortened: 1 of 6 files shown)


Trancever:
Thanks for your contribution @raajnadar ! I think it should be resolved on your end by converting the id to string.

raajnadar:
It will be useful when you directly use the data from GaphQL query. To convert them to int we need an extra loop and store them in local state.

Trancever:
@raajnadar I don't understand it. You always have to loop over some array of values and render RadioButton for each

PR #1964, opened 2020-06-03
docs: paper logo favicon, upgraded components-docs to 0.23.0

marchenk0va:
### Summary

Added paper logo favicon by upgrading `component-docs` to version `0.23.0`, which includes favicon option.

Screenshot 2020-06-04 at 09 53 52

### Test plan

Build and run docs.


thymikee:
How about using a transparent one? :) It would look good with both light and dark themes

marchenk0va:
> How about using a transparent one? :) It would look good with both light and dark themes

As for me, transparent is hard to see on dark theme, what do you think?

Screenshot 2020-06-03 at 21 21 45

thymikee:
Use the one from https://reactnativepaper.com :) 
image

PR #1966, opened 2020-06-04
feat: Created react-navigation guide

haxonadora:


raajnadar:
Can you add an example how to replace left & right element using the Appbar from paper!

When we use header options in the root navigator and when we try to override it from a screen using the `navigation.screenOptions` we need to add all the components like Appbar header, content & action.

I tried using `headerLeft` & `headerRight` but we don't have a wrapper for `Appbar.Header`

eg: -

```
<Appbar.Header>
   {previous ? (
     <Appbar.BackAction onPress={navigation.goBack} />
   ) : (
     <Appbar.Content title="Home" />
   )}
  // Component like Menu from paper
</Appbar.Header>
```

Lets say In few page I don't want menu & I want to replace it with filter Icon, so in that case I need to override the whole header option.

Something like this in my child screen

```
<Appbar.Header>
   {previous ? (
     <Appbar.BackAction onPress={navigation.goBack} />
   ) : (
     <Appbar.Content title="Search user" />
   )}
  // Icon button Filter
</Appbar.Header>
```

docs/pages/8.react-navigation.md
```diff
@@ -0,0 +1,200 @@
...
+const style = StyleSheet.create({
+  container: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+});
+```
+
+Once we have that we can start our app and see what we have.
```
Trancever, line 96:
```suggestion
Once we have that we can run the app and check how Stack looks like.
```

docs/pages/8.react-navigation.md
```diff
@@ -0,0 +1,200 @@
...
+Our result:
+
+//gif
+
+Now we can navigate between two screens, but we are not done yet.
+
+
+### Adding AppBar
+
+We can customize `react-nativation` top bar passing other components:
```
Trancever, line 125:
```suggestion
We can customize Stack's header by passing custom component:
```

docs/pages/8.react-navigation.md
```diff
@@ -0,0 +1,200 @@
...
+        <Appbar.Content title="My awesome app" />
+      )}
+    </Appbar.Header>
+  );
+}
+```
+
+![]('../assets/screenshots/react-nativation-appBar3.gif')
+
+And that's we need. Now you can use Material design `AppBar` with `react-navigation` to ensure your app works and looks well.
```
Trancever, line 197:
```suggestion
And that's what we need. Now you can use Material design `AppBar` with `react-navigation` to ensure your app works and looks well.
```

docs/pages/8.react-navigation.md
```diff
@@ -0,0 +1,200 @@
...
+Now we can check if navigation have previous state. If it has we can render back arrow button, otherwise we will render app name as before.
+
+```js
+function CustomNavigationBar({ navigation, previous }) {
+  return (
+    <Appbar.Header>
+      {previous ? (
+        <Appbar.BackAction onPress={navigation.goBack} />
+      ) : (
+        <Appbar.Content title="My awesome app" />
```
Trancever, line 188: I believe Content should be always visible. It doesn't look good when the header has only back button

arafatzahan:
What is react-nativation ??? Is that a typo or I'm missing something?

thymikee:
@haxonadora mind resolve conflicts? :)

haxonadora:
moved to https://github.com/callstack/react-native-paper/pull/2248

PR #1969, opened 2020-06-04
feat: add numberOfLines prop to button

### LINKED ISSUE #1968: Add prop to control the number of lines visible for a button
For some of the buttons in my project, labels that are too long may sometimes get clipped with an ellipses. This is usually made more difficult by supporting multiple languages - a text size or similar for one language may not work for another.

I've noticed that the `numberOfLines` prop for the Text component inside the Button is hardcoded to be 1. It would be nice to be able to modify this with a prop on the Button component.

**Describe alternatives you've considered**
N/A

**Additional context**
Here is an example of long text in a button:
actual

And here is what it would look like if we had 2 lines instead of 1:
expected

salockhart:
### Summary
Closes #1968 

For some of the buttons in my project, labels that are too long may sometimes get clipped with an ellipses. This is usually made more difficult by supporting multiple languages - a text size or similar change for one language may not work for another. By adding a `numberOfLines` prop to the button, I can choose to make some buttons wrap in order to fit their contents.

### Test plan

1. In the example app, create a new button component with a long label, and pass the prop `numberOfLines={2}` or similar
2. Similarly, buttons that do not have the prop or have a value of `numberOfLines={1}` should appear unchanged.

example

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -78,6 +78,10 @@ type Props = React.ComponentProps<typeof Surface> & {
    */
   labelStyle?: StyleProp<TextStyle>;
   /**
+   * Maximum number of lines for the label.
+   */
+  numberOfLines?: number;
+  /**
    * @optional
    */
   theme: Theme;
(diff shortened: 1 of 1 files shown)


Trancever:
Thanks for your contribution! Unfortunately, MD Guidelines explicitly say that button text can't wrap and that's why we won't accept this PR.

dont wrap

Issue #1893, opened 2020-05-05
The Demo App is not Rendering Example Components


valoricDe:
## Environment

Browser: Firefox latest

## Description

The page https://callstack.github.io/react-native-paper/index.html links to an snack expo demo site. In the web view it gives following error:
![Screenshot 2020-05-05 at 11 27 58](https://user-images.githubusercontent.com/22564188/81058326-81c80d00-8ece-11ea-8722-1034052397fb.png)

## Reproducible Demo

https://snack.expo.io/@satya164/github.com-callstack-react-native-paper:example

Is this a know issue?

Vn-ChemGio:
using expo in IOs has error too,

mMajch:
The snack is outdated - it uses react-native-paper 3.3.0. I created a new one that uses the latest version and everything seems to work https://snack.expo.io/@mmajch/github.com-callstack-react-native-paper:example 

@satya164 I see that you created the current snack. You can export mine and update yours.

mMajch:
The snack is outdated - it uses react-native-paper 3.3.0. I created a new one that uses the latest version and everything seems to work https://snack.expo.io/@mmajch/github.com-callstack-react-native-paper:example

@satya164 I see that you created the current snack. You can export mine and update yours.

valoricDe:
https://github.com/callstack/react-native-paper/pull/1976/files is not merged yet. So yes please stay open

valoricDe:
@Trancever could you reopen this, and or merge this one-liner? https://github.com/callstack/react-native-paper/pull/1976

PR #1976, opened 2020-06-07
Update demo to use most recent npm packages

valoricDe:
This merge request is related to #1893 as @satya164 does not find the time to update the snack it might be easier to update the link in the docu


Trancever:
Fixed in https://github.com/callstack/react-native-paper/pull/2173

PR #1997, opened 2020-06-17
feat: add menu prop for external touch capture options

finnmerlett:
### Summary

Additional optional prop in Menu component, that allows touches outside the menu surface (when the menu is visible) to be ignored, or to trigger a custom function instead of `onDismiss`.

Use cases include using the Menu component as a dropdown tied to a TextInput component, where the dismiss needs to be triggered solely by the TextInput `onBlur` method.

The prop, `handleExternalTouches`, is optional and defaults to capturing touches and triggering `onDismiss` (the original behavior, to avoid a breaking change). Can set the prop to:

-  `'dismiss'` - provides default 'dismiss menu' behaviour
-  `'ignore'` - prevents external touches from being captured
-  `() => void` - custom function to trigger on captured touch, instead of `onDismiss`

### Test plan

Set up the Menu component with `handleExternalTouches` set to:
-  `'ignore'`, and check touches are passed through to other components and the menu is not dismissed.
-  function of type `() => void`, and check touches are captured and trigger the custom function


finnmerlett:
What a joke. I put time and effort into making this feature PR-worthy, and it has just been systematically ignored :angry: . Last time I submit a PR to this repo.

PR #1998, opened 2020-06-17
docs: Update class component examples to be fc's for consistency

NickFoden:
### Summary

After finding a couple usage code examples in the docs that were incorrect/will not compile if someone attempts to run them, started with updating them, then went through and updated class examples to be functional components. Maybe got carried away as was just updating while watching the developer week global talks. I can break into smaller PR's, but I think a reviewer can actually review quite quickly as small components and can compare the diffs. 

Spotted issue in [Data Table Pagination example](https://callstack.github.io/react-native-paper/data-table-pagination.html)

image

And function declaration issue in [Drawer Section](https://callstack.github.io/react-native-paper/drawer-section.html)
image

And currently there are many examples still using class components, but some have in the past been updated to be functional components. This PR updates the remaining class component examples in the docs to use Functional Components, and also removes any unused imports from examples. `{View, Text,` etc unless in use. 

Went from the existing functional component examples as the model for naming conventions, formatting, etc - for instance `Appbar.Action` example was already a FC https://callstack.github.io/react-native-paper/appbar-action.html 

### Test plan

Sort of visual yes, as it is the usage examples in the docs, I think testers can review/view easily in the PR diff or via running the branch and browsing the docs, most will look like this updated view of the Data Table Pagination example

image

I can add more screenshots if that is helpful, please comment and tag me. Txs.

### DIFF
--- a/src/components/Appbar/AppbarHeader.tsx
+++ b/src/components/Appbar/AppbarHeader.tsx
@@ -56,29 +56,24 @@ type Props = React.ComponentProps<typeof Appbar> & {
  * import * as React from 'react';
  * import { Appbar } from 'react-native-paper';
  *
- * export default class MyComponent extends React.Component {
- *   _goBack = () => console.log('Went back');
+ * const MyComponent = () => {
+ *   const _goBack = () => console.log('Went back');
  *
- *   _handleSearch = () => console.log('Searching');
+ *   const _handleSearch = () => console.log('Searching');
  *
- *   _handleMore = () => console.log('Shown more');
+ *   const _handleMore = () => console.log('Shown more');
  *
- *   render() {
- *     return (
- *       <Appbar.Header>
- *         <Appbar.BackAction
- *           onPress={this._goBack}
- *         />
- *         <Appbar.Content
- *           title="Title"
- *           subtitle="Subtitle"
- *         />
- *         <Appbar.Action icon="magnify" onPress={this._handleSearch} />
- *         <Appbar.Action icon="dots-vertical" onPress={this._handleMore} />
- *       </Appbar.Header>
- *     );
- *   }
- * }
+ *   return (
+ *     <Appbar.Header>
+ *       <Appbar.BackAction onPress={_goBack} />
+ *       <Appbar.Content title="Title" subtitle="Subtitle" />
+ *       <Appbar.Action icon="magnify" onPress={_handleSearch} />
+ *       <Appbar.Action icon="dots-vertical" onPress={_handleMore} />
+ *     </Appbar.Header>
+ *   );
+ * };
+ *
+ * export default MyComponent;
  * ```
  */
 class AppbarHeader extends React.Component<Props> {
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -82,40 +82,39 @@ type NativeEvent = {
  * import { Image } from 'react-native';
  * import { Banner } from 'react-native-paper';
  *
- * export default class MyComponent extends React.Component {
- *   state = {
- *     visible: true,
- *   };
+ * const MyComponent = () => {
+ *   const [state, setState] = React.useState({ visible: true });
  *
- *   render() {
- *     return (
- *       <Banner
- *         visible={this.state.visible}
- *         actions={[
- *           {
- *             label: 'Fix it',
- *             onPress: () => this.setState({ visible: false }),
- *           },
- *           {
- *             label: 'Learn more',
- *             onPress: () => this.setState({ visible: false }),
- *           },
- *         ]}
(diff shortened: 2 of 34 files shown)


src/components/Banner.tsx
```diff
@@ -82,40 +82,39 @@ type NativeEvent = {
  * import { Image } from 'react-native';
  * import { Banner } from 'react-native-paper';
  *
- * export default class MyComponent extends React.Component {
- *   state = {
- *     visible: true,
- *   };
+ * const MyComponent = () => {
+ *   const [state, setState] = React.useState({ visible: true });
```
Trancever, line 86: Can we refactor it to `visible` of type boolean instead of object?

src/components/Checkbox/Checkbox.tsx
```diff
@@ -64,21 +64,22 @@ type Props = {
  * import * as React from 'react';
  * import { Checkbox } from 'react-native-paper';
  *
- * export default class MyComponent extends React.Component {
- *   state = {
- *     checked: false,
- *   };
+ * const MyComponent = () => {
+ *   const [state, setState] = React.useState({ checked: false });
```
Trancever, line 68: Same as above, can we refactor it to keep boolean in the state?

src/components/Dialog/Dialog.tsx
```diff
@@ -48,36 +48,32 @@ type Props = {
  * import { View } from 'react-native';
  * import { Button, Paragraph, Dialog, Portal } from 'react-native-paper';
  *
- * export default class MyComponent extends React.Component {
- *   state = {
- *     visible: false,
- *   };
+ * const MyComponent = () => {
+ *   const [state, setState] = React.useState({ visible: false });
```
Trancever, line 52: Same as above

src/components/Dialog/DialogActions.tsx
```diff
@@ -23,28 +23,26 @@ type Props = React.ComponentPropsWithRef<typeof View> & {
  * import * as React from 'react';
  * import { Button, Dialog, Portal } from 'react-native-paper';
  *
- * export default class MyComponent extends React.Component {
- *   state = {
- *     visible: false,
- *   };
+ * const MyComponent = () => {
+ *   const [state, setState] = React.useState({ visible: false });
```
Trancever, line 27: Same as above

Trancever:
@NickFoden Thank you for your contribution! It looks amazing. There is one thing that I would like to change before we merge this. Could you go through all the examples and replace state which is an object to be of type `boolean`, of course wherever it makes sense :)

NickFoden:
Sure thing, was trying to preserve the existing somewhat, but this is also my preference. Will push up today.

src/components/Appbar/AppbarHeader.tsx
```diff
@@ -56,29 +56,24 @@ type Props = React.ComponentProps<typeof Appbar> & {
  * import * as React from 'react';
  * import { Appbar } from 'react-native-paper';
  *
- * export default class MyComponent extends React.Component {
- *   _goBack = () => console.log('Went back');
+ * const MyComponent = () => {
+ *   const _goBack = () => console.log('Went back');
```
satya164, line 60: Can we remove `_`prefix from these? With classes, we had these to signal that they aren't public properties but they are completely encapsulated with functions, so it's not needed anymore.

src/components/BottomNavigation.tsx
```diff
@@ -334,34 +334,34 @@ class SceneComponent extends React.PureComponent<any> {
  *
  * const RecentsRoute = () => <Text>Recents</Text>;
  *
- * export default class MyComponent extends React.Component {
- *   state = {
+ * const MyComponent = () => {
+ *   const [state, setState] = React.useState({
```
satya164, line 338: For this one, I'd suggest splitting it to 2 states, `index` and `routes`. This way, you could directly pass `setIndex` function to `onIndexChange`

src/components/Appbar/AppbarHeader.tsx
NickFoden: For sure.

src/components/BottomNavigation.tsx
NickFoden: Sounds good.

NickFoden: @satya164 so i didn't make this change. The actual component depends on the `navigationState` object, so to update the example will also need to update the component. I have the bandwidth and can take it on, but just an fyi why this change wasn't made. Txs.

src/components/Appbar/AppbarHeader.tsx
NickFoden: Made these changes in latest commit.

PR #2001, opened 2020-06-20
Fix lint offenses

belgamo:
### Summary
Fix lint errors that were breaking on CI.


Trancever:
Thanks for the contribution! This should be already fixed on master

PR #2002, opened 2020-06-21
feat: add ability to extend theme TS type with custom properties

### LINKED ISSUE #1835: Ability to extend Theme TS type.
## Describe the feature

`useTheme` hook should return the correct type that contains user-defined properties.

## Motivation

The type returned by `useTheme` hook is not customizable in any way right now. It always returns a type containing only properties defined by the library. It would be handy if user-defined properties are also properly typed.

@satya164 Would you like to explain how this could be implemented? Or at least post some link where we could see how it is done in other libs?

Trancever:
### Summary

Fix #1835 

### Test plan

Run the example app and try to add a custom property to the theme.

### DIFF
--- a/docs/pages/2.theming.md
+++ b/docs/pages/2.theming.md
@@ -106,6 +106,8 @@ export default function Main() {
 }
 ```
 
+### Typescript
+
 By default it won't work well with typescript, but we can take advantage of `global augmentations` and specify the new properties that we added to the theme:
 
 ```tsx
--- a/docs/pages/2.theming.md
+++ b/docs/pages/2.theming.md
@@ -78,6 +78,83 @@ If you don't use custom theme, Paper will automatically turn animations on/off, 
 
 Otherwise, your custom theme will need to handle it manually, using React Native's [AccessibilityInfo API](https://reactnative.dev/docs/0.54/accessibilityinfo);
 
+## Extending the theme
+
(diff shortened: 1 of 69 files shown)


docs/pages/2.theming.md
```diff
@@ -78,6 +78,83 @@ If you don't use custom theme, Paper will automatically turn animations on/off,
...
+}
+```
+
+There is a one gotcha we need to be aware of. `Global augmentation` works only when we add new unique properties, but it will throw an error when we modify the existing ones. It means we won't be able to e.g. pass another color property to the `colors` object:
+
+```tsx
+declare global {
+  namespace ReactNativePaper {
+    interface Theme {
+      // This will throw, because Paper itself defined `colors` property
```
satya164, line 141: I wonder whether this will work if we change the `colors` property to an interface?

example/src/index.tsx
```diff
@@ -8,17 +8,68 @@ import {
...
-  Theme,
 } from 'react-native-paper';
 import App from './RootNavigator';
 import DrawerItems from './DrawerItems';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 
+// Add new typescript properties to the theme
+declare global {
+  namespace ReactNativePaper {
+    interface Fonts {
```
satya164, line 19: Let's prefix these with `Theme` to make it clear that they are part of the theme

PR #2010, opened 2020-06-24
refactor(Menu): attach listeners earlier and add prop testID

brunohkbx:
### Summary

Problem: Since [react-native-testing-library@2.0.0](https://github.com/callstack/react-native-testing-library) `getByTestId` queries does not return non-native instances. Therefore, we can't trigger the behavior `dismiss` when Menu is dismissed as before to test whether it changed the prop `visible` to `false`.

Solution: Removing and attaching the listeners when the method show is called does the trick as the component [modal#showModal](https://github.com/callstack/react-native-paper/blob/master/src/components/Modal.tsx#L134) does. Now we can trigger `dismiss` with `hardwareBackPress`  as react-native does [here](https://github.com/facebook/react-native/blob/a903d1b86ab56163abcdcb584f335949ba0c85fc/Libraries/Utilities/__mocks__/BackHandler.js) and check the presence of the menu with the `testID` prop (could also be an accessibility prop).

### DIFF
--- a/example/src/Examples/MenuExample.tsx
+++ b/example/src/Examples/MenuExample.tsx
@@ -120,10 +120,7 @@ const MenuExample = ({ navigation }: Props) => {
           <Menu.Item onPress={() => {}} title="Item 3" disabled />
         </Menu>
         <List.Section style={styles.list} title="Contextual menu">
-          <TouchableRipple
-            onPress={() => {}}
-            onLongPress={_handleLongPress}
-          >
+          <TouchableRipple onPress={() => {}} onLongPress={_handleLongPress}>
             <List.Item
               title="List item"
               description="Long press me to open contextual menu"
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -57,6 +57,10 @@ type Props = {
    * @optional
    */
   theme: Theme;
+  /**
+   * testID to be used on tests.
+   */
+  testID?: string;
 };
 
 type Layout = $Omit<$Omit<LayoutRectangle, 'x'>, 'y'>;
@@ -253,6 +257,8 @@ class Menu extends React.Component<Props, State> {
   };
 
   private show = async () => {
+    this.removeListeners();
+    this.attachListeners();
     const windowLayout = Dimensions.get('window');
     const [menuLayout, anchorLayout] = await Promise.all([
       this.measureMenuLayout(),
@@ -291,8 +297,6 @@ class Menu extends React.Component<Props, State> {
         },
       }),
       () => {
-        this.attachListeners();
-
         const { animation } = this.props.theme;
         Animated.parallel([
           Animated.timing(this.state.scaleAnimation, {
@@ -344,6 +348,7 @@ class Menu extends React.Component<Props, State> {
       theme,
       statusBarHeight,
       onDismiss,
+      testID,
     } = this.props;
 
     const {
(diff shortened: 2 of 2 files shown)


thymikee:
Instead of checking "visible" props, verify if the menu is not visible with a `findByText` or so query?

brunohkbx:
@thymikee this works fine to check if the menu is visible. On the other hand, it's not possible to test if the menu is not visible or side effects of the callback `onDismiss` because there's no way to trigger this.

satya164:
> it's not possible to test if the menu is not visible or side effects of the callback

If your menu items are not found, then can't you be sure that menu isn't visible? There's also `waitForElementToBeRemoved`

> Removing and attaching the listeners when the method show is called does the trick as the component modal#showModal does. Now we can trigger dismiss with hardwareBackPress

Can you clarify why removing and reattaching listeners is needed to use it?

brunohkbx:
> If your menu items are not found, then can't you be sure that menu isn't visible? There's also waitForElementToBeRemoved

You are right. However, consider this example below:

```
<Menu
  visible={_getVisible('menu2')}
  onDismiss={_toggleMenu('menu2')}
  anchor={
    <Button mode="outlined" onPress={_toggleMenu('menu2')}>
      Menu with icons
    </Button>
  }
 >
  ... some items
</Menu>
```

You can test whether the menu is visible as you said. but how would you test if `_toggleMenu` is doing its job? You need to trigger `dismiss` somehow, right?

> Can you clarify why removing and reattaching listeners is needed to use it?

That's because there's two way to trigger dismiss. 
The first approach would be adding a testID to this [TouchableWithoutFeedback - L538](https://github.com/callstack/react-native-paper/blob/master/src/components/Menu/Menu.tsx#L538) 

And the second one would be simulating a `hardwareBackPress`, but to do this we need to attach the listeners earlier because of react-native mocking native methods such as `measureInWindow`.  So the original [attachListeners](https://github.com/callstack/react-native-paper/blob/master/src/components/Menu/Menu.tsx#L294) on L294 is never called in a test environment because these two promises do not resolve:

```js
const [menuLayout, anchorLayout] = await Promise.all([
      this.measureMenuLayout(),
      this.measureAnchorLayout(),
    ]);
```

With this changes we would be able to write such kind of test: 

```js
test('onDismiss dismiss the menu', async () => {
      const {
        wrapper: { findByTestId, queryByTestId },
      } = render(some component)

      fireEvent.press(await findByTestId('anchor'));
      await findByTestId('my-awesome-menu');

      await act(async () => BackHandler.mockPressBack());

      await waitForElementToBeRemoved(() =>
        queryByTestId(''my-awesome-menu')
      );
    });
```

Does it make sense? Sorry if my explanation was not clear.

thymikee:
Thanks for explanation. Seems like a proper way to solve it is to properly mock the View's `measureInWindow` method (or figure out why it doesn't resolve, because [it is mocked](https://github.com/facebook/react-native/blob/a2b699daf65dbd720a367d78da75433f041eabe4/jest/MockNativeMethods.js#L14)). Otherwise your tests are likely to break in the future at any time, if the logic around the Menu changes without affecting non-test environments.

brunohkbx:
@thymikee what if instead we add a prop `testID` for the [touchable here](https://github.com/callstack/react-native-paper/blob/master/src/components/Menu/Menu.tsx#L538)? It will solve the problem.

thymikee:
I'm fine with adding `testID` anyway. However resolving this properly would be a nice bonus :)

brunohkbx:
I'll update this PR. Thanks for the feedback :smiley:

PR #2011, opened 2020-06-24
docs: Add a new app and new github icon to showcase page

jvidalv:
Hi, I just want to share with you the open source app I've made using react-native-paper!

https://play.google.com/store/apps/details?id=josep.astrale

Cheers!


jvidalv:
Seems like an update from last week is breaking the linter tests.

I'll create a separate commit to fix that.

jvidalv:
I have added a new icon and a new property in order to be able to share the github repository in open source apps:

Preview

docs/pages/src/Showcase.js
```diff
@@ -104,6 +112,15 @@ export default class Showcase extends React.Component<{}> {
                         {item.name}
                       </AppName>
                       <BadgeContainer>
+                        <a
```
Trancever, line 115: @jvidalv Could you render githubIcon as a last icon instead of it being the first one?

jvidalv: Done!

PR #2016, opened 2020-06-26
feat: handle theme change based on preferences

troZee:
### Motivation
Closes https://github.com/callstack/react-native-paper/issues/1597

Android 10 and iOS 13 added support for dark theme. User (like me ;p) can set dark theme as default theme. Application should adapt to this theme automatically, hence I would like to have automatically theme changing in my app, so I don't need to care about user preferences.
### TODO 

- [x] Add an example, which demonstrates this api  

### Screenshots 

Simulator Screen Shot - iPhone X - 2020-06-27 at 16 00 29
 
Simulator Screen Shot - iPhone X - 2020-06-27 at 15 59 25

### Test plan
* Check, it theme changes based on user preferences (it should be done automatically)
* You can use my private fork to test it https://github.com/troZee/paperExample


src/core/Provider.tsx
```diff
@@ -22,15 +30,41 @@ export default class Provider extends React.Component<Props> {
...
 
   componentWillUnmount() {
     AccessibilityInfo.removeEventListener(
       'reduceMotionChanged',
       this.updateReduceMotionSettingsInfo
     );
+    Appearance.removeChangeListener(this._handleAppearanceChange);
   }
 
+  private _handleAppearanceChange = (
```
satya164, line 44: You can remove the `_` prefix since it's private. Also consider refactoring this component to a function component.

src/core/Provider.tsx
```diff
@@ -22,15 +30,41 @@ export default class Provider extends React.Component<Props> {
...
+  private _getTheme = () => {
+    const { theme: providedTheme } = this.props;
+    const { reduceMotionEnabled } = this.state;
+
+    if (providedTheme) {
+      return providedTheme;
+    } else {
+      const theme =
+        this.state.colorScheme === 'light' ? DefaultTheme : DarkTheme;
+      return Object.assign(theme as ReactNativePaper.Theme, {
```
satya164, line 60: Probably not from your change, but `Object.assign` incorrect here since:

1. It'll mutate our theme
2. Adding `animation` property will override anything else defined in this property. We don't have anything else right now, but might have in future.

src/core/Provider.tsx
```diff
@@ -1,19 +1,27 @@
 import * as React from 'react';
-import { AccessibilityInfo } from 'react-native';
+import { AccessibilityInfo, Appearance, ColorSchemeName } from 'react-native';
```
satya164, line 2: Also I think we might want to check if `Appearance` module actually exists. Expo inits a project with old version of react native web by default so it'll be undefined there I think.

troZee: Good call, I thought, 4.0 will not support older version of react-native/expo

troZee:
Simulator Screen Shot - iPhone X - 2020-06-27 at 16 00 29
 
Simulator Screen Shot - iPhone X - 2020-06-27 at 15 59 25

src/core/Provider.tsx
matkoson: It's my change. This [PR](https://github.com/callstack/react-native-paper/pull/2024) fixes the issues you mentioned @satya164

src/core/Provider.tsx
```diff
@@ -22,15 +30,42 @@ export default class Provider extends React.Component<Props> {
...
+
+  private getTheme = () => {
+    const { theme: providedTheme } = this.props;
+    const { reduceMotionEnabled } = this.state;
+
+    if (providedTheme) {
+      return providedTheme;
+    } else {
+      const theme =
+        this.state.colorScheme === 'light' ? DefaultTheme : DarkTheme;
```
satya164, line 59: Can you destructure this as well along with `reduceMotionEnabled` for consistency?

PR #1999, opened 2020-06-17
fix: menu example

brunohkbx:
### Summary

The example showing how to display a menu using `onLongPress` was not working.

PR #2020, opened 2020-06-27
fix: fix memory leak in TextInput placeholder animation

brunohkbx:
### Summary
I was getting random warning messages like `Can't perform a React state update on an unmounted component.` when testing with jest because the animation wasn't checking whether it was finished.

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -303,7 +303,9 @@ class TextInput extends React.Component<TextInputProps, State> {
         ios: false,
         default: true,
       }),
-    }).start(this.hidePlaceholder);
+    }).start(({ finished }) => {
+      if (finished) this.showPlaceholder();
+    });
   };
 
   private hideError = () => {


brunohkbx:
@jbinda can you please take a look?

jbinda:
@brunohkbx - sure I will take a look.

You receive that warning when running test or working with the example app ?

brunohkbx:
I ran on this when writing tests. it's the same as #1999

src/components/TextInput/TextInput.tsx
```diff
@@ -303,7 +303,9 @@ class TextInput extends React.Component<TextInputProps, State> {
         ios: false,
         default: true,
       }),
-    }).start(this.hidePlaceholder);
+    }).start(({ finished }) => {
+      if (finished) this.showPlaceholder;
```
SleeplessByte, line 307:
```suggestion
      if (finished) this.showPlaceholder();
```

Shouldn't this actually call the function?

brunohkbx: Ops! Good catch, thanks!

src/components/TextInput/TextInput.tsx
```diff
@@ -303,7 +303,9 @@ class TextInput extends React.Component<TextInputProps, State> {
         ios: false,
         default: true,
       }),
-    }).start(this.hidePlaceholder);
+    }).start(({ finished }) => {
+      if (finished) this.showPlaceholder();
```
Trancever, line 307: Why do we call `showPlaceholder` in a place where we were previously calling `hidePlaceholder`. I don't really understand this change :/ Isn't it gonna change how the component works?

brunohkbx: Ops! 
Actually I'll close this PR. turned out to be a bad configuration on jest.

Issue #1996, opened 2020-06-17
Enforce import type for type imports


satya164:
For the next major, we should enforce using the new `import type` feature of TypeScript to import types everywhere. This will let babel drop import statements when only types are used from a file and result in smaller file size/better tree shaking.

To do this, we can add `"importsNotUsedAsValues": "error"` in the `tsconfig.json` and then fix all of the type errors (it'd be just changing `import` to `import type`).

BatDroid:
Can i work on this please?

satya164:
Feel free

BatDroid:
@satya164 
I worked on it but here is what happened.
As you can see [here](https://github.com/prettier/prettier/issues/7263#issuecomment-597882252) `import type` is not supported in until version `2.*` of `prettier`. 
The project's prettier version was `1.19.0` so I updated to the latest version and the errors related to `import type` were gone, but there were bunch of other errors like [this](https://github.com/prettier/prettier/pull/3903#issuecomment-602589307) that there were no options to be included in `.prettierrc` for preventing the error. 
I believe we(I) have to migrate the whole project to prettier with `--fix` command and then refactor the `"importsNotUsedAsValues": "error"`. Is it  ok if make this refactor first or you have other solutions?

satya164:
Yeah, let's make those changes

satya164:
Also make sure to send the PR to 4.0 branch :)

Trancever:
@BatDroid How is it going? Do you need any help with the PR? Maybe someone from paper core team can help you cause ideally, we need this to be merged this week.

BatDroid:
@Trancever Im working on it at the moment. I have applied all the changes but I received error from `Jest` for not recognizing the `import type`.
I updated the jest to the latest version in the project and those errors were all gone.  However, there are errors (that seems to be like warnings) and   more important one of the tests is failing in `BottomNavigation.test.js`. the error is this :
```
 FAIL  src/components/__tests__/BottomNavigation.test.js
  ● Test suite failed to run

    Call retries were exceeded

      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:191:21)
```
I'm gonna keep working on it for a few hours to see if I can resolve it. meanwhile, you can either wait or if you have the solution would appreciate to share :)

satya164:
[...]

PR #2021, opened 2020-06-27
refactor: update prettier to 2.0.5 and refactor the whole project for it

BatDroid:
### Summary

This Pr is the first step for  [this issue](https://github.com/callstack/react-native-paper/issues/1996). As it is described [here](https://github.com/callstack/react-native-paper/issues/1996#issuecomment-650641925), This refactor is inevitable. To reduce the number of files which needs to be refactored, the `arrowParens: "avoid"` is added to the `.prettierrc`.


.prettierrc
```diff
@@ -3,3 +3,4 @@ singleQuote: true
 tabWidth: 2
 trailingComma: "es5"
 useTabs: false
+arrowParens: "avoid"
```
satya164, line 6: I think it's fine to leave to prettier default, since we're already changing a lot of files.

BatDroid:
Since the base branch changed and brought conflicts. I add these changes again in a new pr

PR #2004, opened 2020-06-22
fix: upgrade @types/react-native to fix #1962

arniu:
### Summary

Resolve #1962

PR #2027, opened 2020-06-30
fix: Unable to style chip icon

contactsimonwilson:
#2004 # Summary

Add IconProps prop to Chip component to allow more control over styling (such as color).

### Test plan
1. Add IconProps={{ color: 'red' }} to Chip component consumption.
2. Ensure chip icon color is red

### DIFF
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -11,7 +11,7 @@ import {
   ViewStyle,
 } from 'react-native';
 import color from 'color';
-import Icon, { IconSource } from './Icon';
+import Icon, { IconSource, Props as IconProps } from './Icon';
 import Surface from './Surface';
 import Text from './Typography/Text';
 import TouchableRipple from './TouchableRipple';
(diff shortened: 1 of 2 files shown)


contactsimonwilson:
@Trancever any feedback on this? This would allow us to style action chips in the same way as Google does on many of their applications.

src/components/Chip.tsx
```diff
@@ -33,6 +33,10 @@ type Props = React.ComponentProps<typeof Surface> & {
    * Icon to display for the `Chip`. Both icon and avatar cannot be specified.
    */
   icon?: IconSource;
```
Trancever, line 35: `icon` prop accepts a function. It means you can do something like this: `icon={() => <MaterialCommunityIcon {...passWhateverPropsYouNeed} />}`. It's already fully customizable and we don't need additional prop.

contactsimonwilson:
Thanks @Trancever

PR #2028, opened 2020-06-30
feat(Menu): improve accessibility around dismissing

brunohkbx:
following up #2010

#### Android
1. Check `accessibilityLabel` of the overlay with TalkBack

#### iOS
1. Check whether `onAccessibilityEscape` dismiss the menu with the escape/scrub gesture

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -47,7 +47,7 @@ type Props = {
   /**
    * Accessibility label for the overlay. This is read by the screen reader when the user taps outside the menu.
    */
-  dismissAccessibilityLabel?: string;
+  overlayAccessibilityLabel?: string;
   /**
    * Content of the `Menu`.
    */
@@ -134,7 +134,7 @@ class Menu extends React.Component<Props, State> {
 
   static defaultProps = {
     statusBarHeight: APPROX_STATUSBAR_HEIGHT,
-    dismissAccessibilityLabel: 'Close menu',
+    overlayAccessibilityLabel: 'Close menu',
   };
 
   static getDerivedStateFromProps(nextProps: Props, prevState: State) {
@@ -344,7 +344,7 @@ class Menu extends React.Component<Props, State> {
       theme,
       statusBarHeight,
       onDismiss,
-      dismissAccessibilityLabel,
+      overlayAccessibilityLabel,
     } = this.props;
 
     const {
@@ -537,7 +537,7 @@ class Menu extends React.Component<Props, State> {
         {rendered ? (
           <Portal>
             <TouchableWithoutFeedback
-              accessibilityLabel={dismissAccessibilityLabel}
+              accessibilityLabel={overlayAccessibilityLabel}
               accessibilityRole="button"
               onPress={onDismiss}
             >
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -134,7 +134,7 @@ class Menu extends React.Component<Props, State> {
 
   static defaultProps = {
     statusBarHeight: APPROX_STATUSBAR_HEIGHT,
-    overlayAccessibilityLabel: 'Close menu',
+    dismissAccessibilityLabel: 'Close menu',
   };
 
   static getDerivedStateFromProps(nextProps: Props, prevState: State) {
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -45,6 +45,10 @@ type Props = {
    */
   onDismiss: () => void;
   /**
+   * Accessibility label for the overlay. This is read by the screen reader when the user taps outside the menu.
+   */
+  dismissAccessibilityLabel?: string;
+  /**
    * Content of the `Menu`.
    */
   children: React.ReactNode;
@@ -130,6 +134,7 @@ class Menu extends React.Component<Props, State> {
 
   static defaultProps = {
     statusBarHeight: APPROX_STATUSBAR_HEIGHT,
+    overlayAccessibilityLabel: 'Close menu',
   };
 
   static getDerivedStateFromProps(nextProps: Props, prevState: State) {
@@ -339,6 +344,7 @@ class Menu extends React.Component<Props, State> {
       theme,
       statusBarHeight,
       onDismiss,
+      dismissAccessibilityLabel,
     } = this.props;
 
     const {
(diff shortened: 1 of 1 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -530,7 +535,10 @@ class Menu extends React.Component<Props, State> {
         {this.isAnchorCoord() ? null : anchor}
         {rendered ? (
           <Portal>
-            <TouchableWithoutFeedback onPress={onDismiss}>
+            <TouchableWithoutFeedback
+              onPress={onDismiss}
+              testID={overlayTestID}
```
thymikee, line 540: I preferred the one passing `testID` to the actual view with content, instead of an overlay. Can we bring that back and get rid of `overlayTestID `?

satya164: Is there a need to pass `testID` to the menu content? It should already be possible to pass them to menu items which the user interacts with.

brunohkbx: It's possible to pass to menu items. However, there's no way to trigger dismiss on them.

satya164: I'm not sure I understand what you mean by triggering dismiss on menu items

brunohkbx: @satya164 I added a [test](https://github.com/callstack/react-native-paper/pull/2028/files#diff-871591f2003665c0b2a83202f7617b77R62) to illustrate my point. 
In testing library 2 we could use `waitForElementToBeRemoved` instead of checking whether `onDismiss` was called or not. Either way, we need testID to fire the event `dismiss` on the menu.

satya164: Why is it necessary to trigger `dismiss` event manually when you can press the overlay (with the `overlayTestID` in this PR)?

brunohkbx: @thymikee suggested moving it back to the view :thinking:
I agree with you, pressing the overlay sounds more natural in the user perspective rather than testing the component implementation.

satya164: @brunohkbx after talking with @thymikee, maybe instead of `overlayTestID`, it'd be better to expose a prop `overlayAccessibilityLabel` which defaults to "Close menu" that you can use to press on it?

thymikee: We've concluded that voice users don't have an easy way to dismiss the menu at the moment, so the right thing to do would be empowering them to do so. We think that's the best way forward and it would be lovely if you could make it happen as a part of this PR (rewording it a bit).

Once we're done, we leave the library more accessible and testable, which is a win-win!

brunohkbx: Sounds great! will do.
Thanks for your efforts on this :smiley:

lukewalczak: In terms of accessibility, I can suggest:
* iOS: use [prop](https://reactnative.dev/docs/accessibility#onaccessibilityescape-ios) called `onAccessbilityEscape` where you can pass a function responsible for closing `Menu`. It can help users with disabilities to close it doing `Z` shape gesture. 
* Android: I can recommend to add a listener on hardware back button using the `BackHandler` [API](https://reactnative.dev/docs/backhandler)

src/components/Menu/Menu.tsx
```diff
@@ -530,7 +536,10 @@ class Menu extends React.Component<Props, State> {
         {this.isAnchorCoord() ? null : anchor}
         {rendered ? (
           <Portal>
-            <TouchableWithoutFeedback onPress={onDismiss}>
+            <TouchableWithoutFeedback
+              accessibilityLabel={overlayAccessibilityLabel}
+              onPress={onDismiss}
```
brunohkbx, line 541: @lukewalczak do you think the overlay should have `accessibilityRole` as a button?

src/components/Menu/Menu.tsx
brunohkbx: I think it's done. Menu already has a BackHandler listener, so we are good.

src/components/Menu/Menu.tsx
lukewalczak: I've checked Facebook and Discord app and I think we can add:
 `accessibilityRole` or maybe better `accessibilityLabel="Close Menu button"` to inform the user that overlay is pressable and will close menu.

src/components/Menu/Menu.tsx
```diff
@@ -57,6 +57,10 @@ type Props = {
    * @optional
    */
   theme: Theme;
+  /**
+   * Accessibility label for the overlay. This is read by the screen reader when the user taps outside the menu.
+   */
+  overlayAccessibilityLabel?: string;
```
thymikee, line 63: I'm thinking that we have `onDismiss` prop defined, how about renaming this to something like `dismissAccessibilityLabel` and group it closer to the `onDismiss` prop?

src/components/Menu/Menu.tsx
```diff
@@ -130,6 +134,7 @@ class Menu extends React.Component<Props, State> {
 
   static defaultProps = {
     statusBarHeight: APPROX_STATUSBAR_HEIGHT,
+    overlayAccessibilityLabel: 'Close menu',
```
thymikee, line 137:
```suggestion
    dismissAccessibilityLabel: 'Close menu',
```

brunohkbx: Ops! Thanks.

src/components/Menu/Menu.tsx
```diff
@@ -541,6 +551,8 @@ class Menu extends React.Component<Props, State> {
               accessibilityViewIsModal={visible}
               style={[styles.wrapper, positionStyle, style]}
               pointerEvents={visible ? 'box-none' : 'none'}
+              // @ts-ignore - FIX ME
```
Trancever, line 554: What error is being thrown?

src/components/Menu/Menu.tsx
```diff
@@ -44,6 +44,10 @@ type Props = {
    * Callback called when Menu is dismissed. The `visible` prop needs to be updated when this is called.
    */
   onDismiss: () => void;
+  /**
+   * Accessibility label for the overlay. This is read by the screen reader when the user taps outside the menu.
+   */
+  dismissAccessibilityLabel?: string;
```
Trancever, line 50: I think `overlayAccessibilityLabel` makes more sense. @satya164 What do you think?

src/components/Menu/Menu.tsx
brunohkbx: We are using `"@types/react-native": "^0.57.61"` which does not have `onAccessibilityEscape` defined and the next version `0.62.x` has some breaking changes.

Trancever: Ahh that's right. Types are updated on 4.0 branch only so we have to keep this ts-ignore for now.

PR #2029, opened 2020-06-30
fix: catch empty touches

jaulz:
### Summary

There are situations when the `touches` array is empty and thus it causes errors.
image

As I noticed it especially with `react-native-web@^0.13.0` I checked their code and they have a similar check:
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/hooks/usePressEvents/PressResponder.js#L548-L557

### Test plan

Unfortunately, I cannot provide an example that reproduces the issue but I noticed it once in a while with `react-native-web@^0.13.0`.

### DIFF
--- a/src/components/TouchableRipple/index.tsx
+++ b/src/components/TouchableRipple/index.tsx
@@ -119,12 +119,10 @@ class TouchableRipple extends React.Component<Props> {
       touchX = dimensions.width / 2;
       touchY = dimensions.height / 2;
     } else {
-      const startX = e.nativeEvent.touches
-        ? e.nativeEvent.touches[0].pageX
-        : e.pageX;
-      const startY = e.nativeEvent.touches
-        ? e.nativeEvent.touches[0].pageY
-        : e.pageY;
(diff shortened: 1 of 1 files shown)


src/components/TouchableRipple/index.tsx
```diff
@@ -119,12 +119,10 @@ class TouchableRipple extends React.Component<Props> {
...
     } else {
-      const startX = e.nativeEvent.touches
-        ? e.nativeEvent.touches[0].pageX
-        : e.pageX;
-      const startY = e.nativeEvent.touches
-        ? e.nativeEvent.touches[0].pageY
-        : e.pageY;
+      const { changedTouches, touches } = e.nativeEvent;
+      const touch = touches?.[0] || changedTouches?.[0];
+      const startX = touch.pageX || e.pageX;
```
Trancever, line 124: This will cause bugs if `touch.pageX` is `0`. Can we handle this case?

jaulz: Very good catch 😊  I just fixed it.

PR #2034, opened 2020-07-01
Fix: RadioButtonIOS inverted icon in RTL mode

avirankatzsofi:
### Summary

The radio button displays a checkbox in iOS devices. The checkbox is inverted when using RTL.


Trancever:
Thanks!

PR #2035, opened 2020-07-01
feat: add render prop to Avatar.Image

jaulz:
### Summary

Currently, it is not possible to use another `Image` component within `Avatar.Image`. This change introduces a `render` function which can be used to customize the rendered image.

### Test plan

Use the example provided in the inline docs.


Trancever:
@jaulz Hey, thanks for the PR. Sorry that I took so long to review it.

I am happy with merging it however in the meantime we released 4.0 and replaced master and 4.0 branches and now the PR is kinda broken. Could you recreate it?

Trancever:
Fixed in https://github.com/callstack/react-native-paper/pull/2168

PR #2044, opened 2020-07-07
feat: add ellipsizeMode prop to Chip

brunohkbx:
### Summary
It'd be nice to be able to truncate the chip text when it's too big.

example


Trancever:
@brunohkbx Wasn't it already truncating too long text at the end since the `ellipsizeMode` default is `tail`? Anyway I am ok with merging it 👍

Trancever (approved):

brunohkbx:
@Trancever you are right! I updated the example.

PR #2049, opened 2020-07-09
fix: allow auto change theme

jbinda:
I have add expo configuration that allows to use automatic change theme

Also I have fixed broken DarkTheme switch


troZee:
I have tested on my real device and it LGTM

PR #2057, opened 2020-07-16
chore: bump rn web and add async storage that supports web

Trancever:
This fixes the example app on web.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -13,6 +13,7 @@
   "main": "node_modules/expo/AppEntry.js",
   "dependencies": {
     "@expo/vector-icons": "^10.2.0",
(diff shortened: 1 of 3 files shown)


example/package.json
```diff
@@ -30,7 +31,7 @@
     "react-native-safe-area-context": "3.0.2",
     "react-native-screens": "~2.9.0",
     "react-native-vector-icons": "^6.6.0",
-    "react-native-web": "~0.11.7",
+    "react-native-web": "~0.13.0",
```
SleeplessByte, line 34: This is **not** compatible with Expo 38. It might work for this example, but will unexpectedly break if other packages are added (e.g. `expo-av`).

PR #2064, opened 2020-07-18
fix: allow labelStyle to override all built-in styles

mars-lan:
### Summary

Currently it's not possible to override the text color using `labelStyle` prop due to the internal overriding order.


mars-lan:
@satya164 please also help merge the PR if you have write permission.

Trancever:
@satya164 Isn't it breaking?

mars-lan:
@Trancever it is to some extent, but it seems like the correct expectation that a user-specified prop should override internal default styles? I believe that's how many other paper components work?

Trancever (approved):

PR #2070, opened 2020-07-21
feat(Modal): improve accessibility around dismissing

brunohkbx:
#### Android
1. Check `accessibilityLabel` of the overlay with TalkBack

#### iOS
1. Check whether `onAccessibilityEscape` dismiss the modal with the escape/scrub gesture


Trancever:
Thanks @brunohkbx for improving a11y! That's huge!

PR #2072, opened 2020-07-22
feat: add support for long press to FAB

achuinard:
A few people have requested this and I could use it as well, I believe this should add long press support to the FAB component.


Trancever:
Thanks for your contribution and sorry that it took so long!

achuinard:
Hey I'm just excited to long press my FABs. Thank you!

PR #2078, opened 2020-07-24
feat(TextInput): allow opening a Menu onPress of icon's adornment

brunohkbx:
### Summary

This PR do the following changes:
1. Add the prop `autoFocus` to `TextInputIcon` since we don't want the input to focus when the menu is opened.
2. Pass the event up to the `onPress` prop on the [L68](https://github.com/callstack/react-native-paper/pull/2078/files#diff-5c7a1356b821b108f5584d2b4ea4c02aR68) so we can catch the coordinates.

menu

### Test plan
1. Once you open the `TextInput` section, scroll down to the last input
2. Then click on the chevron icon and check whether the menu has opened.

### DIFF
--- a/src/components/TextInput/Adornment/Icon.tsx
+++ b/src/components/TextInput/Adornment/Icon.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
-import { View, StyleSheet, StyleProp, ViewStyle } from 'react-native';
+import {
+  View,
+  StyleSheet,
+  StyleProp,
+  ViewStyle,
+  GestureResponderEvent,
+} from 'react-native';
 
 import IconButton from '../../IconButton';
 import type { $Omit } from '../../../../src/types';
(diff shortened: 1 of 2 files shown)


src/components/TextInput/Adornment/Icon.tsx
```diff
@@ -48,17 +55,20 @@ export const IconAdornment: React.FunctionComponent<
   );
 };
 
-const TextInputIcon = ({ name, onPress, ...rest }: Props) => {
+const TextInputIcon = ({ name, onPress, autoFocus, ...rest }: Props) => {
```
matkoson, line 58: `autofocus` seems a bit confusing, considering what this value does. `focusable`?

matkoson:
Hey, we think this functionality should be handled by a separate component and we have in plans to add `Select` component in the near future, so thanks for the PR, but it can't be accepted atm.

Issue #2079, opened 2020-07-25
Can't change text color with labelStyle 


Iharidla:
### Current behaviour

When you passing `color` to the `labelStyle` label text color isn't changing

### Expected behaviour

When you passing `color` to the `labelStyle` it should change label text color

### Code sample

[Snack code example](https://snack.expo.io/29DMGsIUs?name=Checkbox.Item&description=https%3A%2F%2Fcallstack.github.io%2Freact-native-paper%2Fcheckbox-item.html&code=import%20*%20as%20React%20from%20%27react%27%3B%0Aimport%20%7B%20View%20%7D%20from%20%27react-native%27%3B%0Aimport%20%7B%20Checkbox%20%7D%20from%20%27react-native-paper%27%3B%0A%0Aconst%20MyComponent%20%3D%20()%20%3D%3E%20(%0A%20%20%3CView%3E%0A%20%20%20%20%3CCheckbox.Item%20label%3D%22Item%22%20status%3D%22checked%22%20%2F%3E%0A%20%20%3C%2FView%3E%0A)%3B%0A%0Aexport%20default%20MyComponent%3B&dependencies=react-native-paper)

### Screenshots (if applicable)
Here text should be red!
![image](https://user-images.githubusercontent.com/44581138/88460013-e11c3480-cea1-11ea-947b-f356f7797011.png)

### What have you tried
In [this](https://github.com/callstack/react-native-paper/blob/2ce47a149215bfe6c7c6cb8391a5096faf8a1266/src/components/Checkbox/CheckboxItem.tsx#L91) row we need move `labelStyle` after color, as it done 
[here](https://github.com/callstack/react-native-paper/blob/2ce47a149215bfe6c7c6cb8391a5096faf8a1266/src/components/RadioButton/RadioButtonItem.tsx#L129).

Trancever:
It should be fixed on the current master. Fix will be available in the next release

simonmasinde:
getting this on android RNP 4.12.5

PR #2080, opened 2020-07-25
fix: place labelStyle after color

Iharidla:
### Summary
#2079

It was quite a simple task so I made a PR immediately.


Trancever:
You can do it via theme prop: `theme={{ colors: { primary: '#BADA55' } }}`

PR #2083, opened 2020-07-27
docs: Add Lyra Collect app to showcase page

alex-fournier:
Hi, we use react-native-paper a lot in our app, we would be proud to be part of the showcase :)


Trancever:
Thanks! Very nice UI

PR #2109, opened 2020-08-05
fix: unable to change Checkbox.Item label color

michalzuk:
### Summary
The order of styles used on `Checkbox.Item` prevented from changing text color using `labelStyle` prop  to make a change.


Trancever:
You can do it via `theme` prop: `theme={{ colors: { primary: '#BADA55' } }}`

PR #2068, opened 2020-07-20
fix: ensure compatible animation drive for cards

SleeplessByte:
### Summary

On dark themes, the default mode is adaptive. This mode will change the `backgroundColor`. This property is not animatable using the nativeDriver. Upstream issue: https://github.com/facebook/react-native/issues/14178

The bug was introduced in https://github.com/callstack/react-native-paper/pull/1787
Fixes https://github.com/callstack/react-native-paper/issues/2063

Issue #2063, opened 2020-07-17
Darkmode Card onPress error with 4.0.x


Keplyx:
### Current behaviour

With darkmode enabled, when clicking on a Card component with onPress prop set, I get the following error:

`Error: Style property 'backgroundColor' is not supported by native animated module, js engine: hermes`

### Expected behaviour

The onPress function should be executed and the Card background should be animated to become lighter.

### Code sample

Just create any project rendering a card with the following props:

```
<Card
   onPress={() => console.log("cc")}
>
   <Text>Hello</Text>
</Card>
```

[Link to a Snack example](https://snack.expo.io/6LbCMqVqs)

### Screenshots (if applicable)
![Screenshot_20200717-164154](https://user-images.githubusercontent.com/23726131/87798659-91f45500-c84c-11ea-81b0-6dfdda5608b7.png)
![Screenshot_20200717_164322](https://user-images.githubusercontent.com/23726131/87798724-a59fbb80-c84c-11ea-945d-9a686a8de18f.png)

### What have you tried

Tried switching to light mode, no issue.
Tried to go back to 3.11.1, no issue.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | android (did not try iOS)
| react-native          | 0.63.1
| react-native-paper    | 4.0.1 and 4.0.0
| react-native-vector-icons    | 7.0.0
| node                  | 12.18.2
| npm or yarn           | npm 6.14.5
| expo sdk              | N/A
| JS engine              | Hermes

ITLight:
I have met same issue, but now I fixed by changing the theme from
const theme = {
  ...DarkTheme,
};
to 
const theme = {
  ...DefaultTheme,
};

wortkotze:
I have the same issue. when I toggle on darkmode > the card module from react native paper throws an error for animated backgrouncColor.

This issue is only available on darkmode not on the normal theme

Keplyx:
> I have met same issue, but now I fixed by changing the theme from
> const theme = {
> ...DarkTheme,
> };
> to
> const theme = {
> ...DefaultTheme,
> };

Yes but this means disabling the built-in DarkTheme, which is not a fix.



[...]

PR #1783, opened 2020-04-04
fix: warning requesting `useNativeDriver` config

fmsouza:
Fixes #1782

After React Native update to version 0.62, a new warning is visible whenever a Button imported from this lib is clicked, requesting to set `useNativeDriver` property explicity in the animation config for this action.

```
Animated: useNativeDriver was not specified. This is a required option and must be explicitly set to true or false
```

### Motivation

This Pull Request sets the `useNativeDriver` flag explicitly to `false` so the warning stops showing.

### Test plan

- Start a new project with React Native version 0.62
- Install `react-native-paper` to the project
- Add the Paper Provider so themes can be available
- Add a Button to the screen with `import { Button } from 'react-native-paper';`
- Run the app and click it
- See the warning in the emulator/simulator/device and in the console

PR #2113, opened 2020-08-08
fix: Card correct animation drivers

### LINKED ISSUE #2138: Card elevation animation is not correct on Dark Adaptive Mode
# PR submitted

fixed with #2113 

### Current behaviour

Currently the press in/out animation of Cards in the Dark Adaptive mode only animates the color with an overlay.

#### Before press
normal

#### Current overlay press
ovelay

### Expected behaviour

Use scale animation.

#### Expected overlay and scale

scale_and_overlay

### Your Environment

| react-native          | 0.62
| react-native-paper    | 4.0.1
| react-native-vector-icons | 7.0.0

PedroBern:
### Summary

#2068 was a partial solution for #2063, bug introduced in #1783.

It did solve in case of toggling the theme while there is no animations running, **but if a card is animating during the process** it may (in case of adaptive mode) receive the animation with native drivers and break in the `Surface` component, just as before of #2068.

**[update]**: Just to clarify the "may receive the animation",  the previous one may be with native drivers `true`, then toggling the theme while animating may change the animation to native drivers `false`, but it is not possible to change native drivers of a an animation during an animation, it does breaks the app.

This PR solves this issue, now you can even toggle the theme with `<Card onPress={toggleTheme}>...` which wasn't possible before (with adaptive mode).

~~**[update 2]** Now it fix #2138 too, this is particularly useful if the card is just an image, as in this case there is no feedback when you press it on dark adaptive mode.~~

### Test plan

Try the following code, before and after this PR. Before, it will break.

```jsx
<Card onPress={toggleTheme}>{children}</Card>
```

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -122,11 +122,10 @@ class Card extends React.Component<Props, State> {
      * toggling the theme will stay at that animated value until
      * the next press-in
      */
-
     const { dark: isDark, mode } = this.props.theme;
     const { dark: wasDark } = prevProps.theme;
     const { elevation, elevationDarkAdaptive } = this.state;
-    if (isDark && mode === 'adaptive' && isDark !== wasDark) {
+    if (isDark && mode === 'adaptive' && !wasDark) {
       // @ts-ignore
       elevation.setValue(this.props.elevation);
       // @ts-ignore
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -49,7 +49,16 @@ type Props = React.ComponentProps<typeof Surface> & {
 };
 
 type State = {
+  /**
+   * Default animated value
+   */
   elevation: Animated.Value;
+
+  /**
+   * Dark adaptive animated value, used in case of toggling the theme,
+   * it prevents animating the background with native drivers inside Surface
+   */
+  elevationDarkAdaptive: Animated.Value;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


Trancever:
@PedroBern Thanks for your contribution! 

I have a question regarding card scaling. Could you point me to a place in MD Guidelines with the spec for scaling surfaces in the dark mode? I've tried looking for it but found only this https://material.io/design/color/dark-theme.html#ui-application and they recommend to only change the surface color.

PedroBern:
@Trancever  you are right, there is no guideline for this, Im sorry! But maybe I can set the default to 1 (no scale) making it optional, or if you want I can remove this conpletely from the pr.

Trancever:
@PedroBern I think it's best to remove it completely from the PR and limit the scope of this PR just to fixing the issue with animation. Thanks in advance!

PedroBern:
@Trancever ok perfect :)

PedroBern:
@Trancever already done, reverted the commits :)

kamui545:
@PedroBern Would it be possible to update the PR to resolve the conflicts? It looks like the component got refactored to FC (v4.2.0).

If you don't have time just let me know, I can work on a new PR (I just don't want to steal the credits).

It would be nice to have this merged, I've been using this patch with `patch-package` for a month and it's working great, I am not seeing any regressions or side effects.

PedroBern:
@kamui545 thanks for the call :) Please go ahead with a new PR then.

kamui545:
Okay, so it looks like I am too retarded to port this fix to React.FC...
No idea why, but it won't work. Questioning my whole career right now... 😂

PedroBern:
@kamui545  It must be something we are not seeing, I did a quick try now and didn't work too! But I don't have time to investigate at the moment... Maybe in the future!

OlimpiaZurek (approved):

tellodaniel:
This will be addressed or merge soon? @OlimpiaZurek 

Thanks!

lukewalczak:
> there are quite a few conflicts, probably not worth fixing it. Feel free to close or reference in new issue/pr if this is still relevant.

Thanks @PedroBern for your work and commitment. Based on your PR I will open a fresh one _(to avoid rebasing and resolving conflicts)_ along with tests and new example.

PR #2143, opened 2020-08-20
feat: add prop forceTextInputFocus to icon adornment

brunohkbx:
### Summary
It would be nice being able to control whether the adornment should focus the input. 

### Test plan
1. Open the example app
2. Click on the eye icon in the password input and verify if the input will show the password without being focused.

### DIFF
--- a/src/components/TextInput/Adornment/Icon.tsx
+++ b/src/components/TextInput/Adornment/Icon.tsx
@@ -10,6 +10,7 @@ type Props = $Omit<
 > & {
   name: string;
   onPress?: () => void;
+  forceTextInputFocus?: Boolean;
   style?: StyleProp<ViewStyle>;
   theme?: ReactNativePaper.Theme;
 };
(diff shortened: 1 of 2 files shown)


src/components/TextInput/Adornment/Icon.tsx
```diff
@@ -10,6 +10,7 @@ type Props = $Omit<
 > & {
   name: string;
   onPress?: () => void;
+  focusable?: Boolean;
```
Trancever, line 13: Can we rename the prop to something like `forceFocus`? I think `focusable` is not a good prop name because it sounds more like whether icon should be focusable or not and in our case it is always focusable.

Trancever (changes requested):

src/components/TextInput/Adornment/Icon.tsx
brunohkbx: Done :+1:

Trancever (approved):

Issue #2056, opened 2020-07-13
iOS Chip Accessibility Issues


kotavy:
### Current behaviour

The Chip component is not accessible as expected. `accessibilityLabel` reads out, but `accessibilityHint` does not. `onClose` cannot be accessed by iOS VoiceOver.

This may be an issue for Android as well, but this project is built solely for iOS so I can't say.

### Expected behaviour

The `accessibilityHint` should be read out in VoiceOver. `onClose` should be accessible, either by enabling selection of both the chip content and the close button, or by propagating taps to the close button.

### Code sample

      <Chip
        accessibilityHint='Hints are currently not read.'
        accessibilityLabel='Labels are read.'
        // onPress offers a workaround, but some users may want both onPress and onClose
        onPress={() => onClose()}
        onClose={() => onClose()}
        mode='outlined'
      >
        My Chip
      </Chip>

### What have you tried

Inspecting the Chip structure, there doesn't seem to be a way to change the accessibility settings of the components that build the Chip itself. `onAccessibilityTap` did not work either.

I would like to update my versions, but this is not an independent project and I would have to show evidence that the issue would be resolved in order to justify the updates.

I searched the repo but could not find a similar issue, either open or closed. A web search also turned up sparse results for Chip accessibility.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | iOS
| react-native          | 0.60.5
| react-native-paper    | 3.5.1
| react-native-vector-icons    | 6.6.0
| node                  | 12.14.1
| npm or yarn           | yarn 1.17.3
| expo sdk              | N/A

brunohkbx:
Hey @kotavy. This should be fixed in the next release :smiley:

PR #2144, opened 2020-08-21
feat: improve chip accessibility

brunohkbx:
### Summary
As described in #2056, the prop `onClose` it's not accessible on iOS.

### Before (Not being possible to focus on the close icon)
Photo 21-08-20 11 22 34

### Fixed
Photo 21-08-20 11 21 01

Photo 21-08-20 11 21 35

### Android kept the same behavior
WhatsApp Image 2020-08-21 at 11 35 33 (1)

WhatsApp Image 2020-08-21 at 11 35 33

### Test plan

1. Open the example app with both Android(TalkBack) and iOS(VoiceOver)
2. Check whether Chip is accessible or not

### DIFF
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -55,6 +55,10 @@ type Props = React.ComponentProps<typeof Surface> & {
    */
   accessibilityLabel?: string;
   /**
+   * Accessibility label for the close icon. This is read by the screen reader when the user taps the close icon.
+   */
+  closeIconAccessibilityLabel?: string;
+  /**
    * Function to execute on press.
    */
   onPress?: () => void;
@@ -121,6 +125,7 @@ class Chip extends React.Component<Props, State> {
     mode: 'flat',
     disabled: false,
     selected: false,
+    closeIconAccessibilityLabel: 'Close',
   };
 
   state = {
@@ -154,6 +159,7 @@ class Chip extends React.Component<Props, State> {
       selected,
       disabled,
       accessibilityLabel,
+      closeIconAccessibilityLabel,
       onPress,
       onLongPress,
       onClose,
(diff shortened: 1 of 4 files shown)


src/components/Chip.tsx
```diff
@@ -288,27 +288,28 @@ class Chip extends React.Component<Props, State> {
...
-                  <Icon source="close-circle" size={16} color={iconColor} />
-                </View>
-              </TouchableWithoutFeedback>
-            ) : null}
           </View>
         </TouchableRipple>
+        {onClose ? (
+          <TouchableWithoutFeedback
+            onPress={onClose}
+            accessibilityTraits="button"
```
Trancever, line 304: Could you add `closeIconAccessibilityLabel` prop with a default value of "Close"? Once it's done it will be good to merge

Trancever (changes requested):

src/components/Chip.tsx
brunohkbx: Done :+1:

Trancever (approved):

src/components/Chip.tsx
```diff
@@ -154,6 +158,7 @@ class Chip extends React.Component<Props, State> {
       selected,
       disabled,
       accessibilityLabel,
+      closeIconAccessibilityLabel,
```
Trancever, line 161: @brunohkbx Sorry, one more thing. Can we have a default value "Close" for this prop? It would be best to set it with `defaultProps`

Trancever (changes requested):

src/components/Chip.tsx
brunohkbx: Ops! Totally forgot.
Should be good now, thanks.

Trancever (approved):

PR #2153, opened 2020-08-24
TouchableRipple: use React Native Web styles + performance improvements

RichardLindhout:
### Summary

Better performance with the ripple. + more react code instead of direct manipulation

### Test plan

https://codesandbox.io/s/magical-dhawan-0jn0r?file=/src/App.tsx


Trancever:
@RichardLindhout I've tried it locally and the animation doesn't seem to be as smooth as before. There is a flicker when you press and immediately release the touchable - all the touchable is highlighted

RichardLindhout:
I agree let's close it for now. I'll look into it later

PR #2165, opened 2020-08-28
fix: fix TouchableRipple crash when pressed with keyboard on web

Trancever:
### Summary

We were trying to access `touches` on the event, but it is undefined when touchable is pressed using keyboard on web.
In such a situation we simply show ripple in the middle of the touchable.

### DIFF
--- a/src/components/TouchableRipple/TouchableRipple.tsx
+++ b/src/components/TouchableRipple/TouchableRipple.tsx
@@ -115,13 +115,13 @@ class TouchableRipple extends React.Component<Props> {
     let touchY;
 
     const { changedTouches, touches } = e.nativeEvent;
+    const touch = touches?.[0] ?? changedTouches?.[0];
 
     // If centered or it was pressed using keyboard - enter or space
-    if (centered || (!changedTouches && !touches)) {
+    if (centered || !touch) {
       touchX = dimensions.width / 2;
       touchY = dimensions.height / 2;
     } else {
-      const touch = touches?.[0] ?? changedTouches?.[0];
       touchX = touch.locationX ?? e.pageX;
       touchY = touch.locationY ?? e.pageY;
     }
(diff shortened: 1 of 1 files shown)


src/components/TouchableRipple/TouchableRipple.tsx
```diff
@@ -111,17 +111,21 @@ class TouchableRipple extends React.Component<Props> {
     const style = window.getComputedStyle(button);
     const dimensions = button.getBoundingClientRect();
 
+    console.log('!@#', { style, dimensions });
```
arekkubaczkowski, line 114: do we need this console log?

arekkubaczkowski (approved):

src/components/TouchableRipple/TouchableRipple.tsx
Trancever: -.-

src/components/TouchableRipple/TouchableRipple.tsx
```diff
@@ -114,11 +114,13 @@ class TouchableRipple extends React.Component<Props> {
...
+
+    // If centered or it was pressed using keyboard - enter or space
+    if (centered || (!changedTouches && !touches)) {
       touchX = dimensions.width / 2;
       touchY = dimensions.height / 2;
     } else {
-      const { changedTouches, touches } = e.nativeEvent;
       const touch = touches?.[0] ?? changedTouches?.[0];
       touchX = touch.locationX ?? e.pageX;
       touchY = touch.locationY ?? e.pageY;
```
thymikee, line 126: seems like `touch` can still be undefined/null here?

arekkubaczkowski: AFAIR pageX and pageY are always available on event object

Trancever: Yes, theoretically it can. I will send a fix for it

thymikee (approved):

NoemiRozpara (approved):

PR #2168, opened 2020-08-31
feat: Add `imageProps` to Avatar.Image

kamui545:
New PR for https://github.com/callstack/react-native-paper/pull/1817

It will allow you to pass any props to the `<Image>` component inside of `Avatar.Image`.
Eg: `defaultSource`, `resizeMode`, etc...

### DIFF
--- a/src/components/Avatar/AvatarImage.tsx
+++ b/src/components/Avatar/AvatarImage.tsx
@@ -13,7 +13,7 @@ const defaultSize = 64;
 
 export type AvatarImageSource =
   | ImageSourcePropType
-  | (props: { size: number }) => React.ReactNode;
+  | ((props: { size: number }) => React.ReactNode);
 
 type Props = React.ComponentPropsWithRef<typeof View> & {
   /**
--- a/src/components/Avatar/AvatarImage.tsx
+++ b/src/components/Avatar/AvatarImage.tsx
@@ -1,11 +1,12 @@
 import * as React from 'react';
 import {
   Image,
+  ImageProps,
+  ImageSourcePropType,
   StyleSheet,
   View,
   ViewStyle,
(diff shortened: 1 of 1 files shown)


src/components/Avatar/AvatarImage.tsx
```diff
@@ -16,6 +17,10 @@ type Props = React.ComponentPropsWithRef<typeof View> & {
    * Image to display for the `Avatar`.
    */
   source: ImageSourcePropType;
+  /**
+   * Props for the `Image`
+   */
+  imageProps?: ImageProps;
```
Trancever, line 23: I think this API is not aligned with the rest of the lib. Can we change `source` prop to accept `ImageSourcePropType` or `({ size: number }) => React.ReactNode`? Then in the render function, we can check if `typeof source === 'function'` and call it or pass it to `Image` just like it is now.

Trancever (changes requested):

src/components/Avatar/AvatarImage.tsx
kamui545: Just made the changes, please have a look when you have time, thanks

Trancever (approved):

src/components/Avatar/AvatarImage.tsx
Trancever: 👍

aprilmintacpineda:
This is misleading, I don't see how I can pass ImageProps to Avatar.Image. The original PR contained these lines:

```diff
+     const { size = defaultSize, source, style, theme, imageProps } = this.props;
```

```diff
+ {...imageProps}
```

which will indeed allow us to pass props to image, such as `onLoad`, `onError`, etc.

PR #2174, opened 2020-09-01
fix: fix chip ripple/highlight effect not covering whole component.

Trancever:
### Summary

Followup to https://github.com/callstack/react-native-paper/pull/2144

By rendering the close button absolutely we ensure that the ripple/highlight effect is displayed on the whole chip and we keep the ability to select only the close button using a screen reader.

### DIFF
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -267,7 +267,7 @@ class Chip extends React.Component<Props, State> {
           accessibilityState={accessibilityState}
           testID={testID}
         >
-          <View style={styles.content}>
+          <View style={[styles.content, { paddingRight: onClose ? 32 : 4 }]}>
             {avatar && !icon ? (
               <View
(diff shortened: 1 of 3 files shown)


src/components/Chip.tsx
```diff
@@ -323,22 +323,24 @@ class Chip extends React.Component<Props, State> {
...
-              <MaterialCommunityIcon
-                name="close-circle"
-                size={16}
-                color={iconColor}
-                direction="ltr"
-              />
-            </View>
-          </TouchableWithoutFeedback>
+          <View style={styles.closeButtonStyle}>
+            <TouchableWithoutFeedback
```
brunohkbx, line 327: @Trancever what do you think about using a `TouchableRipple` here to also display the ripple effect when pressing on the button?

brunohkbx:
Looks good! I was able to select the whole chip with both TalkBack and VoiceOver.

Issue #2155, opened 2020-08-25
Add onLongPress to List.Item


kendallroth:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Add an `onLongPress` handler to the `List.Item` component (to complement existing `onPress` handler).

**Describe alternatives you've considered**
Likely will involve updating the `ListItem` file ([link](https://github.com/callstack/react-native-paper/blob/master/src/components/List/ListItem.tsx#L179)

**Additional context**
N/A

brunohkbx:
`List.Item` [spread](https://github.com/callstack/react-native-paper/blob/master/src/components/List/ListItem.tsx#L180) the other props to the `TouchableRipple`, So you can pass `onLongPress` even though the component does not accept. That may be worth to be specified in the docs.

kendallroth:
@brunohkbx Thank you for mentioning this, I'm not sure how I neglected to notice that while looking at the component source... 😕. At any rate, this was not working for me previously but clearly was the result of some other problem, as this does work as expected now. Updating the docs would be great though!

kendallroth:
Ah @brunohkbx I realize now why I thought it wasn't working... If `onPress={() => {}}` is not also specified (ie. with an empty function), `onLongPress` will be ignored. Is there any documentation around this anywhere (seem to remember coming across it earlier, unsure where)?

brunohkbx:
@kendallroth I had this problem once. It seems like something related to react native itself. You can try the same with a Touchable*. will only work If `onPress={() => {}}` is specified.

kendallroth:
Ah, that is interesting. It would almost be nice (sometimes) to have the `onPress` ripple disabled if not specified with `onLongPress`, but that's just wishful thinking probably (and may seem to indicate to interactive ability). Thanks for the update!

PR #2175, opened 2020-09-01
docs: render TextInput as a section and other updates

brunohkbx:
### Summary
This PR fixes the regex to render nested directories because the components `TextInputAffix` and `TextInputIcon` weren't being rendered. 

Screenshot from 2020-09-01 17-53-46

Besides, I also added the directive `@extends` in the components that are spreading props. Following up this issue #2155


Trancever:
@brunohkbx I see there is issue with sorting - TextInput is a second entry in the sidebar and it should be somewhere in the end. Also, it's not possible to see TextInput docs, when I click it it just expands/collapses the group.

docs/component-docs.config.js
```diff
@@ -64,7 +64,7 @@ function getPages() {
         acc.push({ file, group });
       }
 
-      const match = content.match(/\/\/ @component (.\/\w+\.(js|tsx?))/gm);
+      const match = content.match(/\/\/ @component (.+\/\w+\.(js|tsx?))/gm);
```
brunohkbx, line 67: This is to also match nested folders. such as: `./Adornment/TextInputIcon.tsx`

brunohkbx:
@Trancever I had to rename Affix to TextInputAffix and Icon to TextInputIcon in order to fix that.

Trancever (approved):

PR #2177, opened 2020-09-02
refactor: update snackbar class component to functional with hooks

unikvozm:
### Summary

A class component for Snackbar is rewritten to a functional one with hooks.

### Test plan

Visually the component looks and works the same as before, but 2 snapshots were updated.

### DIFF
--- a/src/components/Snackbar.tsx
+++ b/src/components/Snackbar.tsx
@@ -52,11 +52,6 @@ type Props = React.ComponentProps<typeof Surface> & {
   theme: ReactNativePaper.Theme;
 };
 
-type State = {
-  opacity: Animated.Value;
-  hidden: boolean;
-};
-
 const DURATION_SHORT = 4000;
 const DURATION_MEDIUM = 7000;
 const DURATION_LONG = 10000;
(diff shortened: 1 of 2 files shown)


src/components/Snackbar.tsx
```diff
@@ -109,179 +104,156 @@ const DURATION_LONG = 10000;
...
+  const [opacity, setOpacity] = React.useState<Animated.Value>(
+    new Animated.Value(0.0)
+  );
+  const [hidden, setHidden] = React.useState<boolean>(!visible);
 
-  /**
-   * Show the Snackbar for a medium duration.
-   */
-  static DURATION_MEDIUM = DURATION_MEDIUM;
+  let hideTimeout: ReturnType<typeof setTimeout>;
```
Trancever, line 125: I think we should keep this variable in `useRef` hook. Now even when you set it to something in the next render it will be undefined again.

Trancever (changes requested):

src/components/Snackbar.tsx
unikvozm: fixed

Trancever (changes requested):

src/components/__tests__/__snapshots__/Snackbar.test.js.snap
```diff
@@ -29,7 +29,7 @@ exports[`renders snackbar with Text as a child 1`] = `
         "flexDirection": "row",
         "justifyContent": "space-between",
         "margin": 8,
-        "opacity": 1,
```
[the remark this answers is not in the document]
unikvozm: fixed. useLayoutEffect instead of useEffect hook helped to fix the changes

Trancever (approved):

PR #2179, opened 2020-09-03
refactor: update appBar class components to functional

unikvozm:
### Summary

A class component for AppBar is rewritten to a functional one.

### DIFF
--- a/src/components/Appbar/AppbarBackAction.tsx
+++ b/src/components/Appbar/AppbarBackAction.tsx
@@ -63,16 +63,15 @@ type Props = $Omit<
  * export default MyComponent;
  * ```
  */
-class AppbarBackAction extends React.Component<Props> {
-  static displayName = 'Appbar.BackAction';
+const AppbarBackAction = ({ accessibilityLabel = 'Back', ...rest }: Props) => (
+  // @ts-ignore
+  <AppbarAction
+    accessibilityLabel={accessibilityLabel}
+    {...rest}
+    icon={AppbarBackIcon}
+  />
+);
 
-  static defaultProps = {
-    accessibilityLabel: 'Back',
-  };
-
-  render() {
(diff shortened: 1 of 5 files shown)


src/components/Appbar/AppbarBackAction.tsx
```diff
@@ -63,16 +63,15 @@ type Props = $Omit<
  * export default MyComponent;
  * ```
  */
-class AppbarBackAction extends React.Component<Props> {
-  static displayName = 'Appbar.BackAction';
+const AppbarBackAction = ({ accessibilityLabel = 'Back', ...rest }: Props) => (
+  // @ts-ignore
```
thymikee, line 67: why ts-ignore? Remember to add a reason (error message, link to TS bug, whatever that helps tackle this in the future)

unikvozm: no need anymore=)

OlimpiaZurek (approved):

Trancever (approved):

src/components/Appbar/AppbarHeader.tsx
```diff
@@ -75,58 +75,53 @@ type Props = React.ComponentProps<typeof Appbar> & {
...
-      ...restStyle
-    }: ViewStyle = StyleSheet.flatten(style) || {};
-    const backgroundColor = customBackground
-      ? customBackground
-      : isDarkTheme && mode === 'adaptive'
-      ? overlay(elevation, colors.surface)
-      : colors.primary;
-    // Let the user override the behaviour
-    const Wrapper =
-      typeof this.props.statusBarHeight === 'number' ? View : SafeAreaView;
```
Trancever, line 104: We were accessing `this.props.statusBarHeight` previously which doesn't have a default value assigned. In the new version we are assigning the default value which causes that condition on line 99 always evaluates to true and we don't render SafeAreaView.

PR #2187, opened 2020-09-04
refactor: update FAB class components to functional

unikvozm:
### Summary

Class components for FAB are rewritten to functional ones.


OlimpiaZurek (approved):

src/components/FAB/FABGroup.tsx
```diff
@@ -148,39 +143,44 @@ type State = {
...
-    animations: [],
-  };
+  const { scale } = theme.animation;
 
-  componentDidUpdate(prevProps: Props) {
-    if (this.props.open === prevProps.open) {
-      return;
-    }
+  React.useLayoutEffect(() => {
+    animations.current = actions.map(
```
Trancever, line 170: I think having this logic for updating animations.current in the effect is wrong. According to the [hooks FAQ](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops) getDerivedFromProps can be implemented in the function component by updating the property while rendering. Can we use `useEffect` instead of `useLayoutEffect` and only run the animations inside that effect? Also, can we update the `animations.current` in render whenever the actions prop changes? Please make sure that the scenario where actions prop length changes is tested and works well.

Trancever (changes requested):

PR #2199, opened 2020-09-07
refactor: update helper text from class component to functional

unikvozm:
### Summary

HelperText class component is rewritten to a functional one

### DIFF
--- a/src/components/HelperText.tsx
+++ b/src/components/HelperText.tsx
@@ -42,11 +42,6 @@ type Props = $Omit<
   testID?: string;
 };
 
-type State = {
-  shown: Animated.Value;
-  textHeight: number;
-};
-
 /**
  * Helper text is used in conjuction with input elements to provide additional hints for the user.
  *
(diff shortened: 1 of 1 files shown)


src/components/HelperText.tsx
```diff
@@ -82,109 +77,91 @@ type State = {
...
+  visible = true,
+  theme,
+  onLayout,
+  padding = 'normal',
+  ...rest
+}: Props) => {
+  // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [shown, setShown] = React.useState<Animated.Value>(
+    new Animated.Value(visible ? 1 : 0)
```
thymikee, line 92: we use refs instead of state for animated values, so they persist throughout re-renders and mutations made by Animated.timing are more obvious

unikvozm: Good to know! I'll recheck in other components and fix this.

thymikee: Something like this should do:
```js
const { current: shown } = React.useRef<Animated.Value>(
  new Animated.Value(visible ? 1 : 0)
);
``` 
It's worth to note that when the `visible` prop changes, it won't be reflected by the animated value. This behavior is the same as from the class component, but maybe it's worth to think about fixing it, e.g. through synchronizing the state through `useLayoutEffect` when `visible` changes

thymikee (approved):
Looks good. Please consider the last comment I left as a followup improvement once this is merged

src/components/HelperText.tsx
unikvozm: Can't understand your 2nd point, there is a useEffect which depends on visible and calls show or hide function  respectively.

src/components/HelperText.tsx
```diff
@@ -82,109 +77,90 @@ type State = {
...
-        this.showText();
-      } else {
-        this.hideText();
-      }
-    }
-  }
+  React.useEffect(() => {
+    if (visible) showText();
+    else hideText();
+  }, [visible, textHeight]);
```
thymikee, line 98: If we converted `textHeight` to a ref as well, then we should be able to avoid relying on its state for this hook, right? The interpolation on line 148 should be able to pick it up without component re-render

src/components/HelperText.tsx
```diff
@@ -82,109 +77,90 @@ type State = {
...
   };
 
-  private hideText = () => {
-    const { scale } = this.props.theme.animation;
-    Animated.timing(this.state.shown, {
+  const hideText = () => {
+    const { scale } = theme.animation;
+    Animated.timing(shown, {
       toValue: 0,
       duration: 180 * scale,
```
thymikee, line 113: @Trancever @satya164 is 180 vs 150ms when hiding/showing intended? 🤔 I've seen other components having it consistently on 150.

src/components/HelperText.tsx
```diff
@@ -118,7 +118,7 @@ const HelperText = ({
   const handleTextLayout = (e: LayoutChangeEvent) => {
     //@ts-ignore Animated.Text typings are improved but something is still broken. It thinks onLayout is not callable.
     onLayout?.(e);
-    setTextHeight(e.nativeEvent.layout.height);
+    textHeight.current = e.nativeEvent.layout.height;
```
thymikee, line 121: let's stay consistent and either use `animatedValue.current` or `animatedValue`, just not both :)

unikvozm: If I destructure `const {constant: 'textHeight'} = ... `, I can't assign a new value to the `textHeight` as it is a constant.
Do you mean to rewrite shown in the same way (without destructuring)?

thymikee: you'd need to use `let` then, your call

Trancever (approved):

PR #2202, opened 2020-09-07
refactor: update touchable ripple class components to functional

unikvozm:
### Summary

Touchable Ripple class components are rewritten to functional ones.

### DIFF
--- a/src/components/TouchableRipple/TouchableRipple.tsx
+++ b/src/components/TouchableRipple/TouchableRipple.tsx
@@ -85,18 +85,16 @@ type Props = React.ComponentPropsWithRef<typeof TouchableWithoutFeedback> & {
  * ```
  */
 const TouchableRipple = ({
-  /* eslint-disable @typescript-eslint/no-unused-vars */
   style,
-  background,
+  background: _background,
   borderless = false,
   disabled: disabledProp,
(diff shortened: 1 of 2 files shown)


marchenk0va (approved):

src/components/TouchableRipple/TouchableRipple.tsx
```diff
@@ -84,18 +84,21 @@ type Props = React.ComponentPropsWithRef<typeof TouchableWithoutFeedback> & {
...
+  style,
+  background,
+  borderless = false,
+  disabled: disabledProp,
+  rippleColor,
+  underlayColor,
+  children,
+  theme,
+  ...rest
+}: /* eslint-disable @typescript-eslint/no-unused-vars */
```
thymikee, line 98:
```suggestion
}: /* eslint-enable @typescript-eslint/no-unused-vars */
```

thymikee: Also, it would likely be better to prefix the variable names with `_` to hint ESLint that these are unused but we need them for reasons. I'm not 100% sure if ESLint is configured properly for this here, but it can certainly be.

Trancever (approved):

PR #2204, opened 2020-09-07
refactor: update badge class component to functional

unikvozm:
### Summary

Badge class component is rewritten to a functional one.

### DIFF
--- a/src/components/Badge.tsx
+++ b/src/components/Badge.tsx
@@ -66,16 +66,17 @@ const Badge = ({
     new Animated.Value(visible ? 1 : 0)
   );
 
+  const {
+    animation: { scale },
+  } = theme;
+
   React.useEffect(() => {
-    const {
-      animation: { scale },
-    } = theme;
     Animated.timing(opacity, {
       toValue: visible ? 1 : 0,
       duration: 150 * scale,
       useNativeDriver: true,
     }).start();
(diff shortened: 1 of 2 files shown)


src/components/Badge.tsx
```diff
@@ -58,77 +54,62 @@ type State = {
...
+  children,
+  size = defaultSize,
+  style,
+  theme,
+  visible = true,
+  ...rest
+}: Props) => {
+  // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [opacity, setOpacity] = React.useState<Animated.Value>(
```
thymikee, line 67: useRef

thymikee (approved):

src/components/Badge.tsx
```diff
@@ -75,7 +75,7 @@ const Badge = ({
       duration: 150 * scale,
       useNativeDriver: true,
     }).start();
-  }, [visible]);
+  }, [visible, opacity, theme]);
```
thymikee, line 78: `theme` is a structured object and it may cause re-render even when a small part of it changes (React checks dependencies by reference). Let's move the `scale` variable out of this effect so that it can depend on `scale` itself instead of whole `theme` object

Trancever (approved):

PR #2205, opened 2020-09-07
refactor: update banner class component to functional

unikvozm:
### Summary

Banner class component is rewritten to a functional one.

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -50,14 +50,6 @@ type Props = $RemoveChildren<typeof Surface> & {
   theme: ReactNativePaper.Theme;
 };
 
-type State = {
-  position: Animated.Value;
-  layout: {
-    height: number;
-    measured: boolean;
-  };
-};
-
 type NativeEvent = {
   nativeEvent: {
     layout: {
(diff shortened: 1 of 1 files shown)


src/components/Banner.tsx
```diff
@@ -117,132 +109,123 @@ type NativeEvent = {
...
+  children,
+  actions,
+  contentStyle,
+  style,
+  theme,
+  ...rest
+}: Props) => {
+  // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [position, setPosition] = React.useState<Animated.Value>(
```
thymikee, line 124:
```suggestion
  const { current: position } = React.useRef<Animated.Value>(
```

src/components/Banner.tsx
```diff
@@ -117,132 +109,123 @@ type NativeEvent = {
...
 
-  private toggle = () => {
-    if (this.props.visible) {
-      this.show();
-    } else {
-      this.hide();
-    }
+  const toggle = () => {
+    if (visible) show();
+    else hide();
```
thymikee, line 146: I'd inline the `toggle` function inside the effect it's used. It would make it easier to see that it depends on `visible`.

thymikee (approved):

Trancever (approved):

Issue #2161, opened 2020-08-26
testID on Checkbox


jauggy:
**Is your feature request related to a problem? Please describe.**
I want to be able to set testID for Checkbox. I use testID for testing with Detox.

**Describe the solution you'd like**
The testID should be a Checkbox prop that takes a string and then passes that prop down to the react native Checkbox

PR #2210, opened 2020-09-08
testID property for Checkbox

marchenk0va:
### Summary

#2161


unikvozm (commented):
I think that CheckboxItem needs testID as well since it can be used separately.

unikvozm (approved):

Issue #1943, opened 2020-05-22
TextInput: Using error prop function with placeholder prop set makes name overlap with placeholder


victorbutler:
## Environment

react: 16.11.0
react-native: 0.62.2
react-native-paper: 3.10.1

## Description

When TextInput is used with a dynamic error prop value, the error styles become red like expected, but the placeholder is also shown at the same time. When I set the error prop `error={true}` that field renders correctly.

![TextInput-Error-Styles](https://user-images.githubusercontent.com/1063420/82697109-8b1edc80-9c25-11ea-9d17-b7efd95d1873.png)

## Reproducible Demo

```js
import React, {useState} from 'react';
import {SafeAreaView, StyleSheet, ScrollView, View, Text} from 'react-native';
import {Button, TextInput} from 'react-native-paper';

const TestScreen: () => React$Node = ({navigation}) => {
  const [getEmail, setEmail] = useState('');
  const [hasEmailErrors, setEmailErrors] = useState(false);
  return (
    <SafeAreaView>
      <ScrollView contentInsetAdjustmentBehavior="automatic">
        <View style={styles.halfContainer}>
          <Text>Test TextInput error styles with placeholder set</Text>
          <TextInput
            style={styles.textInput}
            label="Email"
            mode="outlined"
            value={getEmail}
            onChangeText={(email) => setEmail(email)}
            textContentType="username"
            returnKeyType="next"
            keyboardType="email-address"
            autoCapitalize="none"
            error={hasEmailErrors}
            placeholder="user@example.com"
          />
        </View>
        <Button onPress={() => setEmailErrors(!hasEmailErrors)}>
          Toggle Error Styles
        </Button>
      </ScrollView>
    </SafeAreaView>
  );
};

const styles = StyleSheet.create({
  halfContainer: {
    flex: 1,
    padding: 8,
  },
  headline: {
    textAlign: 'center',
  },
  textInput: {
    paddingVertical: 5,
  },
});

export default TestScreen;
```

*Just a note of thanks and gratitude for all the work that goes into this beautiful and easy to use framework. Thank you!*

marchenk0va:
Hi @victorbutler,
Error prop pops up 'on the go', e.g. text validation. In your case we are manually toggling empty inputs, which are correct by default. It's an unexpected behaviour for an error prop. But your observation is a nice opportunity for contribution!

Taym95:
Hi, I don't think we should close this issue, right?

marchenk0va:
[...]

PR #2213, opened 2020-09-09
fix: keep TextInput placeholder when error is true

brunohkbx:
### Summary
Recently I noticed that the component `TextInput` is hiding the placeholder as in the gif below when you change the prop error from false to true. I don't think this makes sense since the placeholder is shown after focusing it again.
ezgif com-video-to-gif (4)

Snack
https://snack.expo.io/@bruno.castro/paper-textinput-bug

How to reproduce:
1. Click on the textinput
2. Click on the button `Add error` while the input is focused

### Test plan
Following up #1943 and #2032 I didn't have any problems regarding overlapping the label and the placeholder.

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -303,7 +303,7 @@ class TextInput extends React.Component<TextInputProps, State> {
         ios: false,
         default: true,
       }),
-    }).start(this.hidePlaceholder);
+    }).start();
   };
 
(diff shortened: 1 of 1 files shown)


soumya-banerjee:
Awesome! Please merge this commit.

brunohkbx:
@Trancever @matkoson could you please also review this one? :smiley:

jbinda (approved):
LGTM

PR #2216, opened 2020-09-09
refactor: update search bar class component to functional

unikvozm:
### Summary

SearchBar class component is rewritten as a functional one.


src/components/Searchbar.tsx
```diff
@@ -94,146 +102,116 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
...
-   * Focuses the input.
-   */
-  focus() {
-    return this.root && this.root.focus();
-  }
+  React.useImperativeHandle(ref, () => ({
+    focus: () => root.current?.focus(),
+    clear: () => root.current?.clear(),
+    setNativeProps: (args: Object) => root.current?.setNativeProps(args),
+    isFocused: () => root.current?.isFocused() || false,
```
thymikee, line 128: Technically this is a change in behavior. Likely better to tackle as a followup

src/components/Searchbar.tsx
```diff
@@ -94,146 +102,116 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
...
+        placeholderTextColor={colors.placeholder}
+        selectionColor={colors.primary}
+        underlineColorAndroid="transparent"
+        returnKeyType="search"
+        keyboardAppearance={dark ? 'dark' : 'light'}
+        accessibilityTraits="search"
+        accessibilityRole="search"
+        ref={(c) => {
+          root.current = c;
+        }}
```
thymikee, line 186:
```suggestion
        ref={root}
```

src/components/Searchbar.tsx
Trancever, line 128: Let's type the isFocused as `isFocused(): boolean | undefined;` and get rid of the fallback

src/components/Searchbar.tsx
```diff
@@ -94,146 +102,116 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
...
-    return this.root && this.root.setNativeProps(args);
-  }
-
-  /**
-   * Returns `true` if the input is currently focused, `false` otherwise.
-   */
-  isFocused() {
-    return this.root && this.root.isFocused();
-  }
+const Searchbar: React.RefForwardingComponent<TextInputHandles, Props> = (
```
Trancever, line 105: Can we pass native TextInput instead of custom type here?

Trancever (changes requested):

src/components/Searchbar.tsx
unikvozm: native TextInput has isFocused: () => boolean;

src/components/__tests__/Searchbar.test.js
```diff
@@ -2,6 +2,8 @@ import * as React from 'react';
 import renderer from 'react-test-renderer';
 import Searchbar from '../Searchbar.tsx';
 
+jest.mock('react-native/Libraries/Animated/src/NativeAnimatedHelper');
```
Trancever, line 5: This can be deleted after rebase with master

Trancever (changes requested):

src/components/__tests__/Searchbar.test.js
MrMuzyk: Done

src/components/Searchbar.tsx
```diff
@@ -94,67 +99,39 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
...
+    }: Props,
+    ref
+  ) => {
+    const root = React.useRef<TextInput>(null);
+
+    React.useImperativeHandle(ref, () => ({
+      focus: () => root.current?.focus(),
+      clear: () => root.current?.clear(),
+      setNativeProps: (args: Object) => root.current?.setNativeProps(args),
+      isFocused: () => root.current?.isFocused() || false,
```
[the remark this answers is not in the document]
MrMuzyk: When its not provided by arrow function anymore there is no need for that

src/components/Searchbar.tsx
```diff
@@ -94,67 +99,39 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
...
       value,
       ...rest
-    } = this.props;
+    }: Props,
+    ref
+  ) => {
+    const root = React.useRef<TextInput>(null);
+
+    React.useImperativeHandle(ref, () => ({
+      focus: () => root.current?.focus(),
```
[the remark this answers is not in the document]
MrMuzyk: done

src/components/Searchbar.tsx
```diff
@@ -94,67 +99,40 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
...
       value,
       ...rest
-    } = this.props;
+    }: Props,
+    ref
+  ) => {
+    const root = React.useRef<TextInput>(null);
+
+    React.useImperativeHandle(ref, () => ({
+      focus: root && root.current!.focus,
```
Trancever, line 123: Why do we need to check if `root` is defined? It's the `current` property on `root` that can be nullish, but root should be always truthy.

Trancever (changes requested):

src/components/Searchbar.tsx
MrMuzyk: If the `current` property can be nullish then we need to provide default values in such case for types to match

Trancever: Yeah, but `root` is not nullish, which means we should be able to simply `focus: root.current!.focus,`

MrMuzyk: ok, fixed that

src/components/Searchbar.tsx
```diff
@@ -94,67 +99,39 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
...
       value,
       ...rest
-    } = this.props;
+    }: Props,
+    ref
+  ) => {
+    const root = React.useRef<TextInput>(null);
+
+    React.useImperativeHandle(ref, () => ({
+      focus: root.current!.focus,
```
Trancever, line 123: Hmm, after looking at it again I think what we are doing here is wrong. We are basically changing the old behaviour and current version will throw an error if `root.current` is nullish for some reason.

I think we should keep the same behaviour as before which means we should only call the function when it is defined and if it is not then we shouldn't (this will throw a typescript error cause return value won't match, but we can silence it with `@ts-ignore`) - `focus: root.current?.focus`

Trancever (changes requested):

Trancever (approved):

PR #2219, opened 2020-09-10
refactor: update progress bar class component to functional

unikvozm:
### Summary

ProgressBar class component is rewritten as a functional one.


src/components/ProgressBar.tsx
```diff
@@ -65,53 +59,45 @@ const { isRTL } = I18nManager;
...
+    // Start animation the very first time when previously the width was unclear
+    if (visible && prevWidth === 0) {
+      startAnimation();
     }
-  }
-
-  private onLayout = (event: LayoutChangeEvent) => {
-    const { visible } = this.props;
-    const { width: previousWidth } = this.state;
+  }, [width]);
```
thymikee, line 92: There are numerous hooks warnings here and there. I see they're not automatically "fixed" due to linter config. Please address these warnings in all cases. They'll often require wrapping functions in `useCallback` or so, but it's better to rely on the rules of hooks instead of manually track dependencies.

thymikee (changes requested):
Rules of Hooks must be obeyed :)

src/components/ProgressBar.tsx
```diff
@@ -79,17 +79,13 @@ const ProgressBar = ({
...
-
   React.useEffect(() => {
     // Start animation the very first time when previously the width was unclear
     if (visible && prevWidth === 0) {
       startAnimation();
-    }
-  }, [width]);
+    } else if (visible) startAnimation();
+    else stopAnimation();
+  });
```
thymikee, line 88: this will re-run with every render, let's add a dependency array :)

src/components/ProgressBar.tsx
```diff
@@ -162,87 +128,90 @@ class ProgressBar extends React.Component<Props, State> {
...
+    if (visible) startAnimation();
+    else stopAnimation();
+  }, [progress, visible, startAnimation, stopAnimation]);
+
+  React.useEffect(() => {
+    // Start animation the very first time when previously the width was unclear
+    if (visible && prevWidth === 0) {
+      startAnimation();
+    }
+  }, [width, prevWidth, startAnimation, visible]);
```
Trancever, line 143: I think we don't need to put `width` in the dependency array

Trancever (changes requested):

src/components/ProgressBar.tsx
```diff
@@ -162,87 +128,90 @@ class ProgressBar extends React.Component<Props, State> {
       useNativeDriver: true,
       isInteraction: false,
     }).start();
-  };
+  }, [fade, scale]);
+
+  React.useEffect(() => {
+    if (visible) startAnimation();
+    else stopAnimation();
+  }, [progress, visible, startAnimation, stopAnimation]);
```
thymikee, line 136: We can remove progress (ESLint should hint you on that) as it's a dependency of `startAnimation`
```suggestion
  }, [visible, startAnimation, stopAnimation]);
```

unikvozm: it's strange, but ESLint hinted only about missing dependencies in useEffect(). I'll check manually other components then

thymikee: Ah yea, I tend to remove dependencies and regenerate them once in a while. Sounds like something to be improved in the plugin :)

Trancever (approved):

PR #2222, opened 2020-09-10
refactor: update cross fade icon class component to functional

unikvozm:
### Summary

CrossFadeIcon class component is rewritten as a functional one.


OlimpiaZurek (approved):

src/components/CrossFadeIcon.tsx
```diff
@@ -23,114 +23,91 @@ type Props = {
...
-              {
-                opacity: opacityPrev,
-                transform: [{ rotate: rotatePrev }],
-              },
-            ]}
-          >
-            <Icon source={this.state.previousIcon} size={size} color={color} />
-          </Animated.View>
-        ) : null}
+  }, [previousIcon, fade, scale]);
```
Trancever, line 52: I am not sure if having `previousIcon` in dependency array is enough, it will only compare the object reference. Previously we were checking the equality with the `isEqualIcon` function. @satya164 What do you think about keeping the previous previous icon in the ref and still comparing it with `isEqualIcon` function?

satya164: @Trancever sounds good to me.

src/components/__tests__/FAB.test.js
```diff
@@ -2,6 +2,9 @@ import * as React from 'react';
 import renderer from 'react-test-renderer';
 import FAB from '../FAB/FAB.tsx';
 
+jest.useFakeTimers();
+jest.mock('react-native/Libraries/Animated/src/NativeAnimatedHelper');
```
satya164, line 6: Maybe we can add this `NativeAnimatedHelper` mock in a global jest setup file instead of in each test file.

MrMuzyk: @unikvozm adviced to separate it into different PR as there are more places in the code thats already on master so here it is https://github.com/callstack/react-native-paper/pull/2309/files

MrMuzyk: therefore if the changes above will be merged to master ill simply remove these mocks from this PR

src/components/CrossFadeIcon.tsx
```diff
@@ -23,114 +23,100 @@ type Props = {
...
+    () => source
+  );
+  const [previousIcon, setPreviousIcon] = React.useState<IconSource | null>(
+    null
+  );
+  const { current: fade } = React.useRef<Animated.Value>(new Animated.Value(1));
+
+  const { scale } = theme.animation;
+
+  const previousPreviousIconRef = React.useRef<IconSource | null>();
```
Trancever, line 37: Can we do:

```suggestion
  const { current: previousPreviousIcon } = React.useRef<IconSource | null>();
```

and remove line 41?

Trancever (changes requested):

src/components/CrossFadeIcon.tsx
MrMuzyk: After changing it to 
```
  let { current: previousPreviousIcon } = React.useRef<IconSource | null>();
  React.useEffect(() => {
    previousPreviousIcon = previousIcon;
  }, [previousIcon]);
```
and removing line 41

Im getting a warning from eslint:
ESLint: Assignments to the 'previousPreviousIcon' variable from inside React Hook React.useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside React.useEffect.(react-hooks/exhaustive-deps)

Trancever: Ok then let's keep the ref as it is now, but let's remove line 41 anyway, since `previousPreviousIcon ` is used only once so there's no need to make it and we can directly use `previousPreviousIconRef.current`

MrMuzyk: Sure

Trancever (approved):

PR #2227, opened 2020-09-11
refactor: update typography class components to functional

unikvozm:
### Summary

Typography class components are rewritten as functional ones

### DIFF
--- a/src/components/Typography/StyledText.tsx
+++ b/src/components/Typography/StyledText.tsx
@@ -23,7 +23,6 @@ const StyledText = ({ theme, alpha, family, style, ...rest }: Props) => {
       style={[
         { color: textColor, ...font, textAlign: 'left', writingDirection },
         style,
-        style,
       ]}
     />
   );
--- a/src/components/Typography/StyledText.tsx
+++ b/src/components/Typography/StyledText.tsx
@@ -12,24 +12,21 @@ type Props = React.ComponentProps<typeof Text> & {
   theme: ReactNativePaper.Theme;
 };
 
-class StyledText extends React.Component<Props> {
-  render() {
-    const { theme, alpha, family, style, ...rest } = this.props;
-    const textColor = color(theme.colors.text).alpha(alpha).rgb().string();
-    const font = theme.fonts[family];
-    const writingDirection = I18nManager.isRTL ? 'rtl' : 'ltr';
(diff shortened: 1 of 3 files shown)


src/components/Typography/StyledText.tsx
```diff
@@ -12,24 +12,21 @@ type Props = React.ComponentProps<typeof Text> & {
...
-    );
-  }
-}
+  return (
+    <Text
+      {...rest}
+      style={[
+        { color: textColor, ...font, textAlign: 'left', writingDirection },
+        style,
+        style,
```
thymikee, line 26: style is duplicated
```suggestion
```

unikvozm: yes, I've noticed that. I wasn't sure to delete it causes tests to fail =( There was `style, this.props.style`.

thymikee: `style` and `this.props.style` seem to refer to the same thing, so if the test reflects that, it can safely be updated to avoid duplicates

src/components/Typography/Text.tsx
```diff
@@ -17,36 +17,32 @@ type Props = React.ComponentProps<typeof NativeText> & {
...
-      />
-    );
-  }
-}
+  return (
+    <NativeText
+      {...rest}
+      ref={(c) => {
+        root = c;
+      }}
```
thymikee, line 35: can we use hook form of ref here?
```js
const textRef = useRef();
<NativeText ref={textRef} />
```

src/components/Typography/Text.tsx
```diff
@@ -17,36 +17,32 @@ type Props = React.ComponentProps<typeof NativeText> & {
...
 
-  /**
-   * @internal
-   */
-  setNativeProps(args: Object) {
-    return this.root && this.root.setNativeProps(args);
-  }
+  React.useImperativeHandle(ref, () => ({
+    setNativeProps: (args: Object) => root?.setNativeProps(args),
+  }));
```
thymikee, line 28: @satya164 wdyt, do we need to restrict what's there in the native ref?

Trancever: I think we should keep the previous behaviour so restricting it is fine. Other methods on ref are not really that important cause e.g. `measure` or `measureInWindow` can be called on a wrapping View and it should cover all use cases.

Trancever (approved):

PR #2228, opened 2020-09-11
chore: deploy @callstack/eslint-config v10

thymikee:
### Summary

v10 comes with upgraded plugins for a11y, import order etc. 

Also:
- optimized Jest husky call to utilize `--changedSince master`
- bumped eslint (eslint-disable-* comments now accept extra text after " -- ", neat), enabled cache – helpful when precommit hook fails
- added missing babel/core peer dep to example
- fixed hook deps in CheckboxAndroid

### Test plan

CI


example/src/DrawerItems.tsx
```diff
@@ -8,6 +6,8 @@ import {
   Colors,
   useTheme,
 } from 'react-native-paper';
+import * as React from 'react';
```
Trancever, line 9: Why is the order changed? I am ok with it but it looks weird :D

thymikee: It's autofixed by eslint-plugin-import/order rule: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md

satya164: Looks weird that "react-native-paper" is imported before "react" & "react-native". What's the logic? If "react-native" comes later because it has more stuff after "react", why doesn't "react-native-paper" 🤔

package.json
```diff
@@ -74,10 +74,8 @@
     "commitlint": "^8.3.4",
     "conventional-changelog-cli": "^2.0.11",
     "dedent": "^0.7.0",
-    "eslint": "^6.7.2",
-    "eslint-config-callstack-io": "^1.1.1",
-    "eslint-plugin-prettier": "^3.0.0",
-    "eslint-plugin-react-native": "^3.5.0",
+    "eslint": "^7.8.1",
+    "eslint-plugin-prettier": "^3.1.4",
```
satya164, line 78: Any reason the callstack config doesn't include this already?

package.json
```diff
@@ -34,7 +34,7 @@
   ],
   "scripts": {
     "typescript": "tsc --noEmit",
-    "lint": "eslint --ext '.js,.ts,.tsx' .",
+    "lint": "eslint . --cache",
```
satya164, line 37: Does `eslint` support these extensions by default now? It didn't support them before

satya164: Though for cross-platform support, we need `eslint \"**/*.{js,ts,tsx}\"` since the `.` doesn't work properly on Windows

src/components/Avatar/AvatarImage.tsx
```diff
@@ -80,6 +80,7 @@ const AvatarImage = ({
       {typeof source === 'function' && source({ size })}
       {typeof source !== 'function' && (
         <Image
+          accessibilityIgnoresInvertColors
```
satya164, line 83: Is this intentional? Looks like a breaking change

src/components/Card/CardCover.tsx
```diff
@@ -68,7 +68,11 @@ const CardCover = ({ index, total, style, theme, ...rest }: Props) => {
 
   return (
     <View style={[styles.container, coverStyle, style]}>
-      <Image {...rest} style={[styles.image, coverStyle]} />
+      <Image
+        accessibilityIgnoresInvertColors
```
satya164, line 72: Looks like a breaking change

src/components/Chip.tsx
```diff
@@ -309,8 +309,6 @@ const Chip = ({
         <View style={styles.closeButtonStyle}>
           <TouchableWithoutFeedback
             onPress={onClose}
-            accessibilityTraits="button"
-            accessibilityComponentType="button"
```
satya164, line 313: I think we need these props for support for older RN. We can remove them in next major

thymikee:
Thanks @satya164 will address the feedback when I find a moment 👍🏼

PR #2247, opened 2020-09-21
Fix/modal styles

unikvozm:
### Summary

#2245

### Test plan

Merge should be done after merging `refactor/modal`

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -43,11 +43,6 @@ type Props = {
   theme: ReactNativePaper.Theme;
 };
 
-type State = {
-  opacity: Animated.Value;
-  rendered: boolean;
-};
-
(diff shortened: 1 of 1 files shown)


src/components/Modal.tsx
```diff
@@ -160,73 +151,69 @@ class Modal extends React.Component<Props, State> {
...
+            [
+              { opacity },
+              styles.content,
+              {
+                borderRadius: theme.roundness,
+                backgroundColor:
+                  theme.dark && theme.mode === 'adaptive'
+                    ? (overlay(MODAL_ELEVATION, theme.colors.surface) as string)
+                    : theme.colors.surface,
+              },
```
unikvozm, line 206: the updates are only here

Issue #2253, opened 2020-09-21
Please, add testID for RadioButton component


felipebergamin:
Please, add `testID` prop for RadioButton components.

```json
"react-native": "0.63.2",
"react-native-paper": "^4.2.0",
"react-native-vector-icons": "^7.0.0"
```

PR #2256, opened 2020-09-22
feat: add testID to RadioButton components

OlimpiaZurek:
### Summary
#2253

### DIFF
--- a/src/components/RadioButton/RadioButton.tsx
+++ b/src/components/RadioButton/RadioButton.tsx
@@ -35,6 +35,10 @@ export type Props = {
    * @optional
    */
   theme: ReactNativePaper.Theme;
+  /**
+   * testID to be used on tests.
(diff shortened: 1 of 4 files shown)


unikvozm:
I think it would be better to add a test that check a custom testID (like in Checkbox component)

src/components/__tests__/RadioButton/RadioButton.test.js
```diff
@@ -59,4 +59,14 @@ describe('RadioButton', () => {
       expect(tree).toMatchSnapshot();
     });
   });
+
+  describe('RadioButton with custom testID', () => {
+    it('rendrs properly', () => {
```
unikvozm, line 64: a small typo: `renders` =)

Trancever (approved):

PR #2259, opened 2020-09-23
[docs] add link to RN's Text.ellipsizeMode

gregfenton:
### Summary

Prior to this change, it is not obvious what valid values are.

### Test plan

(docs review)


Trancever (approved):
Thanks!

PR #2262, opened 2020-09-30
fix: onPress RadioButton not working

anymore1405:
### Summary

Fix prop `onPress` of `RadioButton` and `RadioButton.Item` not working


Trancever:
Can you provide a Snack or repository that reproduces the issue and reopen the PR if still needed? I believe the current implementation works fine.

PR #2263, opened 2020-09-30
feat: improve accessibility updating accessibilityRole and accessibilityState

brunohkbx:
### Summary
Update the following components with a more specific [accessibilityRole](https://reactnative.dev/docs/accessibility#accessibilityrole)

### Test plan
1. Download this branch and open the example app in your phone
2. Test with both TalkBack(Android) and VoiceOver(iOS)

### DIFF
--- a/src/components/RadioButton/RadioButtonGroup.tsx
+++ b/src/components/RadioButton/RadioButtonGroup.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { View } from 'react-native';
 
 type Props = {
   /**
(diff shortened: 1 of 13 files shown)


Trancever (approved):
Looks good @brunohkbx ! If you have some more time then it would be awesome if you could check if Checkbox and RadioButton.Group have proper roles assigned.

src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -66,9 +67,17 @@ export const RadioButtonContext = React.createContext<RadioButtonContextType>(
  */
 const RadioButtonGroup = ({ value, onValueChange, children }: Props) => (
   <RadioButtonContext.Provider value={{ value, onValueChange }}>
-    {children}
+    <View accessible accessibilityRole="radiogroup" style={styles.container}>
```
brunohkbx, line 70: @Trancever do you think it's necessary to have `flex: 1` here? I added just in case.

The `accessible` prop is to make the entire container focusable:
WhatsApp Image 2020-10-09 at 09 36 41 (1)

We can also select individuals elements though
WhatsApp Image 2020-10-09 at 09 36 41

Trancever: @brunohkbx I think we should get rid of that `flex: 1`. Once it's done PR will be ready to be merged! Thanks for your contributions!

PR #2269, opened 2020-10-02
fix: switch colors incorrect on react-native-web

jasperro:
React-native-web uses different switch color syntax from ios/android,
update the code to work with this syntax. See https://necolas.github.io/react-native-web/docs/?path=/docs/components-switch--active-thumb-color

### Summary

This pull request solves that the Switch component wasn't rendered correctly in react-native-web. The color of the thumb remained the default React Native color for the switch component. This may be able to be fixed in react-native-web itself?

### Test plan

I already tested my branch on web, ios and android and it works correctly

Open example app, go to switches, check if switch (thumb) color is correct.
Schermafdruk van 2020-10-02 15-34-24

### DIFF
--- a/src/components/Switch.tsx
+++ b/src/components/Switch.tsx
@@ -117,6 +117,12 @@ class Switch extends React.Component<Props> {
             onTintColor,
             thumbTintColor,
           }
+        : Platform.OS === 'web'
+        ? {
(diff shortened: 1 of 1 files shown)


Trancever (approved):
Thanks!

Issue #2270, opened 2020-10-06
BottomNavigation not animating top correctly- stuck at FAR_FAR_AWAY


no12345678:
### Current behaviour

I copied the example of BottomNavigation from the docs, all working. instead of rendering the 3 dummy components in the example, i changed 1 of them to be my component.
The first tab is loaded and displayed as it should, but when i change tab i see nothing (content is still at top: 9999px, FAR_FAR_AWAY value). it seems i have to visit each tab and only then when i will try to go back to one of those tabs it will now work as it should have.

### Expected behaviour
on tab change, opacity changes from 0 to 1 as it should, but top is stuck at 9999px and not animating to 0px

### Code sample
import React, { useEffect } from "react";
import { Text, View } from "react-native";
import Main from "./Main";
import { BottomNavigation } from "react-native-paper";

const MusicRoute = () => < Main />;

const AlbumsRoute = () => < Text >Albums</ Text >;

const RecentsRoute = () => < Text >Recents</ Text >;

export default function BottomNavigator() {
  const [index, setIndex] = React.useState(0);
  const [routes] = React.useState([
    { key: 'music', title: 'Music', icon: 'queue-music' },
    { key: 'albums', title: 'Albums', icon: 'album' },
    { key: 'recents', title: 'Recents', icon: 'history' },
  ]);

  const renderScene = BottomNavigation.SceneMap({
    music: MusicRoute,
    albums: AlbumsRoute,
    recents: RecentsRoute,
  });
    console.log(index)
  return (
    <BottomNavigation
      navigationState={{ index, routes }}
      onIndexChange={setIndex}
      renderScene={renderScene}
    />
  );
}
### What have you tried

i've tried overriding the top value but i cant seem to get to that css class or override it from the child.
tried looking into the bottom navigation implementation on git and see the code and figure out why wouldn't it change the top value, but had no luck.

### Your Environment
    "react-native": "^0.63.3",
    "react-native-paper": "^4.2.0",
    "npm": "6.14.8",
    "expo: "39.0.3",
    "react-native-vector-icons": "^7.1.0",

no12345678:
after updating everything as suggested, the issue is still there..

no12345678:
It seems to happen when the component displayed has an < Image /> element...

jasperro:
@no12345678, do you think this pull request sufficiently fixes this? Do you happen know if this workaround impacts performance?

no12345678:
[...]

PR #2276, opened 2020-10-08
fix: grey screen on first open of tab on react-native-web

jasperro:
### Summary

On react-native-web, bottom tabs weren't being rendered on the first click (grey screen), probably due to a issue in the way Animated is implemented on react-native-web. This is fixed by this patch by setting FAR_FAR_AWAY to 0px on web, as the components aren't rendered before opening the tab (on web) anyways, I think.

This fixes issue https://github.com/callstack/react-native-paper/issues/2270.

### Test plan

Switch to different tab in material-bottom-tabs in example app.

### DIFF
--- a/src/components/List/ListItem.tsx
+++ b/src/components/List/ListItem.tsx
@@ -82,11 +82,15 @@ type Props = $RemoveChildren<typeof TouchableRipple> & {
    */
   descriptionNumberOfLines?: number;
   /**
-   * Ellipsize Mode for the Title
+   * Ellipsize Mode for the Title.  One of `'head'`, `'middle'`, `'tail'`, `'clip'`.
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
I think for web, we can set `display: 'none'` instead of setting a `top` value

jasperro:
@satya164 Why is there even this top: 9999px, doesn't
```
if (!loaded.includes(index)) {
              // Don't render a screen if we've never navigated to it

              return null;
}
```
do the trick already?

jasperro:
The pull request somehow broke? This is the new one: https://github.com/callstack/react-native-paper/pull/2280. I updated the branch to display:none on web.

PR #2277, opened 2020-10-08
fix: use route keys for lazy loading tabs instead of index

satya164:
### Summary

Previously, we kept an array with indices of the routes we've visited previously. However, if routes were re-ordered/removed, the indices may not represent the actual routes that were visited previously since the routes might have a different index after re-ordering.

Using route keys instead of indices makes sure that we only render screens that were visited previously even when routes are re-ordered.

### Test plan

Run the example app and ensure that lazy lading works correctly.


Trancever (approved):
Thanks Satya!

Issue #2272, opened 2020-10-06
Appearance listeners shouldn't be set when the theme prop is passed.


Trancever:
**Is your feature request related to a problem? Please describe.**
Automatic theme handling is being set up even though the `theme` prop is passed to the `Provider`.

**Describe the solution you'd like**
`Appearance` module shouldn't be used and listeners shouldn't be set when the `theme` prop is passed to the `Provider`. 

Tests should be written to ensure the logic works properly.

PR #2279, opened 2020-10-09
feat: do not set Appearance module when theme prop is passed

OlimpiaZurek:
### Summary
PR resolves issue:  automatic theme handling is being set up even though the theme prop is passed to the Provider  (#2272)
And also contains refactor Provider component to a functional with hooks.

### DIFF
--- a/src/core/Provider.tsx
+++ b/src/core/Provider.tsx
@@ -13,44 +13,61 @@ type Props = {
   settings?: Settings;
 };
 
-type State = {
-  colorScheme: ColorSchemeName;
-  reduceMotionEnabled: boolean;
-};
+const Provider = ({ ...props }: Props) => {
+  const colorSchemeName =
+    (!props.theme && Appearance?.getColorScheme()) || 'light';
 
-export default class Provider extends React.Component<Props, State> {
-  state = {
-    reduceMotionEnabled: false,
-    colorScheme: Appearance?.getColorScheme() || 'light',
-  };
+  const [reduceMotionEnabled, setreduceMotionEnabled] = React.useState<boolean>(
+    false
+  );
+  const [colorScheme, setColorScheme] = React.useState<ColorSchemeName>(
+    colorSchemeName
+  );
 
-  async componentDidMount() {
-    AccessibilityInfo.addEventListener(
-      'reduceMotionChanged',
-      this.updateReduceMotionSettingsInfo
-    );
-    this.updateReduceMotionSettingsInfo();
-    Appearance?.addChangeListener(this.handleAppearanceChange);
-  }
-
-  componentWillUnmount() {
-    AccessibilityInfo.removeEventListener(
-      'reduceMotionChanged',
-      this.updateReduceMotionSettingsInfo
-    );
-    Appearance?.removeChangeListener(this.handleAppearanceChange);
-  }
-
-  private handleAppearanceChange = (
+  const handleAppearanceChange = (
     preferences: Appearance.AppearancePreferences
   ) => {
     const { colorScheme } = preferences;
-    this.setState({ colorScheme });
+    setColorScheme(colorScheme);
   };
 
-  private getTheme = () => {
-    const { theme: providedTheme } = this.props;
-    const { reduceMotionEnabled, colorScheme } = this.state;
+  const updateReduceMotionSettingsInfo = async () => {
+    try {
+      const reduceMotionEnabled = await AccessibilityInfo.isReduceMotionEnabled();
+      if (reduceMotionEnabled) {
+        setreduceMotionEnabled(reduceMotionEnabled);
+      }
+    } catch (err) {
+      console.warn(err);
+    }
+  };
+
+  React.useEffect(() => {
+    AccessibilityInfo.addEventListener(
+      'reduceMotionChanged',
+      updateReduceMotionSettingsInfo
+    );
+
+    updateReduceMotionSettingsInfo();
+
+    return () => {
+      AccessibilityInfo.removeEventListener(
+        'reduceMotionChanged',
+        updateReduceMotionSettingsInfo
+      );
+    };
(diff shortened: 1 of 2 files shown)


src/core/Provider.tsx
```diff
@@ -13,44 +13,61 @@ type Props = {
...
 
-type State = {
-  colorScheme: ColorSchemeName;
-  reduceMotionEnabled: boolean;
-};
+const Provider = ({ ...props }: Props) => {
+  const colorSchemeName =
+    (!props.theme && Appearance?.getColorScheme()) || 'light';
+
+  const [reduceMotionEnabled, setreduceMotionEnabled] = React.useState<boolean>(
```
Trancever, line 20: Should be `setReduceMotionEnabled `

src/core/Provider.tsx
```diff
@@ -13,44 +13,61 @@ type Props = {
...
-    colorScheme: Appearance?.getColorScheme() || 'light',
+  const handleAppearanceChange = (
+    preferences: Appearance.AppearancePreferences
+  ) => {
+    const { colorScheme } = preferences;
+    setColorScheme(colorScheme);
   };
 
-  async componentDidMount() {
+  const updateReduceMotionSettingsInfo = async () => {
```
Trancever, line 34: Let's refactor this component a bit cause it's over-complicated. Instead of passing this function, we can pass `setReduceMotionEnabled ` as the even handler on line 48 and 56 cause the listener will pass the value as an argument to the handler.

src/core/Provider.tsx
```diff
@@ -13,44 +13,61 @@ type Props = {
...
-  }
 
-  componentWillUnmount() {
-    AccessibilityInfo.removeEventListener(
-      'reduceMotionChanged',
-      this.updateReduceMotionSettingsInfo
-    );
-    Appearance?.removeChangeListener(this.handleAppearanceChange);
-  }
+    updateReduceMotionSettingsInfo();
```
Trancever, line 51: This can be removed

src/core/Provider.tsx
```diff
@@ -13,44 +13,61 @@ type Props = {
...
+    } catch (err) {
+      console.warn(err);
+    }
+  };
+
+  React.useEffect(() => {
     AccessibilityInfo.addEventListener(
       'reduceMotionChanged',
-      this.updateReduceMotionSettingsInfo
+      updateReduceMotionSettingsInfo
```
Trancever, line 48: Can we add and remove the `reduceMotionListener` conditionally similarly as we do for the `Appearance`? It would be awesome if we could also add tests for it :D

Trancever (changes requested):

src/core/Provider.tsx
```diff
@@ -13,44 +13,51 @@ type Props = {
...
   };
 
-  private getTheme = () => {
-    const { theme: providedTheme } = this.props;
-    const { reduceMotionEnabled, colorScheme } = this.state;
+  React.useEffect(() => {
+    if (!props.theme) {
+      AccessibilityInfo.addEventListener(
+        'reduceMotionChanged',
+        (reduceMotionEnabled) => setReduceMotionEnabled(reduceMotionEnabled)
```
Trancever, line 38: This creates a new function on every render. Please pass the `setReduceMotionEnabled` without creating an arrow function.

```suggestion
        setReduceMotionEnabled
```

src/core/Provider.tsx
```diff
@@ -13,44 +13,51 @@ type Props = {
...
+      );
+    }
+    return () => {
+      if (!props.theme) {
+        AccessibilityInfo.removeEventListener(
+          'reduceMotionChanged',
+          (reduceMotionEnabled) => setReduceMotionEnabled(reduceMotionEnabled)
```
Trancever, line 45:
```suggestion
          setReduceMotionEnabled
```

src/core/__tests__/Provider.test.js
```diff
@@ -49,12 +52,55 @@ describe('Provider', () => {
...
+    mockAppearance();
+    const { getByTestId } = render(createProvider(null));
+
+    expect(AccessibilityInfo.addEventListener).toHaveBeenCalled();
+    expect(getByTestId('provider-child-view').props.theme).toStrictEqual(
+      DefaultTheme
+    );
+  });
+
+  it('should not set AccessibilityInfo listeners, if there is no theme', async () => {
```
Trancever, line 71:
```suggestion
  it('should not set AccessibilityInfo listeners, if there is a theme', async () => {
```

src/core/__tests__/Provider.test.js
```diff
@@ -49,12 +52,55 @@ describe('Provider', () => {
...
-    Appearance.__internalListeners[0]({ colorScheme: 'dark' });
+    act(() => Appearance.__internalListeners[0]({ colorScheme: 'dark' }));
+    expect(getByTestId('provider-child-view').props.theme).toStrictEqual(
+      DarkTheme
+    );
+  });
+
+  it('should set AccessibilityInfo listeners, if there is no theme', async () => {
+    mockAppearance();
+    const { getByTestId } = render(createProvider(null));
```
Trancever, line 63: can we rerender the component with a theme prop to check if the listener will be removed?

src/core/__tests__/Provider.test.js
```diff
@@ -49,12 +52,55 @@ describe('Provider', () => {
...
+
+    expect(AccessibilityInfo.addEventListener).toHaveBeenCalled();
+    expect(getByTestId('provider-child-view').props.theme).toStrictEqual(
+      DefaultTheme
```
Trancever, line 67: We should call the listener, return true/false, and check if `scale` in the theme has a correct value.

Trancever (changes requested):

Trancever (approved):

PR #2280, opened 2020-10-09
fix: grey screen on first open of tab on react-native-web

jasperro:
Summary

On react-native-web, bottom tabs weren't being rendered on the first click (grey screen), probably due to a issue in the way Animated is implemented on react-native-web. This is fixed by this patch by setting FAR_FAR_AWAY to 0px on web, and setting display:none, like in standard react-navigation bottom tabs.

This fixes issue #2270.
Test plan

Switch to different tab in material-bottom-tabs in example app.

### DIFF
--- a/src/components/BottomNavigation.tsx
+++ b/src/components/BottomNavigation.tsx
@@ -731,7 +731,15 @@ class BottomNavigation extends React.Component<Props, State> {
                   Platform.OS === 'ios' ? navigationState.index !== index : true
                 }
               >
-                <Animated.View style={[styles.content, { top }]}>
+                <Animated.View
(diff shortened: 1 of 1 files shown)


RichardLindhout:
Maybe this is because of the flex bug which is fixed in React Native 0.14.0.

jasperro:
@RichardLindhout The only thing I know is that react-navigation is using the exact same fix for the bugginess, and I just implemented their fix.

jasperro:
@satya164 What do you think of the changes? I would like the bug to be fixed :)

satya164 (approved):

PR #2293, opened 2020-10-15
feat(Appbar): use height based on the platform

brunohkbx:
### Summary
Appbar is using the wrong height on iOS. This was noticed when using Appbar as a custom header in react-navigation.

This implementation still differs from [react-navigation](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/views/Header/HeaderSegment.tsx#L56) because we aren't checking if the orientation is landscape.

### Test plan
1. Open the example app on iOS
2. Check if the header has a height of 44px

### DIFF
--- a/src/components/Appbar/Appbar.tsx
+++ b/src/components/Appbar/Appbar.tsx
@@ -31,7 +31,11 @@ type Props = Partial<React.ComponentPropsWithRef<typeof View>> & {
   style?: StyleProp<ViewStyle>;
 };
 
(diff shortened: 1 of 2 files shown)


Trancever:
@brunohkbx Thanks for the contribution. Unfortunately, this is a breaking change and we need to wait with this change till the next major version. We will keep it open, cause we definitely want to have it merged when the right time comes.

ottob:
Is it possible to work around this with theming or styling?

PR #2296, opened 2020-10-19
feat: add uppercase prop to FAB

haakandev:
### Summary

Similar to the Button component, in some instances you might not want to uppercase the FAB text. This PR adds that option.

### Test plan

An instance of a non-uppercased FAB has been added to the example app.
1. Verify that by default, the FAB still has uppercase text
2. Verify that the FAB with `uppercase={false}` does not force uppercase

### DIFF
--- a/example/src/Examples/FABExample.tsx
+++ b/example/src/Examples/FABExample.tsx
@@ -43,7 +43,14 @@ const ButtonExample = () => {
           visible={visible}
           disabled
         />
-
+        <FAB
+          icon="format-letter-case"
+          label="Mixed case"
+          style={styles.fab}
+          onPress={() => {}}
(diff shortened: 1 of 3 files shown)


Trancever (approved):

Trancever:
@haakandev Thanks for the PR! It would be awesome if you could create another one for the Button component 🙏

haakandev:
> @haakandev Thanks for the PR! It would be awesome if you could create another one for the Button component 🙏

Thanks for the review @Trancever !
Happy to have a look at the Button component as well, but I believe it already has the `uppercase` prop added here. Let me know if I misunderstood what you meant! 😄

Trancever:
Oh yeah, you're right!

Issue #1449, opened 2019-11-05
Support for 'bold' namespace in fonts


robertLichtnow:
## Describe the feature

Support for `bold` namespace in the `fonts` namespace of the theme

## Motivation

The reasons is getting closer to what material design showcases as 'common weights' in here https://material.io/design/typography/understanding-typography.html

## Related Issues

No related issues

satya164:
Can you clarify your use case?

robertLichtnow:
I'm making a design system for web (wrapping material ui) and react native (wrapping react native paper) that supports theming and one of the premises of the theme is that the fonts have the `'light' | 'regular' | 'medium' | 'thin' | 'bold'` configurations.

satya164:
You can put any additional stuff in the theme. I don't think it makes sense for us to add a `bold` to the default theme when we don't use it.

joebochill:
@satya164 I have a similar use case to @robertLichtnow. We are writing a component library that builds off of react-native-paper. We have our own theme file (where we can add bold), but we would like our components to be compatible with the default theme (or custom themes) as well. If we specify bold and it's not there, it defaults to regular, which is less than ideal.

joaovpmamede:
@satya164 I've added the custom fonts to the:
```
namespace ReactNativePaper {
    interface ThemeFonts {
      bold: ThemeFont;
    }
  }
}
```

but now the `theme` object passed to the `<PaperProvider theme={theme}>` is giving me:

```
Types of property 'fonts' are incompatible.
Type 'Fonts' is missing the following properties from type 'ThemeFonts': bold,
```

Any way of augmenting the `Fonts` type?

likon416:
> You can put any additional stuff in the theme. I don't think it makes sense for us to add a `bold` to the default theme when we don't use it.

Doing a theme manager where you can't set bold because they dont use it. Wow, great idea.

lghimfus:
> @satya164 I've added the custom fonts to the:
> 
> ```
> namespace ReactNativePaper {
>     interface ThemeFonts {
>       bold: ThemeFont;
>     }
>   }
> }
> ```
> 
> but now the `theme` object passed to the `<PaperProvider theme={theme}>` is giving me:
> 
> ```
> Types of property 'fonts' are incompatible.
> Type 'Fonts' is missing the following properties from type 'ThemeFonts': bold,
> ```
> 
> Any way of augmenting the `Fonts` type?

did you find a way to fix this?

[...]

PR #2299, opened 2020-10-20
feat: add bold option to font namespace

joebochill:
### Summary

This PR adds `bold` to the type definition for `ThemeFont` (refer to #1449). 

This can also be achieved by extending the global ReactNativePaper namespace to add the additional font weight (as described in the theming docs and also demonstrated in the example with _superLight_). However, someone using this approach loses the ability to swap between the default React Native Paper theme and their custom theme (or they need to add a bunch of conditional checks to check for the existence of the bold font anywhere they want to use it).

The bold weight is a standard of the Material Design type system (https://material.io/design/typography/understanding-typography.html) and so should be included in React Native Paper even though it is not explicitly used in any of the components (yet). 

### Test plan

In `example/src/Examples/TextExample.tsx` add an example that utilizes the bold font family/weight to verify that it is available and works as expected:
```tsx
const TextExample = () => {
  const {
    colors: { background },
    fonts
  } = useTheme();

  return (
    <View style={[styles.container, { backgroundColor: background }]}>
      <Caption style={styles.text}>Caption</Caption>
      <Paragraph style={styles.text}>Paragraph</Paragraph>
      <Subheading style={styles.text}>Subheading</Subheading>
      <Title style={styles.text}>Title</Title>
      <Headline style={styles.text}>Headline</Headline>
      <Headline style={{fontFamily: fonts.bold.fontFamily, fontWeight: fonts.bold.fontWeight}}>I'm Bold</Headline>
    </View>
  );
};
```

### DIFF
--- a/docs/pages/2.theming.md
+++ b/docs/pages/2.theming.md
@@ -67,6 +67,7 @@ A theme usually contains the following properties:
 - `fonts` (`object`): various fonts used throughout different elements.
   - `regular`
   - `medium`
+  - `bold`
   - `light`
   - `thin`
 - `animation` (`object`)
--- a/docs/pages/4.fonts.md
+++ b/docs/pages/4.fonts.md
@@ -63,6 +63,10 @@ const fontConfig = {
       fontFamily: 'sans-serif-medium',
       fontWeight: 'normal',
     },
+    bold: {
+      fontFamily: 'sans-serif-bold',
+      fontWeight: 'normal',
+    },
     light: {
       fontFamily: 'sans-serif-light',
       fontWeight: 'normal',
--- a/src/components/Typography/StyledText.tsx
+++ b/src/components/Typography/StyledText.tsx
@@ -7,7 +7,7 @@ import { withTheme } from '../../core/theming';
 
 type Props = React.ComponentProps<typeof Text> & {
   alpha: number;
-  family: 'regular' | 'medium' | 'light' | 'thin';
+  family: 'regular' | 'medium' | 'bold' | 'light' | 'thin';
   style?: StyleProp<TextStyle>;
   theme: ReactNativePaper.Theme;
 };
--- a/src/components/__tests__/__snapshots__/ListSection.test.js.snap
+++ b/src/components/__tests__/__snapshots__/ListSection.test.js.snap
@@ -31,6 +31,10 @@ exports[`renders list section with custom title style 1`] = `
       },
       "dark": false,
       "fonts": Object {
+        "bold": Object {
+          "fontFamily": "System",
+          "fontWeight": "700",
+        },
         "light": Object {
           "fontFamily": "System",
           "fontWeight": "300",
@@ -363,6 +367,10 @@ exports[`renders list section with subheader 1`] = `
       },
       "dark": false,
       "fonts": Object {
+        "bold": Object {
+          "fontFamily": "System",
+          "fontWeight": "700",
+        },
         "light": Object {
           "fontFamily": "System",
           "fontWeight": "300",
@@ -693,6 +701,10 @@ exports[`renders list section without subheader 1`] = `
       },
       "dark": false,
       "fonts": Object {
+        "bold": Object {
+          "fontFamily": "System",
+          "fontWeight": "700",
+        },
         "light": Object {
           "fontFamily": "System",
           "fontWeight": "300",
--- a/src/styles/fonts.tsx
+++ b/src/styles/fonts.tsx
@@ -11,6 +11,10 @@ const fontConfig = {
       fontFamily: 'Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif',
       fontWeight: '500' as '500',
     },
+    bold: {
+      fontFamily: 'Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif',
+      fontWeight: '700' as '700',
+    },
     light: {
       fontFamily: 'Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif',
       fontWeight: '300' as '300',
@@ -29,6 +33,10 @@ const fontConfig = {
       fontFamily: 'System',
       fontWeight: '500' as '500',
     },
+    bold: {
+      fontFamily: 'System',
+      fontWeight: '700' as '700',
+    },
     light: {
       fontFamily: 'System',
       fontWeight: '300' as '300',
@@ -45,6 +53,10 @@ const fontConfig = {
     },
     medium: {
       fontFamily: 'sans-serif-medium',
+      fontWeight: 'normal' as 'normal',
+    },
+    bold: {
+      fontFamily: 'sans-serif-bold',
       fontWeight: 'normal' as 'normal',
     },
     light: {
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -19,6 +19,7 @@ export type Font = {
 export type Fonts = {
   regular: Font;
   medium: Font;
+  bold: Font;
   light: Font;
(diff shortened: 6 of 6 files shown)


joebochill:
@Trancever @satya164 any thoughts on this?

Trancever:
@joebochill We would prefer not to add properties that are not used internally by the lib if the use case can be implemented on the user side.  Could you elaborate more on the issue you have with extending the global ReactNativePaper namespace?

My understanding is that, if you support multiple themes in your app, all themes should have the same type/shape. Thanks to the same type, you won't need to add any checks property existence.

Let me know if I am missing something :D

joebochill:
@Trancever totally agree with you in the case of an app with multiple themes. In our particular case, we're building a component library that builds on RNP, so it would be used across a variety of different applications. Some of these components need a bold font internally. But since we don't ultimately know what kind of applications these components may end up in, we'd like to be able to support using any themes that we create _plus_ the default RNP theme if possible.

I'm open to extending the global namespace from within our library if that would work (have to wrap my head around where exactly I would put that). But if we extend it to make the bold font required, we break compatibility with the default theme, and if we make it optional, we're back to conditionally applying styles wherever we need bold.

joebochill:
To expand on that, these end-user applications may use:
- Our components + our themes
- Our components + their own custom themes
- Our components + default RNP theme
- Just RNP components + our themes
- or other combinations of the above
 
Ideally we'd like to simplify things by having all of these use cases operating from a single definition/type of `Theme`

SleeplessByte:
> I'm open to extending the global namespace from within our library if that would work (have to wrap my head around where exactly I would put that). 

It likely won't work because it's defined as a _type_ and not an interface. Interfaces can be augmented, Types can not; I'm hopeful that I'm wrong on this.

Here is how I solved this:

```typescript

export const FONTS = {
  web: {
    regular: {
      fontFamily: '"Helvetica Neue", Whatever_400Regular, Arial, sans-serif',
      fontWeight: '400' as const,
    },
    medium: {
      fontFamily: '"Helvetica Neue", Whatever_500Medium, Arial, sans-serif',
      fontWeight: '500' as const,
    },
    light: {
      fontFamily: '"Helvetica Neue", Whatever_300Light, Arial, sans-serif',
      fontWeight: '300' as const,
    },
    thin: {
      fontFamily: '"Helvetica Neue", Whatever_100Thin, Arial, sans-serif',
      fontWeight: '100' as const,
    },
    // @ts-ignore
    semibold: {
      fontFamily: '"Helvetica Neue", Whatever_600SemiBold, Arial, sans-serif',
      fontWeight: '600' as const,
    },
    bold: {
      fontFamily: '"Helvetica Neue", Whatever_700Bold, Arial, sans-serif',
      fontWeight: 'bold' as const,
    },
    extrabold: {
      fontFamily: '"Helvetica Neue", Whatever_800ExtraBold, Arial, sans-serif',
      fontWeight: '800' as const,
    },
  },
  ios: {
    regular: {
      fontFamily: 'System',
      fontWeight: '400' as const,
    },
    medium: {
      fontFamily: 'System',
      fontWeight: '500' as const,
    },
    light: {
      fontFamily: 'System',
      fontWeight: '300' as const,
    },
    thin: {
      fontFamily: 'System',
      fontWeight: '100' as const,
    },
    // @ts-ignore
    semibold: {
      fontFamily: 'System',
      fontWeight: '600' as const,
    },
    bold: {
      fontFamily: 'System',
      fontWeight: 'bold' as const,
    },
    extrabold: {
      fontFamily: 'System',
      fontWeight: '800' as const,
    },
  },
  android: {
    regular: {
      fontFamily: 'Whatever_400Regular',
      fontWeight: '400' as const,
    },
    medium: {
      fontFamily: 'Whatever_500Medium',
      fontWeight: '500' as const,
    },
    light: {
      fontFamily: 'Whatever_300Light',
      fontWeight: '300' as const,
    },
    thin: {
      fontFamily: 'Whatever_100Thin',
      fontWeight: '100' as const,
    },
    // @ts-ignore
    semibold: {
      fontFamily: 'Whatever_600SemiBold',
      fontWeight: '600' as const,
    },
    bold: {
      fontFamily: 'Whatever700Bold',
      fontWeight: 'bold' as const,
    },
    extrabold: {
      fontFamily: 'Whatever_800ExtraBold',
      fontWeight: '800' as const,
    },
  },
  default: {
    regular: {
      fontFamily:
        '"Helvetica Neue", Whatever_400Regular, Arial, Roboto, sans-serif',
      fontWeight: '400' as const,
    },
    medium: {
      fontFamily:
        '"Helvetica Neue", Whatever_500Medium, Arial, Roboto, sans-serif-medium',
      fontWeight: '500' as const,
    },
    light: {
      fontFamily:
        '"Helvetica Neue", Whatever_300Light, Arial, Roboto, sans-serif-light',
      fontWeight: '300' as const,
    },
    thin: {
      fontFamily:
        '"Helvetica Neue", Whatever_100Thin, Arial, Roboto, sans-serif-thin',
      fontWeight: '100' as const,
    },
    // @ts-ignore
    semibold: {
      fontFamily:
        '"Helvetica Neue", Whatever_600SemiBold, Arial, Roboto, sans-serif',
      fontWeight: '600' as const,
    },
    bold: {
      fontFamily:
        '"Helvetica Neue", Whatever_700Bold, Arial, Roboto, sans-serif',
      fontWeight: 'bold' as const,
    },
    extrabold: {
      fontFamily:
        '"Helvetica Neue", Whatever_800ExtraBold, Arial, Roboto, sans-serif',
      fontWeight: '800' as const,
    },
  },
};

export const BASE_FONTS = Platform.select(FONTS);
export const BASE_THEME: Theme = {
  ...DefaultTheme,
  colors: {
    ...DefaultTheme.colors,
    primary: PRIMARY_LIGHT,
    accent: PASTEL_BLUE,
  },
  fonts: configureFonts(FONTS),
};
```

And then somewhere:

```tsx
<Text style={{ fontSize: 24, ..BASE_FONTS.semibold }}>
  Not perfect
</Text>
```

satya164:
@joebochill

> But if we extend it to make the bold font required, we break compatibility with the default theme, and if we make it optional, we're back to conditionally applying styles wherever we need bold.

I think that's necessary compromise you have to do when building upon paper's theme. Bold font is just one of the things, surely you'll need more things in the theme in future, and it won't be possible to add those to the default theme because another library uses them.

Once you start building upon the components and adding things to theme, the best option is to provide your own theme which extends the default theme, and not use the default theme directly.

Another option you might want to consider is to wrap Paper's `Provider` and `useTheme` hooks with your own asbtraction. Then you can transparently add more stuff to the theme.

@SleeplessByte

> It likely won't work because it's defined as a type and not an interface

https://github.com/callstack/react-native-paper/blob/eb421bc16747e89e9bc0a96d55deefab6f42680f/src/types.tsx#L77

SleeplessByte:
(yep, as I expected)

FWIW, in other projects we do exactly what @satya164 suggest: the theme we expose and ingest is called Theme, and the react-native paper theme (and react-navigation theme) are derived from it.

One interface to export (it's a type, just like here) and no need to change any of the libraries.

joebochill:
> I think that's necessary compromise you have to do when building upon paper's theme. Bold font is just one of the things, surely you'll need more things in the theme in future, and it won't be possible to add those to the default theme just because another library uses them.
> 
> Once you start building upon the components and adding things to theme, the best option is to provide your own theme which extends the default theme, and not use the default theme directly.

I think you are right. This is probably the most logical direction to go in given that there may be other things we need to add in the future. I appreciate the suggestions.

PR #2301, opened 2020-10-21
chore: exported Prop types for components

justindra:
### Summary

Often we would like to extend some components from `react-native-paper` and would like to extend the props so that consumer of our custom components will still pass correct props to the component below. To help with this, we would need the Props typing to be exported out. So I have now exported out the Props

### Test plan

* Run tests as per [contribution-guidelines](https://github.com/callstack/react-native-paper/blob/master/CONTRIBUTING.md#linting-and-tests)

* Import the component and the Props and extend them to make sure it works
```tsx
// example-custom-button.tsx
import React from 'react';
import { Button, ButtonProps } from 'react-native-paper';

export const CustomBlueCircleButton: React.FC<ButtonProps> = ({ children, ...props }) => {
  return (
    <Button mode="outlined" compact style={{
      backgroundColor: 'blue'
      borderRadius: 50,
    }} {...props}>{children}</Button>
  )
}

```

### DIFF
--- a/src/components/ActivityIndicator.tsx
+++ b/src/components/ActivityIndicator.tsx
@@ -8,6 +8,7 @@ import {
   View,
   ViewStyle,
 } from 'react-native';
+import type { SetPropAsOptional } from '../types';
 import { withTheme } from '../core/theming';
 
 type Props = React.ComponentPropsWithRef<typeof View> & {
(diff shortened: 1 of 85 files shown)


Trancever:
To access Prop types of any component you can use: `React.ComponentProps<typeof Component>`, so for `Button` it will look like this: `React.ComponentProps<typeof Button>`.

PR #2310, opened 2020-10-23
fix: cross fade icons cr fixes

MrMuzyk:
### Summary

Corrected some code after CR

### DIFF
--- a/src/components/CrossFadeIcon.tsx
+++ b/src/components/CrossFadeIcon.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { Animated, StyleSheet, View } from 'react-native';
-import Icon, { isValidIcon, IconSource } from './Icon';
+import Icon, { isValidIcon, IconSource, isEqualIcon } from './Icon';
 
 import { withTheme } from '../core/theming';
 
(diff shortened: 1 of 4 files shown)


src/components/CrossFadeIcon.tsx
```diff
@@ -34,13 +34,22 @@ const CrossFadeIcon = ({ color, size, source, theme }: Props) => {
 
   const { scale } = theme.animation;
 
+  const previousPreviousIconRef = React.useRef<IconSource | null>();
+  React.useEffect(() => {
+    previousPreviousIconRef.current = previousIcon;
+  });
```
unikvozm, line 40: I think we need to add  `previousIcon` as a dependency, otherwise it will be triggered every render

src/components/CrossFadeIcon.tsx
```diff
@@ -34,13 +34,22 @@ const CrossFadeIcon = ({ color, size, source, theme }: Props) => {
 
   const { scale } = theme.animation;
 
+  const previousPreviousIconRef = React.useRef<IconSource | null>();
```
unikvozm, line 37: we can use destructuring like `let {current: previousPreviousIconRef} = React.useRef<IconSource | null>(null)`

src/components/CrossFadeIcon.tsx
MrMuzyk: Fixed that

unikvozm (approved):
lgtm

PR #2322, opened 2020-10-28
fix: read prop without default value in appbar header

MrMuzyk:
### Summary

statusBarHeight prop is now used without default value in condition whether to use View or SafeAreaView


Trancever (approved):

Trancever:
Looks good, Thanks!

PR #2324, opened 2020-10-29
feat: add accessibility state expanded to fab group

brunohkbx:
### Summary
Following up #2263. I missed this one.

### Test plan
1. Open the example
2. Check whether the screen reader will read the `expanded` state

### DIFF
--- a/src/components/FAB/FAB.tsx
+++ b/src/components/FAB/FAB.tsx
@@ -13,6 +13,7 @@ import { black, white } from '../../styles/colors';
 import { withTheme } from '../../core/theming';
 import type { $RemoveChildren } from '../../types';
 import type { IconSource } from './../Icon';
+import type { AccessibilityState } from 'react-native';
 
 type Props = $RemoveChildren<typeof Surface> & {
   /**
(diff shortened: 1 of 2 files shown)


src/components/FAB/FAB.tsx
```diff
@@ -32,6 +33,10 @@ type Props = $RemoveChildren<typeof Surface> & {
    * Uses `label` by default if specified.
    */
   accessibilityLabel?: string;
+  /**
+   * Accessibility state for the FAB. This is read by the screen reader when the user taps the FAB.
+   */
+  accessibilityState?: AccessibilityState;
```
Trancever, line 39: @brunohkbx Do we really need this prop? What is the use case? What user could pass here?

brunohkbx: @Trancever that's because I have to pass whether the fab group is expanded to the fab, here: https://github.com/callstack/react-native-paper/pull/2324/files#diff-0132f02d391233c0c9c4ae3a00199078daa9184bb0986dc715ab75ce8ddb7086R346

Trancever: ahh ok, makes sense

Trancever (approved):

PR #2333, opened 2020-11-01
fix: fix oversight in BottomNavigation

jasperro:
Replaced index.toString() by route.key, as loaded.includes needs the key string (album, library) instead of index (0, 1)

### Summary

This code worked previously with my own repo, but using the new 4.3.0 release didn't work with bottom navigation. `loaded.includes` needs the key string (album, library) instead of index (0, 1), this pull changes that to be right. Somehow I missed this in my pull request https://github.com/callstack/react-native-paper/pull/2280.

### Test plan

`yarn example web`, open bottom navigation and switch between tabs.

### DIFF
--- a/src/components/BottomNavigation.tsx
+++ b/src/components/BottomNavigation.tsx
@@ -737,9 +737,7 @@ class BottomNavigation extends React.Component<Props, State> {
                     { top },
                     Platform.OS === 'web'
                       ? {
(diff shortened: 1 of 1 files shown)


Trancever (approved):
Thanks!

Trancever (approved):

PR #2337, opened 2020-11-03
Clearer example code

gregfenton:
### Summary

Overloading of `value` in the original example makes it hard for newcomers to truly understand what is happening in this code.

### Test plan

Should work with whatever existing testing there is for this example.

### DIFF
--- a/src/components/RadioButton/RadioButtonGroup.tsx
+++ b/src/components/RadioButton/RadioButtonGroup.tsx
@@ -49,7 +49,7 @@ export const RadioButtonContext = React.createContext<RadioButtonContextType>(
  *   const [value, setValue] = React.useState('first');
  *
  *   return (
- *     <RadioButton.Group onValueChange={value => setValue(value)} value={value}>
+ *     <RadioButton.Group onValueChange={newValue => setValue(newValue)} value={value}>
  *       <View>
  *         <Text>First</Text>
  *         <RadioButton value="first" />


gregfenton:
Another change I suggest is to enhance the `onValueChange` to be a bit more involved, enabling "toggling" of the selected radio value:

```js
const toggleableValue = (newValue) => {
  if (newValue !== value) {
    setValue(newValue);
  } else {
    setValue(null);  // clicked already selected value, so un-check it
  }
};

// ...

<RadioButton.Group onValueChange={newValue => toggleableValue(newValue)} value={value}>
```

Trancever (approved):

Trancever:
@gregfenton Thanks for this PR! I don't really understand your comment about the enhancement. RadioButton.Group can't work as you explained, there should always be at least one radio button selected.

gregfenton:
I disagree with "always be at least one button selected".  If I have a form with a Yes or No question on it, initially the user has nothing selected so there is not one button selected.  If the user presses a button - on purpose or possibly fat-fingering some other field - and they aren't prepared to answer the yes/no question, I would like the option to allow them to de-select.  That would be returning the field response to its original state.

There are several UI libraries that support this use case for their radio buttons.  It is akin to the three-state checkbox: CHECKED, UNCHECKED, and INITIAL/DEFAULT VALUE (typically rendered as a grey box).

PR #2344, opened 2020-11-04
fix: allow bluring of disabled field

haakandev:
### Summary

When trying to blur and set a TextInput as disabled at the same time, the blur does not take effect.
Example use case:
- Form of TextInputs (and other inputs) where fields have their `disabled` prop set to true while submitting
- While a TextInput is focused, user clicks a submit button that sets the above mentioned prop to loading
Expected: The TextInput is blurred
Actual: The TextInput is not properly blurred and left in a semi-focused state (no keyboard and greyed out, but but showing thicker underline)

See https://snack.expo.io/r8Cpj71On for iOS/Android
1. Having the first TextInput focused when clicking the button works fine
2. Focusing the second TextInput and clicking the button leaves the TextInput with a thicker underline as it is still in focused state

This PR fixes this by removing the check on `disabled` prop in `handleBlur`. If this existed for some other purpose, please let me know and maybe there's a different solution.

### Test plan

See Snack above for testing

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -376,7 +376,7 @@ class TextInput extends React.Component<TextInputProps, State> {
   };
 
   private handleBlur = (args: Object) => {
-    if (this.props.disabled || !this.props.editable) {
+    if (!this.props.editable) {
       return;
     }
 


jbinda (approved):
LGTM

PR #2349, opened 2020-11-10
fix: avoid initial animation in (Checkbox|RadioButton).Android

jaulz:
### Summary

Currently, whenever the `Checkbox.Android` is rendered there is a very short animation that is irritating and not necessary.

### Test plan

1. Check any `Checkbox.Android` and see if there is an initial animation (even easier if you set the animation scale in theme to a high number).


brunohkbx:
@jaulz the component RadioButton has the same problem. could you please verify this one as well?

jaulz:
@brunohkbx sure, it's fixed as well now.

Trancever:
Thanks @jaulz !

Trancever (approved):

PR #2350, opened 2020-11-11
fix: set line height to Icon

jaulz:
### Summary

Currently, the icon is slightly shifted to the bottom on web because there is no line height. In this example the size is 24 but the actual height of the rendered icon element is 28:
image

If we pass a line height as well the height of the icon element will be 24 (as expected):
image

### Test plan

1. Check if icon (for example in Appbar) is not shifted to the bottom anymore


Trancever (approved):

Trancever:
@jaulz Thanks!

PR #2352, opened 2020-11-11
Allow DataTable.Cell text to wrap on multiple lines

cmnstmntmn:
This PR will allow text to wrap by specifying the `numberOfLines`

```jsx
<DataTable.Row>
    <DataTable.Cell numberOfLines={3}>
...
```


Trancever:
Hey @cmnstmntmn This is explicitly disallowed by MD Guidelines. If you really need this functionality, then you should use a `View` instead of `DataTable.Cell`.

Issue #2185, opened 2020-09-03
textAlign is not supported for TextInput


exneval:
### Current behaviour

TextInput from react-native-paper doesn't support textAlign

### Expected behaviour

textAlign should work as TextInput of react-native

### What have you tried

1. First importing TextInput from react-native-paper then use it
2. Add textAlign both from props or styles object
3. textAlign doesn't work

Reference from closed issue (by Author): [issue](https://github.com/callstack/react-native-paper/issues/1076)

### Your Environment

| software              | version
| --------------------- | -------
| android        | 9.0 (Pie)
| react-native          | 0.63.2
| react-native-paper    | 4.1.0
| node                  | 12.16.1
| npm          | 6.13.4

gsevla:
Could you provide an snack example or something like that? If it what I'm thinking about, you need to stretch text on textinput to achieve alignment.

bencart:
https://snack.expo.io/FiIAsnjVT

RichardLindhout:
It's because it is overwritten it also should be inside the style prop if I am not mistaken.

https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput/TextInputFlat.tsx#L352
https://github.com/callstack/react-native-paper/blob/master/src/components/TextInput/TextInputOutlined.tsx#L331

Also the behaviour is 'center'?

RichardLindhout:
It should be inside the styles object but React Native Paper does not support it indeed.
Updated snack with working RN and broken RNP example: https://snack.expo.io/ResdgTixY

I'll try to work on a PR next week to provide a solution for this

RichardLindhout:
I have a working solution with tests and tested on all platforms and it works correctly!

PR is up here: https://github.com/callstack/react-native-paper/pull/2369

Kommandat:
Following up on this issue. I looked at [this PR](https://github.com/callstack/react-native-paper/pull/2369) and saw it fixed `textAlign : "center"` but is `textAlign: "right"` still not supported for `react-native-paper` text inputs? Thanks!

PR #2369, opened 2020-11-20
fix: add support for textAlign: center on TextInput

RichardLindhout:
### Summary
fixes: https://github.com/callstack/react-native-paper/issues/2185

### Test plan
It has been testen on iOS, Android, web and works correctly
Added 2 snapshot tests
- Default behaviour 
- Custom style behaviour


.expo/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "hostType": "lan",
+  "lanType": "ip",
+  "dev": true,
```
Trancever, line 4: Could you remove that file from the PR?

Trancever (changes requested):

.expo/settings.json
RichardLindhout: Yes will do, will also add it in the .gitignore so it won't be added for other people too

Trancever (approved):

PR #2375, opened 2020-11-23
feat: let users choose checkbox for Checkbox.Item

tegandbiscuits:
### Summary

This makes it so developers are able to specify which checkbox component should be used for `Checkbox.Item`.

My main motivation for this was that, at least for a single checkbox, the iOS checkbox is very difficult to tell there's a thing to check.

### DIFF
--- a/src/components/Checkbox/CheckboxItem.tsx
+++ b/src/components/Checkbox/CheckboxItem.tsx
@@ -9,6 +9,8 @@ import {
 } from 'react-native';
 
 import CheckBox from './Checkbox';
+import CheckboxAndroid from './CheckboxAndroid';
+import CheckboxIOS from './CheckboxIOS';
 import Text from '../Typography/Text';
 import TouchableRipple from '../TouchableRipple/TouchableRipple';
 import { withTheme } from '../../core/theming';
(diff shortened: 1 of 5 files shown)


src/components/Checkbox/CheckboxItem.tsx
```diff
@@ -54,6 +56,11 @@ type Props = {
    * testID to be used on tests.
    */
   testID?: string;
+  /**
+   * Whether `<Checkbox.Android />` or `<Checkbox.IOS />` should be used.
+   * Left undefined `<Checkbox />` will be used.
+   */
+  mode?: 'android' | 'ios';
```
tegandbiscuits, line 63: Open to renaming this whatever, `mode` just seemed consistent with other components.

I also wasn't sure if it'd be useful to have a third option like `'system'`, mostly just for explicitness. In my mind I can see it maybe being useful if you need/want to have `mode` set, but just want to use the platform default, though I don't know if this is a situation anyone will be in.

Trancever: If such a situation occurs, one can pass undefined and it will default to platform-specific component 👍

Trancever (approved):

Trancever:
@tegandbiscuits Could you submit a similar PR but for the RadioButton.Item component?

src/components/Checkbox/CheckboxItem.tsx
tegandbiscuits: That makes sense.

tegandbiscuits:
@Trancever Sure thing!

PR #2376, opened 2020-11-23
feat: button right side icon

haakandev:
### Summary

Similar to #1370 this adds the ability to put an icon on the right side of buttons. Credit to the author of that PR, I used similar approach but without breaking current functionality for the `icon` prop.

### Test plan

See Button in Example app with icon on the right side.

### DIFF
--- a/example/src/Examples/ButtonExample.tsx
+++ b/example/src/Examples/ButtonExample.tsx
@@ -30,6 +30,14 @@ const ButtonExample = () => {
           <Button loading onPress={() => {}} style={styles.button}>
             Loading
           </Button>
+          <Button
+            icon="camera"
+            onPress={() => {}}
+            style={styles.button}
+            contentStyle={{ flexDirection: 'row-reverse' }}
+          >
+            Icon right
+          </Button>
         </View>
       </List.Section>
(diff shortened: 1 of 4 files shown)


Trancever:
@haakandev Hey, thanks for your contribution! So the are a few problems with this PR:

MD Guidelines doesn't really show an example of displaying a button with an icon on the right side and what is more they call it a "leading" icon in their documentation which suggests it should be displayed only on the left side.

Also, displaying 2 icons at the same time is explicitly disallowed by MD Guidelines.

Anyway, this is a feature that many people requested before and I think we should implement it in a non-breaking way. We currently accept the `contentStyle` prop that is passed to the View that wraps the Icon and text inside the Button. We could check whether it includes `flexDirection: 'row-reverse'` and if that's the case we could adjust the icon style margins.

haakandev:
@Trancever thanks for the feedback!
Appreciate you being open to this due to popular demand despite not being explicitly specified in MD guidelines. 
I've taken your suggestion to use the `flexDirection: 'row-reverse'` style as the deciding factor and updated the PR.

Trancever (approved):

Trancever:
Thank you!

Issue #2381, opened 2020-11-25
Is it possible to add more than one line for a particular DataTable Cell?


jota12x:
**Problem**
Hello, I've been using `react-native-paper` for the past month with satisfying results, but now I am stumbled on a particular use case. I have a `DataTable.Cell` that needs to show more than one line at a time (imagine a list of food ingredients). Is this possible within the current version? 

**Approaches**
I tried to append line breaks without result. I also insert a Text with a `numberOfLines` attribute set to more than one but nothing happened as well. Inspecting the source code for `DataTable.Cell` , I found that there is a `numberOfLines` set to one by default , so I was wondering if that is the cause, because when I change the underlying Text Component on the debugger, it will show appropriately.

![imagen](https://user-images.githubusercontent.com/6154802/100194166-8a4e6680-2f38-11eb-8ceb-33d59483323a.png)

Thanks in advance

Trancever:
@jota12x `DataTable.Cell` is a `View` with just some styles to make it MD Guidelines compliant, but nothing fancy. If you need to make it multiline then use `View` instead.

Could you submit a PR adding this information to the DataTable.Cell docs? That would be awesome 🙏

jota12x:
@Trancever Thanks for the answer. I will submit a PR later this week.

jota12x:
The documentation was updated, so I will be closing the issue.

hasbelcaro:
It works for me changing this value in the DataCell library code
![Screen Shot 2022-04-25 at 9 45 49 AM](https://user-images.githubusercontent.com/7729951/165113735-78a21c74-de6a-4af1-8c4b-83429dce234e.png)

ramirobg94:
https://github.com/callstack/react-native-paper/issues/1038#issuecomment-1186243530

PR #2382, opened 2020-11-25
feat: support for multi-line Text in DataTableCell

match:
### Summary

Text in DataTableCell only support single-line but some use-cases, like #2381, need to support multi-line.
It enable to support multi-line text in DataTableCell.

### Test plan

- See Data Table in Example app
- single line text
image
- multi line text
image


Trancever:
This is explicitly disallowed by MD Guidelines. If you still want to do it, then use `View` instead of `DataTable.Cell`.

PR #2383, opened 2020-11-25
Handle null ListItem title

wholenews:
### Summary

Don't take up empty space if `ListItem` `title` is `null`.  The `description` property may then be used for arbitrary content, as the opinionated `<Text>` wrapper is avoided, which causes trouble on Android.


Trancever:
This is against Material Design Guidelines - Card should always have a title. If you really need this feature then you should use patch-package or fork the lib.

wholenews:
Hm. What about removing the <Text> wrapper around title?  I do want a title, I just don't want the particular style of title demanded.

PR #2384, opened 2020-11-27
docs: explain how to change theme based on device settings

MatthewThompson:
### Summary

In the theming section of the documentation it states that if you don't use a custom theme, then animations will be turned on and off based on device settings.

The same is true for toggling of dark theme based on device settings, react-native-paper does it for the default theme but not if you use a custom theme. It didn't mention this anywhere in the documentation so I added it to the dark theme section


Trancever (approved):

Trancever:
Thanks!

PR #2391, opened 2020-12-02
fix: wrong types for onPress and onLongPress and bug in example app

RichardLindhout:
### Summary
I'm working on a navigation library on top of react-router-dom and react-native-navigation. But the Link component I'm using won't work well with the onPress of react-native-paper in typescript since the event is not defined in typescript (but is provided to the function)

```
  const onPress = React.useCallback(
    (event: GestureResponderEvent) => {
      if (!event.defaultPrevented) {
        event.preventDefault();
        navigation.push(to, params);
      }
    },
    [navigation, to, params],
  );
```

Types of property 'onPress' are incompatible.         Type '(e: GestureResponderEvent) => any' is not assignable to type '() => void'.
Schermafbeelding 2020-12-02 om 23 26 23

I decided to fix it everywhere so I can use links on some places or it just good to provide it to users so can they can provide preventDefault() or things like that.

### DIFF
--- a/example/src/Examples/ToggleButtonExample.tsx
+++ b/example/src/Examples/ToggleButtonExample.tsx
@@ -21,7 +21,7 @@ const ToggleButtonExample = () => {
             icon="android"
             value="android"
             status={status}
-            onPress={(status) =>
+            onPress={() =>
               setStatus(status === 'checked' ? 'unchecked' : 'checked')
             }
           />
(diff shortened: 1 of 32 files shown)


RichardLindhout:
Schermafbeelding 2020-12-02 om 23 43 16

This fixes a bug in the example app too since the android button is not switchable at all it is always 'checked' after first click since the argument is only an event and not the previous status.

lukewalczak:
Closing due to your second PR https://github.com/callstack/react-native-paper/pull/3484 😓 

`ToggleButton`'s I will handle separately.

PR #2395, opened 2020-12-03
Fix example in List.Icon

jaworek:
### Summary

Current example does not wrap multiple elements in container component.
Thus it does not work on Snack out of the box.
I added ```React.Fragment``` to fix it.

### Test plan

This is a documentation update.


Trancever (approved):

Trancever:
Thanks!

PR #2406, opened 2020-12-09
Update AppbarBackIcon.tsx

Svarto:
The AppbarBackIcon is now editable and follows the size set in the component. This fixes the icon being a fixed size, with the "size" prop in the Appbar.BackAction only changing the enveloping view.

### DIFF
--- a/src/components/Appbar/AppbarBackIcon.tsx
+++ b/src/components/Appbar/AppbarBackIcon.tsx
@@ -16,7 +16,7 @@ const AppbarBackIcon = ({ size, color }: { size: number; color: string }) =>
     >
       <Image
         source={require('../../assets/back-chevron.png')}
-        style={[styles.icon, { tintColor: color }]}
+        style={[styles.icon, { tintColor: color, width: size, height: size }]}
       />
     </View>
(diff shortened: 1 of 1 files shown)


src/components/Appbar/AppbarBackIcon.tsx
```diff
@@ -16,7 +16,7 @@ const AppbarBackIcon = ({ size, color }: { size: number; color: string }) =>
     >
       <Image
         source={require('../../assets/back-chevron.png')}
-        style={[styles.icon, { tintColor: color }]}
+        style={[styles.icon, { tintColor: color, width: size, height: size }]}
```
Trancever, line 19: The size was previously set to 21 purposefully. Could you pass `size - 3` instead of just `size`? We don't want to make a breaking change :)

```
const iosIconSize = size - 3;
...
...
{ width: iosIconSize, height: iosIconSize }
```

Trancever (changes requested):
Thanks for the contirbution! Just small think to update and it will be ready to merge!

Please make sure that unit tests are passing as well.

Thanks!

Issue #2402, opened 2020-12-08
TextInput.Flat text is incorrectly aligned when left icon is displayed inside input (Web)


Trancever:
### Current behaviour

The text inside TextInput.Flat with Icon on the left-hand side is incorrectly aligned.

The issue occurs only on the web.

### Expected behaviour

The text should be aligned in the same way as the label is.

### Code sample

```
        <TextInput
          label="Flat input"
          placeholder="Type something"
          value={text}
          onChangeText={(text) => inputActionHandler('text', text)}
          left={
            <TextInput.Icon
              name="heart"
              color={flatLeftIcon}
              onPress={() => {
                changeIconColor('flatLeftIcon');
              }}
            />
          }
        />
```

### Screenshots (if applicable)

![image](https://user-images.githubusercontent.com/18584155/101473428-1bb2d500-394a-11eb-8ab9-c0d031e63d64.png)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | ios
| react-native          | 0.62
| react-native-paper    | 4.4.1
| node                  | 12.18.2
| npm or yarn           | 1.22.1
| expo sdk              | 38

Trancever:
Fixed in v4.5.0

PR #2410, opened 2020-12-10
fix: padding in flat text input on web

unikvozm:
### Summary
#2402


Trancever:
@unikvozm Do you know what is the root of that inconsistency between mobile and web? That issue is web-specific and I am trying to understand if there is a way to fix it in a root instead of adding some padding here and there :D

unikvozm:
@Trancever, yeah, the different logic was for web and mobiles: for web it was padding 36 only, and for mobiles: margin 36 and padding 8
Screenshot 2020-12-10 at 20 21 35

and now it is the same on all platforms.
Screenshot 2020-12-10 at 20 31 54

Trancever (approved):

Trancever:
Thank you @unikvozm

PR #2422, opened 2020-12-13
fix: replace safe-area-view with iphonex-helper

### LINKED ISSUE #2323: Material Tab Navigator not compatible with iPhone 12
The dependency react-native-safe-area-view should be changed to "^2.0.0", which supports the iPhone 12 series.

At the moment, the safe area does not work properly on iPhone 12.

satya164:
Fixes #2323


Trancever (approved):
Thank you, Satya! <3

Issue #2417, opened 2020-12-11
Doubt about the ordering of the datatable


Angelk90:
Hi @Trancever , @unikvozm ,

In the datatable header documentation, [here](https://callstack.github.io/react-native-paper/data-table-header.html).

There is the following image (see below), with its example saying that if `sortDirection = 'descending'` then the arrow is down, but trying the code is opposite the arrow is up.

So I got the doubt, which is the right way?

![](https://callstack.github.io/react-native-paper/screenshots/data-table-header.png)

Expo:

<img width="323" alt="Schermata 2020-12-11 alle 12 28 40" src="https://user-images.githubusercontent.com/20476002/101898455-7186b780-3bac-11eb-9fef-29dacf0b50b2.png">

Angelk90:
Hi @Trancever, @unikvozm: Any news?

haxonadora:
Hey @Angelk90. Thanks for submitting the issue. According to Material design guidelines down arrow should indicate descending sorting as documentation shows. However, as you pointed, that wasn't a case in paper code. Fix for this is already merged and it will be available in next release.

Angelk90:
Hi @haxonadora , 

before making the request I had also checked the material design documentation, which is why I doubted.
Thanks for the correction.

P.s.
You might check out this #2416.

PR #2435, opened 2020-12-15
fix: Changed sorting arrow direction in data table

haxonadora:
Fix related to [issue](https://github.com/callstack/react-native-paper/issues/2417).
According to Material design guidelines: `The downward arrow shows that the schedule column is sorted in descending order.`

Currently `sortDirection` prop in `DataTableTitle` set to `descending` was making the arrow to point up instead of down. Documentation is showing correct behaviour but the code was doing the opposite.


Trancever (approved):

wojteg1337:
Good job

wojteg1337 (approved):

PR #2448, opened 2020-12-18
fix: allow setting accessibility prop for Button component

RafikiTiki:
### Summary

In my current project I have a very specific business requirement: in a web version of application few buttons need to be taken out of keyboard navigation flow (using tab key) while at the same time still being clickable by mouse. Using `disabled` prop will obviously break second part of the requirement so I've investigated a little how other accessibility-related props do work in react-native-web and it turned out that `accessible` prop is exactly what I'm looking for. Unfortunately in Paper's `<Button />` component `accessible` prop is passed to the wrapping `<Surface />` component, not the `<TouchableRipple />` which is then rendering `<TouchableWithoutFeedback />`. That is making `accessible` prop have no effect.

This PR is fixing that issue but for some reason some of the library users may depend on passing `accessible` prop to the wrapping `<Surface />` so it is possibly a **Breaking Change**.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -139,6 +139,7 @@ const Button = ({
   contentStyle,
   labelStyle,
   testID,
+  accessible,
   ...rest
 }: Props) => {
   const { current: elevation } = React.useRef<Animated.Value>(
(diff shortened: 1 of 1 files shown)


Trancever:
@RafikiTiki I'll review and eventually merge early next week, but at first glance, it doesn't look like a breaking change. It's been broken and we are fixing a bug here I think.

Trancever (approved):

PR #2451, opened 2020-12-19
example update expo40

Quantmatic:
### Summary

Updated the example dir with expo40
Appbar, MenuExample, and the /src/index.tsx

### Test plan

cd example
yarn install && yarn start

### Issue

If you run the example on ios/android, better get rid of the metro.config
Causes warning about Constants I'm not sure how to fix
Related to latest expo using 'expo-constants'


raajnadar:
The expo version is already upgraded!

Quantmatic:
Looks like you beat me to it...

PR #2474, opened 2020-12-31
feat: add style prop in Modal

ahce:
### Motivation

This prop is necessary when you need to see content in full screen on iPhone X/11

### Test plan
1. Download this branch
2. Open Modal example
3. Pass style prop with this values (iPhone X/11):
```js
 { marginTop: 0, marginBottom: 0 }
```
### Actual behaviour
image

### Behaviour with  style = { marginTop: 0, marginBottom: 0 }
image

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -18,10 +18,6 @@ import { withTheme } from '../core/theming';
 
 type Props = {
   /**
-   * Determine whether the insets are applied.
-   */
-  disableSafeAreaInsets?: boolean;
-  /**
    * Determines whether clicking outside the modal dismiss it.
    */
   dismissable?: boolean;
@@ -45,6 +41,11 @@ type Props = {
    * Style for the content of the modal
    */
   contentContainerStyle?: StyleProp<ViewStyle>;
+  /**
+   * Style for the wrapper of the modal.
+   * Use this prop to change the default wrapper style or to override safe area insets with marginTop and marginBottom.
+   */
+  style?: StyleProp<ViewStyle>;
   /**
    * @optional
    */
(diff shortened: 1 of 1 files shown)


src/components/Modal.tsx
```diff
@@ -17,6 +17,10 @@ import Surface from './Surface';
 import { withTheme } from '../core/theming';
 
 type Props = {
+  /**
+   * Determine whether the insets are applied.
```
Trancever, line 21:
```suggestion
   * Determines whether the insets are applied.
```

Trancever (changes requested):

src/components/Modal.tsx
```diff
@@ -17,6 +17,10 @@ import Surface from './Surface';
...
+   */
+  disableSafeAreaInsets?: boolean;
```
Trancever, line 23: We would prefer to have `safeAreaInsets` prop instead where users could pass `top` and `bottom` properties to override the default behavior. Could you change the implementation, please?

Trancever (changes requested):

src/components/Modal.tsx
ahce-meli: @Trancever you think is better pass a wrapperStyle?

ahce: you think is better pass a wrapperStyle?

Trancever: Yes, we can even call it simply `style` :) Please provide a nice description for this prop so it's clear what can be achieved using it. Thanks!

ahce: done!

Trancever (approved):

PR #2505, opened 2021-01-14
feat: add prop right to DrawerItem

brunohkbx:
### Summary
It would be nice to be able to render a badge or chip in the Drawer.Item

### Test plan
1. Open the example app and see the Drawer.Item using a badge

example

### DIFF
--- a/src/components/Drawer/DrawerItem.tsx
+++ b/src/components/Drawer/DrawerItem.tsx
@@ -27,6 +27,10 @@ type Props = React.ComponentPropsWithRef<typeof View> & {
    * Accessibility label for the button. This is read by the screen reader when the user taps the button.
    */
   accessibilityLabel?: string;
+  /**
+   * Callback which returns a React element to display on the right side. For instance a Badge.
+   */
+  right?: (props: { color: string }) => React.ReactNode;
   style?: StyleProp<ViewStyle>;
   /**
    * @optional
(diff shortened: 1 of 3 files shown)


src/components/Drawer/DrawerItem.tsx
```diff
@@ -27,6 +27,10 @@ type Props = React.ComponentPropsWithRef<typeof View> & {
    * Accessibility label for the button. This is read by the screen reader when the user taps the button.
    */
   accessibilityLabel?: string;
+  /**
+   * Callback which returns a React element to display on the right side. For instance a Badge.
+   */
+  right?: () => React.ReactNode;
```
brunohkbx, line 33: @Trancever I think it should be just a `React.ReactNode` but I used a callback since most components are following this pattern.

Trancever, line 33: @brunohkbx Can we change the type to `right?: (props: { color: string }) => React.ReactNode;` ? I think it makes sense to passDrawer's content color here, so users can apply it to the right element if they want.

Trancever (changes requested):

brunohkbx:
Done @Trancever :+1: I updated the example as well

lukewalczak (approved):
@brunohkbx really nice and useful enhancement, thanks!

PR #2512, opened 2021-01-19
feat: add card mode outlined

brunohkbx:
### Summary
Implement MD outlined card.

https://material.io/components/cards

Elevated | Outlined
--- | ---
light-elevated | light-outlined

Elevated | Outlined
--- | ---
dark-elevated | dark-outlined

### Test plan
1. Open the example app
2. Check the example showing an outlined card

### DIFF
--- a/example/src/Examples/CardExample.tsx
+++ b/example/src/Examples/CardExample.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Alert, ScrollView, StyleSheet } from 'react-native';
+import { Alert, ScrollView, StyleSheet, View } from 'react-native';
 import {
   Avatar,
   Paragraph,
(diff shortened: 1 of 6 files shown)


src/components/Card/Card.tsx
```diff
@@ -29,6 +29,12 @@ type Props = React.ComponentProps<typeof Surface> & {
    * Function to execute on press.
    */
   onPress?: () => void;
+  /**
+   * Mode of the Card.
+   * - `elevated` - Card with elevation, elevation defaults to 1.
+   * - `outlined` - Card with an outline.
+   */
+  mode?: 'elevated' | 'outlined';
```
Trancever, line 37: I don't like the way it's typed currently. The outlined card shouldn't accept elevation prop. Could you try something like that:

```js
type OutlinedCardProps = {
  mode?: 'outlined';
  elevation: never;
};

type ElevatedCardProps = {
  mode?: 'elevated';
  elevation?: number;
};

type Props = (OutlinedCardProps | ElevatedCardProps) & { Common Props };
```

I haven't tried that, so it might not work, but I would like to explore this way of handling conditional types.

Trancever: I haven't tried that, so it might not work, but I would like to explore this way of handling conditional types.

Trancever (changes requested):

src/components/Card/Card.tsx
brunohkbx: @Trancever I tried your approach and it wasn't working at first. So after some time debugging, I found a problem in the HOC `withTheme`. I removed it and it worked as follow:

Raises error
`<Card mode="outlined" elevation={3} />`

Works
`<Card mode="elevated" elevation={3} />`

After looking up the issue in `withTheme` I found a problem in its type declaration [here](https://github.com/callstack/react-theme-provider/blob/master/typings/index.d.ts#L12)
`$Without` it's not working as expected because `Pick` doesn't distribute over discriminated union types and you can read more about [here](https://github.com/microsoft/TypeScript/issues/28791)

Should we fix the type declaration first?
`type $Without<T, K extends keyof any> = T extends any ? Pick<T, Exclude<keyof T, K>> : never;` 
works for me

Trancever: I think we should first fix the type of the withTheme HOC. Could you submit a PR? 🙏

brunohkbx: Done :+1: https://github.com/callstack/react-theme-provider/pull/96

src/components/Card/Card.tsx
```diff
@@ -171,6 +195,11 @@ const styles = StyleSheet.create({
     flexGrow: 1,
     flexShrink: 1,
   },
+  outlined: {
+    elevation: 0,
+    borderWidth: 1,
+    borderColor: 'rgba(0, 0, 0, 0.12)',
```
lukewalczak, line 201: I guess that `borderColor` should depend on the theme and it should be specified in that way:

```
  const borderColor = color(theme.dark ? white : black)
    .alpha(0.12)
    .rgb()
    .string();
```

However, I didn't check the specification, so please look on it.

example/src/Examples/CardExample.tsx
```diff
@@ -98,6 +98,12 @@ const CardExample = () => {
           </Paragraph>
         </Card.Content>
       </Card>
+      <Card style={styles.card} mode="outlined">
+        <Card.Title title="Outlined card" />
+        <Card.Content>
+          <Paragraph>This is an outlined card</Paragraph>
+        </Card.Content>
+      </Card>
     </ScrollView>
```
lukewalczak, line 107: Can we add one more example with media and buttons?

brunohkbx:
@lukewalczak I added a switch at the top to toggle the mode. WDTY?

lukewalczak:
Hey @brunohkbx, thanks for updating the PR! I like the idea of switch and the whole feature 🎉 . I will test it a bit, ping the theme provider lib one more time and go through the code. Anyway I guess we're close to ship it!

souhe:
@brunohkbx I just published a new version of theme-provider with your fix. Sorry for the lag. 
The new version is now `v3.0.6` I had to bump it a lot since we did a mistake with the previous release ;)

brunohkbx:
@souhe thanks. I will update this PR

brunohkbx:
TS now complains when using mode `outlined` + elevation cc @Trancever @lukewalczak 
Screenshot from 2021-05-12 14-51-14

lukewalczak:
> TS now complains when using mode outlined + elevation

Isn't a correct behavior since you typed:

```
type OutlinedCardProps = {
  mode: 'outlined';
  elevation: never;
};
```

brunohkbx:
@lukewalczak yeah it is. I'm just showing it. Works now with `callstack/react-theme-provider` 3.0.6

lukewalczak:
> yeah it is. I'm just showing it. Works now with callstack/react-theme-provider 3.0.6

Ahhh ok I see, sorry for misunderstanding 😅 thanks @brunohkbx!

lukewalczak (approved):

lukewalczak:
Awesome feature, thanks @brunohkbx for your work!

PR #2515, opened 2021-01-20
feat: set default cursor for TouchableRipple to 'pointer'

RafikiTiki:
### Summary
While I was working on client's app code I've come to a realisation that in every place where Paper's `<TouchableRipple />` component was used we needed to pass additional style to ensure pointer cursor on the web. After some digging I've found that it's caused because TouchabeRipple is using `<TouchableWithoutFeedback />` under the hood and based on this: https://github.com/necolas/react-native-web/issues/1847#issuecomment-747625960 that's exactly what's causing react-native-web to NOT "inject" `cursor: 'pointer'` style on the web.

### Test plan
In the example app:
- [x]  on mobile device/simulator everything should work exactly the same as before this change.

- [x] on the web `<TouchableRipple />` component should have `pointer` cursor

### DIFF
--- a/src/components/TouchableRipple/TouchableRipple.tsx
+++ b/src/components/TouchableRipple/TouchableRipple.tsx
@@ -6,6 +6,7 @@ import {
   StyleSheet,
   StyleProp,
   GestureResponderEvent,
+  Platform,
 } from 'react-native';
 import color from 'color';
 import { withTheme } from '../../core/theming';
@@ -253,6 +254,7 @@ TouchableRipple.supported = true;
 const styles = StyleSheet.create({
   touchable: {
     position: 'relative',
+    ...(Platform.OS === 'web' && { cursor: 'pointer' }),
   },
   borderless: {
     overflow: 'hidden',


src/components/TouchableRipple/TouchableRipple.tsx
```diff
@@ -253,6 +254,7 @@ TouchableRipple.supported = true;
 const styles = StyleSheet.create({
   touchable: {
     position: 'relative',
+    ...(Platform.OS === 'web' && { cursor: 'pointer' }),
```
Trancever, line 257: Do we really need this conditional? What's gonna happen if we pass `cursor: 'pointer'` in RN?

Trancever (approved):

mikehardy:
Ironically, the one place I noticed this goes to 'pointer' from 'default' is in DataTable.Cell, where most of my data is just text (except a couple IconButtons) and they all have pointers making them look clickable so now I need to pass in cursor: default :laughing:

PR #2530, opened 2021-01-27
feat: dropdown component

omneimneh:
### Summary

This PR introduces a new component which is `Dropdown` and `Dropdown.Option`, I noticed that the native `Picker` in react-native isn't very customizable and doesn't look nice with other react-native-paper components.

I found other libraries on npm such as 'react-native-paper-dropdown' and 'react-native-material-dropdown' however I tested these libraries, they appear to have an issue displaying on the web, when developing this I went through the same problem because I was following their approach at first so I had to change a little bit.
I will try to explain it briefly:
If we want to create a dropdown in react (only web), we can use a div with relative position to parent by following this approach [https://stackoverflow.com/a/5209833](https://stackoverflow.com/a/5209833), however we will need to set the z-index to prevent other components down in the page to appear on top of the dropdown.
With react-native it's a little more complicated since all components by default have a relative position a a z-index preset to 0, so even if we increase the z-index of the dropdown we still can't make it appear above other components unless we change the z-index of the parent (and maybe other parents in the dom tree as well) [https://stackoverflow.com/a/35326592](https://stackoverflow.com/a/35326592).
The approach I took involved using a portal and computing the location of the dropdown when the user clicks on the component and displaying it, it works because the portal renders the dropdown at the top of the tree. But in this approach I assumed that the Provider component will provide the `Protal.Host`.

### Test plan
Screenshot_20210127-110722
Screenshot_20210127-110657

I already provided an example in my PR, but I believe this is not ready to merge yet since there are tests and docs to add and maybe some refactoring required. Other features can be added as well like completely custom components (here I'm just using `List.Item`) and making the dropdown searchable... but I'll let you guys be the judge.

I'll keep updating the PR as much as I can until it's ready to merge and if anyone can also contribute with this it would be great


example/src/ExampleList.tsx
```diff
@@ -75,6 +76,7 @@ export const examples: Record<
   toggleButton: ToggleButtonExample,
   touchableRipple: TouchableRippleExample,
   theme: ThemeExample,
+  dropdown: DropdownExample,
```
brunohkbx, line 79: Can you add it below the divider?

src/components/Dropdown/DropdownOption.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * ```
+ */
+class DropdownOption<T> extends React.Component<
+  Props<T> & { theme: ReactNativePaper.Theme }
+> {
+  static displayName = 'Dropdown.Option';
+
+  // eslint-disable-next-line react/require-render-return
+  render(): React.ReactNode {
+    throw new Error('<DropdownOption/> must be a direct child to <Dropdown/>');
```
brunohkbx, line 98: @omneimneh @Trancever I wonder if wouldn't be better to use a context to share props from the dropdown to this component? Then it would render the `List.Item` directly instead of using the `DropdownChild`.

omneimneh: Yes @brunohkbx, I think it's a good idea. The way I've done it is a little hacky I guess and I was thinking of a way to make it better, another approach would be using cloneElement and pass the onPress prop to it. I'm fine with both approaches what do you think?

brunohkbx: I think both are good. kentcdodds wrote a pretty good post about this https://kentcdodds.com/blog/compound-components-with-react-hooks

omneimneh:
@brunohkbx any updates on this? thanks 😃

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,340 @@
...
+  /**
+   * Placeholder for the dropdown.
+   */
+  placeholder?: string;
+  /**
+   * The text to use for the floating label.
+   */
+  label?: string;
+  /**
+   * Required flag removes the possibility for the user to unselect a value once it's selected.
```
brunohkbx, line 29: It's not clear to me how this prop works. I'm still able to change the selected option :thinking:

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,340 @@
...
+  onSelect?: (selected: T | null) => void;
+  /**
+   * The option having this key will be selected
+   * If undefined the currently selected option will remain selected.
+   */
+  selectedValue?: T | null;
+  /**
+   * This message will appear when the required prop is set to true and the number of children is zero.
+   */
+  emptyDropdownLabel?: string;
```
brunohkbx, line 48: What do you think about adding an example with this?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,340 @@
...
+type OptionProps<T> = {
+  value: T | null;
+  renderLabel?: (props: RenderProps) => React.ReactNode;
+  label?: string;
+};
+
+const DEFAULT_EMPTY_DROPDOWN_LABEL = 'No available options';
+const DEFAULT_PLACEHOLDER_LABEL = 'Select an option';
+const DROPDOWN_NULL_OPTION_KEY = 'DROPDOWN_NULL_OPTION_KEY';
+// const DEFAULT_MAX_HEIGHT = Infinity;
```
brunohkbx, line 102: Should we keep this?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,340 @@
...
+ * };
+ *
+ * export default MyComponent;
+ * ```
+ */
+const Dropdown = React.forwardRef(function <T>(
+  {
+    children,
+    emptyDropdownLabel = DEFAULT_EMPTY_DROPDOWN_LABEL,
+    // maxHeight = DEFAULT_MAX_HEIGHT,
```
brunohkbx, line 179: Do we need this?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,340 @@
...
+  null as any
+);
+
+function findSelectedOption<T extends any>(
+  value: T | null,
+  children?: React.ReactNode
+) {
+  return React.Children.toArray(children)
+    .map((child) => (child as any)?.props)
+    .find((props) => props?.value === value);
```
brunohkbx, line 130: I'm wondering if wouldn't be better to filter by the `key` here. just in case when someone uses the same value by mistake.

src/index.tsx
```diff
@@ -44,7 +45,6 @@ export { default as Appbar } from './components/Appbar';
 export { default as TouchableRipple } from './components/TouchableRipple/TouchableRipple';
 export { default as TextInput } from './components/TextInput/TextInput';
 export { default as ToggleButton } from './components/ToggleButton';
-
```
brunohkbx, line 47: No need to remove this line

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,340 @@
...
+  };
+
+  React.useEffect(() => {
+    if (selectedValue) {
+      setSelected(findSelectedOption(selectedValue, children));
+    }
+  }, [children, selectedValue]);
+
+  React.useEffect(() => {
+    if (isMenuOpen) {
```
brunohkbx, line 254: How about creating a custom hook for this? Something like `useOrientations` (maybe you can come up with a better name)
You can create it in two different files: `useOrientations.native.js` and `useOrientations.js`
just to clean up this a little

brunohkbx:
@omneimneh I just realized that there's a missing top/bottom margin in the dropdown. Check it out:

From MD guide: https://material.io/components/menus#specs
Screenshot from 2021-02-23 00-02-25

dropdown

brunohkbx:
Also, can we support this error state?

Screenshot from 2021-02-23 00-34-13

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,340 @@
...
+  renderNoneOption?: (props: {
+    onPress: () => void;
+    style: { color: string };
+    key: React.Key;
+  }) => React.ReactNode;
+  /**
+   * A dropdown can be either be rendered inside a modal or inside a floating menu
+   * The default value is 'floating' if the platform is web and 'modal' otherwise
+   */
+  mode?: 'modal' | 'floating';
```
brunohkbx, line 61: About this `floating` mode, shouldn't this be only available on web? I couldn't find anything in the MD docs regarding this https://material.io/components/menus

src/components/Dropdown/Dropdown.tsx
omneimneh: So the idea is to remove the option that comes with a value of null, if required is set to false, the default null option will be shown as the first option in the dropdown, so to make a dropdown required you should give it a value and set required to true.

Sorry if the docs are not very clear, I will try to improve them as much as I can, and if you think we need to change the behavior we can maybe change the required prop to something like showUnselectOption or make it an iconButton at the right of the textInput instead I don't mind.

src/components/Dropdown/Dropdown.tsx
omneimneh: Hmm, I'm not sure about what the material design guides indicate sorry that I don't have much knowledge about this, I was thinking in case we have a very long dropdown and when displaying this on the web (floating mode) on a large display that we may prefer to limit the height of the dropdown, that was my thoughts when I started working on it but I guess we don't need it anymore so I can remove it if you want

src/components/Dropdown/Dropdown.tsx
omneimneh: Sorry about that too, I wasn't sure what the correct behavior should be based on the material design guidelines, from what I saw there was only a "floating" mode (I didn't know what to call it) but the default Picker of the android appears in a modal (or at least used too, it's been a long time since I used native android). Anyways we I can fix this as soon as we figure out what the correct behavior is.

omneimneh: I believe they call it exposed dropdown menu

avvari-da:
@brunohkbx Hello, please let me know if this is going live

brunohkbx:
hey @lukewalczak can you please take a look when you have some time?

RichardLindhout:
I'm working on a similar component but I only tested it on the web for now but it will improve in the coming month:
https://github.com/web-ridge/react-native-paper-autocomplete

NahumG10:
How do I intall this component in my app?

MartijnGevaert:
+1
I'm using react-native-paper and wanted to use a dropdown picker the Project as well, unfortunately did not find one with good web & react-native support as mentioned at the start of this PR. Looking for other solutions now.

Angelk90:
@lukewalczak , @omneimneh : Are there any news?

jigarzon:
Any news on this PR?

fellenabmb:
Any news on this one? 😄

H-pun:
Any news on this?

omneimneh:
Hello everyone, sorry for not being able to close this PR. A lot of things has changed with the material design M3 and it was very difficult to find a convenient time to work on this but I'll try to work on it in the upcoming days although I can't promise anything due to life/work commitments.
Sorry again for keeping this open for too long and thanks for understanding.

omneimneh:
I moved this to a new PR https://github.com/callstack/react-native-paper/pull/4102

Issue #2150, opened 2020-08-24
How can I change the accordion right 'chevron-up' 'chevron-down' icons to my custom icons?


sathwik77:
Custom icons for accordion

- `react-native-paper 4.0.1`
- `expo 38.0`

NagarajBD:
> Custom icons for accordion
> 
> * `react-native-paper 4.0.1`
> * `expo 38.0`

i am also have same problem, any one can help.

henrik-d:
Maybe one could override the current chevron with a custom component if user passed a renderProp, e.g. like that:
```
right?: (props: {
        color: string;
        expanded: boolean;
    }) => React.ReactNode;
```
Would you accept an PR with this extension?

moedeveloper:
why need answer :(, we really need this feature:(
did u guys solve it anyhow?

samOpenforce:
Ability to customise the right icon (in both expanded and not-expanded states) would be a greatly appreciated feature. I'm not clever enough to create a PR to add this, please take pity :)

PR #2531, opened 2021-01-29
feat: add prop right to List Accordion

Joeao:
### Summary
This pull request addresses issue https://github.com/callstack/react-native-paper/issues/2150. It gives users the flexibility to replace what's rendered on the right-hand-side of the accordion with their own component. The expanded state is passed back to the user so that they can render based on this value.

This feature draws inspiration from the `left` prop on the same component. Instead, this could have been addressed similarly to how Fab Icon is implemented, however the drawback is that the implementation of Fab Icon is a bit more complex when compared to the `left` prop of this component.

### Test plan

Render a `List.Accordion` component and pass in a separate component as the `right` prop. E.g.

Omitting this prop results in the default `MaterialCommunityIcon` element being rendered.

### DIFF
--- a/src/components/List/ListAccordion.tsx
+++ b/src/components/List/ListAccordion.tsx
@@ -29,6 +29,10 @@ type Props = {
    */
   left?: (props: { color: string }) => React.ReactNode;
   /**
+   * Callback which returns a React element to display on the right side.
+   */
+  right?: (props: { isExpanded: boolean }) => React.ReactNode;
+  /**
    * Whether the accordion is expanded
    * If this prop is provided, the accordion will behave as a "controlled component".
    * You'll need to update this prop when you want to toggle the component or on `onPress`.
@@ -127,6 +131,7 @@ type Props = {
  */
 const ListAccordion = ({
   left,
+  right,
   title,
   description,
   children,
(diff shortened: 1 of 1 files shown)


sobiamaredia:
Can someone please merge this PR? This is a great addition.

Joeao:
Hi @lukewalczak.

In my case it's not that I want to use an icon other than Chevron up/down, it's that I want to avoid using `MaterialCommunityIcon`.

The intention of this PR isn't to go against Material Design principals, but to allow for more flexibility in icon choices. I could have mentioned that in my initial comment but I hadn't been thinking about principals. I now see that the issue I referenced suggests going against the principals, for that I condemn it and refuse to have this PR be associated with it any longer!

I know that the [RNP documentation](https://callstack.github.io/react-native-paper/icons.html) says that some components may render incorrectly if not using the default icons, but it'd be nice to be able to reduce that as much as possible which this PR would do in one case.

As to whether or not RNP would like to tackle the issue of icon replacement either in a different way or not at all is up to you and the Callstack team. If this PR is a suboptimal way to address the problem right now then feel free to close. I can say that my project would benefit from it as currently there is no Chevron, just an empty box, which definitely goes against the principals. But my individual case doesn't mean much against the 60+ thousand weekly download users who expect a robust, well supported component library that follows the principals that they know and love.

lukewalczak:
Hey @Joeao thanks for your comment. I understand your point of view and your intention. I think since the `right` prop default is always `MaterialCommunityIcon` we can accept that PR and introduce the prop to allow more flexibility in that matter.

lukewalczak (approved):

Joeao:
Excellent, thank you very much!

lukewalczak:
It's not released yet, in the meantime you can use package from main branch.

serge-deugoue:
Please do you know when it will be released?

lukewalczak:
Hopefully until the end of the week 🤞🏽

lukewalczak:
New version is already release (4.8.0)

PR #2552, opened 2021-02-17
fix: component BottomNavigation

sviande:
Prevent scene blink fix desynchronization between opacity & top

Sometimes the top property is too far and the opacity = 0 so the screen flash.

### Summary

Fix screen flash / blink when switching tabs.

https://github.com/react-navigation/react-navigation/issues/9256

### Test plan

Switch between tabs.

### DIFF
--- a/src/components/BottomNavigation.tsx
+++ b/src/components/BottomNavigation.tsx
@@ -605,10 +605,14 @@ const BottomNavigation = ({
             ? 1
             : 0;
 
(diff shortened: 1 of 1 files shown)


kamui545:
I did not notice any flickering issues while switching tabs since patching with your fix.
I will let you know if it happens again.

Huge thanks, this issue was so troublesome to debug / reproduce.

satya164 (approved):

PR #2557, opened 2021-02-22
refactor(modal): migrate to functional component

crutchcorn:
### Summary

While migrating some components in React Native Paper to "internal" packages (still public, just without deps since we only use one or two low-level components from RNP), we went ahead and refactored the Modal component to be a functional component rather than a class-based one.

@satya164 suggested we open a PR to introduce this change upstream:

https://github.com/callstack/react-native-paper/pull/98#discussion_r580300226

Frankly, however, while this code functions as-expected, we're not particularly happy with how this refactor's code came out. The cyclical nature of the class methods made for a somewhat difficult-to-follow structure that includes some hacky workarounds once migrated to closures where declaration order matters.

We'd love some feedback on how we can improve the readability of this migrated component!

### Test plan

Nothing visual or functionally should have changed. This was tested in my application that uses React Native Paper dialogs without any issues.

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -14,7 +14,7 @@ import {
   getBottomSpace,
 } from 'react-native-iphone-x-helper';
 import Surface from './Surface';
-import { withTheme } from '../core/theming';
+import { useTheme } from '../core/theming';
 
 type Props = {
   /**
@@ -46,15 +46,6 @@ type Props = {
    * Use this prop to change the default wrapper style or to override safe area insets with marginTop and marginBottom.
    */
   style?: StyleProp<ViewStyle>;
-  /**
-   * @optional
-   */
-  theme: ReactNativePaper.Theme;
-};
-
-type State = {
-  opacity: Animated.Value;
-  rendered: boolean;
 };
 
 const DEFAULT_DURATION = 220;
@@ -100,51 +91,41 @@ const BOTTOM_INSET = getBottomSpace();
  * export default MyComponent;
  * ```
  */
-class Modal extends React.Component<Props, State> {
-  static defaultProps = {
-    dismissable: true,
-    visible: false,
-    overlayAccessibilityLabel: 'Close modal',
-  };
+const Modal: React.FC<Props> = ({
+  dismissable = true,
+  visible = false,
+  overlayAccessibilityLabel = 'Close modal',
+  onDismiss,
+  children,
+  contentContainerStyle,
+  style,
+}) => {
+  const visibleRef = React.useRef(visible);
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.visible && !prevState.rendered) {
-      return {
-        rendered: true,
-      };
+  visibleRef.current = visible;
+
+  const { colors, animation } = useTheme();
+
+  const [opacity] = React.useState(new Animated.Value(visible ? 1 : 0));
+
+  const [rendered, setRendered] = React.useState(visible);
+
+  React.useLayoutEffect(() => {
+    if (visible && !rendered) {
+      setRendered(true);
     }
+  }, [visible, rendered]);
 
-    return null;
-  }
+  // Must be done like this in order to handle cyclical function reference
+  // Must be () => () => due to functional API
+  const [handleBack, setHandleBack] = React.useState<() => boolean>(() => () =>
+    false
+  );
 
-  state = {
-    opacity: new Animated.Value(this.props.visible ? 1 : 0),
-    rendered: this.props.visible,
-  };
+  const showModal = React.useCallback(() => {
+    BackHandler.addEventListener('hardwareBackPress', handleBack);
 
-  componentDidUpdate(prevProps: Props) {
-    if (prevProps.visible !== this.props.visible) {
-      if (this.props.visible) {
-        this.showModal();
-      } else {
-        this.hideModal();
-      }
-    }
-  }
-
-  private handleBack = () => {
-    if (this.props.dismissable) {
-      this.hideModal();
-    }
-    return true;
-  };
-
-  private showModal = () => {
-    BackHandler.removeEventListener('hardwareBackPress', this.handleBack);
-    BackHandler.addEventListener('hardwareBackPress', this.handleBack);
-
-    const { opacity } = this.state;
-    const { scale } = this.props.theme.animation;
+    const { scale } = animation;
 
     Animated.timing(opacity, {
       toValue: 1,
@@ -152,13 +133,20 @@ class Modal extends React.Component<Props, State> {
       easing: Easing.out(Easing.cubic),
       useNativeDriver: true,
     }).start();
-  };
 
-  private hideModal = () => {
-    BackHandler.removeEventListener('hardwareBackPress', this.handleBack);
+    return () => {
+      BackHandler.removeEventListener('hardwareBackPress', handleBack);
+    };
+  }, [animation, handleBack, opacity]);
 
-    const { opacity } = this.state;
-    const { scale } = this.props.theme.animation;
+  // Cyclical dep requires me to lazily initialize/call this otherwise we get stuck
+  // in infinite loop
+  const hideModalRef = React.useRef<() => void>(() => () => {});
+
+  const hideModal = React.useCallback(() => {
+    BackHandler.removeEventListener('hardwareBackPress', handleBack);
+
+    const { scale } = animation;
 
     Animated.timing(opacity, {
       toValue: 0,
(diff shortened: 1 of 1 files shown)


src/components/Modal.tsx
```diff
@@ -100,65 +91,62 @@ const BOTTOM_INSET = getBottomSpace();
...
-      }
-    }
-  }
+  const [rendered, setRendered] = React.useState(visible);
 
-  private handleBack = () => {
-    if (this.props.dismissable) {
-      this.hideModal();
+  React.useLayoutEffect(() => {
+    if (visible && !rendered) {
```
satya164, line 114: This can be done in function body instead of in `useEffect` https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

src/components/Modal.tsx
```diff
@@ -100,65 +91,62 @@ const BOTTOM_INSET = getBottomSpace();
  * export default MyComponent;
  * ```
  */
-class Modal extends React.Component<Props, State> {
-  static defaultProps = {
-    dismissable: true,
-    visible: false,
-    overlayAccessibilityLabel: 'Close modal',
-  };
+const Modal: React.FC<Props> = ({
```
satya164, line 94: Avoid `React.FC`. Just annotate the function's props and it's enough.

Also we prefer `export default function Modal`

src/components/Modal.tsx
```diff
@@ -100,65 +91,62 @@ const BOTTOM_INSET = getBottomSpace();
...
-      this.hideModal();
+  React.useLayoutEffect(() => {
+    if (visible && !rendered) {
+      setRendered(true);
     }
-    return true;
-  };
+  }, [visible, rendered]);
+
+  // Must be done like this in order to handle cyclical function reference
```
satya164, line 119: I think a lot of these problems arise due to use of `React.useCallback` as it needs dep array. However we don't need to use `useCallback`, so this is not something we should have to do

src/components/Modal.tsx
```diff
@@ -170,83 +158,90 @@ class Modal extends React.Component<Props, State> {
...
 
-  componentWillUnmount() {
-    BackHandler.removeEventListener('hardwareBackPress', this.handleBack);
-  }
+  hideModalRef.current = hideModal;
 
-  render() {
-    const { rendered, opacity } = this.state;
+  // Handle before initial render to avoid re-rendering issues
+  React.useLayoutEffect(() => {
```
satya164, line 176: Try to avoid `useLayoutEffect` since it warns on SSR and we want to be compatible with SSR. But it shouldn't be needed if you remove usage of `useCallback`

src/components/Modal.tsx
```diff
@@ -170,83 +158,90 @@ class Modal extends React.Component<Props, State> {
...
+  React.useEffect(() => {
+    if (prevVisible.current !== visible) {
+      if (visible) {
+        showModal();
+      } else {
+        hideModal();
+      }
+    }
+    prevVisible.current = visible;
+  }, [prevVisible, visible, showModal, hideModal]);
```
satya164, line 197: The dependency array isn't really necessary here since the checks inside already ensure that the relevant functions only run when the visibility changes

src/components/Modal.tsx
```diff
@@ -100,65 +91,62 @@ const BOTTOM_INSET = getBottomSpace();
...
+}) => {
+  const visibleRef = React.useRef(visible);
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.visible && !prevState.rendered) {
-      return {
-        rendered: true,
-      };
-    }
+  visibleRef.current = visible;
```
satya164, line 105: Don't assign refs during render. They should only be assigned in `useEffect` (except the case of lazy initialization)

src/components/Modal.tsx
```diff
@@ -100,65 +91,62 @@ const BOTTOM_INSET = getBottomSpace();
...
 
-    return null;
-  }
+  const { colors, animation } = useTheme();
 
-  state = {
-    opacity: new Animated.Value(this.props.visible ? 1 : 0),
-    rendered: this.props.visible,
-  };
+  const [opacity] = React.useState(new Animated.Value(visible ? 1 : 0));
```
satya164, line 109: We have a `useAnimatedValue` hook that you can use here

satya164 (changes requested):

src/components/Modal.tsx
crutchcorn: I didn't even know there was a RNW SSR solution out there, admittedly. Will migrate

src/components/Modal.tsx
crutchcorn: I went back on forth on `useCallback`'s usage here, myself. I'm thinking that if I refactor to `function` definition, we can leverage the name hoisting functionality to make the sting of the cyclical calling sting less

src/components/Modal.tsx
crutchcorn: I agree that this is an anti-pattern but my fear here stems from the timing of `useEffect`. This was put in place after I realized that the closure from `hideModal` wasn't updating `visibility` and thus was repeatedly calling `showModal` (at the end of the logic - can't link to LOC cuz on mobile).

If I switch this to a `useEffect`, won't the `.current` assignment occur after the render, causing the same `visibility` issue?

src/components/Modal.tsx
crutchcorn: I've admittedly never had to refactor `getDerivedStateFromProps` to hooks and somehow missed this part of docs. Will update

src/components/Modal.tsx
satya164: Most of the weirdness is because of using `useCallback` and trying to use refs to work around that. I believe this ref isn't needed at all after removing `useCallback`s

satya164: Also refs shouldn't be used during render so this shouldn't be causing any issue

src/components/Modal.tsx
satya164: I'm don't think that's going work with `useCallback`. To best way forward is to remove `useCallback`. There's no reason to be using it here.

crutchcorn:
@satya164 I've removed all instances of `useCallback` and `useLayoutEffect` and I think this is now a fairly acceptable component that's significantly easier to read. (Thanks for helping me think through this! Got a bit lost in the weeds so the fresh perspective helped a lot)

That said, there are still two things that I had to add to make everything work as-expected:

- `function` instead of `const` assignment. This was the only way I could get the cyclical calling of `hideModal` and `handleBack` to work without doing some tricky/hacky `useRef` assignment
- `useRef` for the `handleBack` to bind the `'hardwareBackPress'` event removal. I tried simply using `handleBack` without this `useRef` and a common issue that happened a lot during refactor happened:
    
   The back button would work as-expected, but then when I tried navigating to a different page and pressing the hardware "back" button, it wouldn't work the first time and I'd have to press it twice. 
   
 

I've done my best to document both of these code oddities in the comments, leaving a sufficient explanation of why they're there.

Let me know if there's anything else I can do to get this merged and thanks again for the attentive code review!

src/components/Modal.tsx
```diff
@@ -100,65 +92,73 @@ const BOTTOM_INSET = getBottomSpace();
...
 
-  private showModal = () => {
-    BackHandler.removeEventListener('hardwareBackPress', this.handleBack);
-    BackHandler.addEventListener('hardwareBackPress', this.handleBack);
+  /**
+   * If we don't use a "ref" to keep track of the exact version of the handleBack function we're removing from the event
+   * listener, what will happen is that `handleBack`'s reference in memory will change between `showModal` and `hideModal`
+   * (due to a potential re-render of `visible` being set) being called, leaving the event in-tact and not removing properly
+   */
+  const handleBackToRemove = React.useRef<() => boolean>(() => true);
```
satya164, line 134: I believe `BackHandler.addEventListener` return a subscription that can be stored in the ref and used to unsubscribe. So we should do that instead since it's cleaner.

src/components/Modal.tsx
```diff
@@ -100,65 +92,73 @@ const BOTTOM_INSET = getBottomSpace();
...
 
-    const { opacity } = this.state;
-    const { scale } = this.props.theme.animation;
+  function showModal() {
```
satya164, line 136: Function hoisting shouldn't be needed here. These functions should already be available in scope by the time they will be called - since the component's body is evaluated first where the functions are defined and added to scope, making them available. A function access another variable which is defined below it as long as the function isn't being called before the variable is defined. So `const` would work fine.

src/components/Modal.tsx
```diff
@@ -170,84 +170,76 @@ class Modal extends React.Component<Props, State> {
...
 
-      if (this.props.visible && this.props.onDismiss) {
-        this.props.onDismiss();
+      if (visible && onDismiss) {
+        onDismiss();
       }
 
-      if (this.props.visible) {
-        this.showModal();
+      if (visibleRef.current) {
```
satya164, line 177: I believe the reason for a ref here is because it's being used in a listener and due to that, it may not be uptodate and refer to a stale vallue. But it may not be obvious why on first look.

I'd suggest to store `handleBack` in a ref and then do:

```js
backHandlerSub.current = BackHandler.addEventListener('hardwareBackPress', () => handleBackRef.current());
```

And we can assign the callback in the ref in a `useEffect` (as well as ref creation).

This will fix the issue at the point where it originates so it's more obvious. And also avoids other mistakes (like `visible` state being used the line before).

src/components/Modal.tsx
```diff
@@ -170,84 +170,76 @@ class Modal extends React.Component<Props, State> {
...
+      if (visible) {
+        showModal();
+      } else {
+        hideModal();
+      }
+    }
+    prevVisible.current = visible;
+  });
+
+  if (!rendered) return null;
```
satya164, line 198: It's a minor issue, but we should use `!rendered && !visible` since unlike `getDerivedStateFromProps`, `rendered` won't be correct the first time after `visible` changed until the `setRendered` triggers another render.

Or to make it clear, maybe set a varible when you call `setRendered(true)` above and used it in all places.

satya164 (changes requested):
Thanks for the updates. Left some suggestions which should clarify the points your raised.

src/components/Modal.tsx
crutchcorn: I didn't know that `addEventListener` returned anything - Totally agree. I'll refactor

src/components/Modal.tsx
crutchcorn: 🤦 Thinking back, I already knew this at some point. Don't know why I got "function hoists, const doesn't" stuck in my head so bad this time around. Thanks for the reminder, will update

src/components/Modal.tsx
crutchcorn: I believe the reason for this `ref.current` is so that when the `hideModal` function is called, it can tell if `onDismiss` actually updated the `visible` prop and conditionally either re-run `showModal` if it didn't and `setRendered(false)` if it did.

The rationale behind it being a ref is because if it's not inside of a ref, the value becomes stale and will always call `showModal` when dismissing via `onDismiss`, causing an infinite render in certain conditions (like pressing the scrim)

I'm admittedly unsure how to work around this without a ref, admittedly 😅

src/components/Modal.tsx
crutchcorn: Migrated to `!rendered && !visible`

src/components/Modal.tsx
satya164: I'm not sure I understand. `showModal()` will be called inside a `useEffect`which should already be up-to-date. Why would `visible` be stale here? Only reason I can think is that because it's in a async callback, but also if `visible` changed in middle of animation, this code shouldn't run since animation should be cancelled and `finished` would be `false`.

Also if `visible` is stale here, wouldn't it be stale the line above?

crutchcorn: Well, it's less `visible` is stale, and more checking if `visible` is set to `false` during the next re-render due to `onDismiss` changing the passed in value. This is because `this.props` is able to update in class comps and update the `this.props` object between checking that variable once and then not again.

I generally agree that `visible=false` should be set during `onDismiss`, and therefore we could simply remove this `ref` check. That would allow us to simplify the code a bit but would change behavior from previous if a user did not set `visible` to false in `onDismiss`

crutchcorn:
@satya164 are we able to get this merged in by any chance?

satya164:
cc @lukewalczak

crutchcorn:
Pinging @satya164 @lukewalczak 

Is this good-to-go? I know there's a merge conflict again, but I want to get a sign-off before I resolve

src/components/Modal.tsx
```diff
@@ -100,51 +93,53 @@ const BOTTOM_INSET = getBottomSpace();
...
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.visible && !prevState.rendered) {
-      return {
-        rendered: true,
-      };
-    }
+  React.useEffect(() => {
+    visibleRef.current = visible;
+  });
```
lukewalczak, line 109: It can be removed

src/components/Modal.tsx
```diff
@@ -100,51 +93,53 @@ const BOTTOM_INSET = getBottomSpace();
...
 
-  private showModal = () => {
-    BackHandler.removeEventListener('hardwareBackPress', this.handleBack);
-    BackHandler.addEventListener('hardwareBackPress', this.handleBack);
+  /**
+   * If we don't use a "ref" to keep track of the exact version of the handleBack function we're removing from the event
+   * listener, what will happen is that `handleBack`'s reference in memory will change between `showModal` and `hideModal`
+   * (due to a potential re-render of `visible` being set) being called, leaving the event in-tact and not removing properly
+   */
+  const backHandlerSub = React.useRef<NativeEventSubscription | null>(null);
```
lukewalczak, line 134:
```suggestion
  const handleBackRef = React.useRef(handleBack);

  React.useEffect(() => {
    handleBackRef.current = handleBack;
  });
```

src/components/Modal.tsx
lukewalczak:
```suggestion
      if (visible) {
```

crutchcorn:
@lukewalczak the suggestions you provided still incur incorrect functionality.

https://github.com/callstack/react-native-paper/pull/2557#discussion_r583225491

This comment is still accurate in describing what's happening. Basically, it's that within `hideModal`, `visible` is stale and as a result, `showModal` is called immediately after requesting to hide every time.

Behavior can be seen in action in this GIF:

Showing a dialog close, then immediately reopen again

I agree that the code isn't _pretty_ to do what it is, but I can't think of a way to remove that code without introducing a breaking change of some kind into the modal API.

crutchcorn:
Requesting a re-review. Want to loop back to the contested line-of-code and how to move forward, given that removing said line-of-code drastically breaks functionality (see above)

crutchcorn:
Conflicts resolved! :)

PR #2559, opened 2021-02-23
fix: aligned the icons in the Menu

samuel-sou:
### Summary

Saw an issue where the icons in the menu were not properly aligned. So just added this line of code to fix that.

### Test plan

Before:
Screenshot from 2021-02-23 08-59-22

After: 
Screenshot from 2021-02-23 08-59-57

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -156,6 +156,7 @@ const styles = StyleSheet.create({
   },
   icon: {
     width: iconWidth,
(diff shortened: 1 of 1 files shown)


wojteg1337 (approved):

wojteg1337:
Looks fine, just not sure about it, because material design does not follow icon support for mobile..

Desktop has different horizontal spacing: (24 x 20)

image

@Trancever Maybe we should set the icons margin to be left X+4 and right to X (Left: 16, Right: 12)?

PR #2560, opened 2021-02-23
chore: remove constants.expo.tsx and move logic to constants.tsx

brentvatne:
### Summary

We're removing the .expo extension in SDK 41. In the future we should move towards using react-native-safe-area-context to accurately determine the insets rather than estimating, but that will require a major version bump, so I discussed this with @satya164 that the solution in this PR should be fine until then.

I also updated the expo-cli version in the example project and swapped from react-native's `StatusBar` component to expo-status-bar, because its defaults work better in this context (eg: assumes translucent status bar). 

### Test plan

I ran the example project and verified that it works as expected:

Screen Shot 2021-02-23 at 12 35 12 PM

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -5,10 +5,10 @@
   "author": "",
   "private": true,
   "scripts": {
-    "start": "expo start",
-    "android": "expo android",
-    "ios": "expo ios",
-    "web": "expo start:web"
+    "start": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start",
+    "android": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --android",
+    "ios": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --ios",
+    "web": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --web"
   },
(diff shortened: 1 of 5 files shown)


example/package.json
```diff
@@ -5,10 +5,10 @@
   "author": "",
   "private": true,
   "scripts": {
-    "start": "expo start",
-    "android": "expo android",
-    "ios": "expo ios",
-    "web": "expo start:web"
+    "start": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start",
```
brentvatne, line 8: expo-cli will create a tsconfig.json in your app directory if you have typescript source files but no tsconfig. i'm not sure how typechecking on the example directory is meant to work (it was not working for me in vscode) but i didn't want to mess with it in this pr so i just disabled the auto initialization of tsconfig.json.

satya164 (approved):

lukewalczak:
Thanks @brentvatne for handling this!

AugustinasMK:
@lukewalczak when can we expect this to be available in npm?

lukewalczak:
In the next few day I think

PR #2564, opened 2021-02-25
feat: Update searchbar props

bmburlingame:
onFocus and onBlur are valid props for the searchbar

### DIFF
--- a/src/components/Searchbar.tsx
+++ b/src/components/Searchbar.tsx
@@ -47,11 +47,11 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
   /**
    * Set style of the TextInput component inside the searchbar
    */
-  onFocus?: () => void;
+  onFocus?: (args: any) => void;
   /**
    * Callback that is called when the text input is focused.
    */
-  onBlur?: () => void;
+  onBlur?: (args: any) => void;
   /**
    * Callback that is called when the text input is blurred.
    */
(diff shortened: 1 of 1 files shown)


src/components/Searchbar.tsx
```diff
@@ -47,6 +47,14 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
   /**
    * Set style of the TextInput component inside the searchbar
    */
+  onFocus?: (args: any) => void;
+  /**
+   * Callback that is called when the text input is focused.
+   */
+  onBlur?: (args: any) => void;
```
wojteg1337, line 54: I think this is not a valid type for callback argument

src/components/Searchbar.tsx
```diff
@@ -47,6 +47,14 @@ type Props = React.ComponentPropsWithRef<typeof TextInput> & {
   /**
    * Set style of the TextInput component inside the searchbar
    */
+  onFocus?: (args: any) => void;
```
wojteg1337, line 50: Callback type here is `NativeSyntheticEvent` of `TextInput`

wojteg1337 (changes requested):

lukewalczak:
Hey @bmburlingame I think there is no need to introduce these two props since `Searchbar` already inherits them from `TextInput` on which it's based.

PR #2577, opened 2021-03-02
feat: add accessibilityLabel prop to MenuItem

ahce:
### Summary

Add accessibilityLabel prop to MenuItem

### Test plan

1. Download this branch
2. Create Menu
3. Pass accessibilityLabel to MenuItem


src/components/Menu/MenuItem.tsx
```diff
@@ -41,6 +41,7 @@ type Props = {
    * TestID used for testing purposes
    */
   testID?: string;
+  accessibilityLabel?: string;
```
brunohkbx, line 44: Can you add a description above this prop? Something like:

```
/**
 * Accessibility label for the Touchable. This is read by the screen reader when the user taps the component.
*/
```

ahce: Done!

lukewalczak (approved):

lukewalczak (approved):

Issue #2596, opened 2021-03-09
feat: add numberOfLines props to Snackbar


whoami-shubham:
Right now I couldn't find any props for trimming the snack bar text, I can trim the text before giving it to the snack bar however that wouldn't be very responsive. if the text is very long the snack bar looks like following
![image](https://user-images.githubusercontent.com/28999685/110437915-e306d100-80db-11eb-953b-01d2ba225791.png)

if I provide numberOfLines = 2 at [here](https://github.com/callstack/react-native-paper/blob/975cfa3be4956b629aa8255e226ff9ca068448eb/src/components/Snackbar.tsx#L205)  it looks like following 
![image](https://user-images.githubusercontent.com/28999685/110438381-527cc080-80dc-11eb-95a0-cf3d13a18a55.png)

let me know is it a good idea to work on it.

1. react-native : 0.63.3
2. react-native-paper: 4.2.0
3. react-native-vector-icons:6.6.0

whoami-shubham:
yes

PR #2607, opened 2021-03-16
feat: A new feature numberOfLines props added in snackbar

whoami-shubham:
### Summary

This pull request solves this issue https://github.com/callstack/react-native-paper/issues/2596

### Test plan

in your local project which is using `react-native-paper` create a snackbar component and copy the snackbar contents from pull request snackbar file, and if any component is has missing import in snackbar component, then import it from  `react-native-paper`

### DIFF
--- a/src/babel/__fixtures__/rewrite-imports/output.js
+++ b/src/babel/__fixtures__/rewrite-imports/output.js
@@ -1,11 +1,11 @@
 /* eslint-disable prettier/prettier */
 import { Text } from 'react-native';
-import PaperProvider from "react-native-paper/lib/module/core/Provider";
-import BottomNavigation from "react-native-paper/lib/module/components/BottomNavigation";
-import Button from "react-native-paper/lib/module/components/Button";
-import FAB from "react-native-paper/lib/module/components/FAB";
-import Appbar from "react-native-paper/lib/module/components/Appbar";
-import * as Colors from "react-native-paper/lib/module/styles/colors";
+import PaperProvider from "react-native-paper/lib\\module\\core\\Provider";
+import BottomNavigation from "react-native-paper/lib\\module\\components\\BottomNavigation";
+import Button from "react-native-paper/lib\\module\\components\\Button";
+import FAB from "react-native-paper/lib\\module\\components\\FAB";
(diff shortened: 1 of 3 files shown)


src/components/Snackbar.tsx
```diff
@@ -207,6 +212,7 @@ const Snackbar = ({
             styles.content,
             { marginRight: action ? 0 : 16, color: colors.surface },
           ]}
+          numberOfLines={numberOfLines}
```
lukewalczak, line 215: According to the MaterialDesign guidelines we shouldn't allow to customize the number of lines. Documentation indicates that on mobile number of lines should be up to two and on the web it should be single line. Then basically value should be set conditionally:

`numberOfLines={Platform.OS === 'web' ? 1 : 2}`

src/babel/__fixtures__/rewrite-imports/output.js
```diff
@@ -1,11 +1,11 @@
...
+import BottomNavigation from "react-native-paper/lib\\module\\components\\BottomNavigation";
+import Button from "react-native-paper/lib\\module\\components\\Button";
+import FAB from "react-native-paper/lib\\module\\components\\FAB";
+import Appbar from "react-native-paper/lib\\module\\components\\Appbar";
+import * as Colors from "react-native-paper/lib\\module\\styles\\colors";
 import { NonExistent, NonExistentSecond as Stuff, Theme } from "react-native-paper/lib/module/index.js";
-import { ThemeProvider } from "react-native-paper/lib/module/core/theming";
-import { withTheme } from "react-native-paper/lib/module/core/theming";
+import { ThemeProvider } from "react-native-paper/lib\\module\\core\\theming";
+import { withTheme } from "react-native-paper/lib\\module\\core\\theming";
```
lukewalczak, line 11: Do we need that changes? If not please revert it

src/components/Snackbar.tsx
fabervae: What do you think about setting the value conditionally only if no value was provided? I believe this would achieve the best of both worlds.

```suggestion
          numberOfLines={
            numberOfLines > 0
              ? numberOfLines
              : Platform.OS === 'web' ? 1 : 2
          }
```

PR #2610, opened 2021-03-19
fix(BottomNavigation): selected state now reads properly on voiceover

crutchcorn:
### Summary

Originally, when using TalkBack to read through the bottom nav, it would tell me a tab was selected, even when it was not. This PR fixes that, and matches the older behavior that is still in place with `accessibilityTraits`

### Test plan

Open app that utilized bottom nav, turn on screenreader (I'm using TalkBack on Android to test)


brunohkbx:
@crutchcorn same here. Please update the snapshots :smile:

crutchcorn:
@brunohkbx done for both PRs now

brunohkbx (approved):

PR #2611, opened 2021-03-19
fix(BottomNavigation): migrate to tab role instead of button

crutchcorn:
### Summary

As-of React Native v0.60.0, we now have the `tab` and `tablist` accessibility roles that we're able to use. They seemed like a better fit for the BottomNavigation actings more as tabs than buttons, so the roles were switched.

Checked with @satya164 on Twitter about this beforehand, and he suggested opening a PR, so here we are.

https://twitter.com/satya164/status/1370197157328867329

### Test plan

- Open app with `BottomNavigation` implemented
- Turn on screen-reader
- Listen for ""tab"


brunohkbx:
@crutchcorn thank you for the PR. Could you please update the snapshots as well?

brunohkbx (approved):

src/components/BottomNavigation.tsx
```diff
@@ -759,7 +760,7 @@ const BottomNavigation = ({
                   ? ['button', 'selected']
                   : 'button',
                 accessibilityComponentType: 'button',
-                accessibilityRole: 'button',
+                accessibilityRole: 'tab',
```
satya164, line 763: From my tests, `tab` doesn't seem to work as expected on iOS. So we'd probably need to conditionally set `button` for iOS

brunohkbx: Worked for me in an old iPhone 5s

satya164: Can you clarify what you mean by it worked? VoiceOver doesn't announce it like it does with native tabs. Also the tabs are read as "tab description" which is incorrect.

brunohkbx: Oh I see. I don't know how native is announced that's why I thought it was working when it read "tab"

lukewalczak: [That PR](https://github.com/facebook/react-native/blob/v0.64.0/React/AccessibilityResources/en.lproj/Localizable.strings) is fixing weird behavior related to "tab description" however it was not released in rn version 0.64. As Satya suggest let's go with `button` role for now.

lukewalczak:
Hey @crutchcorn, could you please revert the change ` accessibilityRole: 'tab'`, update the tests again and resolve the conflict?

crutchcorn:
@lukewalczak I assume we only want to do that for iOS conditionally?

lukewalczak:
You're totally correct. I wrote "revert" however forgot that on Android it works properly (_IIRC_)

crutchcorn:
Apologies for the delay - iOS now has a specific conditional and the merge conflict has been fixed

lukewalczak:
Thanks @crutchcorn !

PR #2640, opened 2021-04-01
fix: do not render empty string as section title

karszawa:
## Summary

List.Section tries to render empty string without `Text` when `""` is passed to `title` prop and causes an error:

image

## Test plan

Render this:

```
<List.Section title="" />
```


brunohkbx (approved):

brunohkbx:
Thanks @karszawa

Issue #2598, opened 2021-03-10
TextInput with mode outlined - Outline over label when focused


denniske:
### Current behaviour

When the text input is focused the outline is all around the text input and as such also over the label.

### Expected behaviour

Do not cover the label with the outline.

### Code sample

```
<TextInput
    right={
        <TextInput.Affix
            text={' %'}
        />
    }
    dense={true}
    mode="outlined"
    label="Opacity"
    value={text}
    onChangeText={text => setText(text)}
/>

```
### Screenshots (if applicable)

![image](https://user-images.githubusercontent.com/1076211/110711663-370ad600-8200-11eb-8f31-7d004384bacf.png)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        |  web
| react-native          | 0.63
| react-native-paper    | 4.7.2
| node                  | 12
| npm or yarn           | both
| expo sdk              | 40

aiapps2020:
same issues using
"react": "16.13.1",
"react-native": "0.63.4",
"react-native-vector-icons": "^8.1.0"
"react-native-paper": "4.7.2",

NoumanSakhawat:
same issues using
"react": "16.13.1",
"react-native": "0.63.4",
"react-native-vector-icons": "^8.1.0"
"react-native-paper": "4.7.2",

brunohkbx:
@denniske @NoumanSakhawat  can you please create a snack reproducing this issue?

FelipeACP:
https://snack.expo.io/P1bac2SGL

PR #2654, opened 2021-04-09
fix: remove additional outline from web inputs

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/2598

### Summary

That PR is removing the outline from web text input which is overlapping with component border. 
**NOTE**: we are providing the visual indicator/effect when input if focused, so there is no regression in terms of accessibility.

### Test plan

1. Use text input on the web
2. Focus the text input
3. Expect there is additional `outline`

before | after
--- | ---
image | image


Trancever (approved):
Nice!

PR #2661, opened 2021-04-13
feat: add prop color to TextInputIcon

brunohkbx:
### Summary
This PR adds a new prop `color` to the component `TextInputIcon`. It accepts whether a function or a string, making it easier to change the icon color when the input is focused as in the MD example below.

Screenshot from 2021-04-12 23-09-41

### Test plan
Check the example app

### DIFF
--- a/example/src/Examples/TextInputExample.tsx
+++ b/example/src/Examples/TextInputExample.tsx
@@ -95,7 +95,7 @@ const TextInputExample = () => {
   const _isUsernameValid = (name: string) => /^[a-zA-Z]*$/.test(name);
 
   const {
-    colors: { background, accent },
+    colors: { background, accent, primary },
   } = useTheme();
 
   const inputActionHandler = (type: keyof State, payload: string) =>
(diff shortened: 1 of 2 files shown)


example/src/Examples/TextInputExample.tsx
```diff
@@ -215,6 +215,12 @@ const TextInputExample = () => {
             inputActionHandler('flatDenseText', flatDenseText)
           }
           left={<TextInput.Affix text="#" />}
+          right={
+            <TextInput.Icon
+              name="chevron-up"
+              color={(focused) => (focused ? primary : 'black')}
```
lukewalczak, line 221: I think we shouldn't use hardcoded color since it will not change when switching between themes. I would suggest to use as a fallback `text` color from theme, which is used as a [default](https://github.com/callstack/react-native-paper/blob/main/src/components/IconButton.tsx#L103) in `IconButton` or type that prop as a possibly `undefined` and then fallback to it.

lukewalczak (changes requested):
Left one comment, except it lgtm. Thanks!

brunohkbx:
@lukewalczak thanks for reviewing it. I updated the type to `color?: ((isTextInputFocused: boolean) => string | undefined) | string;` and now is working fine to both light/dark themes.

lukewalczak (approved):

PR #2664, opened 2021-04-13
fix: correct example list background color

lukewalczak:
### Summary

That PR corrects background color on example list.

### Test plan

1. Open example app
2. Switch mode to `dark`
3. Expect example list has correct bg color when bounces

before | after
--- | ---
wrong | correct


brunohkbx (approved):

raajnadar (approved):
Looks good now

PR #2455, opened 2020-12-20
feat: DataTable Pagination

unikvozm:
### Summary
#2416

- [x] add navigation to first and last page
- [x] fix bug with text displaying the last row
- [x] add select for number of rows

### Test plan

Screenshot 2021-01-10 at 17 51 57
Screenshot 2021-01-10 at 17 52 08
Screenshot 2021-01-10 at 17 52 08
Screenshot 2021-01-18 at 10 19 57

PR #2674, opened 2021-04-21
Fixed documentation

macss:
Number of pages was Math.floor() causing component not to work, changed to Math.ceil() which corrects malfunction


brunohkbx:
It was updated in this PR #2455

PR #2721, opened 2021-05-12
fix: example app tweaks

lukewalczak:
### Summary

- [X] reload the app when RTL toggle switched
- [X] correct example list paddings
- [x] create `ScreenWrapper` with safe area
- [x] use `DrawerContentScrollView`


example/src/Examples/ChipExample.tsx
```diff
@@ -1,17 +1,16 @@
...
+import ScreenWrapper from '../ScreenWrapper';
 
 const ChipExample = () => {
   const [visible, setVisible] = React.useState<boolean>(false);
   const { colors } = useTheme();
 
   return (
     <>
-      <ScrollView
-        style={[styles.container, { backgroundColor: colors.surface }]}
```
brunohkbx, line 13: styles.container is not being used anymore

brunohkbx (approved):

dsznajder (approved):
:eyes: Nice refactor here

PR #2734, opened 2021-05-18
feat: Added new prop (iconRight) to drawer item

varunhhhrahul:
### Summary

This pull request adds iconRight prop to drawer item, so that the users can have the option of placing an icon to right of text instead of only left.


lukewalczak:
Actually that prop is already added via [merged PR](https://github.com/callstack/react-native-paper/pull/2505), but it's not released yet.

PR #2735, opened 2021-05-19
fix: correct Button elevation

tasugi:
### Summary

It has a bug that Button elevation is still zero after mode is changed to `contained` from other modes.
This PR fixes this.

### Test plan

1. Render a Button with mode other than `contained`.
2. Change the mode to `contained` mode dynamically(e.g. on button press).
3. Check elevation

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -143,8 +143,9 @@ const Button = ({
   ...rest
 }: Props) => {
   const { current: elevation } = React.useRef<Animated.Value>(
-    new Animated.Value(mode === 'contained' ? 2 : 0)
+    new Animated.Value(0)
   );
+  elevation.setValue(mode === 'contained' ? 2 : 0);
 
   const handlePressIn = () => {
     if (mode === 'contained') {
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -145,7 +145,9 @@ const Button = ({
   const { current: elevation } = React.useRef<Animated.Value>(
     new Animated.Value(0)
   );
-  elevation.setValue(mode === 'contained' ? 2 : 0);
+  React.useEffect(() => {
+    elevation.setValue(mode === 'contained' ? 2 : 0);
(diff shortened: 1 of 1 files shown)


brunohkbx:
related to #2736? cc @lukewalczak

lukewalczak:
@tasugi could you please provide the snack to the issue which you're solving, where we can observe the bug?

tasugi:
@lukewalczak Sure! I created a snack. https://snack.expo.io/@tasugi/reactnativepaper-pr2735

If you press the button, you can change the mode of it `contained`. I think you can see the button is not elevated.( Element inspector may help you to check it)

src/components/Button.tsx
```diff
@@ -143,8 +143,9 @@ const Button = ({
   ...rest
 }: Props) => {
   const { current: elevation } = React.useRef<Animated.Value>(
-    new Animated.Value(mode === 'contained' ? 2 : 0)
+    new Animated.Value(0)
   );
+  elevation.setValue(mode === 'contained' ? 2 : 0);
```
lukewalczak, line 148:
```suggestion
  React.useEffect(() => {
    elevation.setValue(mode === 'contained' ? 2 : 0);
  }, [mode, elevation]);
```

tasugi:
@lukewalczak Thank you for suggestions!

src/components/Button.tsx
```diff
@@ -143,8 +143,11 @@ const Button = ({
   ...rest
 }: Props) => {
   const { current: elevation } = React.useRef<Animated.Value>(
-    new Animated.Value(mode === 'contained' ? 2 : 0)
+    new Animated.Value(0)
```
lukewalczak, line 146: I guess we have to revert that change to initially have elevation set to 2 for contained buttons

tasugi:
Fixed an issue that caused the test to fail.

lukewalczak:
Thanks @tasugi!

Issue #2733, opened 2021-05-18
Unable to remove box-shadow (elevation) on Card component (iOS)


g-hamilton:
"react-native-paper": "^4.8.1"

Hi, I'm trying to remove the box-shadow completely on the Card component. Setting `elevation={0}` seems to work on web, but doesn't appear to be working on iOS (untested on Android).

When I preview on Snack using iOS, I can still see a faint gray line around the card. I've also tried `style={{ boxShadow: "none" }}` but that seems to make the shadow even more visible.

How can I remove all box-shadow so that there's no visible border/shadow around the card?
[card.zip](https://github.com/callstack/react-native-paper/files/6500744/card.zip)

Thanks!

lukewalczak:
Hey @g-hamilton, try to use `shadowColor: 'transparent' within Card styles.

g-hamilton:
Thanks @lukewalczak that seems to work :)

Unfortunately, I'm doing all my styling in styled components (written in CSS syntax) so the `shadowColor` property (JSX) isn't the ideal solution. It would be really nice if `elevation={0}` did that for me ;)

But.. works for now so thanks!

satya164:
> shadowColor property (JSX) 

`shadowColor` is a style

PR #2736, opened 2021-05-19
fix: interpolate shadowOpacity to correct its value for the elevation equal zero

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/2733

### Summary

Fixes the shadow which remains even if the elevation was set to 0.

### Test plan

1. Run the app
2. Set `elevation={0}` on Card component
**3. Expect there is no shadow**


dsznajder (approved):
Tested

brunohkbx (approved):

PR #2743, opened 2021-05-25
fix(android): add importantForAccessibility in Modal component

brunohkbx:
### Summary
It wasn't being possible to focus on the content under the dialog.

Before | After
--- | ---
before | after

### Test plan
1. Open the dialog example
2. Focus on the dialog content using TalkBack


lukewalczak (approved):
LGTM, thanks @brunohkbx!

Issue #1852, opened 2020-04-22
Disable checkmark icon when chip selected


AvneeshD:
I don't want the checkmark to appear when I select a chip, I'm already changing it's background color. How can I do that?

xXFracXx:
Can this be done?

sharmad1205:
@AvneeshD  a workaround for this dont use selected option in your chip

BishoyBishai:
set icon with { source: null }

Tze-Sien:
How can I set the selected "tick" icon to be null ?

shanedaugherty:
The tick looks pretty old school. You can get rid of it by passing a function that returns null. But that doesn't remove the padding so unfortunately, the chip text is off-center. 

```
const emptyIcon = () => null

const MyComponent = () => (
  <Chip icon={emptyIcon}
)
```

<img width="59" alt="Screen Shot 2021-05-30 at 9 08 13 PM" src="https://user-images.githubusercontent.com/32719539/120134451-5aa04580-c18b-11eb-9a8f-ef74c720f60f.png">

gregorojstersek:
@BishoyBishai @shanedaugherty It seems that resolving with null or {source: null} creates an empty space on the left. See picture below (first one is without it and second one has it):

![Screenshot 2022-11-14 at 15 43 35](https://user-images.githubusercontent.com/9784551/201689192-420132ef-0926-456b-ac5f-c85d4f6c503c.png)

sivakarthikeyanrumango:
in Latest version you can use `showSelectedCheck={false}` to hide checkmark

PR #2753, opened 2021-05-31
Chip - optional icon 

shanedaugherty:
### Summary

The `Chip` component has some undesired internal padding when the icon is removed. It looks like this is because the icon was not designed to be optional.

Currently, we can do this (among other methods) to render the chip without an icon:
```
const emptyIcon = () => null

const MyComponent = () => (
  <Chip icon={emptyIcon}
)
```

However, as seen in this screenshot, there is some leftover padding inside the chip that is designed to create space between the icon and the text - even when the icon is not there.

Screen Shot 2021-05-31 at 7 39 35 AM

### Test plan

- render a chip component and pass `null` as the icon, notice that the padding is now gone
- render a chip with an icon and notice that the padding is there
- render a chip with an avatar and notice that the padding is still there

Relates to: https://github.com/callstack/react-native-paper/issues/1852

### DIFF
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -34,7 +34,7 @@ type Props = React.ComponentProps<typeof Surface> & {
   /**
    * Icon to display for the `Chip`. Both icon and avatar cannot be specified.
    */
-  icon?: IconSource;
+  icon?: IconSource|null;
   /**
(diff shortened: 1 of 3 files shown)


shanedaugherty:
Hi

aLLUPS:
This is an essential update. Can we reopen this PR again, plesae?

PR #2758, opened 2021-06-02
feat: banner, add optional callback for on animation complete

daniel112:
### Summary

After using the Banner component, for better flexibility for customization, I find that it could be useful for consumers to hook into the callback for when the default animation completes.
i.e doing some action after the banner is completely gone as opposed to during animation after setting `visible=false`

### Test plan

I've modified the Banner example to log out whenever it completes the `open/close` animation. If that is not a preferred modification to examples then you can test by:
1. modifying `BannerExample.tsx` for `onAnimateShowFinish` and/or `onAnimateHideFinish` callbacks

```
        <Banner
          ...otherprops
          onAnimateShowFinish={() => console.log('Completed opening animation')}
          onAnimateHideFinish={() => console.log('Completed closing animation')}
        >

      </Banner>
```
2. Run the example app
3. toggle the Banner show/hide
4. see that it logs out at the appropriate time

### DIFF
--- a/example/src/Examples/BannerExample.tsx
+++ b/example/src/Examples/BannerExample.tsx
@@ -26,6 +26,8 @@ const BannerExample = () => {
           ]}
           icon={require('../../assets/images/email-icon.png')}
           visible={visible}
+          onAnimateShowFinish={() => console.log('Completed opening animation')}
+          onAnimateHideFinish={() => console.log('Completed closing animation')}
         >
           Two line text string with two actions. One to two lines is preferable
           on mobile.
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -48,6 +48,16 @@ type Props = $RemoveChildren<typeof Surface> & {
    * @optional
    */
   theme: ReactNativePaper.Theme;
+  /**
+   * @optional
+   * Optional function that will be called after the open animation finished running normally
+   */
+  onAnimateShowFinish?: Animated.EndCallback;
+  /**
+   * @optional
+   * Optional function that will be called after the open animation finished running normally
+   */
+  onAnimateHideFinish?: Animated.EndCallback;
 };
 
 type NativeEvent = {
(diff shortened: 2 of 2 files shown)


daniel112:
@brunohkbx

example/src/Examples/BannerExample.tsx
```diff
@@ -26,6 +26,8 @@ const BannerExample = () => {
           ]}
           icon={require('../../assets/images/email-icon.png')}
           visible={visible}
+          onAnimateShowFinish={() => console.log('Completed opening animation')}
```
lukewalczak, line 29: I would suggest to rename those two props to: `onShowAnimationFinished` and `onHideAnimationFinished`

src/components/Banner.tsx
```diff
@@ -48,6 +48,16 @@ type Props = $RemoveChildren<typeof Surface> & {
    * @optional
    */
   theme: ReactNativePaper.Theme;
+  /**
+   * @optional
+   * Optional function that will be called after the open animation finished running normally
```
lukewalczak, line 53:
```suggestion
   * Optional callback that will be called after the opening animation finished running normally
```

src/components/Banner.tsx
```diff
@@ -48,6 +48,16 @@ type Props = $RemoveChildren<typeof Surface> & {
...
+   */
+  onAnimateShowFinish?: Animated.EndCallback;
+  /**
+   * @optional
+   * Optional function that will be called after the close animation finished running normally
```
lukewalczak, line 58:
```suggestion
   * Optional callback that will be called after the closing animation finished running normally
```

lukewalczak (changes requested):
I would rename those

example/src/Examples/BannerExample.tsx
daniel112: will do

lukewalczak (approved):

PR #2771, opened 2021-06-13
feat: add line height prop to flat and outlined input label

akeuk:
### Summary
Line height style not working on InputLabel when we add lineHeight to TextInput style props

### Test plan

 Add lineHeight to flat and outlined TextInput, lineHeight  must be applied to TextInput and InputLabel.


lukewalczak:
Implemented in the library from `5.0.0`.

PR #2786, opened 2021-06-23
fix: TextInput border being cut out

brunohkbx:
### Summary

Following up #2678. `useTheme` doesn't work in the component `LabelBackground` when you pass a custom theme to the TextInput as in the example below:

```js
<TextInput
  mode="outlined"
  label="Test"
  theme={{
    roundness: 25,
  }}
  underlineColor="red"
  value="custom rounded input"
/>
```

The solution was to pass the prop `roundness` along under the `labelProps`

Before | After
--- | ---
before | after

### Test plan
1. Check it out in the example app

### DIFF
--- a/example/src/Examples/TextInputExample.tsx
+++ b/example/src/Examples/TextInputExample.tsx
@@ -116,7 +116,7 @@ const TextInputExample = () => {
   return (
     <TextInputAvoidingView>
       <ScreenWrapper
-        style={styles.container}
+        contentContainerStyle={styles.container}
         keyboardShouldPersistTaps={'always'}
         removeClippedSubviews={false}
       >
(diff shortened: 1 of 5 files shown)


example/src/Examples/TextInputExample.tsx
```diff
@@ -435,6 +435,13 @@ const TextInputExample = () => {
             }}
           />
         </View>
+        <TextInput
```
lukewalczak, line 438: Please wrap new example within the `View` with `styles.inputContainerStyle` to provide the same margin.

lukewalczak (approved):
Thanks for that fix @brunohkbx 🎉 , left one small nit.

rahmanharoon:
How can I move that label to right side? Is their any method to do that

PR #2791, opened 2021-06-25
feat: allows hide a tab in bottom tab bar

baptistemon:
### Summary

allows hide a tab in bottom tab bar

react-navigation pull request : https://github.com/react-navigation/react-navigation/pull/9697


lukewalczak:
According to the discussion within the react-navigation repository and @satya164's [reply](https://github.com/react-navigation/react-navigation/pull/9697#issuecomment-870846112), closing it.

Issue #2730, opened 2021-05-17
TextInput content scrolls in outlined mode on Android


pedrobneto:
### Current behaviour

When height property is set on an outlined input, it becomes scrollable when focused

Edit: only happens on Android

### Expected behaviour

TextInput should not be scrollable

### Code sample

`import * as React from 'react';
import {TextInput} from 'react-native-paper';

export default function App() {
  return (
        <TextInput
          style={{marginTop: 24, height: 40}}
          mode='outlined'
          placeholder='This placeholder will scroll when focused'
        />
  );
}`

### Screenshots (if applicable)

![input-bug](https://user-images.githubusercontent.com/15621292/118539241-94059980-b725-11eb-9f6d-f8d07a3b9f25.gif)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        |
| react-native          | "4.13.0"
| react-native-paper    | "^4.8.1"
| node                  | "14.7.0"
| npm or yarn           | "7.12.0"
| expo sdk              |  "3.24.0"

Snack that reproduces the issue:
[https://snack.expo.io/6tTczRnvw](url)

lukewalczak:
Hey @PedroBern, I will check that, however since you set the height on `40` maybe you can use the [prop](https://callstack.github.io/react-native-paper/text-input.html#dense) `dense` as a workaround?

pedrobneto:
Hi @lukewalczak, thanks for replying. I'm currently using dense, but this happens regardless of the height set, it will still "scroll", which makes it impossible to use inside a SrollView, for example.

lukewalczak:
@PedroBern Without specifying the height and with `dense` prop, I'm not able to observe the input scrolling issue on the [snack](https://snack.expo.io/lBhW9y4Fs). Do I miss something?

pedrobneto:
@lukewalczak, you do need to specify the height to reproduce it, I meant that it doesn't matter what height value you use, the bug still happens. Using the dense prop without height is working for me as well.

sommcz:
same here, on android. If i specify any height, it happens. Calculated paddingTop/Bottom is probably too big. But if i change it to zero (with render prop), it works as expected in my case (text is already aligned in the center so paddingTop/Bottom is not needed for my single line case).

lukewalczak:
@sommcz could you please attach the snack presenting the issue?

sommcz:
https://snack.expo.dev/wV8wWJ9pS

Type some text and drag it up and down.

[...]

PR #2793, opened 2021-06-26
fix: outlined TexInput adjustPaddingOut (#2740)

Kamalnrf:
### Summary

Fixes issue [2740](https://github.com/callstack/react-native-paper/issues/2740). The issue happens to be when we calculate the adjustPaddingOut. 

The change is to not return the centering paddingTop, and paddingBottom in case it is multiline.

### Test plan

Screen Shot 2021-06-30 at 23 10 26

### DIFF
--- a/src/components/__tests__/TextInput.test.js
+++ b/src/components/__tests__/TextInput.test.js
@@ -85,3 +85,17 @@ it('correctly applies textAlign center', () => {
 
   expect(toJSON()).toMatchSnapshot();
 });
+
+it('correctly applies heigh to mutline Outline TextInput', () => {
+  const { toJSON } = render(
+    <TextInput
+      label="Outline Input"
+      placeholder="Type Something"
+      value={'Some test value'}
+      style={{ height: 100 }}
+      multiline
+    />
+  );
+
+  expect(toJSON()).toMatchSnapshot();
+});
--- a/src/components/__tests__/TextInput.test.js
+++ b/src/components/__tests__/TextInput.test.js
@@ -86,7 +86,7 @@ it('correctly applies textAlign center', () => {
   expect(toJSON()).toMatchSnapshot();
 });
 
-it('correctly applies heigh to mutline Outline TextInput', () => {
+it('correctly applies height to multiline Outline TextInput', () => {
   const { toJSON } = render(
     <TextInput
       mode="outlined"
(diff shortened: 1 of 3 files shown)


brunohkbx:
Reviewing this PR I noticed that currently, both flat and outline **multiline** TextInput differ from each other. 

Flat | Outlined
--- | ---
flat | outlined

Per MD specs and in this [example](https://material-components.github.io/material-components-web-catalog/#/component/text-field) it shouldn't be centered.

Screenshot from 2021-06-30 11-54-46

Fixed(iOS) | Fixed(Android)
--- | ---
ios | android

I can confirm this PR fixes #2730 as well. Applying padding makes it scrollable.

brunohkbx:
@Kamalnrf can you update the PR description? It should be an outlined textinput rather than a flat one.

src/components/__tests__/TextInput.test.js
```diff
@@ -85,3 +85,17 @@ it('correctly applies textAlign center', () => {
 
   expect(toJSON()).toMatchSnapshot();
 });
+
+it('correctly applies heigh to mutline Outline TextInput', () => {
+  const { toJSON } = render(
+    <TextInput
```
brunohkbx, line 91: missing `mode="outlined"` here

Kamalnrf: my bad, updating both.

src/components/__tests__/TextInput.test.js
```diff
@@ -85,3 +85,18 @@ it('correctly applies textAlign center', () => {
 
   expect(toJSON()).toMatchSnapshot();
 });
+
+it('correctly applies heigh to mutline Outline TextInput', () => {
```
brunohkbx, line 89:
```suggestion
it('correctly applies height to multiline Outline TextInput', () => {
```

brunohkbx: @Kamalnrf just a typo here

Kamalnrf: Done!

brunohkbx: You need to update the snapshot as well

Kamalnrf: Just pushed that change as well. TIL test names are also added to the snapshot

brunohkbx (approved):

PR #2827, opened 2021-07-20
Feat: allows adding width to outline text input optionally

IlMatveev:
### Summary

Allows adding width to outline text input optionally.


brunohkbx:
Hi @IlMatveev . Can you update your PR description adding a screenshot with an example of how it looks?

PR #2830, opened 2021-07-24
fix: add missing padding to CardTitle

brunohkbx:
### Summary
I just realized that in the Card.Title paddingRight was missing. So the title was a little bit off when centered.

### Test plan

Before | After
--- | ---
before | after

### DIFF
--- a/src/components/Card/CardTitle.tsx
+++ b/src/components/Card/CardTitle.tsx
@@ -176,11 +176,13 @@ const styles = StyleSheet.create({
 
   title: {
     minHeight: 30,
+    paddingRight: 16,
   },
 
   subtitle: {
     minHeight: 20,
     marginVertical: 0,
+    paddingRight: 16,
   },
 });
 


satya164 (approved):

src/components/Card/CardTitle.tsx
```diff
@@ -158,7 +158,7 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     alignItems: 'center',
     justifyContent: 'space-between',
-    paddingLeft: 16,
+    paddingHorizontal: 16,
```
lukewalczak, line 161: Let's keep the `paddingLeft: 16` as it was and please add `paddingRight:16` to the `title` style, to avoid shifting other components.


padding horizontal | padding right
--- | ---
Zrzut ekranu 2021-07-29 o 17 40 42 | Zrzut ekranu 2021-07-29 o 17 40 59

brunohkbx: Good catch! Thanks @lukewalczak, I totally missed that.

PR #2845, opened 2021-08-10
fix: missing Menu.Item theme prop

### LINKED ISSUE #2844: Regression in Menu.Item -`theme` prop disappeared, cannot change color
### Current behaviour

The color of the icon for `Menu.Item` cannot be changed.

This is a regression introduced by https://github.com/callstack/react-native-paper/pull/2589

In the previous version it was possible to supply a `theme` prop to `Menu.Item` and specify a different text color. This color would apply to the icon.

### Expected behaviour

Should allow overriding the color for the icon.

### Code sample

```tsx
<Menu.Item theme={{ colors: { text: AppColors.menuText } }}
```

The `theme` prop doesn't exist

The default color is white for our app, and the menus are not rendering. I could use `titleStyle` but the icon color still remains white.

Screenshot 2021-08-10 at 15 20 41

### Your Environment

andreialecu:
Fixes #2844

The changes closely follow Card: https://github.com/callstack/react-native-paper/blob/main/src/components/Card/Card.tsx

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -20,8 +20,7 @@ import { withTheme } from '../../core/theming';
 import type { $Omit } from '../../types';
 import Portal from '../Portal/Portal';
 import Surface from '../Surface';
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-import MenuItem, { MenuItem as _MenuItem } from './MenuItem';
+import MenuItem from './MenuItem';
 import { APPROX_STATUSBAR_HEIGHT } from '../../constants';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


andreialecu:
I actually wanted to add tests for this but it seems that `Menu` tests are broken altogether.

Take a look at the snapshots in: 

https://github.com/callstack/react-native-paper/blob/main/src/components/__tests__/__snapshots__/Menu.test.js.snap

They are all pretty much identical and none of them contain any menu items, even though `visible` is set to true for some menus.

Seems like a maintainer should take a look at them.

andreialecu:
Before: 

image

After:

Screenshot 2021-08-10 at 16 02 14

brunohkbx (approved):

PR #2870, opened 2021-08-23
header goback button fix

digmel:
### Summary

There was old property named `previous`, for `AppBar` header goBack button which is deprecated. 

### Test plan
Localhost, chrome


lukewalczak:
Closing in favour of: https://github.com/callstack/react-native-paper/pull/2898

PR #2883, opened 2021-09-01
Add custom close icon for the Chip component

essana3:
### Summary
The close button icon in the Chip is currently hard-coded to the MaterialCommunityIcon "close-circle" which is might be undesirable to users who prefer to use their own icon sets.
The PR below adds a **closeIcon** prop which when specified replaces the default "close-circle" icon.

### DIFF
--- a/example/src/Examples/ChipExample.tsx
+++ b/example/src/Examples/ChipExample.tsx
@@ -25,6 +25,15 @@ const ChipExample = () => {
               Close button
             </Chip>
             <Chip
+              onPress={() => {}}
+              onClose={() => {}}
+              closeIcon="arrow-down"
+              style={styles.chip}
+              closeIconAccessibilityLabel="Custom Close icon accessibility label"
+            >
+              Custom Close button
+            </Chip>
+            <Chip
               icon="heart"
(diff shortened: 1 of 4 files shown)


brunohkbx (approved):

example/src/Examples/ChipExample.tsx
```diff
@@ -24,6 +24,15 @@ const ChipExample = () => {
...
               Close button
             </Chip>
+            <Chip
+              onPress={() => {}}
+              onClose={() => {}}
+              closeIcon="arrow-down"
+              style={styles.chip}
+              closeIconAccessibilityLabel="Custom Close icon accessibility label"
+            >
+              Custom Close button
```
lukewalczak, line 34: 1. Please move your example under `Custom chip` section.

2. Please change the label to: `With custom close icon`

lukewalczak (changes requested):
Hey @essana3, thanks for your work. I've left two nits, please look at them.

example/src/Examples/ChipExample.tsx
essana3: Done!

PR #2884, opened 2021-09-02
Update DataTableCell.tsx

asherkuu:


brunohkbx:
@asherkuu looks like this was already updated in the `main` branch
https://github.com/callstack/react-native-paper/blob/main/src/components/DataTable/DataTableCell.tsx#L39

Issue #1677, opened 2020-02-16
Accesibility issues with Modal dialogs, List.Item and more


tasn:
## Environment

Expo 36
react-native-paper@3.6.0

## Description

Slightly related to #1676. I've gotten an accessibility report on my app and when trying to fix the issues I found out there are many accessibility issues with react-native-paper.

1. Cards don't let you choose item separately (e.g. buttons) using a screen reader. Setting "accessibly={false}" fixes it. So I guess Card has accessible set to true for some reason.
2. Same for List.Item (while here it makes sense having it, it prevents things like `right={<Switch ...}` from working.
3. Missing values for `accessibilityRole`. Specifically menu and checkbox (see RN accessibility notes). Why aren't thees just pass-through?
4. Same with `accessibilityStates` that seems to be unsupported.
5. Button on the other hand, should have `accessible` set because otherwise you can click on the button and the content separately.
6. Modals (and other popups such as menu) don't block screen reading of the background which means screen reading can read information behind the popups. This requires using something like https://facebook.github.io/react-native/docs/accessibility#accessibilityelementshidden-ios

In general, try using react-native-paper with a screen reader, and you'll spot these issues and more.

 ## Reproducible Demo

For the Card:
```
        <Card>
          <Card.Title title={props.title} />
          <Card.Content>
            {content}
          </Card.Content>
          <Card.Actions style={{ justifyContent: 'flex-end' }}>
            <Button disabled={loading} onPress={() => true}>{labelOk}</Button>
          </Card.Actions>
        </Card>
```

tasn:
Anyone? :|

hapablap21:
Last week I did a bunch of package updates, and this is all behaving MUCH better now.  I can't claim everything is better, but the modals are functional now.  

I'm not sure which upgrade did it, but I'm up to RN 0.62.2 and RN-paper 3.8.0.

tasn:
I haven't updated RN yet (was encountering issues), though even with RN-paper 3.9.0 the issues are still there. I don't think it's related to RN though.

PR #2893, opened 2021-09-09
fix: enable hiding other content when portal is used

haakandev:
### Summary

I've attempted to fix an issue with Portals/Modals where content under the modal is still discoverable by screen readers.
With this PR, we can pass `isFocused` to a Portal when the content in it should hide other content from screen readers.
This issue has been mentioned before for example point 6 in #1677
Happy to adjust if you think there's a better approach.

### Test plan

1. Add a modal inside a Portal from a screen with some content
2. Turn on a screen reader, eg VoiceOver on iOS and swipe through the content of the screen

### DIFF
--- a/src/components/Portal/Portal.tsx
+++ b/src/components/Portal/Portal.tsx
@@ -16,6 +16,11 @@ type Props = {
    * @optional
    */
   theme: ReactNativePaper.Theme;
+  /**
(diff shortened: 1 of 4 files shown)


lukewalczak:
I'm going to:
* check the a11y issue mentioned in the description
* test changes
* confirm that everything works as expected
* review changes and resolve conflicts

PR #2894, opened 2021-09-09
feat: Styling of labels in FAB.Group

haakandev:
### Summary

This PR adds properties for the FAB.Group actions to allow for custom styling of the FAB.Group labels.

### Test plan

Provide `labelTextColor` and `labelStyle` properties in the FAB.Group's `actions` array.

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -23,7 +23,9 @@ type Props = {
    * - `label`: optional label text
    * - `accessibilityLabel`: accessibility label for the action, uses label by default if specified
    * - `color`: custom icon color of the action item
+   * - `labelTextColor`: custom label text color of the action item
    * - `style`: pass additional styles for the fab item, for example, `backgroundColor`
+   * - `labelStyle`: pass additional styles for the fab item label, for example, `backgroundColor`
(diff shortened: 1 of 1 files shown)


andreialecu:
I recently ended up using patch-package to add this and I was just about to open pretty much the exact same PR.

Please consider merging this.

hefberk:
Hi,
Thank you for this patch, but I can't make it work.
I replace react-native-paper version in package.json like this :
"react-native-paper": "https://github.com/haakandev/react-native-paper.git#feat/fab-group-styling"
In my node_modules I can see the patch.
But when i do 
<FAB.Group labelTextColor={'#FF0000'}/>
Text is always black.
Don't work with labelStyle too.
I don't understand, your code seems ok, does it work for you ?
It would be great to implement selectable=false on label to have good cursor on web.

hefberk:
property fontSize doesn't seems working

haakandev:
> property fontSize doesn't seems working

Hi @hefberk - the label component as is just uses a standard react-native-paper `Text` element. The `labelStyle` is actually for the wrapping `View` component. If desired, the change could be modified to take a `labelTextStyle` prop instead of `labelTextColor` to allow further customising of the label text.

lukewalczak (approved):
@haakandev thanks!

PR #2898, opened 2021-09-10
Docs: React Navigation Integration - update previous prop to back for 6.x

CodingItWrong:
### Summary

The docs page [Integrate AppBar with react-navigation](https://callstack.github.io/react-native-paper/integrate-app-bar-with-react-navigation.html#adding-appbar) describes using the `previous` prop to detect if a back button should be shown. This is correct for React Navigation 5.x as described here: <https://reactnavigation.org/docs/5.x/stack-navigator/#header>

But in React Navigation 6.x, there is no longer a `previous` prop; instead, there is a `back` prop, as described here: <https://reactnavigation.org/docs/stack-navigator/#header>

As a result, when using React Navigation 6.x with the current instructions, the back button never appears. If you switch to using the `back` prop, the back button starts working again. Repro: <https://github.com/CodingItWrong/rn-paper-navigation>

This PR currently updates the instruction page to be correct for React Navigation 6.x

Additional changes I would be happy to make if you'd like:

- I could specify that these instructions are for React Navigation 6.x
- I could add a note that `previous` should be used instead for React Navigation 5.x (and I could investigate if it would work in 4.x and below as well, and note that)

### Test plan

Can be manually tested in the repro repo: <https://github.com/CodingItWrong/rn-paper-navigation>

### DIFF
--- a/docs/pages/9.react-navigation.md
+++ b/docs/pages/9.react-navigation.md
@@ -176,7 +176,7 @@ Another interesting pattern that can be implemented with `react-native-paper` an
 - Pass `Appbar.Action` to the anchor prop
 - Add a state to control `Menu` visibility
 
-We also want the menu to appear only on `HomeScreen`, which means we will render it conditionally based on the `previous` prop.
+We also want the menu to appear only on `HomeScreen`, which means we will render it conditionally based on the `back` prop.
 
 ```js
 function CustomNavigationBar({ navigation, back }) {
--- a/docs/pages/9.react-navigation.md
+++ b/docs/pages/9.react-navigation.md
@@ -155,7 +155,7 @@ export default function App() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


lukewalczak:
Hello @CodingItWrong, thanks for pointing it out!

> Additional changes I would be happy to make if you'd like:
> 
> I could specify that these instructions are for React Navigation 6.x
> I could add a note that previous should be used instead for React Navigation 5.x (and I could investigate if it would work in 4.x and below as well, and note that)

I would be more than happy if you can adjust your suggestion since current change can be confusing for user with navigation 5.x

CodingItWrong:
@lukewalczak OK, I've updated it to say that `back` is in React Navigation 6.x and `previous` is in 5.x. 4.x does have a `previous` prop but the API to create a Stack is different, so it didn't seem like a good idea to go back to 4.x in the docs.

lukewalczak (approved):
Thanks @CodingItWrong!

PR #2938, opened 2021-10-18
Title button with Linear Gradient

iqrakhaliqch:
It supports Linear Gradient in the background of **Title** round button.

The existing component does not allow the user to add Linear Gradient to the Title expandable button. We can use component with or without Linear Gradient in the background of the button with border radius.

In the **List.Accordion** the following options will change the appearance of the **Title** button:
1- titleGradientComponent --> It will allow the user to use gradient component, default is false.
2- gradientColorPrimary --> It will take 2(Two) colors array, default colors will be white and gray.
3- gradientStyle --> It will apply styling to the Title button.

button change


lukewalczak:
Hey @iqraKhaliq, thanks for your work and effort. Unfortunately, looks like MD color system doesn't support gradients, that's why we are not going to merge your PR into the library.

PR #2951, opened 2021-10-26
chore: upgrade expo to 44

jaworek:
RN Paper uses quite old version of `expo` in the example app. This PR updates it to the latest version.

### Summary

This is a maintenance update to keep used packages up to date.

### Test plan

Pass all pipelines. Check if all components display correctly.


example/package.json
```diff
@@ -13,27 +13,27 @@
...
-    "react": "16.13.1",
-    "react-dom": "16.13.1",
-    "react-native": "https://github.com/expo/react-native/archive/sdk-40.0.0.tar.gz",
-    "react-native-gesture-handler": "~1.8.0",
-    "react-native-reanimated": "~1.13.0",
-    "react-native-safe-area-context": "3.1.9",
-    "react-native-screens": "~2.15.0",
+    "react": "17.0.1",
+    "react-dom": "17.0.1",
+    "react-native": "0.64.2",
```
lukewalczak, line 30: please update the react-native version also in root package.json

example/metro.config.js
```diff
@@ -1,7 +1,7 @@
 // /* eslint-disable import/no-commonjs */
 
 const path = require('path');
-const blacklist = require('metro-config/src/defaults/blacklist');
+const blacklist = require('metro-config/src/defaults/exclusionList');
```
jaworek, line 4: Required due to rename:
https://forums.expo.dev/t/cannot-find-module-metro-config-src-defaults-blacklist/53819/2

src/components/BottomNavigation.tsx
```diff
@@ -9,7 +9,6 @@ import {
   Platform,
   ViewStyle,
 } from 'react-native';
-import { getBottomSpace } from 'react-native-iphone-x-helper';
```
jaworek, line 12: This package seems to not work properly with newer RN versions.
To fix this issue, I propose to make the consumer of this component handle this.
Example application uses `react-native-safe-area-context` to pass [inset](https://github.com/callstack/react-native-paper/pull/2951/files#diff-538c2624d10d36acafd30efc9c71aec4748e7ef3b634804e5375d3c1db791975R36) when required.

Unfortunately, this will be a breaking change.

lukewalczak: Hey @jaworek, I think to avoid breaking changes, we can write that helper by ourselves. Basically, we need to collect the list of ios devices with notch and compare them with model and brand from `Device` sdk available on expo.

jaworek: It turns out that the problem with this library was due to change of Jest preset in RN 0.64.

Now it is required to set `transformIgnorePatterns ` to be able to correctly parse ES6 modules in Jest:
https://jestjs.io/docs/tutorial-react-native#transformignorepatterns-customization

I still think that removing use of `react-native-iphone-x-helper` from RN Paper is desired, but it should only be done as a breaking change in the next major version. Instead users consuming components that require setting safe areas, should handle it themselves as it gives more flexibility for custom solutions.

lukewalczak (approved):
@jaworek thanks for an update! please resolve the conflicts and we'll be ready to go 🎉

Issue #1670, opened 2020-02-14
Typescript issue in Banner component when trying to customize an action button


gevorgsahakyann:
## Environment
"react-native": "0.61.5",
"react-native-paper": "^3.6.0",
"typescript": "^3.7.5"

## Description
I tried to customize the action button for the Banner component and got the typescript issue. Can't pass other properties except for the label and onPress.

<img width="497" alt="Screen Shot 2020-02-14 at 18 21 07" src="https://user-images.githubusercontent.com/45065751/74541310-d6f12400-4f5a-11ea-9d25-f150cd988ddd.png">

Please set the button's typescript type for banner actions.
<img width="624" alt="Screen Shot 2020-02-14 at 18 20 43" src="https://user-images.githubusercontent.com/45065751/74541323-db1d4180-4f5a-11ea-9392-f6b352b9e650.png">

kamui545:
I noticed the same issue, however, I was wondering: Would it make sense to have the same kind of API as we have for Card?

I think it would be more flexible. But maybe I am missing something / it's not really a good fit / we could just use the Card components...

```html
<Banner>
  <Banner.Content>
    There was a problem processing a transaction on your credit card.
  </Banner.Content>
  <Banner.Actions>
    <Button>Fix it</Button>
    <Button>Learn more</Button>
  </Banner.Actions>
</Banner>
```

reutg:
I am having the same issue.

mesamhaid:
Experiencing this exact same issue. Can we reopen this issue please?

mister-conor:
I think the proposal from @kamui545 would work very well here.

ericledonge:
I'd like to PR a simple solution that I've done in a project with a patch npm, but for that, I also need to modify the file lib/typescript/components/Banner.d.ts and I don't understand how to do it. I am new in open source and typescript worlds. Any help, please?

PR #2970, opened 2021-11-08
fix: typescript issue in banner when customizing buttons

ericledonge:
### Summary

This PR fixing this issue: https://github.com/callstack/react-native-paper/issues/1670

With this fix, I can customize button you can pass other props that button component takes, without having TS crying, as described in the doc: https://callstack.github.io/react-native-paper/banner.html#actions

Maybe, the way I've submitted the modification in lib/typescript/components/Banner.d.ts is not good. I just want to change the line 32. I am new to both open source and typescript worlds. So let me know how to submit that correctly.

### Test plan

In my IDE, everything is ok:

Capture d’écran, le 2021-11-07 à 21 30 07


src/components/Banner.tsx
```diff
@@ -33,10 +33,12 @@ type Props = $RemoveChildren<typeof Surface> & {
...
    * To customize button you can pass other props that button component takes.
    */
-  actions: Array<{
-    label: string;
-    onPress: () => void;
-  }>;
+  actions: Array<
+    {
+      label: string;
+      onPress: () => void;
```
Drakeoon, line 39:
```suggestion
```

I think this line could be removed, since `onPress()` is already covered by `Button` interface

Drakeoon (changes requested):
@ericledonge I think the way you also updated lib files here is fine. @lukewalczak any preferences about it?

I left a quick note, please apply it and I'll give a green light on this one. Imo a really good change 👍🏻

Drakeoon:
Oh, I actually thought we had the lib folder in our repo and it's just an update. I think you should also remove the lib changes, as we compile them for every release on our own @ericledonge

ericledonge:
@Drakeoon Oki, let me know if it's correct now.

Drakeoon (approved):
@lukewalczak looks good to me 👍🏻  Thanks @ericledonge

Issue #2972, opened 2021-11-08
Chip does not have a height


dardinier:
### Current behaviour

Chip does not have height.

![Screenshot_2021-11-08-16-19-30-706_host exp exponent](https://user-images.githubusercontent.com/24563898/140774220-4e7a07bc-2eaf-43d1-a5fc-2878c1092044.jpg)

### Expected behaviour

Chip should have a height and be visible.

![Screenshot_2021-11-08-16-19-50-355_host exp exponent](https://user-images.githubusercontent.com/24563898/140774251-99a7c7d0-27e5-406d-9b12-d49f856f1b2e.jpg)

### Code sample

https://snack.expo.io/o8csskeeVo

On mobile only. The web version shows the correct result.

### What have you tried

Having the `Chip` component inside a `ScrollView` seems to fix the bug.

Example : https://snack.expo.io/CXZOrbGgY

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android 10 (API level 29)
| react-native          | 0.63.4
| react-native-paper    | 4.10.0
| node                  | 14.11.0
| npm or yarn           | yarn 1.22.5
| expo sdk              | 43.0.0

dardinier:
I provided two samples, they are hosted under `snack.expo.dev` instead of `snack.expo.io` from now.

lukewalczak:
Similar to: https://github.com/callstack/react-native-paper/issues/2959. It's on my radar, will fix it soon.

Issue #2959, opened 2021-11-05
Chip styles


root-io:
### Current behaviour

The size of the 3 chips in my example are shrink, we can't see the text

### Expected behaviour

The size of the 3 chips in my example should display a full height

### Code sample

```javascript
import * as React from 'react';
import { View, StyleSheet } from 'react-native';
import { Chip, List } from 'react-native-paper';

const ChipExample = () => {
  return (
    <>
      <List.Section title="Flat chip">
        <View style={styles.row}>
          <Chip selected onPress={() => { }} style={styles.chip}>
            Simple 1
          </Chip>
          <Chip selected onPress={() => { }} style={styles.chip}>
            Simple 2
          </Chip>
          <Chip selected onPress={() => { }} style={styles.chip}>
            Simple 3
          </Chip>
        </View>
      </List.Section>
    </>
  );
};

const styles = StyleSheet.create({
  row: {
    flexDirection: 'row',
    flexWrap: 'wrap',
    paddingHorizontal: 12,
  },
  chip: {
    margin: 4,
  },
});

export default ChipExample;
```

### Screenshots (if applicable)

![screenshot](https://user-images.githubusercontent.com/6101308/140511138-2de3cafe-bf12-4e9a-aa36-33bb3bf0afb7.jpg)

### What have you tried

Broken since 4.10.0, was working as expected in 4.9.2

### Your Environment

| software              | version
| --------------------- | -------
| android        | 12
| react-native          | 0.66.2
| react-native-paper    | 4.10.0
| node                  | 16.13.0
| npm           | 8.1.0
| expo sdk              | no

lukewalczak:
Hello @root-io, please put your bug reproduction on [snack](snack.expo.io)

jvincent-dev:
> Hello @root-io, please put your bug reproduction on [snack](snack.expo.io)

I added his code to a snack but changed the color of the Chip to make it more visible [here](https://snack.expo.dev/hYObt69RY)

lukewalczak:
Thanks, I will be working on the fix in the upcoming days!

Issue #2804, opened 2021-07-06
Full width Chips are not focusable anywhere on the chip


dardinier:
### Current behaviour

I have a Chip displayed in full width :
```
<Chip onPress={...}>Hello</Chip>
```

Sample : https://snack.expo.io/qeXIhkV_-

![image](https://user-images.githubusercontent.com/24563898/124610530-e3885c00-de70-11eb-95c7-83fdc9907a1a.png)

The chip is only pressable on the text and gets the focused state only around the text :

![image](https://user-images.githubusercontent.com/24563898/124610660-fdc23a00-de70-11eb-9434-a7814ad6f356.png)

### Expected behaviour

It should work like in previous versions :

https://snack.expo.io/ne1ESaMId

Here, when I press anywhere on the chip, the chip is focused and all the chip gets the focused state :

Default :

![image](https://user-images.githubusercontent.com/24563898/124611021-58f42c80-de71-11eb-968a-da239248beb5.png)

Pressed :

![image](https://user-images.githubusercontent.com/24563898/124610953-44b02f80-de71-11eb-859b-5f4a0e0dae90.png)

### What have you tried

I tried to give a full width View as children with the text in it but it didn't work too.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android 10 (API level 29)
| react-native          | 0.63.4
| react-native-paper    | 4.7.2
| node                  | 14.11.0
| npm or yarn           | yarn 1.22.5

dardinier:
Still doesn't work in the latest version `4.9.2` of `react-native-paper` :

https://snack.expo.io/iEnHfXuTH

I don't use `react-native-vector-icons`
I don't use `npm`

PR #2976, opened 2021-11-09
fix: correct Chip touchable style

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/2972, https://github.com/callstack/react-native-paper/issues/2959

### Summary

Setting the `flexGrow` to `Touchable` within `Chip`

Android | ios | web
-- | -- | --
Zrzut ekranu 2021-11-9 o 19 24 33 | Zrzut ekranu 2021-11-9 o 19 31 52 | Zrzut ekranu 2021-11-9 o 19 31 58

at the same time it's not breaking the previous issue: https://github.com/callstack/react-native-paper/issues/2804

Android | ios | web
-- | -- | --
previous_bug | bug_prev_ios | bug_prev_web

### Test plan

Run the code below from one of the issues and **expect** Chips are displayed properly:

```javascript
import { Chip, List } from 'react-native-paper';

const ChipExample = () => {
  return (
    <>
      <List.Section title="Flat chip">
        <View style={styles.row}>
          <Chip selected onPress={() => { }} style={styles.chip}>
            Simple 1
          </Chip>
          <Chip selected onPress={() => { }} style={styles.chip}>
            Simple 2
          </Chip>
          <Chip selected onPress={() => { }} style={styles.chip}>
            Simple 3
          </Chip>
        </View>
      </List.Section>
    </>
  );
};

  row: {
    flexDirection: 'row',
    flexWrap: 'wrap',
    paddingHorizontal: 12,
  },
  chip: {
    margin: 4,
  },
});

export default ChipExample;
```

### DIFF
--- a/example/src/Examples/ChipExample.tsx
+++ b/example/src/Examples/ChipExample.tsx
@@ -216,6 +216,9 @@ const ChipExample = () => {
               With custom close icon
             </Chip>
           </View>
+          <Chip mode="outlined" onPress={() => {}} style={styles.fullWidthChip}>
+            Full width chip
+          </Chip>
         </List.Section>
       </ScreenWrapper>
       <Snackbar
@@ -252,6 +255,10 @@ const styles = StyleSheet.create({
   },
   bigTextStyle: {
     flex: -1,
+  },
+  fullWidthChip: {
+    marginVertical: 4,
+    marginHorizontal: 12,
   },
 });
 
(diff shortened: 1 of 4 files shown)


brunohkbx:
Thanks @lukewalczak. I didn't realize that using `flex: 1` in the touchable would cause problems.
WDYT about adding a test to track this regression described in #2972?

lukewalczak:
@brunohkbx sure, good point! Will add it!

lukewalczak:
After the research I guess we are not able to test the regression without e2e tests. Snapshot indicates only change from `flex` to `flexGrow`. Also the component is always available within the tree, so firing event on it will always pass as well.

brunohkbx (approved):

Issue #2799, opened 2021-07-03
[iOS Only] mutline Textinput label is cover by text


KrifaYounes:
<img width="438" alt="Capture d’écran 2021-07-03 à 17 05 33" src="https://user-images.githubusercontent.com/3875788/124358503-e1c85980-dc20-11eb-8c7d-1566d3bb8e68.png">

here's a snack to reproduce the issue
https://snack.expo.io/@aknad/textinput

Only on iOS
It's ok on Android

latest version react-native-paper and react-native

nicholascm:
Hey @KrifaYounes - I was trying to reproduce this and I noticed that this only happens when there is a fixed height of the input (as you did in the snack). Is that a valid material design scenario? If you just swap `minHeight` for `height` you can start the TextInput at the height you had specified, 100. 

In the [examples I see here ](https://material.io/archive/guidelines/components/text-fields.html#text-fields-field-types), multiline inputs are growing with the length of the content.

KrifaYounes:
> Hey @KrifaYounes - I was trying to reproduce this and I noticed that this only happens when there is a fixed height of the input (as you did in the snack). Is that a valid material design scenario?
> 
> If you just swap `minHeight` for `height` you can start the TextInput at the height you had specified, 100, and then as I see it, the input grows with the content properly without covering the label on iOS.
> 
> In the [examples I see here ](https://material.io/archive/guidelines/components/text-fields.html#text-fields-field-types), multiline inputs are growing with the length of the content.

Hi @nicholascm ,
In android it's working fine and the textInput not cover the inputLabel.
It's not possible to reproduce the same result in iOS.
Thanks

nicholascm:
> I have a form with multiple fields and I don't want to increase height dynamically (minHeight), it's a bad idea if user input text is long for a mobile device.

Got it - and to be clear I'm not a maintainer or anything - just was trying to help with an open issue bug fix and wanted to be sure it was intended to be used that way before trying to fix it...I did see the exact behavior you mentioned on my iOS device, so it is definitely able to be reproduced.

lukewalczak:
Hi, I've created a PR fixing it. You can try to install `react-native-paper` from PR's branch to check whether that fixes your issuse.

KrifaYounes:
> Hi, I've created a [PR](https://github.com/callstack/react-native-paper/pull/2979) fixing it. You can try to install `react-native-paper` from PR's branch to check whether that fixes your issuse.

it's working perfectly
thanks

PR #2979, opened 2021-11-12
fix: add patching container for multiline input on ios

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/2799

### Summary

PR adds the absolutely positioned container which is kind of the patch for a TextInput to avoid covering input label with its value.

ios_multiline

### Test plan

1. Use `TextInput` with prop `multiline`
2. Set `height` in style
3. Type some text within input, which contains more space that set height
4. Expect: input's value should not cover the label


lukaszKowal-datasite:
Hi @lukewalczak 
I think setting https://github.com/callstack/react-native-paper/pull/2979/files#diff-e4a748a69d38ea7aa878ad43d572a9e9b451e1cf72a5635c7e71112437e1a195R344
will manipulate theme colours?

I am having following result on iOS multiline input:
Simulator Screen Shot - iPhone 12 - 2021-12-15 at 16 18 59

Obviously the grey bar should not be there.

lukewalczak:
@lukaszKowal-datasite thanks for feedback. Could you please send simple snack, to make it easier for me to debug it?

Svarto:
@lukaszKowal-datasite @lukewalczak same issue for me, when there is no label on a TextInput then there is a grey box covering the actual text.

Issue #2366, opened 2020-11-19
react-native-web createMaterialBottomTabNavigator delivers clicks to background tabs


paour:
### Current behaviour

- I have a material bottom tab navigator where each tab contains a stack of screens
- on mobile it works fine
- if I use a non-Paper bottom tab `createBottomTabNavigator`, it works fine on the
- with a Paper bottom tab, on react-native-web, once they have been rendered once, background tabs get clicks

### Expected behaviour

- I expect only the frontmost tab to get clicks

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | web
| react-native          | 0.62.2
| react-native-paper    | 4.4.0 (which contains a recent fix for getting tabs to show up the first time the user clicks to them)
| node                  | 10.17.0
| yarn           | 1.22.10
| react-native-vector-icons | 6.6.0
| expo sdk              | no Expo

jasperro:
Seems I overlooked this in my patch. I have no idea what is causing this to happen, maybe something with the z-index?

paour:
If it was the z-index, wouldn't the elements be visible? In my case, the screen looks fine, but clicks are intercepted by (it seems) the last tab (by tab index) that has already been displayed, even if it's not currently the selected tab.

jasperro:
It seems that in switching a tab, the code only changes the opacity of the current screen to 0 and the active one 1. This means everything is still being rendered, just totally transparently. Replacing this with display: none and display:flex removes the issue, I will see if I can make a pull request.

paour:
Awesome, thanks for jumping on this!

jasperro:
You can now test my fix by cloning my fix branch `git clone https://github.com/jasperro/react-native-paper/ -b fix-tab-clickthrough` And running the instructions in https://github.com/callstack/react-native-paper/pull/2370

paour:
I was able to verify that this fixes the issue I reported (I used `yarn link` rather than a simple `cp` but thanks for the detailed instructions).

I hope this can be merged before the next release.

Trancever:
@paour @jasperro Could you provide a repo/snack that reproduces the issue? Does it happen in every browser for you? Or only some specific ones?

Besides changing the opacity, we also set `pointerEvents` to 'auto' or 'none' based on the currently selected tab. Theoretically, this should prevent not selected tabs against capturing clicks.

Trancever:
[...]

PR #2548, opened 2021-02-11
fix: BottomNavigation clickthrough in tabs on web

jasperro:
Fixes https://github.com/callstack/react-native-paper/issues/2366

### Summary

Fixes #2366 by setting `display: none` on the tab when it's not focused. Continuation of broken pull request: https://github.com/callstack/react-native-paper/pull/2370. The commit just uses the already-existing "focused" attribute that the rest of the code uses. I tested it in my own application and the branch works (with `yarn link`)

### Test plan

Clone the branch, install dependencies with `yarn bootstrap` and `yarn link`. Test expo snack online with instructions in the app and see the issue. Download expo snack to a folder (https://snack.expo.io/@jasperro/react-native-paper-tab-clickthrough-test) and run `yarn` and `yarn link react-native-paper`, and it should be fixed.

PR #2985, opened 2021-11-25
fix: correct clickthrough on web tabs within BottomNavigation

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/2366

The PR bases on the approach started within https://github.com/callstack/react-native-paper/pull/2548, however fixes the problem with animation during the transtions. 

Test steps are the same as in the linked PR above. 

### Summary

Kapture 2021-11-24 at 22 48 48

### Test plan

Clone the branch, install dependencies with `yarn bootstrap` and `yarn link`. Test expo snack online with instructions in the app and see the issue. Download expo snack to a folder (https://snack.expo.io/@jasperro/react-native-paper-tab-clickthrough-test) and run `yarn` and `yarn link react-native-paper`, and it should be fixed.


src/babel/__fixtures__/rewrite-imports/output.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable prettier/prettier */
 import { Text } from 'react-native';
 import PaperProvider from "react-native-paper/lib/module/core/Provider";
-import BottomNavigation from "react-native-paper/lib/module/components/BottomNavigation";
+import BottomNavigation from "react-native-paper/lib/module/components/BottomNavigation/BottomNavigation";
```
p-syche, line 4: Why not rename the new file to `index.tsx` so you would avoid this change?

p-syche (approved):
LGTM 😄 

I left a question about naming convention - why not use `index.tsx`? - but that's no reason to hold off on approving this.

src/babel/__fixtures__/rewrite-imports/output.js
lukewalczak: Hey @p-syche, sorry for the late response. Basically, I keep the convention where each file has the component name and `index.ts` is used when we are assigning sub components.

PR #2991, opened 2021-11-28
feat: The `label` prop for `TextInput` can now take a component (`ReactElement`) as well as `string`

smbadiwe:
### Summary

The PR provides a way to solve the problem being discussed at https://github.com/callstack/react-native-paper/discussions/2591. Currently there's no way to customize the label on `TextInput`. A common use-case is adding a red asterisk to indicate that a field is required. This PR provides for the `label` field to also accept a component, thereby solving the problem.

### Test Plan

Here's an example of how a custom component can be passed in:
```js
<TextInput
      label={<Text style={{ color: 'red'}}>Must Fill</Text>}
      placeholder='You must fill out this field'
    />
```

Of course the old usage still works, so this is not a breaking change.
```js
<TextInput
      label='May Fill'
      placeholder='You may not fill out this field'
    />
```


p-syche (commented):
hi @smbadiwe , thank you very much for this change. It's great you were able to introduce a new functionality without breaking the old.

Your code is good, however I would love to see some PR improvements before approving. Namely I would love to see the new feature added to the `TextInput` examples and I would appreciate a test checking the function you added for checking label change.

src/components/TextInput/TextInput.tsx
```diff
@@ -477,4 +476,96 @@ class TextInput extends React.Component<TextInputProps, State> {
   }
 }
 
+export function areLabelsEqual(
```
lukewalczak, line 479: @smbadiwe please extract this function into the file `helpers` located next to the `TextInput`

p-syche (approved):
You wrote very extensive tests 💪  great job!

LGTM

MatthiasD0:
This change is causing `Converting circular structure to JSON` error here.
Before the "feature" being added here was already working on our side.
You can not run JSON.stringify on the ReactComponent seen in the screenshot. This is causing the issue here.

Bildschirmfoto 2022-02-04 um 15 24 33

lukewalczak:
I think the issue can be fixed on the `4.12.0-alpha.0`

mfds:
Quick question: was this feature tested with the `outlined` mode? Doesn't look like there's an example on the `example` folder and I'm getting weird "glitches" when I try:

|Label before animation|Label after animation|
|-|-|
|image|image|

If you zoom the second picture, you can see the `Label` label repeated twice. The second one is squished :/

PR #3004, opened 2021-12-11
feat(a11y): button: add accessibility hint prop

arnaudbzn:
### Summary

Add `accessibilityHint` prop to `Button` component.

### Test plan

- set `accessibilityHint` prop
- Add missing `accessibilityLabel` prop test


lukewalczak (approved):
Thanks for your contribution @arnaudbzn! 🎉

Issue #2984, opened 2021-11-20
EventEmitter.removeListener('change', ...): Method has been deprecated. Warning


pushpender-singh-ap:
### Warning
EventEmitter.removeListener('change', ...): Method has been deprecated. Please instead use `remove()` on the subscription returned by `EventEmitter.addListener`

### Code sample
import { Menu } from 'react-native-paper';
        <Menu
                            visible={menu}
                            onDismiss={() => setmenu(false)}
                            anchor={<Ionicons
                                onPress={() => setmenu(true)}
                                name={"settings-sharp"}
                                size={24}
                                color={"#222"}
                            />}>
                            <Menu.Item onPress={() => { }} title="Download" />
                            <Menu.Item onPress={() => { }} title="Report" />
                            <Menu.Item onPress={() => { }} title="Cancel" />
                        </Menu>

### Screenshots (if applicable)
![image](https://user-images.githubusercontent.com/73298854/142730101-712c30f4-5652-49d9-a432-5ee50d0d756c.png)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        |  android
| react-native          |  0.66.2
| react-native-paper    |   ^4.10.1
| node                  |     v14.18.1
| npm or yarn           |    npm
| expo sdk              |   no

GITHUB REPO:- https://github.com/pushpender-singh-ap/sampleRepo

lukewalczak:
Hey @pushpender-singh-ap, what version of `@types/react-native` are you using?

pushpender-singh-ap:
> Hey @pushpender-singh-ap, what version of `@types/react-native` are you using?
i am not using this dep

lukewalczak:
Are you able to create simple repo reproducing your issue?

pushpender-singh-ap:
> Are you able to create simple repo reproducing your issue?

yes, sure I will make and inform you later.

pushpender-singh-ap:
@lukewalczak sample repo clone and press Paper Button and press hardware back you automatically get the warning.

Repo Link:- https://github.com/pushpender-singh-ap/sampleRepo

lukewalczak:
Hey @pushpender-singh-ap, I quickly looked at your example and ran it, however I was not able to observe the warning. Since you are facing it only on Android platform, I assume something can be wrong with your environment/setup, especially that [PR](https://github.com/callstack/react-native-paper/pull/2918) fixing that behavior was merged couple weeks ago.

[...]

PR #3007, opened 2021-12-12
EventEmitter.removeListener('change', ...): Method has been deprecate…

pushpender-singh-ap:
…d Issue Fixed

https://github.com/callstack/react-native-paper/issues/2984

### Summary

if (this.backHandlerSubscription?.remove) {
      this.backHandlerSubscription.remove();
    } else {
      BackHandler.removeEventListener('hardwareBackPress', this.handleDismiss);    <----   this event trigger when the user does not click on the menu option and the user press Hardware Back Button so no need to do this because BackHandler Listener is not subscribed.
    }

    if (this.dimensionsSubscription?.remove) {
      this.dimensionsSubscription.remove();
    } else {
      Dimensions.removeEventListener('change', this.handleDismiss);     <----   this event trigger when the user does not click on the menu option and the user press Hardware Back Button so no need to do this because BackHandler Listener is not subscribed.
    }

If you wanna see whole process recording click below links:
https://user-images.githubusercontent.com/73298854/145720046-2e3628d5-d94c-409d-9785-287d8432cd87.mp4

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -258,13 +258,13 @@ class Menu extends React.Component<Props, State> {
     if (this.backHandlerSubscription?.remove) {
       this.backHandlerSubscription.remove();
     } else {
-      BackHandler.removeEventListener('hardwareBackPress', this.handleDismiss);
+      this.handleDismiss;
     }
 
     if (this.dimensionsSubscription?.remove) {
       this.dimensionsSubscription.remove();
     } else {
-      Dimensions.removeEventListener('change', this.handleDismiss);
+      this.handleDismiss;
     }
 
     this.isBrowser() &&


src/components/Menu/Menu.tsx
```diff
@@ -258,13 +258,13 @@ class Menu extends React.Component<Props, State> {
     if (this.backHandlerSubscription?.remove) {
       this.backHandlerSubscription.remove();
     } else {
-      BackHandler.removeEventListener('hardwareBackPress', this.handleDismiss);
+      this.handleDismiss;
```
lukewalczak, line 261: We cannot handle that problem in that way since we are losing the backward compatibility.

I can suggest to rather revert your changes and create an `if statement` within `componentWillUnmount` and check whether `backHandlerSubscription` and `dimensionsSubscription` are defined:

```
 componentWillUnmount() {
    if (this.backHandlerSubscription && this.dimensionsSubscription) {
      this.removeListeners();
    }
  }
```

lukewalczak (changes requested):
We cannot handle that problem in that way since we are losing the backward compatibility.

I can suggest to rather create an `if statement` within `componentWillUnmount` and check whether `backHandlerSubscription` and `dimensionsSubscription` are defined:

```
 componentWillUnmount() {
    if (this.backHandlerSubscription && this.dimensionsSubscription) {
      this.removeListeners();
    }
  }
```

pushpender-singh-ap:
@lukewalczak but why do we need this even we are not subscribing backhandler, but user clicks on the menu option private attachListeners trigger, and BackHandler Listeners are started Listening so in case we need to unsubscribe the Listeners.

If you see the video of the process when coming on the menu page and without touching the menu icon click on Hardware back to in case we don't need to unsubscribe the Backhandler (Because Backhandler Listener is not listening).

lukewalczak:
We are subscribing into the `BackHandler` and `Dimensions` within `attachListeners` when opening the `Menu` and accordingly we have to remove event listener when closing it e.g. by pressing android hardware back button.

In your case presented on the video, in fact, we are not subscribing to any of them, because `Menu` was not opened, that's why I'm suggesting to add the _if statement_ in `componentWillUnmount`.

Please consider the case when you've opened the `Menu` component, so you are subscribed to both listeners. When closing the component to remove listeners we need `BackHandler.removeEventListener(...)` and the same for `Dimensions` for the RN < 0.64.

pushpender-singh-ap:
ok, wait for this I will check in RN 0.63 version.

pushpender-singh-ap:
I make new repo with RN 0.63 version and install the same dep and change the same it's working fine.

we are not subscribing to any of them, because Menu was not opened and we are trying to remove listeners even we are not subscribed

For more information see this below video:

https://user-images.githubusercontent.com/73298854/145801430-e3c61522-f5cf-4132-b28c-2d4d0c05ba91.mp4

lukewalczak:
After deeper investigation my proposed solution cannot work on the previous RN, since both subscriptions will be undefined. It looks like I need to dive into listeners and make some refactor in all places where we are adding and removing them and create a separate PR for it.

lukewalczak:
@pushpender-singh-ap I will close it, because it's not the correct approach, however it solves the issue for you – I can suggest you to use patch-package library or base on fork, to keep that solution for you.
In the meantime I will work on more generic fix for all listeners and will let you know once it's ready.

Thanks for your effort and putting more light on that issue.

pushpender-singh-ap:
check one time this commit is also working on  0.64+ & below 0.64 I tested on 0.63 & as well as 0.66.

Maybe this help you if you think that way it is better than the previous one inform me I will again generate PR.

https://github.com/pushpender-singh-ap/react-native-paper/commit/943d743272451840d25506372678c817f5e2ec81#diff-79e0d805785bbe8a0f585efebc274478388e7ce83323d87a7492930efd924de3

pushpender-singh-ap:
@lukewalczak  check one time this commit is also working on  0.64+ & below 0.64 I tested on 0.63 & as well as 0.66.

Maybe this help you if you think that way it is better than the previous one inform me I will again generate PR.

https://github.com/pushpender-singh-ap/react-native-paper/commit/943d743272451840d25506372678c817f5e2ec81#diff-79e0d805785bbe8a0f585efebc274478388e7ce83323d87a7492930efd924de3

lukewalczak:
Hello @pushpender-singh-ap, I'm looking for more generic fix solution for all listeners and already created initial solution. It's still WIP, since currently I'm checking whether I can make it even more generic. 
You can check the result on the [commit](https://github.com/callstack/react-native-paper/commit/1fb7d5e07c83af0c86fa64db039318b2867bcd06), maybe even install the library in your project from that branch and inform me if you're still observing any warnings.

pushpender-singh-ap:
@lukewalczak okk sure i will test this asap

pushpender-singh-ap:
@lukewalczak I am sharing with you the recording, I think the back handler removeEventListener not working if you see in the recording I use console.log for verification and logs return only first and last but not middle 2 logs.

https://user-images.githubusercontent.com/73298854/146017741-fabd84f7-eeb9-4624-b8d3-d2585ed3325e.mp4

lukewalczak:
Thanks for testing! I will check that in my free time, when will can work on it.

pushpender-singh-ap:
@lukewalczak no issue when you update something on this inform me I will test it.

lukewalczak:
Hey @pushpender-singh-ap, I think you were testing on the latest RN version and it's correct that you're not observing "inside" logs, since new `addEventListener` util works in the way where for:
* rn > 0.64 - return a subscription, so object with remove function is not going to be used
* rn < 0.63  - doesn't return subscription, so object is used and you should observe inside logs

pushpender-singh-ap:
@lukewalczak It's working fine in both versions.

Issue #2987, opened 2021-11-26
Transparent underline changes label to transparent


p-syche:
### Current behaviour

When activeUnderline (or activeOutline) is set to transparent, the label turns transparent as well.

### Expected behaviour

The label should use the theme color.

### Code sample

[Link to snack](https://snack.expo.dev/@p-syche/textinput-transparent-underline) with a repro.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android 10
| react-native          | 0.64
| react-native-paper    | 4.10.0
| node                  |
| npm or yarn           |
| expo sdk              |

### Shoutouts

This issue was discovered by @r6rob and described in a [comment](https://github.com/callstack/react-native-paper/pull/1797#issuecomment-978731341) to the PR introducing custom underline color, by @belgamo .

devnido:
If you want to make the under line transparent to make it disappear, you can do this:

` underlineStyle={{  display:'none' }}`

gasbriones:
> If you want to make the under line transparent to make it disappear, you can do this:
> 
> ` underlineStyle={{ display:'none' }}`

Love you dude!

PR #3010, opened 2021-12-14
fix: active input label color no longer overwritten by border color

r6rob:
### Summary

Resolving an [issue outlined here](https://github.com/callstack/react-native-paper/issues/2987) that is caused by the activeUnderlineColor/activeOutlineColor overriding activeColor for the TextInput component. This makes it impossible to set an activeUnderlineColor/activeOutlineColor without also overriding the input label color.

Now the active underline/outline color is set as it's passed to the outline/underline component and does not affect the label.

### Test plan

Ensure the border, underline, and input label colors can all be changed independently of each other, (using outlineColor, underlineColor, activeOutlineColor, activeUnderlineColor, and primary theme color for input label).

Screen Shot 2021-12-14 at 7 31 09 pm
Screen Shot 2021-12-14 at 7 31 15 pm

### DIFF
--- a/src/components/BottomNavigation/BottomNavigation.tsx
+++ b/src/components/BottomNavigation/BottomNavigation.tsx
@@ -323,6 +323,8 @@ const SceneComponent = React.memo(({ component, ...rest }: any) =>
  *
  * const RecentsRoute = () => <Text>Recents</Text>;
  *
+ * const NotificationsRoute = () => <Text>Notifications</Text>;
+ *
  * const MyComponent = () => {
  *   const [index, setIndex] = React.useState(0);
  *   const [routes] = React.useState([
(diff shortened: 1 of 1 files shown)


johanjq:
can we merge this one? is there something I can do to help out and get this one in?

r6rob:
> can we merge this one? is there something I can do to help out and get this one in?

r6rob:
@p-syche I'm also happy to help

p-syche:
@r6rob thank you for your PR 🏅 
However I'm not sure your changes are the best solution to the problem described in the issue. The problem described in issue #2987 concerns transparent underlines only. 
I understand the way you're thinking: now the colors are not linked and everyone can set whatever they want. But I'm afraid the library users want to have the colors set "automagically" if you will, and your change breaks that behavior. Do you think you could find a solution to only fix the transparent underline?

medarzr (approved):

medarzr (approved):

PR #3030, opened 2022-01-10
chore: update typescript and prettier versions

lukewalczak:
### Summary

* updates typescript to `4.5.4`
* updates prettier to `2.5.0`

### Test plan

1. CI should be green (lint, tests, typescript)
2. App mustn't crash

### DIFF
--- a/example/src/RootNavigator.tsx
+++ b/example/src/RootNavigator.tsx
@@ -19,7 +19,7 @@ export default function Root() {
               <Appbar.Action
                 icon="menu"
                 onPress={() =>
-                  ((navigation as any) as DrawerNavigationProp<{}>).openDrawer()
+                  (navigation as any as DrawerNavigationProp<{}>).openDrawer()
                 }
               />
             ) : null}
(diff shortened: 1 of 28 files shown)


example/src/RootNavigator.tsx
```diff
@@ -19,7 +19,7 @@ export default function Root() {
               <Appbar.Action
                 icon="menu"
                 onPress={() =>
-                  ((navigation as any) as DrawerNavigationProp<{}>).openDrawer()
+                  (navigation as any as DrawerNavigationProp<{}>).openDrawer()
```
jaworek, line 22: It would be good to verify if double cast is necessary, but it can be handled separately.

jaworek (approved):

example/src/RootNavigator.tsx
lukewalczak: Looks like it has to be doubled. We are aiming to update navigation to v6, so then we can check it one more time.

PR #3034, opened 2022-01-12
chore: convert docs to TS

jaworek:
### Summary

Removes all references and setup of Flow. Converts docs to use TypeScript.

### Test plan

Verify that docs display and work correctly.


jaworek:
I think this PR can be closed in favour of #3375.

PR #3049, opened 2022-01-26
fix: cleanup modal backhandler listener on unmount

dgruseck:
### Summary

If a modal is getting unmounted and the visible prop is not changed to false beforehand, the hardwareBackPress listener is not getting removed.
This bug was introduced in version 3.10.0, when the change to the functional component happened (#2557). 

### Test plan

1. Show the modal
2. Unmount component
3. The hardwareBackPress is still listening, you can add another hardwareBackPress listener in your parent component to see if it is getting called.

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -148,13 +148,16 @@ export default function Modal({
     }).start();
   };
 
-  const hideModal = () => {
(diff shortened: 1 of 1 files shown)


lukewalczak (approved):
LGTM, thanks!

PR #3062, opened 2022-01-31
feat: replace old variable references

Drakeoon:
### Summary

These changes fix other parts of application when theme V3 would be used instead of V2. Changes:
- [x] each ReactNativePaper.Theme got replaced with exported Theme type
- [x] each way of accessing theme colors, like `colors.background`, were replaced with optional chaining or dirty defaults (to be resolved later on, during the works on individual components)

### DIFF
--- a/example/src/Examples/AnimatedFABExample/AnimatedFABExample.tsx
+++ b/example/src/Examples/AnimatedFABExample/AnimatedFABExample.tsx
@@ -23,9 +23,7 @@ type Item = {
 };
 
 const AnimatedFABExample = () => {
-  const {
-    colors: { background },
-  } = useTheme();
+  const { colors } = useTheme();
 
   const isIOS = Platform.OS === 'ios';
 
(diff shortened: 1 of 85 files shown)


src/components/BottomNavigation/BottomNavigation.tsx
```diff
@@ -743,7 +744,7 @@ const BottomNavigation = ({
                 : 7;
 
               // We render the active icon and label on top of inactive ones and cross-fade them on change.
-              // This trick gives the illusion that we are animating between active and inactive colors.
+              // This trick gives the illusion that we are animating between active and inactive colors?.
```
lukewalczak, line 747: That can be reverted

example/src/Examples/AnimatedFABExample/AnimatedFABExample.tsx
```diff
@@ -117,10 +115,10 @@ const AnimatedFABExample = () => {
         keyExtractor={_keyExtractor}
         onEndReachedThreshold={0}
         scrollEventThrottle={16}
-        style={[styles.flex, { backgroundColor: background }]}
+        style={[styles.flex, { backgroundColor: colors?.background || '#000' }]}
```
lukewalczak, line 118: As we discussed internally, fallback colors are temporary to not block the PR. All the cases will be refactored in the way: `isV3 ? appropriate color from V3 : appropriate color from V2`

PR #3066, opened 2022-02-01
feat: adjust typography component, fix theme types

Drakeoon:
### Summary

This PR introduces a new way of approaching typography - a `<Text>` component with multiple `variants` that are Material 3 compliant

### Usage

`<Text variant="heading-large" />`

### Test plan

1. Open the example app
2. Navigate to Typography
3. You will see old examples at the top (still supported)
4. You will see new examples at the bottom (`<Text variant="...">` usage)

image


example/src/index.tsx
```diff
@@ -34,39 +34,6 @@ declare global {
...
-    ...DefaultTheme.fonts,
-    superLight: { ...DefaultTheme.fonts['light'] },
-  },
-  userDefinedThemeProperty: '',
-  animation: {
-    ...DefaultTheme.animation,
-    customProperty: 1,
-  },
-};
-
```
Drakeoon, line 69: @lukewalczak We need to keep that in mind and decide whether we still want to bring it back or maybe instead of this include a section in our docs on how to override themes

example/src/Examples/TextExample.tsx
```diff
@@ -1,29 +1,99 @@
...
   Caption,
   Headline,
   Paragraph,
   Subheading,
+  Text,
   Title,
 } from 'react-native-paper';
 import ScreenWrapper from '../ScreenWrapper';
 
+const Divider = ({ children }: React.PropsWithChildren<{}>) => {
```
lukewalczak, line 13: Rename it to `SectionHeader`, since we have the component called `Divider` and it can be misleading

lukewalczak (approved):
On the next iteration we need to:
* refine the process of overriding themes
* fix Expo Preview jo

PR #3074, opened 2022-02-03
feat: rename old colors to MD2Colors and introduce MD3Colors

lukewalczak:
PR is:
* introducing new colors related to Material You called `MD3Colors`
* renamed old Colors to `MD2Colors`


Drakeoon (approved):
Looks good to me 👍🏻

PR #3078, opened 2022-02-04
chore: web example support

lukewalczak:
### Summary

PR moves back example web support after expo update:

* adds web to  supported platforms in `app.json`
* adds `babel-loader` to devDependencies
* make using expo updates conditionally only for mobile platforms

### Test plan

1. go to example
2. `yarn start`
3. choose web platform
4. expect web example is working properly

### DIFF
--- a/example/src/index.tsx
+++ b/example/src/index.tsx
@@ -101,6 +101,8 @@ export default function PaperExample() {
     React.useState<ReactNativePaper.Theme>(CustomDefaultTheme);
   const [rtl, setRtl] = React.useState<boolean>(I18nManager.isRTL);
 
+  const isWeb = Platform.OS === 'web';
+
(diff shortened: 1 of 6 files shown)


example/src/index.tsx
```diff
@@ -101,6 +101,8 @@ export default function PaperExample() {
     React.useState<ReactNativePaper.Theme>(CustomDefaultTheme);
   const [rtl, setRtl] = React.useState<boolean>(I18nManager.isRTL);
 
+  const isWeb = Platform.OS === 'web';
```
jaworek, line 104: This util is also used in DrawerItems. Maybe it is a good idea to move is to separate file and make it accessible everywhere in the project?
Other solution would be to just use it inline everywhere.

jaworek (approved):

PR #2484, opened 2021-01-05
chore: remove deprecated accessibility props

brunohkbx:
### Summary
`accessibilityTraits` and `accessibilityComponentType`  were deprecated and removed.

Resolve #2442

PR #3079, opened 2022-02-04
feat: adjust Dialog components, introduce DialogIcon

lukewalczak:
### Summary

PR:
* adjusts `Dialog` components to Material You
* introduces `DialogIcon` component

### DIFF
--- a/example/src/Examples/Dialogs/DialogWithLongText.tsx
+++ b/example/src/Examples/Dialogs/DialogWithLongText.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Dimensions, ScrollView, StyleSheet } from 'react-native';
 import { Button, Portal, Dialog } from 'react-native-paper';
-import { TextComponent } from './utils';
+import { TextComponent } from './DialogTextComponent';
 
 const DialogWithLongText = ({
   visible,
--- a/example/src/Examples/Dialogs/DialogWithLongText.tsx
+++ b/example/src/Examples/Dialogs/DialogWithLongText.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Dimensions, ScrollView, StyleSheet } from 'react-native';
-import { Paragraph, Button, Portal, Dialog } from 'react-native-paper';
+import { Button, Portal, Dialog } from 'react-native-paper';
+import { TextComponent } from './utils';
 
 const DialogWithLongText = ({
   visible,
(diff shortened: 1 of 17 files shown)


src/components/Dialog/DialogTitle.tsx
```diff
@@ -50,21 +51,25 @@ type Props = React.ComponentPropsWithRef<typeof Title> & {
...
+const DialogTitle = ({ children, theme, style, ...rest }: Props) => {
+  const { isV3 } = theme;
+
+  const TextComponent = isV3 ? Text : Title;
+  const textColor = isV3 ? theme.colors.onSurface : theme.colors?.text;
+
+  return (
+    <TextComponent
+      variant="headlineSmall"
+      // @ts-expect-error We keep old a11y props for backwards compat with old RN versions
```
jaworek, line 63: Do we need to keep this? New version will include some breaking changes so maybe this is a good opportunity to stop supporting older RN versions? Although, it depends on how old they are.

example/src/Examples/Dialogs/utils.tsx
```diff
@@ -0,0 +1,34 @@
...
+import type { MD3TypescaleKey } from '../../../../src/types';
+
+type Props = React.ComponentProps<typeof NativeText> & {
+  style?: StyleProp<TextStyle>;
+  variant?: keyof typeof MD3TypescaleKey;
+  children: React.ReactNode;
+  isSubheading?: boolean;
+};
+
+export const TextComponent = ({ isSubheading = false, ...props }: Props) => {
```
jaworek, line 19: I think it would be better to name this file the same as component, unless you're planning to have more utils. Although, in that case you could have a folder named `utils` with each util in a separate file.

example/src/Examples/Dialogs/DialogWithLongText.tsx
```diff
@@ -56,7 +57,7 @@ const DialogWithLongText = ({
             conveying how objects move, interact, and exist in space and in
             relation to each other. Realistic lighting shows seams, divides
             space, and indicates moving parts.
-          </Paragraph>
+          </TextComponent>
         </ScrollView>
       </Dialog.ScrollArea>
       <Dialog.Actions>
```
jaworek, line 63: I noticed small inconsistency that is not related to this PR. All Dialogs use uppercase first letter and rest lowercase. Long text dialog is the only one that has "OK" instead of "Ok".

src/components/Dialog/DialogTitle.tsx
lukewalczak: The plan is to remove it along with the new version, however, it will be tracked in the separated PR.

example/src/Examples/Dialogs/utils.tsx
lukewalczak: Yup, I've renamed it, since actually, it's just a component instead of util.

jaworek (approved):

src/components/Dialog/DialogTitle.tsx
lukewalczak: https://github.com/callstack/react-native-paper/pull/2484

kwereszczynski (approved):
Looks good to me :)

PR #3084, opened 2022-02-08
feat: adjust theme structure

Drakeoon:
### Summary

After internal discussion, we decided to simplify our approach and make the new theme closer to the old theme in terms of structure. This PR delivers that, plus it has all the necessary `theme.isV3 checks` in components

### Test plan

Run the project locally:

`yarn example start`


example/src/Examples/ButtonExample.tsx
```diff
@@ -4,7 +4,9 @@ import { Button, List, useTheme } from 'react-native-paper';
 import ScreenWrapper from '../ScreenWrapper';
 
 const ButtonExample = () => {
-  const { colors } = useTheme();
+  const theme = useTheme();
+
+  const accentColor = theme.isV3 ? theme.colors.primary : theme.colors.accent;
```
lukewalczak, line 9: Suggestion: I think we can rename the variable to `color` and for `v3` use secondary as an _equivalent_ of accent from `v2`

example/src/ScreenWrapper.tsx
```diff
@@ -24,15 +24,15 @@ export default function ScreenWrapper({
...
+  const { colors, getToken, isV3 } = useTheme();
 
   const insets = useSafeAreaInsets();
 
   const containerStyle = [
     styles.container,
     {
       backgroundColor: isV3
-        ? (md('md.sys.color.background') as string)
+        ? (getToken?.('md.sys.color.background') as string)
```
lukewalczak, line 35: Do we need that `as string` casting?

src/components/Badge.tsx
```diff
@@ -86,8 +86,12 @@ const Badge = ({
     }).start();
   }, [visible, opacity, scale]);
 
-  const { backgroundColor = theme.colors?.notification, ...restStyle } =
-    (StyleSheet.flatten(style) || {}) as TextStyle;
+  const {
+    backgroundColor = theme.isV3
+      ? theme.colors.primary
```
lukewalczak, line 91: Could you please use `error` color for `v3`? I know that we'll be refactoring other colors as well, just catch it.

src/components/Drawer/DrawerItem.tsx
```diff
@@ -82,7 +82,10 @@ const DrawerItem = ({
     : 'transparent';
   const contentColor = active
     ? colors?.primary
```
lukewalczak, line 84: Shouldn't be `theme.colors`?

src/components/Drawer/DrawerItem.tsx
```diff
@@ -82,7 +82,10 @@ const DrawerItem = ({
...
-    : color(colors?.text).alpha(0.68).rgb().string();
+    : color(theme.isV3 ? theme.colors.onSurface : theme?.colors?.text)
```
lukewalczak, line 85: I think in most cases we'll not wrap colors from `v3` into `color()`, since we have specifically described colors for inactive state in MD guidelines. Anyway, you can keep it for now, since we will do another round of colors refactor later.

src/components/Typography/Text.tsx
```diff
@@ -64,8 +72,9 @@ const Text: React.ForwardRefRenderFunction<{}, Props> = (
       ref={root}
       style={[
         {
-          ...fonts?.regular,
-          color: colors?.text,
+          ...theme.fonts?.regular,
+          color:
+            theme.version === 3 ? theme.colors?.onSurface : theme.colors.text,
```
lukewalczak, line 77: Can we use `theme.isV3`?

example/src/ScreenWrapper.tsx
Drakeoon: For now the type resolving is not yet implemented. I removed getToken usage, but I think we could keep it in the codebase, as it probably makes sense in the long run for the MD3 transition

src/components/Drawer/DrawerItem.tsx
Drakeoon: In cases where properties overlap i skipped the check, since it would be something like `theme.isV3 ? theme.colors.primary : theme.colors.primary`. In cases where it does matter, I use the full `theme.colors.key` syntax

src/components/Drawer/DrawerItem.tsx
Drakeoon: Yeah, that's what I was thinking too. Ideally we won't need color() anymore, since we don't need to create new colors dynamically in the app

src/components/Typography/Text.tsx
Drakeoon: fixed

Issue #3099, opened 2022-02-23
Bottom Navigation: Labels are cut off when system font size is set to large


ptreitler:
### Current behaviour

When I set the system font size to "Large" or "Largest" on Android, the labels of the Bottom Navigation are cut off at the bottom (see screenshots below). I tried increasing the size of the bar (as per https://github.com/callstack/react-native-paper/discussions/2428#discussioncomment-430524), but that didn't help either.

### Expected behaviour

The Bottom Navigation bar should accomodate the larger label.

### Screenshots (if applicable)

![image](https://user-images.githubusercontent.com/1749537/155363606-deb7f898-2d31-492d-bc30-ed240ecebdcd.png)

-->

![image](https://user-images.githubusercontent.com/1749537/155363449-b512ca2e-2455-483f-8a36-a5b7f0bcabaa.png)

### What have you tried

- increasing the size of the bar (see above)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android 12
| react-native          | 0.66.3
| react-native-paper    | 4.10.1
| node                  | 16.14.0
| npm or yarn           | 7.20.3
| expo sdk              | -

lukewalczak:
Hello @ptreitler, [PR](https://github.com/callstack/react-native-paper/pull/3100) should fix your issue. Would you like to install the library from the PR's commit or branch in order to check if it solves your problem?

ptreitler:
Thanks @lukewalczak, I'll try it out in the next couple of days

PR #3100, opened 2022-02-23
feat: introduce new prop labelMaxFontSizeMultiplier for BottomNavigation

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3099

### Summary

PR introduces new prop for `BottomNavigation` called `labelMaxFontSizeMultiplier` which specifies the largest possible scale a label font can reach in the similar way as rn text prop [maxFontSizeMultipliers](https://reactnative.dev/docs/text#maxfontsizemultiplier)

### Test plan

before | after - `labelMaxFontSizeMultiplier={2}` | after - `labelMaxFontSizeMultiplier={1}`
--- | --- | ---
Zrzut ekranu 2022-02-23 o 22 16 52 | Zrzut ekranu 2022-02-23 o 22 16 19 | Zrzut ekranu 2022-02-23 o 22 16 27

### DIFF
--- a/src/components/BottomNavigation/BottomNavigation.tsx
+++ b/src/components/BottomNavigation/BottomNavigation.tsx
@@ -226,6 +226,10 @@ type Props = {
    * ```
    */
   barStyle?: StyleProp<ViewStyle>;
+  /**
+   * Specifies the largest possible scale a label font can reach.
+   */
+  labelMaxFontSizeMultiplier?: number;
   style?: StyleProp<ViewStyle>;
   /**
    * @optional
@@ -344,6 +348,7 @@ const BottomNavigation = ({
   onIndexChange,
   shifting = navigationState.routes.length > 3,
   safeAreaInsets,
+  labelMaxFontSizeMultiplier,
 }: Props) => {
   const { scale } = theme.animation;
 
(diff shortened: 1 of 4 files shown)


jaworek (approved):

src/components/BottomNavigation/BottomNavigation.tsx
```diff
@@ -344,6 +348,7 @@ const BottomNavigation = ({
   onIndexChange,
   shifting = navigationState.routes.length > 3,
   safeAreaInsets,
+  labelMaxFontSizeMultiplier = 1,
```
dratwas, line 351: Do we really need this default?

```
* null/undefined: inherit from the parent node or the global default (0)
* 0: no max, ignore parent/global default
* >= 1: sets the maxFontSizeMultiplier of this node to this value
```
I think the default should be `undefined` or `null` to not override `maxFontSizeMultiplier` if the `labelMaxFontSizeMultiplier` is not set.

lukewalczak: Thanks for your suggestion @dratwas! Some time ago I've made a research and checked how the apps with bottom navigation behave in terms of the bigger font size. I've noticed that most of them are scaling label font size to some point or not doing it at all. Based on that I've decided to introduce a new prop and set a default value to 1, for the case when someone is not handling it in the app.
I totally understand your point of view, however, I believe that from the library perspective we should take care of the default value for everyone who is handling it within the app, so at least there is protection from our side.

PR #3108, opened 2022-03-07
feat: adjust TextInput

lukewalczak:
### Summary

PR adjusts `TextInput` into Material You. There are no changes in API, only visual changes.

### Test plan

* checked `TextInput` on:
  * both themes version
  * both color themes
  * both platforms
* snapshots indicates there are no changes in the previous version

### DIFF
--- a/src/components/TextInput/Label/InputLabel.tsx
+++ b/src/components/TextInput/Label/InputLabel.tsx
@@ -71,14 +71,13 @@ const InputLabel = (props: InputLabelProps) => {
     ],
   };
 
-  let textColor;
+  let textColor = placeholderColor;
 
   if (error && errorColor) {
     textColor = errorColor;
-  } else if (isV3 && parentState.value && mode !== 'outlined') {
+  }
+  if (isV3 && parentState.value && mode !== 'outlined') {
     textColor = activeColor;
-  } else {
-    textColor = placeholderColor;
   }
 
   return label ? (
--- a/src/components/TextInput/Label/InputLabel.tsx
+++ b/src/components/TextInput/Label/InputLabel.tsx
@@ -71,6 +71,16 @@ const InputLabel = (props: InputLabelProps) => {
     ],
   };
 
+  let textColor;
+
+  if (error && errorColor) {
+    textColor = errorColor;
+  } else if (isV3 && parentState.value && mode !== 'outlined') {
+    textColor = activeColor;
+  } else {
+    textColor = placeholderColor;
+  }
+
   return label ? (
     // Position colored placeholder and gray placeholder on top of each other and crossfade them
     // This gives the effect of animating the color, but allows us to use native driver
@@ -129,12 +139,7 @@ const InputLabel = (props: InputLabelProps) => {
           labelStyle,
           paddingOffset,
           {
-            color:
-              error && errorColor
-                ? errorColor
-                : parentState.value && mode !== 'outlined'
-                ? activeColor
-                : placeholderColor,
+            color: textColor,
             opacity: placeholderOpacity,
           },
           isV3 && styles.md3TextLine,
--- a/src/components/TextInput/Label/InputLabel.tsx
+++ b/src/components/TextInput/Label/InputLabel.tsx
@@ -5,8 +5,7 @@ import AnimatedText from '../../Typography/AnimatedText';
 import type { InputLabelProps } from '../types';
 
 const InputLabel = (props: InputLabelProps) => {
-  const { parentState, labelBackground } = props;
-
+  const { parentState, labelBackground, mode } = props;
   const {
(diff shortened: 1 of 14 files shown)


src/components/TextInput/Label/InputLabel.tsx
```diff
@@ -70,6 +71,16 @@ const InputLabel = (props: InputLabelProps) => {
...
+  let textColor;
+
+  if (error && errorColor) {
+    textColor = errorColor;
+  } else if (isV3 && parentState.value && mode !== 'outlined') {
+    textColor = activeColor;
+  } else {
+    textColor = placeholderColor;
+  }
+
```
p-syche, line 83: I don't love an `if/else` statement where the conditions are different from one another 😬
Is there any way this could be broken up into separate `if/else` checks and treat your last value as the default?

```suggestion
  let textColor = placeholderColor;

  if (error && errorColor) {
    textColor = errorColor;
  }
  if (!error && isV3 && parentState.value && mode !== 'outlined') {
    textColor = activeColor;
  } 

```

src/components/TextInput/Label/LabelBackground.tsx
```diff
@@ -58,23 +61,34 @@ const LabelBackground = ({
...
+              ? {
+                  width:
+                    parentState.labelLayout.width -
+                    placeholderStyle.paddingHorizontal,
+                }
+              : {
+                  maxWidth:
+                    parentState.labelLayout.width -
+                    2 * placeholderStyle.paddingHorizontal,
+                },
```
p-syche, line 91: Would it be possible to put the `transform` inside the `isV3` check? May be more succinct 🤔

src/components/TextInput/TextInputOutlined.tsx
```diff
@@ -378,19 +376,14 @@ const Outline = ({
       {
         backgroundColor,
         borderRadius: theme.roundness,
-        borderWidth: focused ? 2 : 1,
+        borderWidth: (theme.isV3 ? hasActiveOutline : focused) ? 2 : 1,
```
p-syche, line 379: Hm, a double condition... Looks funny 😂 But I don't have any good suggestion to change it 🤷‍♀️

src/components/TextInput/helpers.tsx
```diff
@@ -287,3 +310,179 @@ export const calculateFlatInputHorizontalPadding = ({
...
+      underlineColorCustom = 'transparent';
+    } else {
+      inputTextColor = theme.colors?.text;
+      placeholderColor = theme.colors?.placeholder;
+      underlineColorCustom = underlineColor || theme.colors.disabled;
+      backgroundColor = theme.dark
+        ? color(theme.colors?.background).lighten(0.24).rgb().string()
+        : color(theme.colors?.background).darken(0.06).rgb().string();
+    }
+  }
```
p-syche, line 373: The `if (theme.isV3)` block doesn't have the `placeholderColor`. Is that intentional?

src/components/TextInput/helpers.tsx
```diff
@@ -287,3 +310,179 @@ export const calculateFlatInputHorizontalPadding = ({
...
+
+  // The same for both themes
+  activeColor = error
+    ? theme.colors?.error
+    : activeOutlineColor || theme.colors?.primary;
+  errorColor = theme.colors?.error;
+
+  if (theme.isV3) {
+    if (disabled) {
+      inputTextColor = activeColor = theme.colors.onSurfaceDisabled;
```
p-syche, line 411: InputTextColor equals activeColor equals theme.colors? This looks like a typo, unless I'm missing something?

p-syche (approved):
@lukewalczak - nicely done 👍 This is no trivial change 💪 Even though only 12 files were changed it looks like you had to think about a lot of moving parts.

I left a few comments and questions. However, as I checked on the Expo app, the TextInput works absolutely fine. So you can treat my comments as you like 🙇

src/components/TextInput/helpers.tsx
lukewalczak: Looks like [there](https://github.com/callstack/react-native-paper/pull/3108/files#diff-3291fa058b0085e51724d27a5942d2ea16e44f505f55dfec849c9201dce2a191R353) is `placeholderColor`. 
According to the figma file, in the case of `flat` mode, input text color and placeholder color are the same.

Drakeoon, line 373: I suggest cleaning it up with aproppriate guard clauses or separate functions for every property - it's very hard to understand what the data flow for each property is

Drakeoon (approved):

kwereszczynski (approved):
Tested on Android and Ios. Looks good to me :)

PR #3113, opened 2022-03-08
feat: adjust Chip

lukewalczak:
### Summary

PR:

* adjusts `Chip` component into Material You.
* introduces new props called:
  * `compact` for smaller horizontal paddings when there is only a label
  * `elevated` whether chip should have elevation
  * `showSelectedOverlay` indicating displaying an overlay on the selected chip

### Test plan

* check application in both themes, if `Chip` works and have correct colors in both themes on both platforms
* snapshots indicate that `Chip` from theme version 2 wasn't changed.

### DIFF
--- a/example/src/Examples/ChipExample.tsx
+++ b/example/src/Examples/ChipExample.tsx
@@ -65,7 +65,7 @@ const ChipExample = () => {
               Avatar (disabled)
             </Chip>
             {isV3 && (
-              <Chip dense style={styles.chip}>
+              <Chip dense style={styles.chip} onPress={() => {}}>
                 Dense chip
               </Chip>
             )}
@@ -134,7 +134,12 @@ const ChipExample = () => {
               Avatar (disabled)
             </Chip>
             {isV3 && (
-              <Chip mode="outlined" dense style={styles.chip}>
+              <Chip
+                mode="outlined"
+                dense
+                onPress={() => {}}
+                style={styles.chip}
+              >
                 Dense chip
               </Chip>
             )}
--- a/example/src/Examples/ChipExample.tsx
+++ b/example/src/Examples/ChipExample.tsx
@@ -6,7 +6,8 @@ import ScreenWrapper from '../ScreenWrapper';
 
 const ChipExample = () => {
   const [visible, setVisible] = React.useState<boolean>(false);
-  const { colors, isV3 } = useTheme();
+  const theme = useTheme();
+  const customColor = theme.isV3 ? theme.colors.error : theme.colors.primary;
 
   return (
     <>
@@ -16,6 +17,21 @@ const ChipExample = () => {
             <Chip selected onPress={() => {}} style={styles.chip}>
               Simple
             </Chip>
+            {theme.isV3 && (
+              <Chip
+                selected
+                showSelectedOverlay
+                onPress={() => {}}
+                style={styles.chip}
+              >
+                With selected overlay
+              </Chip>
+            )}
+            {theme.isV3 && (
+              <Chip elevated onPress={() => {}} style={styles.chip}>
+                Elevated
+              </Chip>
+            )}
             <Chip
               onPress={() => {}}
               onClose={() => {}}
@@ -64,9 +80,9 @@ const ChipExample = () => {
             >
               Avatar (disabled)
             </Chip>
-            {isV3 && (
-              <Chip dense style={styles.chip} onPress={() => {}}>
-                Dense chip
+            {theme.isV3 && (
+              <Chip compact style={styles.chip} onPress={() => {}}>
+                Compact chip
               </Chip>
             )}
           </View>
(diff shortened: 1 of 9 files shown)


example/src/Examples/ChipExample.tsx
```diff
@@ -16,6 +17,21 @@ const ChipExample = () => {
...
+                style={styles.chip}
+              >
+                With selected overlay
+              </Chip>
+            )}
+            {theme.isV3 && (
+              <Chip elevated onPress={() => {}} style={styles.chip}>
+                Elevated
+              </Chip>
+            )}
```
p-syche, line 34: Why are these in 2 separate conditions? Wouldn't it be possible to leave one condition and put both `Chips` inside a React.Fragment?

example/src/Examples/ChipExample.tsx
```diff
@@ -64,13 +80,39 @@ const ChipExample = () => {
...
+            {theme.isV3 && (
+              <Chip
+                mode="outlined"
+                elevated
+                onPress={() => {}}
+                style={styles.chip}
+              >
+                Elevated
+              </Chip>
+            )}
```
p-syche, line 115: Same question as above :)

example/src/Examples/ChipExample.tsx
```diff
@@ -128,10 +170,46 @@ const ChipExample = () => {
...
+                  compact
+                  avatar={
+                    <Image source={require('../../assets/images/avatar.png')} />
+                  }
+                  style={[styles.chip, styles.customBorderRadius]}
+                >
+                  Compact with custom border radius
+                </Chip>
+              </>
+            )}
```
p-syche, line 212: Hey, this is what I was suggesting XD 😂

example/src/Examples/ChipExample.tsx
```diff
@@ -6,7 +6,8 @@ import ScreenWrapper from '../ScreenWrapper';
 
 const ChipExample = () => {
   const [visible, setVisible] = React.useState<boolean>(false);
-  const { colors } = useTheme();
+  const theme = useTheme();
+  const customColor = theme.isV3 ? theme.colors.error : theme.colors.primary;
```
p-syche, line 10: So for a custom color, if the user is on V3 they will get the error color? I mean, I checked the app and it looks good, this condition just got me curious 🤔

src/components/Chip/Chip.tsx
```diff
@@ -140,84 +155,62 @@ const Chip = ({
...
+  const isOutlined = mode === 'outlined';
+
   const handlePressIn = () => {
     const { scale } = theme.animation;
     Animated.timing(elevation, {
-      toValue: 4,
+      toValue: isV3 ? (elevated ? 2 : 0) : 4,
       duration: 200 * scale,
-      useNativeDriver: false,
+      useNativeDriver: true,
```
p-syche, line 177: Nice 😎

src/components/Chip/helpers.tsx
```diff
@@ -0,0 +1,126 @@
...
+  theme,
+  isOutlined,
+  disabled,
+}: {
+  selectedColor?: string;
+  showSelectedOverlay?: boolean;
+  customBackgroundColor?: ColorValue;
+  theme: Theme;
+  isOutlined: boolean;
+  disabled?: boolean;
```
p-syche, line 19: This is a total nitpick 🤏 , but I like to put the required items first in the list, and then the optional ones.

src/components/Chip/helpers.tsx
```diff
@@ -0,0 +1,126 @@
...
+    } else {
+      const customTextColor =
+        selectedColor !== undefined ? selectedColor : theme.colors.text;
+      textColor = color(customTextColor).alpha(0.87).rgb().string();
+      iconColor = color(customTextColor).alpha(0.54).rgb().string();
+      underlayColor = selectedColor
+        ? color(selectedColor).fade(0.5).rgb().string()
+        : selectedBackgroundColor;
+    }
+  }
```
p-syche, line 116: The `ifV3` block is nicely organized between `if disabled` and `else`. The bottom block seems kind of unroganized with a few values first and then an `if/else disabled` later.

p-syche (approved):
Whew! A lot of complicated conditions here... But it seems like you got it all working fine 🤩  I checked the examples on the Expo app and they look good in both themes.

I left a few questions and comments, and one nitpick that's probably the nitpickiest nitpick ever 😅

src/components/Chip/helpers.tsx
lukewalczak: Basically in `!ifV3` block two colors: `selectedBackgroundColor` and `borderColor` are above `if/else disabled` since those two should always remain the same no matter of _disablement_.

example/src/Examples/ChipExample.tsx
lukewalczak: Organized them into one fragment.

src/components/Chip/helpers.tsx
lukewalczak: That rule makes sense to me, applied it!

Drakeoon (approved):

PR #3114, opened 2022-03-08
feat: adjust Button

lukewalczak:
### Summary

PR:
* adjusts `Button` into Material You.
* introduces two new modes: `'elevated' and 'contained-tonal'` - along with theme version 3, `contained` button doesn't have elevation, so suggesting to use `elevated`
* deprecated `color` prop and introduces instead `textColor` and `buttonColor`

### Test plan

* check application in both themes, if `Button` works and have correct colors in both themes on both platforms
* snapshots indicate that `Button` from theme version 2 wasn't changed.


OlimpiaZurek (approved):

src/components/Button/helpers.tsx
```diff
@@ -0,0 +1,123 @@
...
+  | 'elevated'
+  | 'contained-tonal';
+
+export const getButtonColors = ({
+  buttonColor,
+  theme,
+  mode,
+  disabled,
+  dark,
+}: {
```
Drakeoon, line 19: @lukewalczak Consider refactoring this function to use guard clauses. It would be more readable and maintainable. We could always asume to return `{
    backgroundColor,
    borderColor,
    textColor,
    borderWidth,
  };` and have getters for each variable put in a separate function

Drakeoon (changes requested):

src/components/Button/helpers.tsx
lukewalczak: Thanks @Drakeoon for pointing it out. When I started refactoring it I've noticed that we have to introduce two new props for background and text color separately. I will work on it right now.

lukewalczak: Done @Drakeoon!

Drakeoon (approved):

PR #3118, opened 2022-03-11
feat: adjust Menu

lukewalczak:
### Test plan

…ription (#3082)

* refactor: adjust AnimatedText, update component and props description

* refactor: correct prop description

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -16,6 +16,7 @@ import {
   findNodeHandle,
   NativeEventSubscription,
 } from 'react-native';
+import color from 'color';
 
 import { withTheme } from '../../core/theming';
 import type { $Omit } from '../../types';
@@ -529,7 +530,7 @@ class Menu extends React.Component<Props, State> {
     const shadowMenuContainerStyle = {
       opacity: opacityAnimation,
       transform: scaleTransforms,
-      borderRadius: theme.roundness,
+      borderRadius: theme.isV3 ? 4 : theme.roundness,
       ...(scrollableMenuHeight ? { height: scrollableMenuHeight } : {}),
     };
 
@@ -571,6 +572,13 @@ class Menu extends React.Component<Props, State> {
                     [
                       styles.shadowMenuContainer,
                       shadowMenuContainerStyle,
+                      theme.isV3 && styles.md3ShadowMenuContainer,
+                      theme.isV3 && {
+                        backgroundColor: color(theme.colors.surface)
+                          .mix(color(theme.colors.primary), 0.08)
+                          .rgb()
+                          .string(),
+                      },
                       contentStyle,
                     ] as StyleProp<ViewStyle>
                   }
@@ -596,6 +604,9 @@ const styles = StyleSheet.create({
     opacity: 0,
     paddingVertical: 8,
     elevation: 8,
+  },
+  md3ShadowMenuContainer: {
+    elevation: 3,
   },
 });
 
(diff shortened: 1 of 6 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -529,7 +530,7 @@ class Menu extends React.Component<Props, State> {
     const shadowMenuContainerStyle = {
       opacity: opacityAnimation,
       transform: scaleTransforms,
-      borderRadius: theme.roundness,
+      borderRadius: theme.isV3 ? 4 : theme.roundness,
```
drplauska, line 533: `theme.roundness` is already equal to 4, and wouldn't it be better to put values to `DefaultTheme` anyway?

src/components/Menu/Menu.tsx
```diff
@@ -571,6 +572,13 @@ class Menu extends React.Component<Props, State> {
                     [
                       styles.shadowMenuContainer,
                       shadowMenuContainerStyle,
+                      theme.isV3 && styles.md3ShadowMenuContainer,
+                      theme.isV3 && {
```
drplauska, line 576: this might be moved to `shadowMenuContainerStyle` ?

src/components/Menu/Menu.tsx
lukewalczak: Base `roundness` will be the same for both themes (4), so used `theme.roundness`. thanks!

src/components/Menu/Menu.tsx
lukewalczak: Refactored

src/components/Menu/MenuItem.tsx
```diff
@@ -14,19 +14,36 @@ import { black, white } from '../../styles/themes/v2/colors';
...
+   * Leading icon to display for the `MenuItem`.
+   */
+  leadingIcon?: IconSource;
+  /**
+   * @supported Available in v3.x with theme version 3
+   *
+   * Trailing icon to display for the `MenuItem`.
    */
-  icon?: IconSource;
+  trailingIcon?: IconSource;
```
p-syche, line 36: If there's a note for `icon` being renamed to `leadingIcon` in v3.x maybe it would be good to add a note here about the `trailingIcon` being added?

p-syche (approved):
Looking good 😎 

I left one nitpick suggestion about leaving a comment in the code, take it or leave it ;)

src/components/Menu/MenuItem.tsx
lukewalczak: I've added three docs annotation types for indicating props changes: 

`@supported`  for totally new one
`@deprecated` for those which won't be available in the new version
`@renamed` for those which are renamed.

In case of `trailingIcon` you can notice there is an annotation `@supported` which then in docs is displayed in the form of badge:

image

src/components/Menu/MenuItem.tsx
```diff
@@ -91,79 +110,142 @@ const MenuItem = ({
...
+      }
+    }
+
+    return {
+      minWidth,
+      maxWidth,
+    };
+  }, [theme.isV3, leadingIcon, trailingIcon]);
+
+  const { minWidth, maxWidth } = getContentWidth;
```
Drakeoon, line 166: To clean up the main implementation I suggest moving this function to a local helper hook for MenuItem that would be consumed like:

`const { minWidth, maxWidth } = useMenuItemContentWidth({ leadingIcon, trailingIcon });`

It can get `theme` inside from a hook and can also be unit tested to make sure all scenarios are giving you the `minWidth` and `maxWidth` you expect

Drakeoon (approved):
My requested change is mostly for readability purposes, it doesn't prevent merging the PR

PR #3120, opened 2022-03-12
feat: adjust Checkbox and RadioButton

lukewalczak:
### Summary

PR adjusting Checkbox and RadioButton to the new Material Design, mostly in terms of theming.


p-syche:
@lukewalczak the code looks good to me 🏅  Setting the component conditionally with a check for `isV3` is an interesting pattern. It reads nicely 👩‍💻 
I checked through the Expo app and everything is working fine too 💪 
The only thing that cought my eye is that in Radio Button Group, in the first section (With RadioButton), the labels are not clickable. It may not be affected by your changes though.

I'm also curious about the fact that you removed fallback colors, for example here: https://github.com/callstack/react-native-paper/pull/3120/files#diff-a38bb8b37129bdb5d5b77ddc5d09028229bde793dd358d6c29aa081d9b4c5573L144
Are you really *really* sure they are not needed?

lukewalczak:
Thanks @p-syche for reviewing and testing changes!

> The only thing that cought my eye is that in Radio Button Group, in the first section (With RadioButton), the labels are not clickable. It may not be affected by your changes though.

I will check what's going on there.

> I'm also curious about the fact that you removed fallback colors, for example here: [#3120 (files)](https://github.com/callstack/react-native-paper/pull/3120/files#diff-a38bb8b37129bdb5d5b77ddc5d09028229bde793dd358d6c29aa081d9b4c5573L144)

_Long story short_ on the beginning of adjusting paper to the new theme, we had some issues with TS complaining about the colors, so to avoid blocking ourselves, we have added the fallback color. You can notice that on `main` [branch](https://github.com/callstack/react-native-paper/blob/main/src/components/RadioButton/RadioButtonAndroid.tsx#L135) there were no any _fallback colors_. Right now when the theme is adjusted properly in terms of TS, we don't need those colors and they can be removed ✌🏽

Drakeoon (approved):

example/src/Examples/CheckboxExample.tsx
```diff
@@ -13,11 +16,14 @@ const CheckboxExample = () => {
...
+
   return (
     <ScreenWrapper style={styles.container}>
       <TouchableRipple onPress={() => setCheckedNormal(!checkedNormal)}>
         <View style={styles.row}>
-          <Paragraph>Normal</Paragraph>
+          <TextComponent>Normal</TextComponent>
           <View pointerEvents="none">
             <Checkbox status={checkedNormal ? 'checked' : 'unchecked'} />
           </View>
```
Drakeoon, line 29: I think having dynamic components like `TextComponent` makes the example harder to understand for Paper newcomers who are the main target audience of examples - It makes it harder to just take what is in the example and use it on their project. I don't think this is a change that should be considered in this PR, but we should discuss the approach and what examples should achieve in the end

Drakeoon (approved):

kwereszczynski (approved):
Tested on Android and IOS. Looks good to me :)

Issue #3144, opened 2022-04-05
TextInput value is ignored (web)


seahorsepip:
### Current behaviour
When you for example filter the input state, the unfiltered input state is rendered.

### Expected behaviour
The current value passed to the value prop is always rendered.

### Code sample
```tsx
const [value, setValue] = useState('');

// Renders react-native-paper instead of my-filtered-value when user types react-native-paper
const render = () => <TextInput value={value} onChangeText={(text) => setValue('my-filtered-value')} />
```

### What have you tried

This seems to still work correctly in older versions (tested in 4.11.1).

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | web
| react-native          | latest
| react-native-paper    | latest

seahorsepip:
Let me know if this issue has been fixed already or a fix is in the works. Else I'll make a PR.

seahorsepip:
Ended up being a small change related to controlled and uncontrolled inputs. Previously the props value was synced with local state in a `useLayoutEffect` but this causes an additional render and can get out of sync in cases where input is filtered. 

In my usecase there was a props value that didn't change while the input entered by the end user did change, the `useLayoutEffect` won't be called and the local state gets out of sync with the props value. 

One example would be `onChangeText{text => text.replace(/./g, 'b')}` and a user typing `aaaaaaaaabbbaaaaabbaaa`.

lukewalczak:
Fixed by: https://github.com/callstack/react-native-paper/pull/3145

PR #3145, opened 2022-04-06
fix: controlled text input

seahorsepip:
Use value from props instead of local state for controlled `TextInput`.

### Summary

See #3144

### Test plan

Test both a controlled and non controlled input, check if value shown in input is correct.


src/components/TextInput/TextInput.tsx
```diff
@@ -220,9 +220,12 @@ const TextInput = React.forwardRef<TextInputHandles, TextInputProps>(
     const [placeholder, setPlaceholder] = React.useState<string | undefined>(
       ''
     );
-    const [value, setValue] = React.useState<string | undefined>(
-      validInputValue
-    );
+    let [value, setValue] = React.useState<string | undefined>(validInputValue);
+    if (isControlled) {
+      // Use value from props instead of local state when input is controlled
+      value = rest.value;
```
drplauska, line 226: useState value shouldn't be reassigned like this.

drplauska, line 226: useState value shouldn't be reassigned like this.

seahorsepip: I don't think it makes much of a difference but I've changed the code t use a separate variable `value`, the state `value` is renamed to `uncontrolledValue`.

src/components/TextInput/TextInput.tsx
```diff
@@ -220,9 +220,10 @@ const TextInput = React.forwardRef<TextInputHandles, TextInputProps>(
     const [placeholder, setPlaceholder] = React.useState<string | undefined>(
       ''
     );
-    const [value, setValue] = React.useState<string | undefined>(
-      validInputValue
-    );
+    let [uncontrolledValue, setUncontrolledValue] = React.useState<string | undefined>(validInputValue);
```
lukewalczak, line 223:
```suggestion
    const [uncontrolledValue, setUncontrolledValue] = React.useState<
      string | undefined
    >(validInputValue);
```

lukewalczak (approved):

Issue #3146, opened 2022-04-08
Breaking changes on TextInput ref between 4.12 and 4.11.x


lucaswitch:
### Current behaviour 4.12

When setting a ref inside a <TextInput/> element exposes in ref the following:
      focus: ()
      clear: ()
      setNativeProps: (args: Object) 
      isFocused: ()
      blur: ()
      forceFocus: ()

### Expected behaviour 4.11

In older versions than **4.11.x it was possible to access <TextInput/> private component methods like, setFocused() and others**, allowing devs to change the behavior of TextInput more than the exposed in 4.12.0.

seahorsepip:
Here's a gist I've been using to share an internal ref while also using it internally: https://gist.github.com/pie6k/b4717f392d773a71f67e110b42927fea

lukewalczak:
Hey @lucaswitch could you please present on the snack (since it's the easiest way to present the problem using different library versions) the difference in `TextInput` between the latest version and the previous one?

lucaswitch:
Ok, i will do.

lucaswitch:
Hello, any news?

goxr3plus:
It's causing this problem https://github.com/callstack/react-native-paper/issues/3141#issuecomment-1147695589

PR #3147, opened 2022-04-08
Allow progamatically update focused TextInput state through ref.

lucaswitch:
Given a <TextInput ref={inputRef} />

We could later do when needed.
inputRef.current.setFocused(true);


drplauska:
is `setFocused` even a legit method? Can't seem to find it in types

lucaswitch:
@drplauska  is actually being discussed since the 4.12 update in #3146

PR #3154, opened 2022-04-16
feat: adjust elevation usage in components based on Surface

lukewalczak:
* After introducing the new `Surface` functionality according to the doc comment saying:

>   Note: In version 2 the `elevation` prop was accepted via `style` prop i.e. `style={{ elevation: 4 }}`. It's no longer supported with theme version 3 and you should use `elevation` property instead.

I've moved the `elevation` style in  Surface to be at the end, so there won't be an option to override it via styles.

* In the components based on `Surface` where previously there was an option to use different elevations, I've introduced a dedicated prop.

* There are couple more components based on `Surface` but elevation will be adjusted in them separaterly.

* With theme version `2` nothing has been changed.

More notes:

* `Banner`, `Searchbar` and `Snackbar` received `elevation` prop
* `Chip`, `Card`, `Button`, will receive a new _mode_ with elevation
* `Appbar` and `FAB` will receive elevated _(boolean)_ prop to enable it on demant
* `BottomNavigation` won't have elevation with theme version 3
* `Menu`, `AnimatedFAB` and `FAB.Group` should always have an elevation

### Test plan

* snapshots are not updated which means there is no regression

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -7,7 +7,6 @@ import Icon, { IconSource } from './Icon';
 import { withTheme } from '../core/theming';
 import type { $RemoveChildren, Theme } from '../types';
 
-const ELEVATION = 1;
 const DEFAULT_MAX_WIDTH = 960;
 
 type Props = $RemoveChildren<typeof Surface> & {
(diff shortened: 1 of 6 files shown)


Drakeoon (approved):
This looks good to me. I'm personally against overriding/supporting custom elevation in elements like Cards, but this change requires a bigger shift in how we approach components, so let's leave it like that for now

PR #3083, opened 2022-02-07
feat: adjust BottomNavigation

lukewalczak:
### Summary

PR:
* adjusts `BottomNavigation` into Material You.
* introduced prop `compact` indicating whether tabs should be spread across the entire width.
* couple changes in `navigationState.routes`:
  * renamed from `icon` to `focusedIcon`
  * introduced `unfocusedIcon`
  * deprecated `color`

### Test plan

* check application in both themes, if `BottomNavigation` works and have correct colors in both themes on both platforms
* rotate the phone to confirm whether `BottomNavigation` works properly in horizontal mode
* snapshots indicate that `BottomNavigation` from theme version 2 wasn't changed.

PR #3158, opened 2022-04-21
feat: adjust Card

OlimpiaZurek:
### Summary
PR: 
- adjust `Card` into Material You:
- introduces new mode `filled` ( without elevation and border), 
- removes elevation from `outlined` mode
- moves buttons to right side
- rounds the corners of the Image

### Test plan
 - check application in both them, if Card works and have correct elevation, button placement roundness  and color in both themes on both platforms


src/components/Card/Card.tsx
```diff
@@ -189,6 +198,17 @@ const Card = ({
...
     .rgb()
     .string();
+
+  const darkModeBackgroundColor =
+    dark && isAdaptiveMode
+      ? overlay(elevation, theme.colors.surface)
+      : theme.colors.surface;
+
+  const backgroundColor =
+    isV3 && cardMode === 'filled'
```
lukewalczak, line 208: Please create a helper `isMode` similar to the one in [button](https://github.com/callstack/react-native-paper/pull/3114/files#diff-ce687691fe458ef8d9a85cc00e06bf778a4272a39b854ce5c89e598eecb531e3R155-R160) and use it in all places where is mode comparison.

src/components/Card/Card.tsx
```diff
@@ -47,9 +55,10 @@ type Props = React.ComponentProps<typeof Surface> & {
   /**
    * Mode of the Card.
    * - `elevated` - Card with elevation.
+   * - `filled` - Card with without outline and elevation.
```
lukewalczak, line 58: Add docs annotation: `@supported Available in v3.x with theme version 3`. Please take a look at a similar usage in `BottomNavigation` for the [reference](https://github.com/callstack/react-native-paper/pull/3083/files#diff-ac8f9da8cbc036504635455603d1faacb546641d2e8cb7f8d85cbc84551bd9b4R87)

src/components/Card/CardActions.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { StyleSheet, StyleProp, View, ViewStyle } from 'react-native';
+import type { Theme } from '../../types';
 
 type Props = React.ComponentPropsWithRef<typeof View> & {
   /**
    * Items inside the `CardActions`.
    */
   children: React.ReactNode;
   style?: StyleProp<ViewStyle>;
+  theme?: Theme;
```
lukewalczak, line 11: No need to introduce theme prop, you can use the hook `useTheme`

example/src/Examples/CardExample.tsx
```diff
@@ -14,22 +14,42 @@ import { PreferencesContext } from '..';
...
-  const mode = isOutlined ? 'outlined' : 'elevated';
-
+  const isV2mode = isOutlined ? 'outlined' : 'elevated';
+  const [isV3mode, setIsV3Mode] = React.useState('elevated') as any;
   const preferences = React.useContext(PreferencesContext);
+  const mode = isV3 ? isV3mode : isV2mode;
 
   return (
     <ScreenWrapper contentContainerStyle={styles.content}>
+      {isV3 && (
```
lukewalczak, line 26: Can we please use three `Chip`s instead of `Button`s, so we can indicate which mode is selected by chip selection and remove that text?

src/components/Card/Card.tsx
OlimpiaZurek: Done

src/components/Card/Card.tsx
OlimpiaZurek: done

src/components/Card/CardActions.tsx
OlimpiaZurek: done

example/src/Examples/CardExample.tsx
OlimpiaZurek: Done

kwereszczynski (approved):
Tested on IOS and Android. Looks good to me :)

PR #3215, opened 2022-06-21
fix: supporting empty labels

VictorioMolina:
### Summary

When using the current `TextInputFlat.js`, I get the error:

> Text strings must be rendered within a <Text> component.

which __breaks my components tree__.

This is how I am using the `TextInput` component:

```javascript
<TextInput
  ref={textInputRef}
  label="" <---- Try with an empty label
  placeholder="Enter a description..."
  ...
/>
```

The problem is in the __line 311__ of the `TextInputFlat.js` module. If you just do the following:

`{!isAndroid && multiline && Boolean(label) && (`

the error disappears.

### Test plan

Just pass an empty label to the `TextInput` component. Something which, in my opinion, should be accepted.

You can also check this little [snack](https://snack.expo.dev/-QUJzBClx), where the current bug is reproduced.

### DIFF
--- a/src/components/TextInput/TextInputFlat.tsx
+++ b/src/components/TextInput/TextInputFlat.tsx
@@ -308,7 +308,7 @@ const TextInputFlat = ({
           },
         ]}
       >
-        {!isAndroid && multiline && Boolean(label) && (
+        {!isAndroid && multiline && !!label && (
           // Workaround for: https://github.com/callstack/react-native-paper/issues/2799
           // Patch for a multiline TextInput with fixed height, which allow to avoid covering input label with its value.
           <View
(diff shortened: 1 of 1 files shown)


VictorioMolina:
It might not be really useful to support empty labels BTW, but I consider this a bug. Also, notice that you can pass empty placeholders to the current version without getting errors.

VictorioMolina:
@lukewalczak yes.

src/components/TextInput/TextInputFlat.tsx
```diff
@@ -308,7 +308,7 @@ const TextInputFlat = ({
           },
         ]}
       >
-        {!isAndroid && multiline && label && (
+        {!isAndroid && multiline && Boolean(label) && (
```
lukewalczak, line 311:
```suggestion
        {!isAndroid && multiline && !!label && (
```

PR #3256, opened 2022-07-15
Add reference to icons list to Avatar.icon

meatnordrink:
### Summary

Users going straight to the `Avatar.icon` section of the navigation may not realize there's a section dedicated to Icons above; I didn't. If they don't, it's unclear where to find what icons you can use with this prop.

This adds an explicit reference to the Icons section.

I also edited the grammar in another sentence on the page.

### Test plan

This changes the text on [avatar-icon.html](https://callstack.github.io/react-native-paper/avatar-icon.html); visual inspection is an appropriate test.

### DIFF
--- a/src/components/Avatar/AvatarIcon.tsx
+++ b/src/components/Avatar/AvatarIcon.tsx
@@ -29,13 +29,15 @@ type Props = React.ComponentPropsWithRef<typeof View> & {
 };
 
 /**
- * Avatars can be used to represent people in a graphical way.
+ * Avatars can be used to graphically represent users or actions.
  *
  * <div class="screenshots">
  *   <figure>
  *     <img class="medium" src="screenshots/avatar-icon.png" />
  *   </figure>
  * </div>
+ *
+ * You can see a list of available icons in the [Icons](icons.html) section.
  *
  * ## Usage
  * ```js


src/components/Avatar/AvatarIcon.tsx
```diff
@@ -29,14 +29,16 @@ type Props = React.ComponentPropsWithRef<typeof View> & {
...
- * Avatars can be used to represent people in a graphical way.
+ * Avatars can be used to graphically represent users or actions.
  *
  * <div class="screenshots">
  *   <figure>
  *     <img class="medium" src="screenshots/avatar-icon.png" />
  *   </figure>
  * </div>
  *
+ * You can see a list of available icons in the [Icons](icons.html) section.
```
lukewalczak, line 40: This is only one of the possible ways to specify the icon. The type of the `icon` prop is [`IconSource`](https://github.com/callstack/react-native-paper/blob/main/src/components/Icon.tsx#L15), so you can also pass there a function with other icon component if needed, like on the presented [snack](https://snack.expo.dev/bWhgAkSV5)

meatnordrink: Thanks for the comment, @lukewalczak. (And again, sorry this fell off my radar.) Do you feel that, because there are other ways to add an icon, it's not useful to let devs know where they can find a list of those available via this method; or do you think the multiple ways should be noted as well?

lukewalczak: > Do you feel that, because there are other ways to add an icon, it's not useful to let devs know where they can find a list of those available via this method; or do you think the multiple ways should be noted as well?

I just have a feeling that it's not required here to mention the list of available icons, especially since we don't have the same explanation in other components having an `icon` prop. I think we should rather link `IconSource` prop with the section `Using the icon prop` from [icons doc](https://callstack.github.io/react-native-paper/icons.html).

meatnordrink: Hm, that probably makes more sense, yes.

lukewalczak: Closing the PR in favour of: https://github.com/callstack/react-native-paper/pull/3350

meatnordrink: Great, thanks @lukewalczak! #3350 looks great.

PR #3260, opened 2022-07-18
fix: correct changing Dialog backdrop color (#3259)

lukewalczak:
### Summary

PR is adjusting changes included in the [PR](https://github.com/callstack/react-native-paper/pull/3259) to the `main` branch.

To override color backdrop color in `FAB.Group` the only way is to use `backdropColor` (supported on v4.x as well), since that color is different than typical backdrop used in `Modal` or `Dialog`.

At the same time, to the theme there was added `backdrop` color.

### Test plan

Added tests.

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -10,16 +10,14 @@ import {
   View,
   NativeEventSubscription,
 } from 'react-native';
-import color from 'color';
 import {
   getStatusBarHeight,
   getBottomSpace,
 } from 'react-native-iphone-x-helper';
 import Surface from './Surface';
-import { useTheme } from '../core/theming';
+import { withTheme } from '../core/theming';
 import useAnimatedValue from '../utils/useAnimatedValue';
 import { addEventListener } from '../utils/addEventListener';
-import { MD3Colors } from '../styles/themes/v3/tokens';
 
 type Props = {
   /**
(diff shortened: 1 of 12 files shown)


src/components/Modal.tsx
```diff
@@ -218,12 +222,11 @@ export default function Modal({
         importantForAccessibility="no"
       >
         <Animated.View
+          testID={`${testID}-backdrop`}
```
slavikdenis, line 225: Should we handle the case when `testID` is undefined?

src/components/Modal.tsx
```diff
@@ -51,6 +49,11 @@ type Props = {
    * Use this prop to change the default wrapper style or to override safe area insets with marginTop and marginBottom.
    */
   style?: StyleProp<ViewStyle>;
+  /**
+   * @optional
+   */
+  theme: ReactNativePaper.Theme;
+  testID?: string;
```
slavikdenis, line 56: Missing comment like so

```
  /**
   * testID to be used on tests.
   */
```

slavikdenis (changes requested):

src/components/Modal.tsx
lukewalczak: Sure, there should be a default `testID`.

slavikdenis (approved):

Issue #3271, opened 2022-07-22
keyboardHidesNavigationBar not work in BottomNavigation


cervisebas:
### Current behaviour
Opening the keyboard does not hide the bottom bar

### Expected behaviour
Hide the bottom bar after touching a TextInput or Searchbar

### Code sample
`<BottomNavigation
            navigationState={{
                index: this.state.indexNav,
                routes: this.routesnav
            }}
            sceneAnimationEnabled={true}
            sceneAnimationType={'shifting'}
            keyboardHidesNavigationBar={true}
            onIndexChange={(index)=>this.setState({ indexNav: index })}
            renderScene={this.renderScene}
/>`

### Screenshots (if applicable)
![image](https://user-images.githubusercontent.com/52644268/180485530-61afd6ed-5258-4cb0-b1bf-c2c4932182da.png)

### What have you tried
-

### Your Environment

| software              | version
| --------------------- | -------
| android        | 11
| react-native          | 0.69.1
| react-native-paper    | 5.0.0-rc.3
| node                  | 16.15.1
| yarn           | 1.22.19
| expo sdk              | none

lukewalczak:
Hey @cervisebas, could you please check the comment above and attach the expo snack or minimal repo presenting the issue?

PR #3275, opened 2022-07-25
fix: pass missing props to surface container

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3271

### Summary

PR passes missing spread props into the android Surface's container. 

### Test plan

Test added.

### DIFF
--- a/src/components/Surface.tsx
+++ b/src/components/Surface.tsx
@@ -31,6 +31,10 @@ type Props = React.ComponentPropsWithRef<typeof View> & {
    * @optional
    */
   theme?: Theme;
+  /**
+   * TestID used for testing purposes
+   */
+  testID?: string;
 };
 
 const MD2Surface = ({
@@ -105,6 +109,7 @@ const Surface = ({
   children,
   theme: overridenTheme,
   style,
+  testID,
   ...props
 }: Props) => {
   const theme = useTheme(overridenTheme);
@@ -137,7 +142,7 @@ const Surface = ({
     return (
       <Animated.View
         {...props}
-        testID={'surface-container'}
+        testID={testID}
         style={[
           { backgroundColor },
           elevation ? shadow(elevation, theme.isV3) : null,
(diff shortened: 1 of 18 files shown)


src/components/Surface.tsx
```diff
@@ -137,6 +137,7 @@ const Surface = ({
     return (
       <Animated.View
         {...props}
+        testID={'surface-container'}
```
slavikdenis, line 140: Are we sure we want to have the `testID` prop hard coded here?
What if the users want to define their own `testID`?
This approach will override it every time.

src/components/Surface.tsx
```diff
@@ -171,6 +172,8 @@ const Surface = ({
 
     return (
       <Animated.View
+        {...props}
+        testID={'surface-container'}
```
slavikdenis, line 176: Same as mentioned above.

src/components/Surface.tsx
```diff
@@ -239,7 +242,11 @@ const Surface = ({
         style={[getStyleForAnimatedShadowLayer(0), absoluteStyles]}
       >
         <Animated.View style={getStyleForAnimatedShadowLayer(1)}>
-          <Animated.View {...props} style={sharedStyle}>
+          <Animated.View
+            {...props}
+            testID={'surface-container'}
```
slavikdenis, line 247: Same as mentioned above.

src/components/Surface.tsx
```diff
@@ -262,7 +269,11 @@ const Surface = ({
   return (
     <Animated.View style={[getStyleForShadowLayer(0), absoluteStyles]}>
       <Animated.View style={[getStyleForShadowLayer(1)]}>
-        <Animated.View {...props} style={sharedStyle}>
+        <Animated.View
+          {...props}
+          testID={'surface-container'}
```
slavikdenis, line 274: Same as mentioned above.

slavikdenis (changes requested):

src/components/Surface.tsx
lukewalczak: Looks like it can be left over from the testing stage, which was to be re-factored later, however, this did not happen :|
Thanks for your input, I've adjusted it to your comment.

slavikdenis (approved):
Thanks for the changes. Looks good :)

Issue #3274, opened 2022-07-25
Card.Actions does not regard the props of Button


tc-imba:
### Current behaviour

In Card.Actions, we have

```ts
return React.isValidElement(child)
          ? React.cloneElement(child, {
              compact: !isV3 && child.props.compact !== false,
              mode: isV3 && (i === 0 ? 'outlined' : 'contained'),
              style: isV3 && styles.button,
            })
          : child;
```

This cause the v3 theme to make the first button always outlined and the rest always contained.
The mode and style defioned in the props of the Button component are overwritten by in Card.Actions, 

### Expected behaviour

The priority of these props on the Button components should be higher, so the users can overwrite the default behavior of Card.Actions.
Typically, if there is only one button, ususally it means OK, not cancel, its style should be contained, but there is no way to set the style now. 

### Code sample

https://snack.expo.dev/Ybtm13hN4

```ts
  <Card>
    <Card.Actions>
      <Button mode="contained">Ok</Button>
    </Card.Actions>
  </Card>
```

And the mode of OK is still outlined.

### What have you tried

Haven't tried to modify the library, but adding a condition whether the props are defined can solve the problem.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        |  *
| react-native          |  0.68.2
| react-native-paper    | 5.0.0-rc.3
| node                  |  16.2.0
| npm or yarn           |  1.22.4
| expo sdk              |  45.0.0

lukewalczak:
Hey @tc-imba, thanks for providing the context and explanation. I agree that button mode should be customizable. Based on the MD3 documentation I can observe different variants of mixed button's mode:

both contained | single outlined | single contained
--- | --- | ---
<img width="549" alt="image" src="https://user-images.githubusercontent.com/22746080/180742135-7edd19bd-f76d-4149-aac5-9d95e79f71bf.png"> | <img width="545" alt="image" src="https://user-images.githubusercontent.com/22746080/180742426-54a47c3b-ea65-457a-9499-de23ed61887f.png"> | <img width="545" alt="image" src="https://user-images.githubusercontent.com/22746080/180742557-6307b20b-4379-49d3-b04c-f735c9535d7b.png">

PR #3276, opened 2022-07-25
feat: allow to specify button's mode in CardAction

lukewalczak:
Solves: https://github.com/callstack/react-native-paper/issues/3274

### Summary

PR extending button within card action by adding condition whether the button's mode in child props is provided.

ios | android
--- | ---
Zrzut ekranu 2022-07-25 o 11 43 36 | Zrzut ekranu 2022-07-25 o 11 43 39

### Test plan

Added test covering change.

### DIFF
--- a/src/components/Card/CardActions.tsx
+++ b/src/components/Card/CardActions.tsx
@@ -49,7 +49,9 @@ const CardActions = (props: Props) => {
         return React.isValidElement(child)
           ? React.cloneElement(child, {
               compact: !isV3 && child.props.compact !== false,
-              mode: isV3 && (i === 0 ? 'outlined' : 'contained'),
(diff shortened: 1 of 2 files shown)


tc-imba:
Looks great! And how about the "style" props? It seems that the Button style will be overwritten as well, is it meant to be customizable or not?

lukewalczak:
> Looks great! And how about the "style" props? It seems that the Button style will be overwritten as well, is it meant to be customizable or not?

I think there is no need for that - border radius is controlled by roundness from the theme, additionally there are dedicated props for handling colors. If I miss some example of styled button in accordance with MD, please let me know.

slavikdenis (approved):
Looks good!

Issue #3185, opened 2022-05-25
Adding `pressIn` and `pressOut` events to Button - how?


esbenvb:
What is the easiest way to make a button that looks/behaves like Paper's  `<Button>` but also supports `onPressIn` and `onPressOut` events?

frannale:
> What is the easiest way to make a button that looks/behaves like Paper's `<Button>` but also supports `onPressIn` and `onPressOut` events?

This is not a issue bro

lukewalczak:
It's not released yet (will be later today).

esbenvb:
Thanks - so which version will contain that feature?

lukewalczak:
Version `5.0.0-rc.4`

PR #3277, opened 2022-07-25
feat: extend Button props by adding onPressIn and onPressOut

lukewalczak:
Solves: https://github.com/callstack/react-native-paper/issues/3185

### Summary

PR extends `Button` component by adding  `onPressIn` and `onPressOut` props.
Refines button's elevation.

### Test plan

New props are covered by tests.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -86,6 +86,16 @@ type Props = React.ComponentProps<typeof Surface> & {
    */
   onPress?: () => void;
   /**
+   * @supported Available in v5.x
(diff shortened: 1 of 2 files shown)


slavikdenis (commented):
@lukewalczak What do you think about using `toHaveBeenCalledTimes(1)` instead of `toHaveBeenCalled()` to ensure, that the function doesn't get called multiple times?

lukewalczak:
@slavikdenis sure, that make sense

slavikdenis (approved):

PR #3283, opened 2022-07-30
feat: add SegmentedButtons component

okwasniewski:
This PR introduces new component named: `SegmentedButton`. 

### Summary

General overview of the component: https://m3.material.io/components/segmented-buttons/overview

Highlights:
- single or multi-selection
- segmented buttons can have 2-5 segments, segments cannot wrap into a new line
- icons can be used in place of labels
- icon next to the label is optional

<details>
<summary><h2> Screenshots </h2></summary>

### Version 3  
ios | android
--- | ---
Screenshot 2022-08-08 at 11 32 58 | Screenshot 2022-08-08 at 11 32 52
 

### Version 2
ios | android
--- | ---
Screenshot 2022-08-08 at 11 33 23 | Screenshot 2022-08-08 at 11 33 21
</details>

### Test plan

Write tests for this component.


src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,269 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Accessibility label for the `SegmentedButton`. This is read by the screen reader when the user taps the button.
+   */
+  accessibilityLabel?: string;
+  /**
+   * Function to execute on press.
+   */
+  onPress?: (value?: GestureResponderEvent | string) => void;
```
satya164, line 42: I think it's confusing that it can give `GestureResponderEvent` and `string`

satya164 (changes requested):

src/components/SegmentedButton/SegmentedButton.tsx
okwasniewski: You are right, I've removed the `string` type.

example/src/Examples/SegmentedButtonExample.tsx
```diff
@@ -0,0 +1,217 @@
...
+            value="transit"
+          />
+          <SegmentedButton
+            style={styles.button}
+            icon="car"
+            label="Driving"
+            value="drive"
+          />
+        </SegmentedButton.Group>
+      </List.Section>
```
Drakeoon, line 45: @okwasniewski Consider creating separate components for individual examples. It would be a little easier to digest if someone wants to look at source

okwasniewski: Done ✅

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,269 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Accessibility label for the `SegmentedButton`. This is read by the screen reader when the user taps the button.
+   */
+  accessibilityLabel?: string;
+  /**
+   * Function to execute on press.
+   */
+  onPress?: (value?: GestureResponderEvent) => void;
```
slavikdenis, line 42: `event?: GestureResponderEvent` would be more descriptive.
What do you think?

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,269 @@
...
+    throw new Error(
+      "<SegmentedButton/> can't be used without <SegmentedButton.Group/> wrapper."
+    );
+  }
+
+  const checkScale = React.useRef(new Animated.Value(0)).current;
+
+  const checked =
+    (context && Array.isArray(context.value)
+      ? context.value.includes(value || '')
```
slavikdenis, line 110: I'm wondering if this would be worthy of wrapping with `useMemo` for a case where there would be a bunch of Segmented Buttons

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,269 @@
...
+  };
+
+  const buttonStyle: ViewStyle = {
+    backgroundColor,
+    borderColor,
+    borderWidth,
+    borderRadius,
+    ...segmentBorderRadius,
+  };
+  const paddingVertical = density
```
slavikdenis, line 167: Shouldn't the condition here check for `isNumber(density)` instead of just truthy checking?

slavikdenis (changes requested):

src/components/SegmentedButton/SegmentedButton.tsx
okwasniewski: Density has default value as `0` so i think it's safe to do it like so

src/components/SegmentedButton/SegmentedButton.tsx
okwasniewski: You are right, I will fix it 👍🏻

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,269 @@
...
+   * @optional
+   */
+  theme: Theme;
+  /**
+   * testID to be used on tests.
+   */
+  testID?: string;
+};
+
+const SegmentedButton = ({
```
lukewalczak, line 82: Please add there a short component description and usage example, like you already did for `SegmentedButtonGroup`.

src/components/SegmentedButton/SegmentedButtonGroup.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+export const SegmentedButtonGroupContext =
+  React.createContext<SegmentedButtonContextType>(
+    {} as SegmentedButtonContextType
+  );
+
+/**
+ * Segmented button group allows to control a group of segmented buttons.</br>
+ *
+ * ## Usage
```
lukewalczak, line 50: Please adjust example usage similarly to the Appbar [example](https://callstack.github.io/react-native-paper/appbar.html) in the following way:

```
 ## Usage
 ### Single select
 ...
 ### Multiselect
 ```

File: src/components/SegmentedButton/SegmentedButton.tsx (hunk 4 above)
Comment (okwasniewski) [reply src/components/SegmentedButton/SegmentedButton.tsx:110]: I'm not sure but I think it wouldn't give any performance improvements in our case. It would still re-run the compute function (in `useMemo`) on every re-render because of context as a dependency.

File: src/components/SegmentedButton/SegmentedButton.tsx [hunk 8]
```diff
@@ -0,0 +1,277 @@
...
+          ? [...context.value.filter((val) => value !== val)]
+          : [...context.value, value]
+      );
+    } else {
+      context.onValueChange(!checked ? value : null);
+    }
+  };
+
+  return (
+    <View style={[buttonStyle, [styles.button, style]]}>
```
lukewalczak, line 211:
```suggestion
    <View style={[buttonStyle, styles.button, style]}>
```

satya164 (changes requested):
I'm not sure about the API of `SegementedButton` and `SegmentedButton.Group` given that `SegementedButton` cannot be used on its own similar to other components which have individual components as well as a group component.

It's more similar to the `FAB.Group` component. Probably it's better to have a single component that takes an array of items instead of having a group and children.

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,277 @@
...
+   */
+  onPress?: (event?: GestureResponderEvent) => void;
+  /**
+   * Value of button.
+   */
+  value?: string;
+  /**
+   * Label text of the button.
+   */
+  label?: string;
```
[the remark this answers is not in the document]
okwasniewski: You can use it only with icons

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,277 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Accessibility label for the `SegmentedButton`. This is read by the screen reader when the user taps the button.
+   */
+  accessibilityLabel?: string;
+  /**
+   * Function to execute on press.
+   */
+  onPress?: (event?: GestureResponderEvent) => void;
```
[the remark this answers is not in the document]
okwasniewski: Yes, you don't need to pass `onPress` as checking functionality is handled by context.

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  const { roundness, isV3 } = theme;
+  const { borderColor, textColor, borderWidth, backgroundColor } =
+    getSegmentedButtonColors({
+      checked,
+      theme,
+      disabled,
+    });
+
+  const borderRadius = (isV3 ? 5 : 1) * roundness;
```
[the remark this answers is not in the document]
okwasniewski: There checks are copied directly from other components. For example: https://github.com/callstack/react-native-paper/blob/b027079eed9d15d71d5fc2aa2399e98c0ac83cfc/src/components/Button/Button.tsx#L232

src/components/SegmentedButton/SegmentedButton.tsx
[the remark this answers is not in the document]
satya164: I'm talking about the event parameter

src/components/SegmentedButton/SegmentedButtonGroup.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export type Props = {
+  /**
+   * Function to execute on selection change.
+   */
+  onValueChange: (item: string | string[] | null) => void | null;
+  /**
+   * Value of the currently selected segmented button.
+   */
+  value: string | string[] | null;
```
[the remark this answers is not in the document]
okwasniewski: None, fixed it 👍🏻

src/components/SegmentedButton/SegmentedButton.tsx
```diff
@@ -0,0 +1,277 @@
...
+   */
+  multiselect?: boolean;
+  /**
+   * Status of button.
+   */
+  status?: 'checked' | 'unchecked';
+  /**
+   * Button segment.
+   */
+  segment?: 'first' | 'last';
```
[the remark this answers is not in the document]
okwasniewski: This depends on index inside map: 
https://github.com/callstack/react-native-paper/blob/659fb6187c2c947bf50a101d0e8700a9b2a3fa61/src/components/SegmentedButton/SegmentedButtonGroup.tsx#L120-L121

I'm not sure how it should be set using context

okwasniewski:
@satya164 Thanks for your feedback 👍 I have a few questions about what to do next with this component.

> I'm not sure about the API of `SegementedButton` and `SegmentedButton.Group` given that `SegementedButton` cannot be used on its own similar to other components which have individual components as well as a group component.
> 
> It's more similar to the `FAB.Group` component. Probably it's better to have a single component that takes an array of items instead of having a group and children.

You are right, this component is not consistent with others (like `ToggleButton` and `ToggleButton.Group`). There are two options: allow it to be used without `SegmentedButton.Group` or refactor the API to pass array of items. 

First option is pretty straight forward just remove the context check. This approach maybe useful for people who want to customize and implement their own logic for handling the checked state.

Second approach would make the component easier to use, however less customizable - I'm not sure which one is better.

Regarding the second approach implementation: 
Should I just change the `Children.map` inside `SegmentedButton.Group` to map an array of passed items? Then rename the `SegmentedButton.Group` to a `SegmentedButton` and remove the current `SegmentedButton` export (to use it only internally) or how you would go about it?  

cc: @lukewalczak @Drakeoon - What do you guys think?

lukewalczak:
I can agree with @satya164, that we can change the `API` since `SegmentedButton` isn't an individual component.
Imo I would prefer the approach with removing the `SegmentedButton.Group` and `context` check rather than passing an array with items.

okwasniewski:
After internal discussion we decided to refactor the component API to pass the items array. Using the `SegmentedButton` component without group doesn't make sense. Current `SegmentedButton` export will be removed and used only internally and `SegmentedButton.Group` will be renamed to `SegmentedButtons`.

lukewalczak (approved):
Awesome work 🎉 Thanks for cooperation and adjusting comments to the feedback!

src/components/SegmentedButtons/SegmentedButtons.tsx
```diff
@@ -0,0 +1,184 @@
...
+   */
+  buttons: {
+    value: string;
+    icon?: IconSource;
+    disabled?: boolean;
+    accessibilityLabel?: string;
+    onPress?: (event: GestureResponderEvent) => void;
+    label?: string;
+    showSelectedCheck?: boolean;
+    density?: 0 | -1 | -2 | -3;
```
satya164, line 64: I think this option is a bit confusing. As a user, looking at the negative numbers doesn't tell me anything. MD docs sometimes have such confusing APIs. I'd avoid using magic numbers and use a union type of strings instead.

I'm not sure about the names, but for some other components, we have a prop like `compact` etc.

Another thing is that it probably doesn't make sense to provide different densities for each button, so I'd move this prop to the `SegmentedButtons` component itself.

src/components/SegmentedButtons/SegmentedButtons.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+SegmentedButtons.displayName = 'SegmentedButtons';
+
+const styles = StyleSheet.create({
+  row: {
+    flexDirection: 'row',
+  },
+});
+
+export default withTheme(SegmentedButtons);
```
satya164, line 181: This should use `useTheme` hook instead.

Unrelated to this PR, but I see `useTheme` isn't getting `props.theme` in other places in the code - which makes the API inconsistent with components using `withTheme` that accept a `theme` prop to override `theme` locally. Should probably drop this feature altogether in next paper version or fix it for existing usage of `useTheme`.

src/components/__tests__/SegmentedButton.test.js
```diff
@@ -0,0 +1,281 @@
...
+            { disabled: true },
+            { disabled: false },
+            { disabled: true },
+          ],
+          index,
+        })
+      ).toMatchObject({ borderRightWidth: 1 });
+    });
+  });
+});
```
satya164, line 236: imo these tests are testing implementation details by checking for exact style properties. setting up some visual integration testing in the repo would be much better.

src/components/SegmentedButtons/SegmentedButtons.tsx
```diff
@@ -0,0 +1,184 @@
...
+              {...item}
+            />
+          );
+        })}
+      </View>
+    </SegmentedButtonGroupContext.Provider>
+  );
+};
+
+SegmentedButtons.displayName = 'SegmentedButtons';
```
satya164, line 173: This shouldn't be necessary. A component's name is automatically inferred from the function/class name. This is only needed in case the function is anonymous.

src/components/SegmentedButtons/SegmentedButtons.tsx
```diff
@@ -0,0 +1,184 @@
...
+  buttons,
+  multiSelect,
+  style,
+  theme,
+}: Props) => {
+  if (buttons.length < 2 || buttons.length > 5) {
+    console.warn(
+      'Segmented buttons are best used for selecting between 2 and 5 choices. If you have more than five choices, consider using another component, such as chips.'
+    );
+  }
```
satya164, line 139: Logging during render can result in a lot of duplicate warnings due to re-renders. Consider logging in an effect or using something like `warn-once` to show the warning only once.

Though I'm not sure such warning is needed here, considering we don't print such warnings anywhere else in the lib.

src/components/SegmentedButtons/SegmentedButtons.tsx
```diff
@@ -0,0 +1,184 @@
...
+};
+
+export const SegmentedButtonGroupContext =
+  React.createContext<SegmentedButtonContextType>({
+    value: '',
+    onValueChange: () => {},
+  });
+
+/**
+ * @supported Available in v5.x
```
satya164, line 88: Seems unnecessary to repeat that it's available for 5.x when I'm reading the docs for 5.x.

satya164 (changes requested):

src/components/SegmentedButtons/SegmentedButtons.tsx
lukewalczak: I can agree, it's a kinda _misuse_, however, let's keep it for now. Maybe it's useful for someone who comparing v4 and v5 🤔 
Anyway, since that label is used also in the other two components, if we decide to remove it, I'll handle it in a separate PR.

src/components/SegmentedButtons/SegmentedButtons.tsx
lukewalczak: We have `console.warn` in 2-3 other places. From the perspective of a developer not so familiar with MD, it's worth showing the warning, but as you suggested, let's introduce and use `warn-once` _(will handle it in other places in the separate PR then)_.

src/components/SegmentedButtons/SegmentedButtons.tsx
lukewalczak: Noted. Will contact @Drakeoon , since he's working on theme refactor.

src/components/SegmentedButtons/SegmentedButtons.tsx
lukewalczak: > I think this option is a bit confusing. As a user, looking at the negative numbers doesn't tell me anything. MD docs sometimes have such confusing APIs. I'd avoid using magic numbers and use a union type of strings instead.
> I'm not sure about the names, but for some other components, we have a prop like compact etc.

Maybe:

```
density?: 'regular' | 'small' | 'medium' | 'high'
```

> Another thing is that it probably doesn't make sense to provide different densities for each button, so I'd move this prop to the SegmentedButtons component itself.

Makes sense

src/components/SegmentedButtons/SegmentedButtons.tsx
satya164: @lukewalczak the other logs for invalid usage, this is more like a guideline so not really the same thing

src/components/__tests__/SegmentedButton.test.js
okwasniewski: This can be addressed in other PR

src/components/SegmentedButtons/SegmentedButtons.tsx
okwasniewski: You are right, I will remove this `console.warn` and the `warn-once` for other components can be addressed in separate PR

src/components/SegmentedButtons/SegmentedButtons.tsx
okwasniewski: I will change this to `useTheme()`

src/components/SegmentedButtons/SegmentedButtons.tsx
okwasniewski: I will leave it for now, it can be addressed in separate PR.

PR #3293, opened 2022-08-05
fix: white screen flickering in dark theme.

SectionTN:
### Summary

SafeAreaProvider (and GestureHandlerRootView) backgrounds are white as default, but to remove that white flickering (or flash) when navigating you must set it color to black.

### Test plan

yarn start && yarn android and test, this change is simple and it does not affect any files.

### DIFF
--- a/example/src/index.tsx
+++ b/example/src/index.tsx
@@ -154,7 +154,7 @@ export default function PaperExample() {
 
   return (
     <PaperProvider theme={theme}>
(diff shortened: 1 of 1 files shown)


lukewalczak:
The issue with flickering white screen was fixed in the following [commit](https://github.com/callstack/react-native-paper/commit/ea22a49a4de567913a9429db0d13474b278b9cc0)

Issue #3290, opened 2022-08-04
Not able to set custom BottomNavigation backgroundColor with v3 theme.


DomiR:
As you can see in the reference below, if the v3 theme is used it uses theme surface color no matter what backgroundColor one passes through the `barStyle` prop.

The [NavigationBar Guidelines](https://m3.material.io/components/navigation-bar/guidelines) says:
> The container can have a color fill to provide separation from the content behind it.

https://github.com/callstack/react-native-paper/blob/6afa4fb98c6826c4d6b80db055ef8f21ed07b59a/src/components/BottomNavigation/BottomNavigation.tsx#L602-L605

Also the elevation should probably be adjustable. According to the [NavigationBar Spec](https://m3.material.io/components/navigation-bar/specs#23b08792-3a52-484a-b76c-c37d0b81bb45) it should actually have a default elevation of 2 but it is currently hardcoded to 0:

https://github.com/callstack/react-native-paper/blob/6afa4fb98c6826c4d6b80db055ef8f21ed07b59a/src/components/BottomNavigation/BottomNavigation.tsx#L728

One can barely see it in the example picture:
<img width="387" alt="Bildschirmfoto 2022-08-04 um 12 37 25" src="https://user-images.githubusercontent.com/1834664/182827539-563edda8-4a5f-456b-b3ac-dd183afccb33.png">

lukewalczak:
Hey @DomiR, thanks for your feedback!

If you want to use a custom background color you can always override it by the `theme` prop:

```
      <BottomNavigation
		...
        theme={{ colors: { surface: 'aquamarine' } }}
      />
 ```
 
>  Also the elevation should probably be adjustable. According to the [NavigationBar Spec](https://m3.material.io/components/navigation-bar/specs#23b08792-3a52-484a-b76c-c37d0b81bb45) it should actually have a default elevation of 2 but it is currently hardcoded to 0:

In that case, the `elevation` prop applies background color and shadows, which in terms of `BottomNavigation` is hardcoded to `0`, since there should be no shadow:

<img width="878" alt="Zrzut ekranu 2022-08-4 o 12 55 18" src="https://user-images.githubusercontent.com/22746080/182830576-f07b3f4b-ebee-4131-9ec6-68ae615e038b.png">

However, based on that information you provided and figma kit, the default background color should be `theme.colors.elevation.level2`, which I will correct in the next week.

DomiR:
Hey @lukewalczak. Thanks for looking into this 😊

PR #3298, opened 2022-08-08
fix: correct BottomNavigation background color

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3290

### Summary

PR fixes background color, which is now set accordingly to the docs and the value is `theme.colors.elevation.level2`.
Also it unblocks a possibility to override background color by `barStyle` prop.

* default theme colors

ios | android
--- | ---
Zrzut ekranu 2022-08-8 o 10 34 58 | Zrzut ekranu 2022-08-8 o 10 35 20

* custom colors 

ios | android
--- | ---
Zrzut ekranu 2022-08-8 o 10 35 55 | Zrzut ekranu 2022-08-8 o 10 35 57

### Test plan

Added tests.

### DIFF
--- a/src/components/BottomNavigation/BottomNavigation.tsx
+++ b/src/components/BottomNavigation/BottomNavigation.tsx
@@ -255,6 +255,10 @@ type Props = {
    * @optional
    */
   theme: Theme;
+  /**
+   * TestID used for testing purposes
+   */
+  testID?: string;
 };
 
 const MIN_RIPPLE_SCALE = 0.001; // Minimum scale is not 0 due to bug with animation
@@ -374,6 +378,7 @@ const BottomNavigation = ({
   safeAreaInsets,
   labelMaxFontSizeMultiplier = 1,
   compact = !theme.isV3,
+  testID = 'bottom-navigation',
 }: Props) => {
   const { scale } = theme.animation;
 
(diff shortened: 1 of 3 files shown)


src/components/BottomNavigation/BottomNavigation.tsx
```diff
@@ -758,7 +771,10 @@ const BottomNavigation = ({
         }
         onLayout={onLayout}
       >
-        <Animated.View style={[styles.barContent, { backgroundColor }]}>
+        <Animated.View
+          style={[styles.barContent, { backgroundColor }]}
+          testID={`${testID}-bar-content`}
```
slavikdenis, line 776: This will fail, in case user provides `testID={undefined}` or `testID={null}`.
Test IDs should be extended when the extended value is truthy.
What do you think?

slavikdenis (changes requested):

src/components/BottomNavigation/BottomNavigation.tsx
lukewalczak: Well, in terms of `undefined` it will pass since there is a default value for `testID`. When it comes to the `null`, I don't think we should not add a check for `testID` value – I believe it's up to the user to pass the correct value or not at all.

slavikdenis (approved):

PR #3300, opened 2022-08-08
chore: update react-native-testing-library to v11

lukewalczak:
### Summary

PR updates testing library `@testing-library/react-native` (renamed in the meantime from `react-native-testing-library`) to v11.

### Test plan

CI should pass.


thymikee (approved):
Looks good 👍🏼

mdjastrzebski:
I've got too late for review, just wanted to leave small hint about RNTL assertions:

**Assertion that element is present (can be any predicate, byText, byLabelText, etc):**
```ts
// Don't do this
expect(() => getByLabelText('page-first')).not.toThrow();

// Do this
expect(getByLabelText('page-first')).toBeTruthy();
```

**Assertion that element is not present (can be any predicate, byText, byLabelText, etc):**
```ts
// Don't do this
expect(() => getByLabelText('Options Select')).toThrow();

// Do this
expect(queryByLabelText('Options Select')).ToBeFalsy();
```

PR #3311, opened 2022-08-12
chore: testing library adjustments

lukewalczak:
### Summary

PR introduces some adjustments according to the [comment](https://github.com/callstack/react-native-paper/pull/3300#issuecomment-1212871400) from the previously merged PR.

### Test plan

Test and CI should pass.

### DIFF
--- a/src/components/__tests__/BottomNavigation.test.js
+++ b/src/components/__tests__/BottomNavigation.test.js
@@ -290,15 +290,10 @@ it('should have appropriate display style according to the visibility', () => {
 
   const wrapper = getByTestId('RouteScreen: 0');
 
-  expect(wrapper.props.style).toEqual(
-    expect.arrayContaining([expect.objectContaining({ display: 'flex' })])
-  );
+  expect(wrapper).toHaveStyle({ display: 'flex' });
 
   rerender(<BottomNavigationRouteScreen visibility={0} index={0} />);
-
-  expect(wrapper.props.style).toEqual(
-    expect.arrayContaining([expect.objectContaining({ display: 'none' })])
-  );
+  expect(wrapper).toHaveStyle({ display: 'none' });
 });
 
 it('should have labelMaxFontSizeMultiplier passed to label', () => {
@@ -333,7 +328,5 @@ it('renders custom background color passed to barStyle property', () => {
   );
 
   const wrapper = getByTestId('bottom-navigation-bar-content');
-  expect(wrapper.props.style).toEqual(
-    expect.objectContaining({ backgroundColor: red300 })
-  );
+  expect(wrapper).toHaveStyle({ backgroundColor: red300 });
 });
--- a/src/components/__tests__/BottomNavigation.test.js
+++ b/src/components/__tests__/BottomNavigation.test.js
@@ -334,8 +334,6 @@ it('renders custom background color passed to barStyle property', () => {
 
   const wrapper = getByTestId('bottom-navigation-bar-content');
   expect(wrapper.props.style).toEqual(
-    expect.arrayContaining([
-      expect.objectContaining({ backgroundColor: red300 }),
-    ])
+    expect.objectContaining({ backgroundColor: red300 })
   );
 });
--- a/src/components/__tests__/Button.test.js
+++ b/src/components/__tests__/Button.test.js
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { StyleSheet } from 'react-native';
 import renderer from 'react-test-renderer';
-import { fireEvent, render } from 'react-native-testing-library';
+import { fireEvent, render } from '@testing-library/react-native';
 import color from 'color';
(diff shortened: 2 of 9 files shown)


src/components/__tests__/Button.test.js
```diff
@@ -128,19 +128,21 @@ it('renders button with an accessibility hint', () => {
 
 it('should execute onPressIn', () => {
   const onPressInMock = jest.fn();
+  const onPress = jest.fn();
```
mdjastrzebski, line 131: Wy adding `onPress` if not used in test? Same question for `onPressOut` test

src/components/__tests__/Card/Card.test.js
```diff
@@ -34,15 +34,15 @@ describe('CardActions', () => {
...
-            Agree
-          </Button>
+        <Card.Actions testID="card-actions">
+          <Button mode="contained">Agree</Button>
         </Card.Actions>
       </Card>
     );
 
-    expect(getByTestId('card-action-button').props.mode).toBe('contained');
+    expect(getByTestId('card-actions').props.children[0].props.mode).toBe(
```
mdjastrzebski, line 43: Try to avoid nesting props checks like this. Not sure what that tests tries to achieve, as it looks you are checking `contained` prop on a composite element, not a host element. You probably would want to check some styles applied on button, but checking that will depend on your element structure. RNTL is not really suited for such "visual" checks, we rather excel as check business logic from users perspective.

src/components/__tests__/Modal.test.js
```diff
@@ -17,9 +17,7 @@ describe('Modal', () => {
     );
 
     expect(getByTestId('modal-backdrop').props.style).toEqual(
```
mdjastrzebski, line 19: Rather use custom matcher from `@testing-library/jest-native` for such cases.

```suggestion
    expect(getByTestId('modal-backdrop')).toHaveStyle({ backgroundColor: 'transparent' });
```

src/components/__tests__/BottomNavigation.test.js
```diff
@@ -334,8 +334,6 @@ it('renders custom background color passed to barStyle property', () => {
 
   const wrapper = getByTestId('bottom-navigation-bar-content');
   expect(wrapper.props.style).toEqual(
```
mdjastrzebski, line 336: Custom matcher from `@testing-library/jest-native`:
```suggestion
  expect(wrapper).toHaveStyle({ backgroundColor: red300 });
```

mdjastrzebski (commented):
Overall looks good, I've added some suggestions.

Two general suggestions:
1) Try to use custom jest matchers from `@testing-library/jest-native` these should allow you to make more high-level assertions
2) Gradually migrate to use `screen` object to make queries: 

```ts
import { render, screen } from '@testing-library/react-native';
import { QuestionsBoard } from '../QuestionsBoard';

test('should verify two questions', () => {
  render(<QuestionsBoard {...props} />);
  const allQuestions = screen.queryAllByRole('header');
  expect(allQuestions).toHaveLength(2);
});
```

src/components/__tests__/Button.test.js
lukewalczak: That's something I've discovered recently, so basically `Button` component is based on `TouchableRipple` which is [disabled](https://github.com/callstack/react-native-paper/blob/main/src/components/TouchableRipple/TouchableRipple.native.tsx#L43) when `onPress` is not passed. I'm going to refactor it, however the current logic force me to add `onPress`

src/components/__tests__/Card/Card.test.js
lukewalczak: To justify myself here, I was on the edge between the approach which I've implemented and testing the styles which are the result of passing `contained` mode. Finally, I've decided to do the nesting props, to confirm that `child` (in that case `Button`) is receiving the `mode` property, according to the code from `CardActions`:

```
      {React.Children.map(props.children, (child, i) => {
        return React.isValidElement(child)
          ? React.cloneElement(child, {
              compact: !isV3 && child.props.compact !== false,
              mode:
                child.props.mode ||
                (isV3 && (i === 0 ? 'outlined' : 'contained')),
              style: isV3 && styles.button,
            })
          : child;
      })}
  ```

lukewalczak:
> Two general suggestions:
> 1) Try to use custom jest matchers from `@testing-library/jest-native` these should allow you to make more high-level assertions
> 2) Gradually migrate to use `screen` object to make queries: 

Thanks for the feedback, both of them make sense. I will iterate on it in the separated PRs.

src/components/__tests__/Card/Card.test.js
mdjastrzebski: That's just a suggestion no need to justify :-) I think you should rather check some styles that will be applied to the button element using `toHaveStyle`, but you may find the test as is work for your case.

Issue #3272, opened 2022-07-23
KeyboardAvoidingView not working with header and material-bottom-tabs on iOS


raajnadar:
Reported on react-navigation https://github.com/react-navigation/react-navigation/issues/10583

OP solved it by switching to bottom tabs so it seems the issue lies in the material bottom tabs from paper https://stackoverflow.com/a/72561415/5519872

### Current behaviour

KeyboardAvoidingView does not avoid the keyboard correctly, if you use a header and material-bottom-tabs on iOS. For Android, everything works out of the box.

### Expected behaviour

It should obviously avoid the keyboard.

### Code sample

https://github.com/tomwaitforitmy/keyboardAvoidingViewAndNavigation

https://snack.expo.dev/@tomwaitforitmy/keyboardavoidingview

### Your Environment

| software              | version
| --------------------- | -------
| ios        |  N/A
| react-native          | 0.68.1
| react-native-paper    | 4.12.3
| node                  | 16.13.1
| npm or yarn           | 1.22.19
| expo sdk              | 45.0.0

tomwaitforitmy:
>     react-native (found: 0.68.1, latest: 0.69.2) 
> Can you verify that the issue still exists after upgrading to the latest versions of these packages?

According to [documention](https://docs.expo.dev/versions/latest/) Expo SDK would only be compatible with react-native 0.68.2

raajnadar:
It is a bot you can ignore that, next Expo SDK will support RN 0.69.3 which is the latest version, It will be available today or tomorrow

lukewalczak:
Fixed via: https://github.com/callstack/react-native-paper/pull/3316. Will be released in `5.0.0-rc.5`

tomwaitforitmy:
@lukewalczak I recently updated my minimal not working example to react-native-paper 5.2.0. See here:
https://snack.expo.dev/@tomwaitforitmy/keyboardavoidingview

The bug is still there. Did I do anything wrong or did you not fix it?

PR #3316, opened 2022-08-14
Fix KeyboardAvoidingView not working with BottomNavigation

okwasniewski:
### Summary

Fixes issue: #3272 

This was introduced here: https://github.com/callstack/react-native-paper/pull/680 and it seems like it was solving only android bug, so it should be safe to disable it on iOS.

Here is a snack with the issue: https://snack.expo.dev/@tomwaitforitmy/keyboardavoidingview

https://user-images.githubusercontent.com/52801365/184548970-6345eae2-82ad-4775-a471-f37e8c81fedf.mp4

### Test plan

Check if this fixes the repro.

### DIFF
--- a/src/components/BottomNavigation/BottomNavigation.tsx
+++ b/src/components/BottomNavigation/BottomNavigation.tsx
@@ -364,7 +364,7 @@ const BottomNavigation = ({
   getTestID = ({ route }: { route: Route }) => route.testID,
   activeColor,
   inactiveColor,
-  keyboardHidesNavigationBar = true,
+  keyboardHidesNavigationBar = Platform.OS === 'android',
   barStyle,
   labeled = true,
   style,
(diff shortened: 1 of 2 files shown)


satya164 (changes requested):
> it seems like it was solving only android bug

It's not solving a bug. The screen resizes by default on Android - where this feature is useful. The same resizing effect could be achieved on iOS as well (e.g. with `KeyboardAvoidingView` or keyboard listeners).

This change isn't correct since not making the bar absolute-positioned will leave a gap at the bottom when the navigation bar is hidden as per the comment - which is not apparent with default iOS behavior but can be visible if the screen is resized.

Instead of the platform check for absolutely positioning the tab bar, you could:

- Default `keyboardHidesNavigationBar` to `false` similar to `@react-navigation/bottom-tabs`
- Default `keyboardHidesNavigationBar` to `false` on iOS & `true` on Android

satya164 (approved):

PR #3321, opened 2022-08-16
feat: add icon prop in RadioButton.IOS

obdulia-losantos:
### Summary

Add `icon` property to the `RadioButton.IOS` component to display a custom icon whether it's checked.

```javascript
<RadioButton.IOS
  value="custom-ios"
  color={isV3 ? MD3Colors.error70 : MD2Colors.blue500}
  status={checked === 'custom-ios' ? 'checked' : 'unchecked'}
  icon="check-underline-circle"
/>
```

<table><tr><td>

</td><td>

</td></tr></table>

### Test plan

Adapted existing test case with new property.

### DIFF
--- a/example/src/Examples/RadioButtonExample.tsx
+++ b/example/src/Examples/RadioButtonExample.tsx
@@ -11,7 +11,7 @@ import {
 } from 'react-native-paper';
 import ScreenWrapper from '../ScreenWrapper';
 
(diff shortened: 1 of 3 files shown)


lukewalczak:
Hey @obdulia-losantos, unfortunately, we cannot merge the PR at the moment, since the `icon` in `RadioButton` is fixed and shouldn't be replaced with another one.

Issue #3225, opened 2022-06-26
[Feature Request] `SearchBar` loading property


fabriziocucci:
**Is your feature request related to a problem?**
No, but it would be nice if the `SearchBar` component could expose a `loading` property.

**Describe the solution you'd like**
Ideally, when `loading={true}`, an `ActivityIndicator` could be shown in place of the clear icon (like Google Maps does).

**Describe alternatives you've considered**
N/A

**Additional context**
Screenshot from Google Maps below:

![image](https://user-images.githubusercontent.com/8156463/175827667-547d64d6-b554-4800-a870-9d6a77d7574a.png)

lukewalczak:
Hey @fabriziocucci, I think that makes sense. Would you like to add the `loading` property to the `Searchbar` _(e.g. similarly to the one existing in `Button`)_ covered by the unit test?

gustavohbeg:
I installed the package in the most up-to-date version but I'm still seeing the old material design design, how do I see the updated material design you ?

lukewalczak:
Solved by: https://github.com/callstack/react-native-paper/pull/3322

PR #3322, opened 2022-08-17
feat: added loading icon in Searchbar

dakshbhardwaj:
### Summary
Added activity indicator in the searchBar. added two props `loading` and `loaderStyle` props for customizing activity indicator.

#3225 

### Test plan

<image src= "https://user-images.githubusercontent.com/22423684/185056075-25fa0dcd-f719-46de-b090-d62bfda2df47.png" height="600" width="300"/>

### DIFF
--- a/src/components/Searchbar.tsx
+++ b/src/components/Searchbar.tsx
@@ -18,6 +18,9 @@ import { withTheme } from '../core/theming';
 import type { IconSource } from './Icon';
 import type { Theme } from '../types';
 import MaterialCommunityIcon from './MaterialCommunityIcon';
+import ActivityIndicator, {
+  Props as ActivityIndicatorProps,
+} from './ActivityIndicator';
 
 export type Props = React.ComponentPropsWithRef<typeof TextInput> & {
   /**
@@ -70,6 +73,14 @@ export type Props = React.ComponentPropsWithRef<typeof TextInput> & {
    * Custom icon for clear button, default will be icon close
    */
   clearIcon?: IconSource;
+  /**
+   * Custom flag for replacing clear button with activity indicator.
+   */
+  loading?: Boolean;
+  /**
+   * Custom styles for the activity indicator.
+   */
+  loaderStyle?: Omit<ActivityIndicatorProps, 'theme'>;
 };
 
 type TextInputHandles = Pick<
@@ -122,6 +133,8 @@ const Searchbar = React.forwardRef<TextInputHandles, Props>(
       style,
       theme,
       value,
+      loading = false,
+      loaderStyle,
       ...rest
     }: Props,
     ref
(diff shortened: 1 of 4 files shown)


src/components/Searchbar.tsx
```diff
@@ -218,26 +231,30 @@ const Searchbar = React.forwardRef<TextInputHandles, Props>(
...
-                name="close"
-                color={color}
-                size={size}
-                direction={I18nManager.isRTL ? 'rtl' : 'ltr'}
-              />
-            ))
-          }
-          accessibilityRole="button"
-        />
+        {loading ? (
```
lukewalczak, line 234: Please cover new functionality with unit tests.

dakshbhardwaj: @lukewalczak i have added the test cases.

lukewalczak: Thanks for adding a snapshot test, but It would be good to add also unit test case, confirming rendering `ActivityIndicator` when the `loading` prop is passed.

dakshbhardwaj: @lukewalczak added the unit test cases when loading prop is passed.

src/components/Searchbar.tsx
```diff
@@ -218,26 +231,30 @@ const Searchbar = React.forwardRef<TextInputHandles, Props>(
...
+          <ActivityIndicator {...loaderStyle} style={styles.loader} />
```
lukewalczak, line 235: The approach of adding spread props in from of `loaderStyle` to the loading component is inconsistent with other similar examples, hence please remove it.

I can see that you passed `testID` to the `loaderStyle` for testing. Since `loaderStyle` should be removed, please add a `testID` prop to the `ActivityIndicator` with the default value `activity-indicator` and update your test case.

dakshbhardwaj: Hi @lukewalczak , Thanks for reviewing, I have one question if we remove `loaderStyle` how can the user customize the activity indicator style? 

or can we do something like this, accept `loadingIcon` directly as props for customization?

 what do you suggest?

lukewalczak: I have the feeling like that component doesn't require additional customization. I was looking for some MD examples related to that functionality, however, didn't find any. I would suggest to keep as not customizable.

src/components/Searchbar.tsx
```diff
@@ -70,6 +73,14 @@ export type Props = React.ComponentPropsWithRef<typeof TextInput> & {
    * Custom icon for clear button, default will be icon close
    */
   clearIcon?: IconSource;
+  /**
+   * Custom flag for replacing clear button with activity indicator.
```
lukewalczak, line 77:
```suggestion
	 * @supported Available in v5.x
   * Custom flag for replacing clear button with activity indicator.
```

dakshbhardwaj: added the line

src/components/Searchbar.tsx
dakshbhardwaj: @lukewalczak I have made the changes. Please review it

lukewalczak (approved):
Thanks @dakshbhardwaj for your contribution! 🎊

Issue #3303, opened 2022-08-09
TouchableRipple onLongPress doesn't work


hbriese:
### Current behaviour

`onLongPress` doesn't work when passed as a prop to `TouchableRipple`

### Expected behaviour

`onLongPress` is executed, as specified by your documentation

### Code sample

The snack is for RNP v4, but this issue also occurs on v5
https://snack.expo.dev/@haydenbriese/touchableripple---onlongpress

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        |
| react-native          | 0.68.2
| react-native-paper    | 4.12.4
| node                  |
| npm or yarn           |
| expo sdk              | 45.0.6

lukewalczak:
As far as I know, there is a need to pass also `onPress` to make `onLongPress` work properly.

hbriese:
@lukewalczak Thank you, I will use that in the meantime.

lukewalczak:
I think that it shouldn't work in that way, so will investigate the problem.

PR #3328, opened 2022-08-22
fix: fixed onLongPress not working on touchableRipple

dakshbhardwaj:
### Summary
if we don't pass the `onPress` props then `onLongPress` was also not working in the `TouchableRipple` component.

Fixes: #3303 

### Test plan
I checked it by running the `yarn run test`, this change is simple and does not affect any files.

### DIFF
--- a/src/components/TouchableRipple/TouchableRipple.tsx
+++ b/src/components/TouchableRipple/TouchableRipple.tsx
@@ -229,7 +229,7 @@ const TouchableRipple = ({
     });
   };
 
-  const disabled = disabledProp || !rest.onPress;
+  const disabled = disabledProp || (!rest.onPress && !rest.onLongPress);
 
   return (
     <TouchableWithoutFeedback
--- a/src/components/TouchableRipple/TouchableRipple.native.tsx
+++ b/src/components/TouchableRipple/TouchableRipple.native.tsx
@@ -40,7 +40,7 @@ const TouchableRipple = ({
   theme,
   ...rest
 }: Props) => {
-  const disabled = disabledProp || !rest.onPress;
+  const disabled = disabledProp || (!rest.onPress && !rest.onLongPress);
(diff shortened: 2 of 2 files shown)


src/components/TouchableRipple/TouchableRipple.tsx
```diff
@@ -229,7 +229,7 @@ const TouchableRipple = ({
     });
   };
 
-  const disabled = disabledProp || !rest.onPress;
+  const disabled = disabledProp || (!rest.onPress && !rest.onLongPress);
```
lukewalczak, line 232: `TouchableRipple` can accept `onPressIn` and `onPressOut` as well, which won't work also without `onPress` or `onLongPress`. I'm wondering if there's a nicer solution to fix that issue

dakshbhardwaj: @lukewalczak if we don't pass touch events like `onPress` `onLongPress` `onPressIn` `onPressOut` do we need to disable the touch events?

lukewalczak: I have to investigate that problem, why the `disabled` prop is set in that way. Will let you know

dakshbhardwaj: @lukewalczak did you get any chance to go through it?

lukewalczak: I think it's a quite complicated case. I guess touchable was initially disabled when `onPress` callback is not defined, to avoid running custom functionality handled by `onPressIn`e.g. in `Button` [component](https://github.com/callstack/react-native-paper/blob/main/src/components/Button/Button.tsx#L208). Probably the same applies to the `Chip` component.
I will try to find some middle ground to cover all the cases related to that issue.

lukewalczak:
Closing in favor of: https://github.com/callstack/react-native-paper/pull/3803

Issue #3132, opened 2022-03-29
Incorrect textinput label position in arabic locale


nevskij2022:
### Current behaviour
We use the latest versions of libraries.
When the localization is changed to Arabic, an indent appears for the TextInput label.
![image](https://user-images.githubusercontent.com/102587316/160619691-c04cd247-3078-4631-8250-b32d5eb540e9.png)

### Expected behaviour

Should be the same behavior as with in English localization

![image](https://user-images.githubusercontent.com/102587316/160619622-304458d5-d433-4202-9312-a31506112f41.png)
### Code sample 
[Code sample](https://snack.expo.dev/@nevskij2022/label-react-native-paper)
### Screenshots (if applicable)
![Анимация4](https://user-images.githubusercontent.com/102587316/160816006-dbd0b44a-b475-4b54-a2f6-8b99add2da54.gif)

### What have you tried
These changes in the 'TechInputFlat.tsx' component solve the problem
![image](https://user-images.githubusercontent.com/102587316/160603268-41b895b9-a19a-4a95-9cf9-65eb6e00cbdf.png)

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | android
| react-native          | 0.67.4
| react-native-paper    | 4.11.2
| node                  | 16.14.0
| npm or yarn           | npm 8.5.5
| expo sdk              |

lukewalczak:
Hey @nevskij2022, please check if the issue exists on the latest version.

PR #3330, opened 2022-08-22
fix: correct padding offset in TextInput on Android when RTL

lukewalczak:
### Summary

Fixes: https://github.com/callstack/react-native-paper/issues/3132

PR corrects padding passed to the `InputLabel` within `TextInput` on Android in the RTL mode.

before | after
--- | ---
Zrzut ekranu 2022-08-22 o 14 29 16 | Zrzut ekranu 2022-08-22 o 14 28 15

### Test plan

Added tests.

### DIFF
--- a/src/components/__tests__/TextInput.test.js
+++ b/src/components/__tests__/TextInput.test.js
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import color from 'color';
-import { StyleSheet, Text, Platform } from 'react-native';
+import { StyleSheet, Text, Platform, I18nManager } from 'react-native';
 import { fireEvent, render } from '@testing-library/react-native';
 import TextInput from '../TextInput/TextInput';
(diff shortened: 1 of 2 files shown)


MrRefactor (approved):

src/components/__tests__/TextInput.test.js
```diff
@@ -218,6 +218,54 @@ it('renders input placeholder initially with an empty space character', () => {
   expect(getByTestId('text-input-flat').props.placeholder).toBe(' ');
 });
 
+it('correctly applies padding offset to input label on Android when LTR', () => {
+  Platform.OS = 'android';
+  I18nManager.isRTL = false;
```
Drakeoon, line 223: Doesn't this change leak to the rest of the tests?

lukewalczak: Pushed the change with resetting the value after test.

Issue #3138, opened 2022-04-02
TextInput's placeholder is not being displayed initially with multiline set to true on iOS


dylanrandle:
### Current behaviour

On iOS, if multiline is true and label and placeholder both have a value, the placeholder is not displayed when you click on the input field. If you type something in, selected it all, and delete it, then the placeholder is displayed and works as expected. If you enter two lines of text into the text input and then erase them, then you will see the placeholder. See the video for an example of this behavior.

![image](https://user-images.githubusercontent.com/14814084/161403967-f177c49c-b57f-4502-bb30-b8680edaea6b.png)

### Expected behaviour

The placeholder should show up.

### Code sample

```
<TextInput
    label="Label"
    placeholder="Placeholder"
    multiline
 />
```

### Screenshots (if applicable)

![image](https://user-images.githubusercontent.com/14814084/161403964-eac5a336-3564-407d-b2a6-3054abd7324a.png)

### What have you tried

This is still happening even though #1483 was closed.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | ios
| react-native          | 0.64.3
| react-native-paper    | 4.11.1
| node                  | 16
| npm or yarn           | yarn
| expo sdk              | 44.0

alexpchin:
Any update on this issue? Thanks

mehdinourollah:
` "react-native-paper": "^4.12.3"`
It's still open ...and for the love of god someone answer this issue without closing it automatically !

raajnadar:
To keep the issue open a repro is required or else the bot will close it!!

If you still have the issue create a new one with a repro!!

lukewalczak:
I will do my best to investigate that issue.

lukewalczak:
The first investigation attempt presented that I was able to reproduce the problem on the fresh `TextInput` coming from `react-native`. Please check the following [snack](https://snack.expo.dev/ilKMCyNa8) with simplified code, based on the input code from the library.
It looks like there is an issue related to setting placeholder value coming from props in correlation with `multiline` prop, no matter if wrapped in timeout.

lukewalczak:
Created a workaround: https://github.com/callstack/react-native-paper/pull/3331

icgarlin:
> Created a workaround: #3331

Workaround doesn't work at all

Issue #1483, opened 2019-11-19
TextInput's placeholder is not being displayed initially with multiline set to true on iOS


ImTheC:
### Current behaviour

On iOS, when using <TextInput/>, if multiline is true and label and placeholder both have a value, the placeholder is not displayed when you click on the input filed. If you type something in, selected it all, and delete it, then the placeholder is displayed and works as expected.

### Expected behaviour

The placeholder should be displayed after clicking on the input field.

### Code sample

https://snack.expo.io/S1_Ags-nB

### Screenshots (if applicable)

N/A

### What have you tried

At first I was passing in the values with props, so I tried setting them directly as strings, but it didn't seem to matter either way.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | ios
| react-native          | 0.60.5
| react-native-paper    |  3.1.1
| node                  | 11.15.0
| npm or yarn           | npm
| expo sdk              | 35.0.0

Trancever:
@ImTheC Can you check if it still occurs in paper v3.3.0?

kotavy:
@Trancever This behavior still exists as of v3.5.1, running the [TextInput example](https://callstack.github.io/react-native-paper/text-input.html#placeholder) on Snack and only adding multiline and a placeholder string. Placeholder works as expected without multiline.

jeffjvick:
We are experiencing the same issue on iOS.  Paper v 3.6.0

If you enter two lines of text into the text input and then erase them, then you will see the placeholder.  See the video for an example of this behavior.

![RNP_multiline_textinput_placeholder](https://user-images.githubusercontent.com/203289/75285085-cfd3dc80-57ca-11ea-985a-e7b7e559a124.gif)

ovbm:
Having this same issue. Was anyone able to resolve it?

kotavy:
@ovbm My team has not prioritized finding a workaround, but I can confirm the issue remains. I tested again using the example Snack, running 3.10.1.

Biciato:
Still present

JulienRst:
We might want to re-open this issues as it still occurs ?

enagorny:
This is still the issue in latest version

yhbyOff:
+1

blixit:
Still having this issue

angel-venchev-toptal:
I am having the same issue

HassanHaiderIX:
in IOS build I am facing same issue Placeholder is missing, can some one help me?

[...]

PR #3331, opened 2022-08-22
fix: adjust initial placeholder text value

lukewalczak:
### Summary

Solves: https://github.com/callstack/react-native-paper/issues/3138 https://github.com/callstack/react-native-paper/issues/1483

PR introduces a workaround to fix the problem with multiline iOS text input.

**Description**: Changing the placeholder text value dynamically, within multiline input on iOS, doesn't work properly – the placeholder is not displayed initially.

**Root cause**:   Placeholder initial value, which has length 0.

**More context:**  The issue was also reproduced in react-native, using its own TextInput, there is a snack: https://snack.expo.dev/ilKMCyNa8?platform=ios

**Workaround**:   Set an empty space character in the default value.

### Test plan

* Tested manually on ios, text input placeholder is displayed.
* Added test case and updated snapshots.


raajnadar:
This is a nice trick 👌, great job 👏

PR #3340, opened 2022-08-29
fix: adjust avatar text line height to font scale

rbcjacques:
### Summary

Similarly to #3264 , this PR adjusts the `Avatar.Text` component's text line height to the font scale. Fixes the following bug in `Avatar.Text`:

Screen Shot 2022-08-29 at 7 34 51 AM

### Test plan

Updated test snapshots.


lukewalczak (approved):
Thanks @rbcjacques for your contribution!

PR #3351, opened 2022-09-05
fix: using theme in Text's components, renaming typescales to fonts

lukewalczak:
### Summary

Fixes: https://github.com/callstack/react-native-paper/issues/3305 https://github.com/callstack/react-native-paper/issues/3369

PR:

* renames `typescale` to `fonts` to avoid confusion between them in version 5
* correct `Text` component to accept `fontFamily`
* enable the possibility to use theme fonts in components using `Text` with variant, such as: 
	* `AppbarContent`
	* `BottomNavigation`

	* `Chip`
	* `DialogTitle`
	* `DrawerCollapsedItem`
	* `FABGroup`
	* `MenuItem`
	* `SegmentedButtonItem`
	* `TextInput`
	

### Test plan

* added tests for `configureFonts`
* update snapshots

PR #3378, opened 2022-09-20
fix Text fontFamily (Material 3)

hosseinmd:
### Summary
Text doesn't have default font family

In this PR text have bodyMedium as default


lukewalczak:
Thanks @hosseinmd , however It will be tackled in the already opened PR related to using theme in `Text`s component: https://github.com/callstack/react-native-paper/pull/3351, hence closing it.

Issue #3387, opened 2022-09-27
Ripple does not respect Border Radius on buttons in 5.0.0-rc.6


NathanC:
### Current behaviour

In 5.0.0-rc.6, buttons are rounded by default. I don't like the rounded look personally, so tried to set the border radius to 0 (or low at least).

The button border radius changes (for example, for a contained button), but **the ripple is still rounded, looking strange when pressed**

This is the inverse of an old bug: https://github.com/callstack/react-native-paper/issues/1342

### Expected behaviour
The ripple on the button follows the `borderRadius` style

### Reproduction

https://snack.expo.dev/A4FEbJrFh

### Your Environment

| software                      | version
| --------------------- | -------
| android                        | 13
| react-native                | 0.69.5
| react-native-paper    | 5.0.0-rc.6
| expo sdk                     | 46.0.9

NathanC:
Thanks @dakshbhardwaj for the quick fix! Always fun to see nullish coalescing in the wild.

dakshbhardwaj:
> Thanks @dakshbhardwaj for the quick fix! Always fun to see nullish coalescing in the wild.

Thank you @NathanC

PR #3388, opened 2022-09-28
fix: button ripple radius

dakshbhardwaj:
### Summary
Fix: #3387 

In the case when the` borderRadius` was `0` the condition was treating it as a falsy value.
Replaced it  with a check of `Nullish coalescing operator` so that it check only for null and undefined values

### Test plan
Checked manually by applying a style with borderRadius : 0 in the button component


lukewalczak:
Thanks @dakshbhardwaj for fixing the bug! 👏🏽

lukewalczak (approved):

PR #3398, opened 2022-10-03
feat: add createDynamicThemeColors utility

lukewalczak:
### Summary

PR introduces a `createDynamicThemeColors` utility function that allows for generating two color schemes `lightScheme` and `darkScheme`, based on the provided source color.

#### Example

```js
import {
  MD3LightTheme as DefaultTheme,
  Provider as PaperProvider,
} from 'react-native-paper';
import App from './src/App';
const { lightScheme } = createDynamicThemeColors({ sourceColor: '#FFFF00' });
const theme = {
  ...DefaultTheme,
  colors: lightScheme,
};
export default function Main() {
  return (
    <PaperProvider theme={theme}>
      <App />
    </PaperProvider>
  );
}
```

### Test plan

New functionality covered by unit tests.


src/core/__tests__/theming.test.js
```diff
@@ -0,0 +1,183 @@
...
+
+  [('darkScheme', 'lightScheme')].forEach((mode) => {
+    it(`should contain correct elevation colors for ${mode}`, () => {
+      const dynamicThemeColors = createDynamicThemeColors({
+        sourceColor,
+      });
+
+      const elevations = ['transparent', 0.05, 0.08, 0.11, 0.12, 0.14];
+      const elevation = elevations.reduce(
+        (a, v, index) => ({
```
jaworek, line 130: It would be good to have more descriptive variable names.

src/core/__tests__/theming.test.js
```diff
@@ -0,0 +1,183 @@
...
+  });
+
+  [('darkScheme', 'lightScheme')].forEach((mode) => {
+    it(`should contain correct elevation colors for ${mode}`, () => {
+      const dynamicThemeColors = createDynamicThemeColors({
+        sourceColor,
+      });
+
+      const elevations = ['transparent', 0.05, 0.08, 0.11, 0.12, 0.14];
+      const elevation = elevations.reduce(
```
jaworek, line 129: This is a copy from original source. If it changes there and we forget to update it here, test will still pass even though it might be broken.

src/core/__tests__/theming.test.js
```diff
@@ -0,0 +1,183 @@
...
+        (a, v, index) => ({
+          ...a,
+          [`level${index}`]:
+            index === 0
+              ? v
+              : color(dynamicThemeColors[mode].surface)
+                  .mix(color(dynamicThemeColors[mode].primary), v)
+                  .rgb()
+                  .string(),
```
jaworek, line 138: I would prefer to extract it and use regular `if-else` for better readability.

src/core/__tests__/theming.test.js
lukewalczak: Sure, make sense.

src/core/__tests__/theming.test.js
lukewalczak: Renamed them then.

src/core/__tests__/theming.test.js
lukewalczak: Imo it's not a complicated ternary expression, so there is no need to introduce if-else

jaworek (approved):

lukewalczak:
@jaworek thanks for review!

PR #3408, opened 2022-10-11
fix: BottomNavigation warn when shifting with the wrong number of tabs

brunohkbx:
### Summary

Recently when working on a new feature I realized that the component BottomNavigation crashes when the prop `shifting` is true but the number of tabs is < 2 as in the following screenshot

Error message

### Test plan

1. Open the example APP
2. Go to the example/src/Examples/BottomNavigationExample.tsx
3. Pass the prop `shifting` to the BottomNavigation component
4. Modify the routes to have a single one instead
5. See the warning popping up

Warning message

### DIFF
--- a/src/components/BottomNavigation/BottomNavigation.tsx
+++ b/src/components/BottomNavigation/BottomNavigation.tsx
@@ -65,6 +65,7 @@ export type Props = {
    *
    * By default, this is `false` with theme version 3 and `true` when you have more than 3 tabs.
    * Pass `shifting={false}` to explicitly disable this animation, or `shifting={true}` to always use this animation.
+   * Note that you need at least 2 tabs be able to run this animation.
    */
   shifting?: boolean;
   /**
@@ -386,6 +387,14 @@ const BottomNavigation = ({
   testID = 'bottom-navigation',
 }: Props) => {
   const { scale } = theme.animation;
+
+  if (shifting && navigationState.routes.length < 2) {
+    shifting = false;
+
+    console.warn(
+      'BottomNavigation needs at least 2 tabs to run shifting animation'
+    );
+  }
 
   const focusedKey = navigationState.routes[navigationState.index].key;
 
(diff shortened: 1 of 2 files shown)


brunohkbx:
Hey @lukewalczak  is been a while since my last contribution 😄. I wasn't sure if I should point this PR to the branch main or 4.0

src/components/BottomNavigation/BottomNavigation.tsx
```diff
@@ -352,6 +352,14 @@ const BottomNavigation = ({
 }: Props) => {
```
brunohkbx, line 352: I'm wondering if we can change `shifting = navigationState.routes.length > 3` to `shifting = navigationState.routes.length >= 2`
Would that be considered a breaking change?

lukewalczak: > Would that be considered a breaking change?

Partially it's because it affects users with two tabs, who didn't want `shifting`style to be applied. Why would you like to change?

lukewalczak:
Hey @brunohkbx, thanks for your contribution! 

> I wasn't sure if I should point this PR to the branch main or 4.0

Generally, please target the `main` branch. I will then port that commit to the next `4.x` release, once gather more bug fixes applied to both versions.

lukewalczak:
Hey @brunohkbx, you have a lot of conflicts after changing the target branch. I think the easiest way will be to push your changes again on that branch however having `main` as a base branch.

Please let me know if you need some assist on that ✌🏽

src/components/BottomNavigation/BottomNavigation.tsx
brunohkbx: Just thought it would make more sense since the interpolation animation expects at least 2 items. So having 2 tabs would be perfectly fine to enable the animation.
I'm totally ok keeping it as it is 😃

brunohkbx:
@lukewalczak Fixed it, sorry I couldn't get back to it earlier. While rebasing this PR I found another issue, and opened #3415 addressing it.

lukewalczak:
@brunohkbx no problem, I will check another PR as well. Please correct the tests since CI is failing on that check.

lukewalczak (approved):

PR #3410, opened 2022-10-12
fix: typescript definition for Button's onPress should have event as argument

abumalick:
typescript definition for Button's onPress should have event as argument


RichardLindhout:
This should be done for all components inclusing list.Item e.g. everything with onPress

abumalick:
I am ok to make more fixes if I get positive feedback from maintainers.

I am a bit surprised that a simple change like this does not get any feedback for a month.

Maybe I did something wrong?

lukewalczak:
Hey @abumalick don't worry, I will review the PR in my spare time

lukewalczak:
> This should be done for all components inclusing list.Item e.g. everything with onPress

@RichardLindhout would you like to handle that in the separate PR?

RichardLindhout:
@lukewalczak done!

lukewalczak (approved):
Thanks @abumalick

PR #3413, opened 2022-10-14
fix: portal visible after immediately unmount

### LINKED ISSUE #3395: Portal stays visible indefinitely if unmounted immediately after mounting
### Current behaviour
When trying to remove a portal immediately after mounting it, it stays visible indefinitely. See the failing test below.
* same issue and description as #2606

### Expected behaviour
The portal is unmounted and does not stay visible.

### Code sample
Here's a minimal failing test case (something equivalent happened to me in more complicated, real world code)

https://snack.expo.dev/qbqHB1abQ

```tsx
import React, { useEffect, useState } from 'react';
import { Portal, Text } from 'react-native-paper';
import { render, waitForElement } from 'react-native-testing-library';

const TestComponent = () => {
  const [isPortalVisible, setIsPortalVisible] = useState(true);
  useEffect(() => {
    // Unmount the portal immediately after mounting it
    setIsPortalVisible(false);
  }, []);

  return (
    <Portal.Host>
      {isPortalVisible && (
        <Portal>
          <Text>{"I'm here"}</Text>
        </Portal>
      )}
    </Portal.Host>
  );
};

test('portal is properly unmounted', async () => {
  const component = render(<TestComponent />);

  // This expectation is failing, the portal is never unmounted, despite the useEffect
  await waitForElement(() => {
    const portalContent = component.queryByText("I'm here");
    if (portalContent) throw new Error('Portal is still visible');
  });
});
```

### What have you tried
The cause seems to be with the `await` here:

https://github.com/callstack/react-native-paper/blob/975cfa3be4956b629aa8255e226ff9ca068448eb/src/components/Portal/PortalConsumer.tsx#L10-L17

The await in `componentDidMount` causes a race condition with the synchronous unmounting:

https://github.com/callstack/react-native-paper/blob/975cfa3be4956b629aa8255e226ff9ca068448eb/src/components/Portal/PortalConsumer.tsx#L25-L29

The portal gets (not really) unmounted before being mounted → it stays registered with the portal manager.

One possible fix is to make componentWillUnmount async too:

```diff
-  componentWillUnmount() {
+  async componentWillUnmount() {
     this._checkManager();

+    await Promise.resolve();
     this.props.manager.unmount(this._key);
   }
```
### Your Environment

| ios                                | 15.6.1
| android                        | 11
| react-native                | 0.69.5
| react-native-paper    | 4.12.4
| node                            | 16.16.0
| npm or yarn                | 8.11.0
| expo sdk                     | 46.0.8

hurali97:
### Summary
Fixes #3395

Safely removes the `async ComponentDidMount` and `await Promise.resolve`, which was causing the issues when Portal is unmounted immediately. 

### Test plan
Unit tests coverage.

### DIFF
--- a/src/components/Portal/PortalConsumer.tsx
+++ b/src/components/Portal/PortalConsumer.tsx
@@ -8,11 +8,8 @@ type Props = {
 };
 
 export default class PortalConsumer extends React.Component<Props> {
-  async componentDidMount() {
+  componentDidMount() {
     this.checkManager();
-
-    // Delay updating to prevent React from going to infinite loop
-    await Promise.resolve();
 
     this.key = this.props.manager.mount(this.props.children);
   }
--- a/src/components/__tests__/Menu.test.js
+++ b/src/components/__tests__/Menu.test.js
@@ -5,6 +5,7 @@ import renderer from 'react-test-renderer';
 
 import Button from '../Button/Button.tsx';
 import Menu from '../Menu/Menu.tsx';
+import Portal from '../Portal/Portal.tsx';
 
 const styles = StyleSheet.create({
   contentStyle: {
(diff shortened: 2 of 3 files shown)


lukewalczak (approved):
Thanks @hurali97 for fixing the issue!

PR #3415, opened 2022-10-18
fix(regression): BottomNavigation can't render a single tab

brunohkbx:
### Summary

After rebasing #3408 I noticed that this regression has been introduced in V5. The BottomNavigation breaks when you try to render it with a single tab.

Issue

### Test plan

1. Open the example APP
2. Go to the example/src/Examples/BottomNavigationExample.tsx
4. Modify the routes to have a single one instead
5. Should render a single tab as in the following screenshot

Fixed

### DIFF
--- a/src/components/__tests__/BottomNavigation.test.js
+++ b/src/components/__tests__/BottomNavigation.test.js
@@ -339,7 +339,7 @@ it('renders custom background color passed to barStyle property', () => {
 });
 
 it('renders a single tab', () => {
-  const { getByTestId } = render(
+  const { queryByTestId } = render(
     <BottomNavigation
       shifting={false}
       navigationState={createState(0, 1)}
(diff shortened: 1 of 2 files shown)


src/components/__tests__/BottomNavigation.test.js
```diff
@@ -337,3 +337,17 @@ it('renders custom background color passed to barStyle property', () => {
...
+    <BottomNavigation
+      shifting={false}
+      navigationState={createState(0, 1)}
+      onIndexChange={jest.fn()}
+      renderScene={({ route }) => route.title}
+      testID={'bottom-navigation'}
+    />
+  );
+
+  getByTestId('bottom-navigation')
```
lukewalczak, line 352: That test doesn't have any assertions

lukewalczak:
@brunohkbx thanks for the fix, totally overlooked it. please only adjust the test and it will be ready to be shipped!

brunohkbx:
@lukewalczak fixed ✅

lukewalczak (approved):
Thanks for the fix @brunohkbx!

PR #3433, opened 2022-10-25
feat: getLazy prop for bottom-navigation

### LINKED ISSUE #3085: Add an option to prevent lazy on @react-navigation/material-bottom-tabs.
I'm currently a bit frustrated having nested routes not being indexed as the route containing the routes not being loaded, by having a lazy prop I'd allow me to load these routes :) 

As `import { createBottomTabNavigator } from "@react-navigation/bottom-tabs";` has it, allowing to specify if a route should be lazy loaded or not

**Describe alternatives you've considered**
I'm currently NOT using the material-bottom-tabs for this reason.
.

hurali97:
### Summary

Fixes #3085 

Introduces `getLazy` callback prop which allows the user to set lazy for each screen as they want leveraging the `route` parameter received in the callback. If user wants to set lazy for every screen, they can simply return `true` from the callback.

Usage:
```
      <BottomNavigation
        {...otherProps}
        getLazy={({ route }) => route.key !== 'album'}
      />
```

### Test plan

Unit test coverage 🧪

### DIFF
--- a/src/components/BottomNavigation/BottomNavigation.tsx
+++ b/src/components/BottomNavigation/BottomNavigation.tsx
@@ -262,6 +262,10 @@ export type Props = {
    * TestID used for testing purposes
    */
   testID?: string;
+  /**
+   * Get lazy for the current screen. Uses true by default.
+   */
+  getLazy?: (props: { route: Route }) => boolean;
 };
 
 const MIN_RIPPLE_SCALE = 0.001; // Minimum scale is not 0 due to bug with animation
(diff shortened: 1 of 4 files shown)


lukewalczak:
That PR will need the following changes in the [material-bottom-tabs](https://github.com/react-navigation/react-navigation/tree/main/packages/material-bottom-tabs) repository once the stable version is released.

lukewalczak (approved):

src/components/BottomNavigation/BottomNavigation.tsx
```diff
@@ -262,6 +263,10 @@ export type Props = {
    * TestID used for testing purposes
    */
   testID?: string;
+  /**
+   * Get lazy for the current screen. Uses true by default.
```
lukewalczak, line 267:
```suggestion
   * @supported Available in v5.x
   * Get lazy for the current screen. Uses true by default.
```

lukewalczak (approved):

Issue #3435, opened 2022-10-25
[5.0.0-RC-8] Use custom font as default


bombillazo:
### Current behaviour
Currently, there is no way to set the default app font to a specific fontFamily. You either have to use variants or pass the fontFamily as a style prop.

### Expected behaviour
There should be a theme property that sets the default fontFamily to use and propagates it to ALL RNP elements.

### How to reproduce?
N/A

### Preview

Currently, when using themeV3, if you don't choose variant, there is a token object with default fonts that gets used to set the `brandRegular` property.

```tsx

  if (theme.isV3 && variant) {
  } else {
    const { brandRegular, weightRegular } = tokens.md.ref.typeface;
```

This causes all Text objects that do not use variant to simply use the default OS font. We should be able to somehow override this.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x
| android                        | x
| react-native                | 0.69.6
| react-native-paper    | 5.0.0-RC-8
| node                            | x.x.x
| npm or yarn                | x.x.x
| expo sdk                     | x.x.x

bombillazo:
Im currently getting around this using this patch and extending fonts to include a default variant `MD3Type`:

```tsx
       />
     );
   } else {
-    const { brandRegular, weightRegular } = tokens.md.ref.typeface;
     const font = theme.isV3
       ? {
-          fontFamily: brandRegular,
-          fontWeight: weightRegular,
+          fontFamily: theme.fonts.default.fontFamily,
+          fontWeight: theme.fonts.default.fontWeight,
         }
       : theme.fonts?.regular;
     const textStyle = {

```

lukewalczak:
Hey @bombillazo, unfortunately looks like I overlooked that case :|

What is the `theme.fonts.default` in your patch?

bombillazo:
Its just another `MD3Type` object like `headlineSmall`, `bodyMedium`... etc. I then set it to use the custom fontFamily

```
  default: {
    fontFamily: 'MyFont',
    fontSize: 0,
    fontWeight: '400',
    letterSpacing: 0,
    lineHeight: 0,
  },
```

lukewalczak:
Could you please if that PR https://github.com/callstack/react-native-paper/pull/3436 is fixing the problem?

bombillazo:
@lukewalczak sorry for the delay, yup! that looks good to me

PR #3436, opened 2022-10-25
fix: add default property to the v3 theme fonts

lukewalczak:
### Summary

Fixes https://github.com/callstack/react-native-paper/issues/3435

Support the possibility to override the default `Text`s component font styles, via theme property `theme.fonts.default`

#### Related issue

- #3435 

### Test plan

Updated snapshots

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -173,11 +173,7 @@ export type MD3Typescale =
   | {
       [key in MD3TypescaleKey]: MD3Type;
     } & {
-      ['default']: {
-        fontFamily: string;
-        fontWeight: Font['fontWeight'];
-        letterSpacing: number;
-      };
+      ['default']: Omit<MD3Type, 'lineHeight' | 'fontSize'>;
     };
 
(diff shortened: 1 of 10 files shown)


src/types.tsx
```diff
@@ -169,9 +169,16 @@ export type MD3Type = {
...
+export type MD3Typescale =
+  | {
+      [key in MD3TypescaleKey]: MD3Type;
+    } & {
+      ['default']: {
+        fontFamily: string;
+        fontWeight: Font['fontWeight'];
+        letterSpacing: number;
+      };
+    };
```
Drakeoon, line 181: Wouldn't this work?
```suggestion
      default: MD3Type
```

Drakeoon (approved):
@lukewalczak Left just one comment, looks good otherwise

src/types.tsx
lukewalczak: `MD3Type` contains also `lineHeight` and `fontSize` which I don't want to specify in the `default` Text, however, I think I can just go MD3Type and omits those two properties 🤔

PR #3441, opened 2022-10-27
feat: create Tooltip component

brunohkbx:
### Summary

This PR implements the Tooltip component described [here](https://m2.material.io/components/tooltips) in MD2 docs.

Tooltips display informative text when users hover over, focus on, or tap an element.
When activated, tooltips display a text label identifying an element, such as a description of its function.

This covers only the minimum functionality to make it usable. Desktop support and animated Transitions described in MD docs are not implemented. Future work can be done and is welcome to support all use cases.

#### Behavior
A tooltip is displayed upon tapping and holding a screen element or component (on mobile) or hovering over it (desktop). Continuously display the tooltip as long as the user long-presses or hovers over the element.

#### Timing
Display the tooltip for 1.5 seconds. If the user takes another action before that time ends, the tooltip will disappear.

Show case

#### Placement
The tooltip can overflow to the left / right or bottom depending on where the wrapped element is placed on the screen.

The purpose of the following test cases is to show that the tooltip text does not goes offscreen.

1. Overflow To Left
Test 1

2. Overflow To Right
Test 2

3. Overflow To Bottom
Test 3

4. Normal placement
Test 4


src/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+        </Portal>
+      )}
+      <View
+        onTouchStart={handleTouchStart}
+        onTouchEnd={handleTouchEnd}
+        onTouchCancel={handleTouchEnd}
+        style={wrapperStyle}
+      >
+        {React.cloneElement(children, { ...rest, ref: childrenWrapperRef })}
+      </View>
```
brunohkbx, line 250: 🧵 
At first, I tried to use this View to measure the wrapped element as in this change

```suggestion
      <View
        onTouchStart={handleTouchStart}
        onTouchEnd={handleTouchEnd}
        onTouchCancel={handleTouchEnd}
        style={wrapperStyle}
        collapsable={false}
        ref={childrenWrapperRef}
      >
        {React.cloneElement(children, { ...rest })}
      </View>
```

However, I ran into some issues with absolute positioned elements such as a FAB. The only I found to work around this issue was to pass a ref to the wrapped elements as MUI does, see [here](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Tooltip/Tooltip.js#L724) and more explanation [here](https://mui.com/material-ui/guides/composition/#caveat-with-refs)

src/components/Appbar/AppbarAction.tsx
```diff
@@ -72,39 +72,45 @@ export type Props = React.ComponentPropsWithoutRef<typeof IconButton> & {
  * export default MyComponent;
  * ```
  */
-const AppbarAction = ({
```
brunohkbx, line 75: See my comment https://github.com/callstack/react-native-paper/pull/3441#discussion_r1006410447 to understand this change

src/components/Appbar/AppbarAction.tsx
```diff
@@ -72,39 +72,45 @@ export type Props = React.ComponentPropsWithoutRef<typeof IconButton> & {
...
-  color: iconColor,
-  icon,
-  disabled,
-  onPress,
-  accessibilityLabel,
-  isLeading,
-  ...rest
-}: Props) => {
-  const theme = useInternalTheme();
+const AppbarAction = React.forwardRef<{}, Props>(
```
brunohkbx, line 75: Is this type declaration enough? Can I replace this `{}` with something more specific? I didn't know what to use

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+/**
+ * Tooltips display informative text when users hover over, focus on, or tap an element.
+ *
+ * Plain tooltips, when activated, display a text label identifying an element, such as a description of its function. Tooltips should include only short, descriptive text and avoid restating visible UI text.
+ *
+ * <div class="screenshots">
+ *   <figure>
+ *     <img src="screenshots/TODO.png" />
+ *     <figcaption>TODO</figcaption>
```
brunohkbx, line 133: @lukewalczak could you please add some screenshots here? 😄 
I don't have a mac to run it in the emulator. I think would be better to keep all screenshots in the same format

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+};
+
+const styles = StyleSheet.create({
+  text: {
+    color: MD2Colors.white,
+  },
+  tooltip: {
+    alignSelf: 'flex-start',
+    justifyContent: 'center',
+    backgroundColor: 'rgba(97, 97, 97, 0.9)',
```
brunohkbx, line 262: @lukewalczak I hard-coded this value because I didn't know where to put it. Should this be in both theme versions (v2/v3)?

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+      >
+        {React.cloneElement(children, { ...rest, ref: childrenWrapperRef })}
+      </View>
+    </>
+  );
+};
+
+const styles = StyleSheet.create({
+  text: {
+    color: MD2Colors.white,
```
brunohkbx, line 257: Same

lukewalczak:
Hello @brunohkbx, thanks for adding new component 🔥 I will answer your questions/comments in the upcoming days 🙌🏽

src/components/Tooltip.tsx
lukewalczak: Sure, I will handle it as the last step, just before merging it

src/components/Tooltip.tsx
lukewalczak: For the text color I would suggest using: `theme.colors.surface` for both versions.

src/components/Tooltip.tsx
lukewalczak: For the v3 we can go with `theme.colors.onSurface` for v2 we can add that new color into the theme with the name `tooltip`.

gbelgamo-godaddy (approved):

src/components/Appbar/AppbarAction.tsx
lukewalczak: I guess there should be the type of the ref

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,275 @@
...
+ * const MyComponent = () => (
+ *   <Tooltip title="Selected Camera">
+ *     <IconButton icon="camera" selected size={24} onPress={() => {}} />
+ *   </Tooltip>
+ * );
+ *
+ * export default MyComponent;
+ * ```
+ */
+const Tooltip = ({
```
lukewalczak, line 151: Can we split that file into two, for the component and utils?

example/src/Examples/AppbarExample.tsx
```diff
@@ -67,10 +68,14 @@ const AppbarExample = ({ navigation }: Props) => {
                 <Appbar.Action icon="calendar" onPress={() => {}} />
               )}
           {showSearchIcon && (
-            <Appbar.Action icon="magnify" onPress={() => {}} />
```
lukewalczak, line 70: Can we add a dedicated example for the Tooltip? 
I mean, someone who is checking only the example app won't have knowledge that the component exists.

src/components/Tooltip.tsx
brunohkbx: Done 👍🏻 what do you think?

MD2 Light | MD2 Dark
--- | ---
Light Theme |  Dark Theme

MD3 Light | MD3 Dark
--- | ---
Light Theme |  Dark Theme

brunohkbx: https://github.com/callstack/react-native-paper/pull/3441/commits/e2e0e0e6a44ea2fae3691ed416276064109c3a02

src/components/Appbar/AppbarAction.tsx
brunohkbx: Do you think that looks right? https://github.com/callstack/react-native-paper/pull/3441/commits/12a645af55ab38f33c3b869fdee6adeec8479d46

example/src/Examples/AppbarExample.tsx
brunohkbx: Good call! What else do you think we can have on this screen?
Commit: https://github.com/callstack/react-native-paper/pull/3441/commits/7be6e3ad203d0447712e95bf4207c7b5c99efa8b

Example

example/src/Examples/IconButtonExample.tsx
```diff
@@ -34,8 +40,12 @@ const ButtonExample = () => {
     <ScreenWrapper contentContainerStyle={styles.v3Container}>
       <List.Section title="Default">
         <View style={styles.row}>
-          <IconButton icon="camera" size={24} onPress={() => {}} />
-          <IconButton icon="camera" selected size={24} onPress={() => {}} />
+          <Tooltip title="Camera">
```
lukewalczak, line 43: Since we have a dedicated example, those changes can be reverted.

src/components/Tooltip/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './Tooltip';
```
lukewalczak, line 1: No need for that, in `src/index.tsx` please point directly to the component file:

```
export { default as Tooltip } from './components/Tooltip/Tooltip';
```

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,176 @@
+import * as React from 'react';
+import { Dimensions, View, LayoutChangeEvent, StyleSheet } from 'react-native';
+
+import { useInternalTheme } from '../../core/theming';
+import { addEventListener } from '../../utils/addEventListener';
+import Portal from '../Portal/Portal';
+import Text from '../Typography/Text';
+import { getTooltipPosition, Measurement } from './utils';
+
+export type Props = {
```
lukewalczak, line 10: Please also in `src/index.tsx` add types export:

```
export type { Props as TooltipProps } from './components/Tooltip/Tooltip';
```

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,176 @@
...
+    tooltip: {},
+    measured: false,
+  });
+  const showTooltipTimer = React.useRef<NodeJS.Timeout>();
+  const hideTooltipTimer = React.useRef<NodeJS.Timeout>();
+  const childrenWrapperRef = React.useRef() as React.MutableRefObject<View>;
+
+  React.useEffect(() => {
+    return () => {
+      if (showTooltipTimer.current) clearTimeout(showTooltipTimer.current);
```
lukewalczak, line 75: Could you please add the curly braces around `if` condition in your whole changes? There is nothing wrong with your approach, I just would like to have consistency in the codebase ✌🏽

src/components/Tooltip.tsx
brunohkbx: @lukewalczak I'm wondering if we can have a prop called `forwardRef` for cases when the element is absolutely positioned such as a FAB. Then we wouldn't need to make the components that I changed forwardable. What do you think?

lukewalczak: I would say to keep your current approach, I think are no benefits of having the prop `forwardRef` and it can be confusing for users.

lukewalczak (approved):
LGTM! Thanks @brunohkbx 🚀

Issue #3439, opened 2022-10-26
How to change List.Accordion rippleColor and underlayColor ?


cyhnkckali:

suleymanbariseser:
May I contribute to the project by implementing this feature?

lukewalczak:
hey @suleymanbariseser, sure, let's try to add `rippleColor` in the similar way as it's currently done in `Button` or `AnimatedFAB` components.

cyhnkckali:
Thank you @suleymanbariseser

lukewalczak:
Closed via: https://github.com/callstack/react-native-paper/pull/3448

PR #3448, opened 2022-10-31
feat: make ripple color dynamic for ListAccordion

suleymanbariseser:
### Summary

It makes rippleColor dynamic in ListAccordion as it is asked in the issue #3439 

### Test plan

Tested in the example application with different text colors and everything seems okay.


src/components/List/ListAccordion.tsx
```diff
@@ -187,6 +187,11 @@ const ListAccordion = ({
   const isExpanded = groupContext
     ? groupContext.expandedId === id
     : expandedInternal;
+
+  const titleTextColor = isExpanded ? theme.colors?.primary : titleColor;
+
+  const rippleColor = color(titleTextColor).alpha(0.12).rgb().string();
```
lukewalczak, line 193: Could you please add a **unit** test covering `rippleColor`?

suleymanbariseser: sure but i think to test it I need to create a utils file and move it there, right? Additionally, I saw tests are missing for titleColor and descriptionColor so if it is okay I can add the tests for them too

suleymanbariseser: @lukewalczak tests are added

lukewalczak (approved):
Thanks @suleymanbariseser!

PR #3452, opened 2022-11-03
fix: unwanted closing of contextMenu

joaovitorqueiroz:
### Summary

The TextInput component is having an unexpected behavior. When the user tries to use the contextMenu to paste a text, it's closing itself.

text-input

### Test plan

To test the bug, just press for a long time text input for the context menu to appear and then close.

To test the solution, just press for a long time on an input where the placeholder was not specified.

text-input-fixed

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -317,12 +317,16 @@ const TextInput = React.forwardRef<TextInputHandles, Props>(
       // We don't show placeholder if there's a label because the label acts as placeholder
       // When focused, the label moves up, so we can show a placeholder
       if (focused || !rest.label) {
-        // Set the placeholder in a delay to offset the label animation
-        // If we show it immediately, they'll overlap and look ugly
-        timer.current = setTimeout(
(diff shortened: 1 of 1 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -317,12 +317,16 @@ const TextInput = React.forwardRef<TextInputHandles, Props>(
...
       if (focused || !rest.label) {
-        // Set the placeholder in a delay to offset the label animation
-        // If we show it immediately, they'll overlap and look ugly
-        timer.current = setTimeout(
-          () => setPlaceholder(rest.placeholder),
-          50
-        ) as unknown as NodeJS.Timeout;
+        // If the user wants to use the contextMenu, when changing the placeholder, the contextMenu is closed
+        // This is a workaround to mitigate this behavior in scenarios where the placeholder is not specified.
+        if (rest.placeholder) {
```
joaovitorqueiroz, line 322: This fix doesn't solve all scenarios, but I already consider it a good solution because it solves scenarios where the placeholder was not specified.

lukewalczak (approved):

PR #3463, opened 2022-11-10
feat: add prop pointerEvents for ListAccordion

brunohkbx:
### Summary

This component can receive a custom prop `right`, so the following code is allowed: 
```js
const renderRight = ({ isExpanded }) =>
(
    <IconButton
      icon={isExpanded ? 'chevron-up' : 'chevron-down'}
      size={24}
      onPress={myCustomOnPress}
    />
);
```

However, the onPress passed to the IconButton won't be triggered since the View container has a hard-coded `pointerEvents="none"`

### Test plan

N/A

### DIFF
--- a/src/components/List/ListAccordion.tsx
+++ b/src/components/List/ListAccordion.tsx
@@ -6,6 +6,7 @@ import {
   StyleSheet,
   TextStyle,
   View,
+  ViewProps,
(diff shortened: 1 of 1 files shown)


lukewalczak (approved):
Thanks @brunohkbx!

PR #3487, opened 2022-11-23
fix: automatically close snackbar by duration

### LINKED ISSUE #3486: Snackbar duration bug
### Current behaviour
Duration prop does not work in Snackbar. Timer resets whenever onDismiss function updates. It means even when a state changes it will reset the timer.

### Expected behaviour
Duration prop must be independent from any other state and prop.

### How to reproduce?
[Here](https://snack.expo.dev/@suleymanbariseser/react-native-paper-snackbar-issue) is a snack where this issue can be reproducable. 

The problem about this is adding onDismiss into dependency array of the useEffect which handles the timer in the Snackbar component. However, this is needed because onDismiss must be triggered whenever time will be over and we must trigger the latest version of onDismiss function. The problem can be solved by moving the logic inside the useEffect into event functions and trigger them in the useEffect whenever `visible` updates. The problem and the solution is described [here](https://beta.reactjs.org/learn/separating-events-from-effects#extracting-non-reactive-logic-out-of-effects) in detail. In the detailed url, it is solved by useEvent but useEvent is still an experimental hook. That's why, an alternative hook([useEventCallback](https://www.npmjs.com/package/use-event-callback)) can be used to achieve the same goal.

### Your Environment

suleymanbariseser:
### Summary

Fixes #3486


lukewalczak:
Thanks @suleymanbariseser, I've resolved conflicts on your behalf.

package.json
```diff
@@ -47,7 +47,8 @@
   },
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.8",
-    "color": "^3.1.2"
+    "color": "^3.1.2",
+    "use-event-callback": "^0.1.0"
```
brunohkbx, line 51: @lukewalczak I wonder if it wouldn't be possible to solve this issue without adding a new dependency for RNP. Shouldn't we rethink this solution?

[use-event-callback](https://www.npmjs.com/package/use-event-callback) was published 4 years ago and hasn't been updated ever since. I don't like the idea of adding an unmaintained library to solve a particular issue

lukewalczak: I would treat using that library as a temporary workaround until `useEvent` is released. I think it's a quite basic util library without any issues _(on github)_ at least, that's why I decided to accept it. Do you have some ideas to refactor it? Writing our implementation of that hook?

PR #3495, opened 2022-11-28
fix: correct tooltip behaviour on mobile

lukewalczak:
### Summary

- [x] - add `Tooltip` into migration guide

* Passes `ref` to the web `Surface` so `Tooltip` can be displayed on that platform as well.
* Fixes `Tooltip` behaviour which allows calling the `children` press action once the finger was released from the tapped element/component. The video present on the MD documentation indicates that element action should not be fired.

_Video from MD:_

https://user-images.githubusercontent.com/22746080/204304666-956686a8-9743-4246-8169-21c30ddee678.mp4


docs/pages/10.migration-guide-to-5.0.md
```diff
@@ -570,6 +570,19 @@ The property `name` was renamed to `icon`, since the scope and type of that prop
 + <TextInput.Icon icon="magnify" />
 ```
 
+## Tooltip
+
+Component displayed upon tapping and holding a screen element or component used to present an informative text label identifying an element, such as a description of its function.
+
+![shiftingAnimation](screenshots/tooltip.gif)
```
brunohkbx, line 577: I think you forgot to update this text here

```suggestion
![tooltip](screenshots/tooltip.gif)
```

src/components/Tooltip/Tooltip.tsx
```diff
@@ -106,10 +116,10 @@ const Tooltip = ({
       clearTimeout(hideTooltipTimer.current);
     }
 
-    showTooltipTimer.current = setTimeout(
-      () => setVisible(true),
-      enterTouchDelay
-    ) as unknown as NodeJS.Timeout;
+    showTooltipTimer.current = setTimeout(() => {
+      setTouched(true);
```
brunohkbx, line 120: Adding another setState under this setTimeout will cause another rerender for React < 18.00 due to how the automatic batching works. 
Performance wise I don't think it would be a problem but this could be easily fixed with a ref, what do you think?

We can replace `const [touched, setTouched] = React.useState(false);` with `const touched = useRef(false)` and then:

```suggestion
      touched.current = true
```

lukewalczak: Will investigate that, thanks for hint!

brunohkbx: For reference: https://reactjs.org/blog/2022/03/29/react-v18.html#new-feature-automatic-batching

lukewalczak: Done, thanks!

Issue #3508, opened 2022-12-06
Support `label` prop for FAB.Group


ramonfabrega:
**Is your feature request related to a problem? Please describe.**
`FAB` has an optional `label` prop that is not exposed to `FAB.Group`

**Describe the solution you'd like**
an optional prop to `FAB.Group`, there was an open PR a while back:
https://github.com/callstack/react-native-paper/pull/2343

**Describe alternatives you've considered**
Make my own `FAB.Group` component that is just a fork from the code + the optional prop

PR #3510, opened 2022-12-07
feat: Add label prop to FAB.Group

guilherme-ferraz1:
### Summary

Allows the `label` prop to work using `FAB.Group` as requested in #3508 

### Test plan

Not sure if we need to add this to the examples since it's such a small and logical change but I can do if needed.

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -105,6 +105,10 @@ export type Props = {
    */
   theme: InternalTheme;
   /**
+   * Optional label for `FAB`.
+   */
+  label?: string;
+  /**
    * Pass down testID from Group props to FAB.
    */
   testID?: string;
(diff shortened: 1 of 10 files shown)


src/components/FAB/FABGroup.tsx
```diff
@@ -423,6 +428,7 @@ const FABGroup = ({
           accessibilityState={{ expanded: open }}
           style={[styles.fab, fabStyle]}
           visible={visible}
+          label={label}
```
lukewalczak, line 431: Could you please add a unit test for covering new prop usage?

gmferraz: Sure, just added 😃

guilherme-ferraz1: Sure, just added 😃

lukewalczak (approved):

PR #3514, opened 2022-12-07
feat: numberOfLines prop added to Button

thijmen-exa:
### Summary
Allows developers to specify how many lines of text can be displayed by a Button component. 

Wasn't able to figure out an easier method of displaying multiple lines of text inside of a button.

### Test plan
By default, nothing visibly changes. When `numberOfLines` is set, it behaves as expected (see screenshot).

Example of what it looks like with `numberOfLines={3}`
image

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -120,6 +120,10 @@ export type Props = React.ComponentProps<typeof Surface> & {
    * testID to be used on tests.
    */
   testID?: string;
+  /**
+   * Number of lines the text in the button should display.
+   */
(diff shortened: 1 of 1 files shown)


lukewalczak:
Hey @thijmen-exa, I've checked the MD documentation and it indicates _(in both MD generations)_ that the text within `Button` should not be wrapped into two lines: 

> Don’t wrap text. For maximum legibility, label text should remain on a single line.

Although we are open to the customization of components, in this case, I think it's very rare to extend the API by adding a new prop. 

Thanks for the effort, but currently we won't merge it.

PR #3521, opened 2022-12-10
fix: fix position of menu when dimensions change

### LINKED ISSUE #3515: Menu position is wrong when changing size of window
### Current behaviour
The position of the menu does not update when changing size of the browser window.

### Expected behaviour
Menu position should render at the anchor that I have specified in the code.

### How to reproduce?
Go to the snack below. Choose web as client, make the preview show in a separate window.
Show the menu, close the menu, make the window more wider. Try to open the menu again. It should render at "show menu" but is instead rendered at the position the "show menu" button first was rendered.

https://snack.expo.dev/@fyranollfyra/faulty-menu-position-when-changing-size

(the same goes for the height also actually. Take the same separate browser window and make it the smallest height. Then reload the window, expand the height as much as you can, then open the menu again. Now the height of the menu is "cropped" and you have to scroll inside the menu to get to all options).

This also happens on iOS for example. Run the snack in Expo Go with your device in portrait mode. Turn it to landscape and the menu renders in the wrong place.

### Preview

Width:
image
image

Height: 
image
image

It works fine with 5.0.0-rc.8, but after that version this could be reproduced by following the above steps.

### Your Environment

| react-native                | 0.70.5
| react-native-paper    | 5.0.1
| node                            | 16.13.1
| yarn                             | 1.22.9
| expo sdk                     | 47.0.0

mathias-berg:
### Summary

The dimensions was set outside the component and therefore didn't change when the screen size was adjusted.
This was introduced in in rc9 so I just moved it back to where it was before when it worked.

Fixes #3515

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -82,8 +82,6 @@ export type Props = {
 
 type Layout = $Omit<$Omit<LayoutRectangle, 'x'>, 'y'>;
 
-type WindowLayout = Omit<ScaledSize, 'scale' | 'fontScale'>;
-
 type State = {
   rendered: boolean;
   top: number;
@@ -92,7 +90,7 @@ type State = {
   anchorLayout: Layout;
   opacityAnimation: Animated.Value;
   scaleAnimation: Animated.ValueXY;
-  windowLayout: WindowLayout;
+  windowLayout: Layout;
 };
 
 // Minimum padding between the edge of the screen and the menu
@@ -101,6 +99,8 @@ const SCREEN_INDENT = 8;
 const ANIMATION_DURATION = 250;
 // From the 'Standard easing' section of https://material.io/design/motion/speed.html#easing
 const EASING = Easing.bezier(0.4, 0, 0.2, 1);
+
+const WINDOW_LAYOUT = Dimensions.get('window');
 
 /**
  * Menus display a list of choices on temporary elevated surfaces. Their placement varies based on the element that opens them.
@@ -178,7 +178,10 @@ class Menu extends React.Component<Props, State> {
     anchorLayout: { width: 0, height: 0 },
     opacityAnimation: new Animated.Value(0),
     scaleAnimation: new Animated.ValueXY({ x: 0, y: 0 }),
-    windowLayout: Dimensions.get('window'),
+    windowLayout: {
+      width: WINDOW_LAYOUT.width,
+      height: WINDOW_LAYOUT.height,
+    },
   };
 
   componentDidMount() {
@@ -393,18 +396,18 @@ class Menu extends React.Component<Props, State> {
 
   private keyboardDidShow = (e: RNKeyboardEvent) => {
     const keyboardHeight = e.endCoordinates.height;
-    const windowLayout = Dimensions.get('window');
+    const { height, width } = Dimensions.get('window');
 
     this.setState({
       windowLayout: {
-        height: windowLayout.height - keyboardHeight,
-        width: windowLayout.width,
+        height: height - keyboardHeight,
+        width,
       },
     });
   };
 
   private keyboardDidHide = () => {
-    const height = Dimensions.get('window').height;
+    const { height } = Dimensions.get('window');
 
     this.setState((state) => ({
       windowLayout: {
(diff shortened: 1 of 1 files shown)


mathias-berg:
@lukewalczak 
Is solution good for you or should I change it to an event listener and set the values in state instead?

lukewalczak:
Hey @mathias-berg, thanks for fix! We can stick with the current approach, however I would like to confirm whether it doesnt interfere with the previously fixed issue.

lukewalczak:
Unfortunately, the current approach with updating window dimensions breaks the `Menu` example when opening the keyboard, from the previous issue, presented on that [snack](https://snack.expo.dev/@react-native-paper/menu-with-keyboard). Looks like we need to find another solution which covers both cases.

mathias-berg:
I will look at this tonight to see if I find a solution.

mathias-berg:
Did a lot of investigating on this last night and found a solution that works. I will do some more testing and then update this PR.

mathias-berg:
Changes are up now, please have a look.

lukewalczak:
Thanks @mathias-berg! I will review & test in my spare time

src/components/Menu/Menu.tsx
```diff
@@ -176,6 +178,7 @@ class Menu extends React.Component<Props, State> {
     anchorLayout: { width: 0, height: 0 },
     opacityAnimation: new Animated.Value(0),
     scaleAnimation: new Animated.ValueXY({ x: 0, y: 0 }),
+    windowLayout: Dimensions.get('window'),
```
lukewalczak, line 181: Can we narrow `windowLayout` state only to `width` and `height`?

src/components/Menu/Menu.tsx
```diff
@@ -384,11 +392,30 @@ class Menu extends React.Component<Props, State> {
   };
 
   private keyboardDidShow = (e: RNKeyboardEvent) => {
-    this.keyboardHeight = e.endCoordinates.height;
+    const keyboardHeight = e.endCoordinates.height;
+    const windowLayout = Dimensions.get('window');
```
lukewalczak, line 396:
```suggestion
    const { width, height } = Dimensions.get('window');
```

src/components/Menu/Menu.tsx
```diff
@@ -384,11 +392,30 @@ class Menu extends React.Component<Props, State> {
...
+
+    this.setState({
+      windowLayout: {
+        height: windowLayout.height - keyboardHeight,
```
lukewalczak, line 400:
```suggestion
        height: height - keyboardHeight,
```

src/components/Menu/Menu.tsx
```diff
@@ -384,11 +392,30 @@ class Menu extends React.Component<Props, State> {
...
+        width: windowLayout.width,
```
lukewalczak, line 401:
```suggestion
        width,
```

src/components/Menu/Menu.tsx
```diff
@@ -384,11 +392,30 @@ class Menu extends React.Component<Props, State> {
...
+      },
+    });
   };
 
   private keyboardDidHide = () => {
-    this.keyboardHeight = 0;
+    const height = Dimensions.get('window').height;
```
lukewalczak, line 407:
```suggestion
    const { height } = Dimensions.get('window');
```

src/components/Menu/Menu.tsx
```diff
@@ -384,11 +392,30 @@ class Menu extends React.Component<Props, State> {
...
+    this.setState((state) => ({
+      windowLayout: {
+        width: state.windowLayout.width,
+        height,
+      },
+    }));
+  };
+
+  private onWindowChangeHandler = ({ window }: { window: WindowLayout }) => {
+    this.setState({ windowLayout: window });
```
lukewalczak, line 418: Accordingly to the first comment, please update there only `width` and `height`

src/components/Menu/Menu.tsx
mathias-berg: Of course, fixing it!

src/components/Menu/Menu.tsx
mathias-berg: Fixed!

src/components/Menu/Menu.tsx
mathias-berg: Fixed!

src/components/Menu/Menu.tsx
mathias-berg: Fixed!

src/components/Menu/Menu.tsx
mathias-berg: Fixed!

src/components/Menu/Menu.tsx
mathias-berg: Fixed!

lukewalczak (approved):
Hey @mathias-berg, thanks for the help with fixing the issue! 🎉

PR #3543, opened 2022-12-21
fix: Made ToggleButton.Group TS compatible

gedu:
fix: Now you can use `ToggleButton.Group` with Union Types and Enums (default string)

### Summary

Before you couldn't use Union types or enums without TS complaining about the type

```
type Fruits = 'watermelon' | 'strawberries' | 'apple';
```
ts_error_type_string

with this new change, TS will get the right type and won't complain

new_toggleButton_unionT_function

### Test plan

- Use different types, like string, enums and union types
- Check if the IDE get the right type

### DIFF
--- a/src/components/ToggleButton/ToggleButtonGroup.tsx
+++ b/src/components/ToggleButton/ToggleButtonGroup.tsx
@@ -1,26 +1,27 @@
 import * as React from 'react';
 
-export type Props<TData = string> = {
+export type Props<Value = string> = {
   /**
    * Function to execute on selection change.
    */
-  onValueChange: (value: TData) => void | null;
+  onValueChange: (value: Value) => void;
   /**
    * Value of the currently selected toggle button.
    */
-  value: TData | null;
+  value: Value | null;
   /**
    * React elements containing toggle buttons.
    */
   children: React.ReactNode;
 };
 
-type ToggleButtonContextType<TData = any> = {
-  value: TData | null;
-  onValueChange: (item: TData) => void | null;
+type ToggleButtonContextType<Value> = {
+  value: Value | null;
+  onValueChange: (item: Value) => void;
 };
 
 export const ToggleButtonGroupContext =
+  //@ts-ignore: so we avoid using `any` into the type
   React.createContext<ToggleButtonContextType>(null as any);
 
 /**
@@ -54,11 +55,11 @@ export const ToggleButtonGroupContext =
  * export default MyComponent;
  *```
  */
-const ToggleButtonGroup = <TData = string,>({
+const ToggleButtonGroup = <Value = string,>({
   value,
   onValueChange,
   children,
-}: Props<TData>) => (
+}: Props<Value>) => (
   <ToggleButtonGroupContext.Provider
     value={{
       value,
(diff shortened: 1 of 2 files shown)


src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -1,23 +1,23 @@
 import * as React from 'react';
 
-export type Props = {
+export type Props<TData = string> = {
```
satya164, line 3: We don't use `T` prefix for types. For regular types that user will encounter, prefer using just the name of the thing it represents (e.g. `Value`), or for generic types like this one just `T` is preferable.

src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -1,23 +1,23 @@
...
   /**
    * Function to execute on selection change.
    */
-  onValueChange: (value: string) => void | null;
+  onValueChange: (value: TData) => void | null;
```
satya164, line 7: Unrelated to this specific change, but `void | null` seems wrong. It shouldn't matter whether `onValueChange` returns `null` or something else

src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -1,23 +1,23 @@
...
   /**
    * React elements containing toggle buttons.
    */
   children: React.ReactNode;
 };
 
-type ToggleButtonContextType = {
-  value: string | null;
-  onValueChange: (item: string) => void | null;
+type ToggleButtonContextType<TData = any> = {
```
satya164, line 18: I don't think it's possible to infer this generic for context so probably better not to use generic. We should also avoid using `any` as it makes it easy to spread and make code unsafe. Prefer ignoring the type error explicitly when necessary instead.

satya164 (changes requested):

src/components/ToggleButton/ToggleButtonGroup.tsx
gedu: This one is hard to infer but I need the generic to handle the Consumer function, if not I get a type error. Agreed to avoid `any`, if ignoring is the preferred way I will add it.

satya164: > I need the generic to handle the Consumer function, if not I get a type error

The type error is valid since there's no way to ensure that the type from the group will match the type of value in the individual buttons. In this case, suppressing the error with a comment will be more explicit.

src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -1,26 +1,27 @@
...
-type ToggleButtonContextType = {
-  value: string | null;
-  onValueChange: (item: string) => void | null;
+type ToggleButtonContextType<Value> = {
+  value: Value | null;
+  onValueChange: (item: Value) => void;
 };
 
 export const ToggleButtonGroupContext =
+  //@ts-ignore: so we avoid using `any`, TS can't ensure the type from Group to children
```
jaworek, line 24:
```suggestion
  //@ts-expect-error: TS can't ensure the type from Group to children
```

lukewalczak (approved):

jaworek (approved):

PR #3546, opened 2022-12-22
fix: menu position fix when dimensions change

mathias-berg:
### Summary

The previous fix that I did that was recently merged (https://github.com/callstack/react-native-paper/pull/3521) worked fine when I tested it (both with the example app and with the code from this snack: https://snack.expo.dev/@fyranollfyra/menu-with-keyboard).
But when I tested the released version with my own app where I originally discovered this bug, the bug was still there when changing width of my window.

So I started digging into the code some more and made it a bit simpler and removed unnecessary listeners since we now get the window layout when opening the menu (the same time as we get the anchor and menu layout as well). 

I also cleaned up some of my previous code (`windowLayout.width` instead of `this.state.windowLayout.width`) to be more in line how we use variables from state like we do for `menuLayout` and `anchorLayout`. 

### Test plan

Make sure the menu position changes when changing the width and height of a window or when going from landscape to portrait on phone or tablet.

Also make sure the menu doesn't get behind the keyboard (on phone or tablet) with the snack above, and that the position of the menu is correct after keyboard is dismissed.

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -12,7 +12,6 @@ import {
   LayoutRectangle,
   NativeEventSubscription,
   Platform,
-  ScaledSize,
   ScrollView,
   ScrollViewProps,
   StyleProp,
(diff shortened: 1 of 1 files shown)


mathias-berg:
@lukewalczak, my previous PR that just got merged still had some issues when changing dimensions (width). Did some changes that now works better.

lukewalczak (approved):
From the code perspective changes LGTM. Tested on mobile with current examples and the one linked in the description.

PR #3571, opened 2023-01-03
fix: Surface to correctly set start and end on iOS

### LINKED ISSUE #3542: Absolute positioning with start/end properties is broken on iOS
### Current behaviour
Using a `<Card>` component and `absolute`-positioning it relative to the `start` or `end` of it's parent doesn't work as expected. It will position relative to an `AnimatedContentWrapper` component which is an implementation detail of `react-native-paper`.

Note: Positioning with `right` or `left` as opposed to `start` or `end` works as intended, as well as `start` and `and` working as intended on Android and Web.

### Expected behaviour
The expected behavior in this case (assuming LTR layout) is for the `<Card>` to be right-aligned relative to its parent.
Please refer to the docs: https://reactnative.dev/docs/layout-props#end
<br/>The behavior should be identical to what happens with right alignment.

### How to reproduce?
A minimal repro would be just nesting a `<Card>` inside a `<View>`, with absolute-positioning, and end property set, such as:

```javascript
<View style={{ flex: 1 }}>
    <Card style={
        {
          position: 'absolute',
          end: 0
        }}>
          <Text>I would like to be end-aligned!</Text>
    </Card>
</View>
```

Please refer to a similar repro in my repo: https://github.com/szaboopeeter/rnp-card-absolute-end-ios-bug/blob/main/CardBug/App.tsx

### Preview
End-alignment with `end: 0` does not work:

Demonstration of buggy end-alignment

Because the `<AnimatedComponentWrapper>` in the component hierarchy to drops the `end` style:

Demonstration of buggy end-alignment causing AnimatedComponentWrapper to drop the end property

The expected behavior is the same as it is with `right: 0`:

Demonstration of working right-alignment

I was able to track this down to the current implementation of the `<Surface>` component for iOS in `react-native-paper`, where some of the relevant properties are skipped from `absoluteStyles`, and only later attached to the component hierarchy as `restStyle`: https://github.com/callstack/react-native-paper/blob/v5.1.0/src/components/Surface.tsx#L224

Seems related to this change: https://github.com/callstack/react-native-paper/pull/3239/files

This causes the `<AnimatedComponentWrapper>` in the component hierarchy to drop the `end` style that was passed to the `Card`. See the style props of the component on screenshot below:

Demonstration of buggy end-alignment causing AnimatedComponentWrapper to drop the end property

I'd be happy to open a PR if you can confirm that this is an actual bug, and we can agree on a solution - which seems to be to just pass additional props that are relevant to absolute-positioning to `absoluteSytles` (considering the shadow-layering seems like an intentional difference on iOS).

### Your Environment

| ios                                | 16.2
| android                        | n/a
| react-native                | 0.70.6
| react-native-paper    | 5.1.0
| node                            | 16.15.1
| yarn                             | 1.22.19
| expo sdk                     | n/a

szaboopeeter:
Fixes #3542 

### Summary

- Added `start` and `end` into absolute styles that get applied to the top level shadow layer of `<Surface>` on iOS, to correctly apply these styles to Surface-based components.
- Added negative tests to ensure these properties do not get applied on the bottom of the component hierarchy as well as positive tests to make sure they do appear on the 0th shadow layer - both of these would fail without the changes introduced in this PR.
- Components that compose `<Surface>` had their snapshots updated due to the inclusion of these properties.

### Test plan

As described in #3542, in current version applying start/end for example to `<Card>` will result in incorrect behavior, see screenshot:

After these changes, this behavior gets fixed, and the appropriate styles get added to the first `<AnimatedComponentWrapper>`, resulting in the component being right-aligned as visible on screenshot below (it's the same component added to the `/example` project after making the changes on this branch):

### DIFF
--- a/src/components/Surface.tsx
+++ b/src/components/Surface.tsx
@@ -218,10 +218,10 @@ const Surface = React.forwardRef<View, Props>(
 
     const shadowColor = '#000';
 
-    const { position, alignSelf, top, left, right, bottom, ...restStyle } =
+    const { position, alignSelf, top, left, right, bottom, start, end, ...restStyle } =
       (StyleSheet.flatten(style) || {}) as ViewStyle;
 
-    const absoluteStyles = { position, alignSelf, top, right, bottom, left };
+    const absoluteStyles = { position, alignSelf, top, right, bottom, left, start, end };
     const sharedStyle = [{ backgroundColor }, restStyle];
 
     if (isAnimatedValue(elevation)) {
--- a/src/components/__tests__/Appbar/__snapshots__/Appbar.test.js.snap
+++ b/src/components/__tests__/Appbar/__snapshots__/Appbar.test.js.snap
@@ -7,6 +7,7 @@ exports[`Appbar does not pass any additional props to Searchbar 1`] = `
     Object {
       "alignSelf": undefined,
       "bottom": undefined,
+      "end": undefined,
       "left": undefined,
       "position": undefined,
       "right": undefined,
@@ -17,6 +18,7 @@ exports[`Appbar does not pass any additional props to Searchbar 1`] = `
       },
       "shadowOpacity": 0,
       "shadowRadius": 0,
+      "start": undefined,
       "top": undefined,
     }
   }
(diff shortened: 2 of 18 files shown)


lukewalczak (approved):
Awesome, thanks @szaboopeeter! 🎉

Issue #3572, opened 2023-01-03
`Appbar.Header.title` renders custom components high on Android


thehale:
### Current behaviour

The `Appbar.Header`'s `title` prop does accept custom components but does not render them vertically centered in the app bar on Android (iOS and Web seem to work fine). Re-rendering the Appbar via a manual press fixes the issue.

### Expected behaviour

The Appbar should render the custom title component vertically centered on the first render.

### How to reproduce?

https://snack.expo.dev/@thehale/react-native-paper-issue-3572

### Preview

https://user-images.githubusercontent.com/47901316/210291700-ad034b99-40c2-4258-9107-07b975681a08.mp4

#### Wrapping the Button in a View with `flex` settings
Wrapping my Button in a separate View with a red background color reveals that the Button is filling the full space available to it. As such, adjusting flexbox alignment/justifications have no effect on the Button's initially rendered position.

#### Automatically triggering a re-render
Also, automatically forcing a re-render does not appear to fix the button's position. Only a manual press seems to do the trick.
- See the commented line of code in the Snack.

#### Changing the `mode` prop on `Appbar.Header`
This alignment issue shows up on `center-aligned` and `small` modes, but not on `medium` or `large`.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 15.1+
| android                        | 10+
| react-native                | 0.68
| react-native-paper    | 5.1.2
| node                            | 18.10.0
| npm or yarn                | 1.22.19
| expo sdk                     | 45+

lukewalczak:
I think the problem arises from the fact that `Button` passed into the `title` prop is then rendered within `Text` [component](https://github.com/callstack/react-native-paper/blob/main/src/components/Appbar/AppbarContent.tsx#L151). By default `Text` has variant `titleLarge` which has line-height is equal to "28" and accommodates that space, however`Button`'s height is `40`. Probably that's why it jumps, so I would suggest to add `titleStyle={{ height: 40 }}` into `Appbar.Content`.

thehale:
I updated the [linked Snack](https://snack.expo.dev/@thehale/react-native-paper-issue-3572) with your comment, and while `titleStyle={{ height: 40 }}` successfully stops the jumping around, it left the button positioned high.

The style settings that kept the button correctly centered without jumping are `titleStyle={{ height: 40, top: 12 }}`. The extra `top` style pushes the button down the extra 12-pixel height difference between the `Button` and the `Text`.
[...]

PR #3596, opened 2023-01-15
fix: appbar content title container adjustments

lukewalczak:
### Summary

PR introduces adjustments for handling the `title` prop in `Appbar.Content` to wrap **only** the title string within the default `Text` component. Other types passed into `title` are rendered directly in the `View` container.

#### Related issue

- #3572 

### Test plan

Added unit tests.

### DIFF
--- a/src/components/Appbar/AppbarContent.tsx
+++ b/src/components/Appbar/AppbarContent.tsx
@@ -18,17 +18,22 @@ import type { $RemoveChildren, MD3TypescaleKey, ThemeProp } from '../../types';
 import Text from '../Typography/Text';
 import { modeTextVariant } from './utils';
 
+type TitleString = {
+  title: string;
+  titleStyle?: StyleProp<TextStyle>;
+};
+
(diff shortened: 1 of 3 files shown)


krozniata:
LGTM! 
It would be also nice to add example to `Usage` using component as a title (maybe the one from tests)

krozniata (commented):
LGTM!
It would be nice to add example to `Usage` using component as a title (maybe the one from tests)

gedu:
if `title` is a string using the `titleStyle` doesn't make sense, what do you think to split the type into something like: 
```
type TitleString = {
  /**
   * Text for the title.
   */
  title: string;
  titleStyle?: never;
};

type TitleElement = {
  /**
   * Element for the title.
   */
  title: React.ReactElement;
  /**
   * Style for the title.
   */
  titleStyle?: StyleProp<TextStyle>;
};

// remove title and titleStyle from Props
export type Props = $RemoveChildren<typeof View> & { ... } & (TitleElement | TitleString);
```
So if you try to add `title` as string and `titleStyle`, TS warns you.

Issue #3589, opened 2023-01-10
`TextInput.Affix` prop `onLayout` is not passed to component


mfds:
### Current behaviour
Prop `onLayout` for `TextInput.Affix` (https://github.com/callstack/react-native-paper/blob/main/src/components/TextInput/Adornment/TextInputAffix.tsx#L24) is not being passed to the Affix component (https://github.com/callstack/react-native-paper/blob/main/src/components/TextInput/Adornment/TextInputAffix.tsx#L117)

### Expected behaviour
To be able to pass a callback that gets called when layout changes.

### How to reproduce?
```jsx
<RNPTextInput.Affix onLayout={(e) => console.log(e)} text="Hello" />;
```

### Preview
N/A

lukewalczak:
Thanks @mfds, I will check the issue ✌🏽

mfds:
Thanks

PR #3597, opened 2023-01-15
fix: calls onLayout on input affix

lukewalczak:
### Summary

Fixes missing possibility to invoke `onLayout` on `TextInput.Affix` which is mentioned in the documentation.

#### Related issue

- #3589 

### Test plan

Added unit test.


brunohkbx (approved):

mfds:
Thanks for looking into this. Would it be backported to 4.x?

lukewalczak:
> Thanks for looking into this. Would it be backported to 4.x?

Hey @mfds, it won't be, we are not planning to release patches for v4 anymore.

Issue #3595, opened 2023-01-13
React Native V0.71.0 no longer relies on @types/react-native


stakez:
### Current behaviour
First, thanks for all the great work that you are doing on the Library. With the upgrade to React Native V0.71.0 which no longer uses @types/react-native, the IDE is dumping typescript linting errors everywhere especially around missing style properties. Do you recommend any works arounds or is this a future effort?

| software                      | version
| --------------------- | -------
| react-native                |0.70.0
| react-native-paper    |^5.1.2

Issue #3517, opened 2022-12-08
How to avoid default undefined parameters in various components?


psk-dreampay:
Checkbox, Fab, HelperText, IconButton, RadioButton, TextInput, AppBar.BackAction and TextInput.Icon

lukewalczak:
Please check if adding `@types/react-native@0.70.7` will solve your issue.

psk-dreampay:
It's working now thanks a lot @lukewalczak

elilabes:
@lukewalczak I have tried installing `@types/react-native@0.70.7` and the current release `@types/react-native@0.71.0`. Neither version fixes this issue for me. It started for me after upgrading from 5.0.0-rc9 to 5.1.3.

Any ideas? :)

raajnadar:
@elilabes did you try deleting & regenerating the lock (yarn.lock or package.lock) file?

elilabes:
@raajnadar yes - same issue

warrenfalk:
I am also seeing this issue: react-native-paper@^5.1.3, @types/react-native@0.70.9 (also occurs with 0.71.0)

The issue is that `react-native-paper/lib/typescript/components/TextInput/TextInput.d.ts` does a `Pick<>` on `import("react-native").TextInputProps` with `"accessibilityLabelledBy"` and `"accessibilityLanguage"` which do not exist on that type

warrenfalk:
It looks like these two properties were only merged into the react-native typescript definitions a few hours ago https://github.com/facebook/react-native/pull/35883 and https://github.com/facebook/react-native/pull/35883

I'm not sure how react-native-paper already had them before react-native did, nor why a search in react-native-paper github does not find these in the code.

warrenfalk:
After updating to 5.1.4, I no longer see the issue 👍️🙏

elilabes:
Working now! Thanks @DimitarNestorov

PR #3603, opened 2023-01-18
fix: TypeScript bloat

DimitarNestorov:
### Summary

Related to #3595 #3517

The `lib/typescript` folder is filled with a lot of bloat (as shown in screenshot below). This PR removes all instances of that kind of bloat and adds a script to prevent a regression.
image

For example: the source code doesn't contain any `accessibilityLabelledBy` but there are 40 occurrences in `lib/typescript`. This causes an issue in React Native 0.71 where this prop has a type of unknown and is required (the prop is missing from the type definitions in 0.71.0).

|Before|After|
|-|-|
|image|image|
|image|image|

### Test plan

Ran `yarn test`, `yarn lint`, and `yarn typescript` with no errors.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,7 @@ jobs:
           name: Build package
           command: |
             yarn prepare
+            node ./scripts/typescript-output-lint
 
   build-docs:
     executor: default
--- a/scripts/typescript-output-lint.js
+++ b/scripts/typescript-output-lint.js
@@ -1,0 +1,113 @@
+const { promises: fs } = require('fs');
+const path = require('path');
+
+const root = path.resolve(__dirname, '..');
+const output = path.join(root, 'lib', 'typescript');
+
+/**
+ * List of React Native props not used by React Native Paper.
+ * Feel free to delete props from this list when
+ * React Native Paper has defined/uses one with the same name.
+ * This script was originally created to detect if TypeScript
+ * has exported a large union of props from ViewProps.
+ * More info: https://github.com/callstack/react-native-paper/pull/3603
+ */
+const unusedViewProps = [
+  'nativeID',
+  'accessibilityActions',
+  'accessibilityRole',
+  'accessibilityValue',
+  'onAccessibilityAction',
+  'accessibilityLabelledBy',
+  'accessibilityLiveRegion',
+  'accessibilityLanguage',
+  'accessibilityViewIsModal',
+  'onAccessibilityEscape',
+  'onAccessibilityTap',
+  'onMagicTap',
+  'accessibilityIgnoresInvertColors',
+  'hitSlop',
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
   ],
   "scripts": {
     "typescript": "tsc --noEmit",
-    "lint": "eslint --ext '.js,.ts,.tsx' . --fix",
+    "lint": "eslint --ext '.js,.ts,.tsx' . --fix && node ./scripts/typescript-output-lint",
```
satya164, line 37: Since the `output-lint` is checking built files that only get generated for publishing and not when devs are working on a component, this will check outdated files most of the time. I'd suggest moving this check to CI after package is built - if it really needs to be run against built files.

scripts/typescript-output-lint.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+async function main() {
+  for await (const file of getFiles(output)) {
+    const content = await fs.readFile(file);
+    for (const prop of unusedViewProps) {
+      if (content.includes(prop)) {
+        throw new Error(
+          `Found text '${prop}' in '${file}'. Please use the wrapped 'fetchRef' in 'src/utils/fetchRef.ts', export some return types, or modify 'scripts/typescript-output-lint.js'`
```
satya164, line 100: Do you mean `forwardRef`?

package.json
DimitarNestorov: Files are generated after `yarn install` but yeah, I'll move it
image

scripts/typescript-output-lint.js
DimitarNestorov: Yes 😄

src/utils/forwardRef.tsx
```diff
@@ -0,0 +1,15 @@
...
+  PropsWithoutRef,
+  RefAttributes,
+  ForwardRefExoticComponent,
+} from 'react';
+
+export type ForwarRefComponent<T, P = {}> = ForwardRefExoticComponent<
+  PropsWithoutRef<P> & RefAttributes<T>
+>;
+
+export const forwardRef: <T, P = {}>(
```
lukewalczak, line 13: @DimitarNestorov could you please explain (maybe add even the comment there) why do we need the custom`forwardRef`?

.circleci/config.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           name: Typecheck files
           command: |
             yarn typescript
+            node ./scripts/typescript-output-lint
```
satya164, line 68: `yarn typescript` doesn't generate the `dist` folder. this needs to be moved to `build-package` job

lukewalczak (commented):
Thanks @DimitarNestorov! 🎉

PR #3583, opened 2023-01-09
fix: Lint warning fixes

gedu:
### Summary

ESLint showed many errors, I fixed most of them, and some were suppressed. 

Screen Shot 2023-01-09 at 13 29 48

### Test plan

Run `yarn lint` and you should get no lint errors

PR #3605, opened 2023-01-19
refactor: Fix exhaustive deps linting errors

tjaniczek:
### Summary

As an addition to https://github.com/callstack/react-native-paper/pull/3583 this PR fixes suppressed `exhaustive deps` errors in:  ProgressBar, Banner, Modal

### Test plan

There should be no functional changes at all after this refactor. Added unit tests to check for regression where applicable.

### Notes

After https://github.com/callstack/react-native-paper/pull/3583 gets merged there will be some conflicts which I'll happily resolve.

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -149,6 +149,18 @@ const Banner = ({
     measured: false,
   });
 
+  const showCallback = React.useRef<Props['onShowAnimationFinished']>(
+    onShowAnimationFinished
+  );
+  const hideCallback = React.useRef<Props['onHideAnimationFinished']>(
+    onHideAnimationFinished
+  );
+
+  React.useEffect(() => {
+    showCallback.current = onShowAnimationFinished;
+    hideCallback.current = onHideAnimationFinished;
+  });
+
   const { scale } = theme.animation;
 
   React.useEffect(() => {
@@ -158,14 +170,14 @@ const Banner = ({
         duration: 250 * scale,
         toValue: 1,
         useNativeDriver: false,
-      }).start(onShowAnimationFinished);
+      }).start(showCallback.current);
     } else {
       // hide
       Animated.timing(position, {
         duration: 200 * scale,
         toValue: 0,
         useNativeDriver: false,
-      }).start(onHideAnimationFinished);
+      }).start(hideCallback.current);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [visible, position, scale]);
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { Animated, StyleProp, StyleSheet, View, ViewStyle } from 'react-native';
 
+import useEventCallback from 'use-event-callback';
+
 import { useInternalTheme } from '../core/theming';
 import type { $RemoveChildren, ThemeProp } from '../types';
 import Button from './Button/Button';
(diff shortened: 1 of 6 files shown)


src/components/Modal.tsx
```diff
@@ -189,6 +160,27 @@ function Modal({
     });
   };
 
+  React.useEffect(() => {
+    if (!visible) {
```
jaworek, line 164: Does this effect does not need dependency array? It seems that it should have `[visible]`, but maybe I'm missing something. There's second effect in this file that also does not have it, so it seems intentional, but I don't understand its purpose.

tjaniczek: This effect uses `visible`, `dismissable` & `hideModal`. As you can see we create `hideModal` on every render so no point of putting these three into dependency array - it would rerun on each render anyway.

We need this subscription in `useEffect` though. Some operations (like subscriptions, assigning to refs etc) have to wait until React "does it's thing". That's because in Fiber React may halt/skip/rerun rendering of any component at any time. That's why body of a component needs to be a pure function, with an _escape hatch_ of `useEffect` to introduce some impurity if required.

At least that's how I understand it 😅

src/components/Banner.tsx
```diff
@@ -148,6 +148,18 @@ const Banner = ({
...
+    onShowAnimationFinished
+  );
+  const hideCallback = React.useRef<Props['onHideAnimationFinished']>(
+    onHideAnimationFinished
+  );
+
+  React.useEffect(() => {
+    showCallback.current = onShowAnimationFinished;
+    hideCallback.current = onHideAnimationFinished;
+  });
```
satya164, line 161: the project has a dependency `use-event-callback` which could be used for this as well

src/components/Modal.tsx
satya164: I'd add a dependency array and put `hideModal` to a event callback/ref to avoid effect run when it changes.

Generally, it's fine since it's just a subscription, however backehandler listeners depend on the order of the listener being added: return value from  event listeners added later would override the return value from events added previously, so it's important not to resubscribe the event while modal is being displayed - otherwise user won't be able to override the back button behaviour consistently depending on whether there was a rerender.

jaworek: Thank you both for thorough explanation!

tjaniczek:
Hello @satya164, thanks for the feedback! `Modal` refactored as requested. Also added a couple of tests to cover all the quirks:
```
    when open
      if closed via touching backdrop
        ✓ will run the animation but not fade out (511 ms)
        ✓ should invoke the onDismiss function after the animation (486 ms)
      if closed via Android back button
        ✓ will run the animation but not fade out (479 ms)
        ✓ should invoke the onDismiss function after the animation (480 ms)
    when open as non-dismissible modal
      if closed via touching backdrop
        ✓ will run the animation but not fade out (240 ms)
        ✓ should not invoke onDismiss (234 ms)
      if closed via Android back button
        ✓ will run the animation but not fade out (240 ms)
        ✓ should not invoke onDismiss (245 ms)
    when visible prop changes
      from false to true (closed to open)
        ✓ should run fade-in animation on opening (248 ms)
      from true to false (open to closed)
        ✓ should run fade-out animation on closing (236 ms)
        ✓ should not invoke onDismiss (236 ms)
        ✓ should close even if the dialog is not dismissible (236 ms)
    when visible prop changes again during the open/close animation
      while closing, back to true (visible)
        ✓ should keep the modal open (244 ms)
      while opening, back to false (hidden)
        ✓ should keep the modal closed (242 ms)
```

Please note `it will...` instead of `it should...` - these specify quirks from previous implementation that are probably bugs. I tried to keep the behaviour identical to avoid any regressions. We can address this issues sometime in the future 👍

lukewalczak (approved):

PR #3606, opened 2023-01-19
refactor: replace withInternalTheme HOC with useInternalTheme

boguszkaszowski:
### Summary

PR replaces use of withInternalTheme HOC with useInternalTheme across all components. Changes shouldn't introduce any functional/visual changes.

### Test plan

Maybe it's worth to go through all components and check if they behave like they used to.


lukewalczak (approved):
Thanks @boguszkaszowski for your contribution!

Issue #3618, opened 2023-01-24
AnimatedFAB Animation


iM-GeeKy:
### Ask your Question

This question is two fold and I'm happy to move this to a discussion if that is the best fit. The first question is a typescript issue that was introduced in the last release and the other is more of an implementation question.

1. The first issue is since the last release `5.1.4` the typescript compiler appears to complain about the following style on the `AnimatedFAB`. The code is in the following [snack](https://snack.expo.dev/@fitzwabs/efe599).
```
style={[
  styles.position,
  { transform: [{ scale: animation }] }, // TYPESCRIPT ERROR HERE (Type 'Value' is not assignable to type 'number')
]}
```
I've had this in my personal project for many previous versions and it never complained until now. Is this an issue or should I restructure the animation? I can use `animation as any`, as a temporary resolution, but wouldn't like to settle on that long term.

2. Long story short, I'm trying to mock the [moving across tabs](https://m3.material.io/components/floating-action-button/guidelines#4dd3a071-5250-4b4b-97b8-78c7ca9e808f) behavior for the FAB suggested by Material Design.

An example is provided in the following [snack](https://snack.expo.dev/@fitzwabs/efe599). Note if you toggle the boolean on the following line

```
extended={isFabExtended} // Toggle this to false to see wrongly positioned start animation.
```

and switch tabs you'll notice the `extended={false}` FAB animates from what looks like the center of the `extended={true}` FAB. When the FAB is `extended={true}` is seems to behave properly.

### Preview

<img src="https://user-images.githubusercontent.com/7604441/214392942-2b3daffa-d30f-4374-a15b-b8b9e8b8c5bc.png" width="240" height="500">

Note: If I inspect the `AnimatedFAB` it looks like it still retains it's `extended={true}` width which might be why the `extended={false}` FAB still animates from the same location. Is this an implementation issue or do you have any ideas why this might be misbehaving?

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 16.2
| android                        | x
| react-native                | 0.70.5
| react-native-paper    | 5.1.4
| node                            | 16.16.0
| npm or yarn                | yarn
| expo sdk                     | 47.0.13

lukewalczak:
@DimitarNestorov could you please take a look on the first point from the issue, related to the TypeScript?

lukewalczak:
[...]

PR #3622, opened 2023-01-25
fix: `AnimatedFAB['style']` type

DimitarNestorov:
### Summary

Related to #3618

`AnimatedFAB['style']` didn't accept `Animated.Value`.
The props are [already an intersection](https://github.com/callstack/react-native-paper/blob/2854b639ffbe140233d08741857708d47195ffb6/src/components/FAB/AnimatedFAB.tsx#L35) of the `Surface` props [which have a correct style type](https://github.com/callstack/react-native-paper/blob/2854b639ffbe140233d08741857708d47195ffb6/src/components/Surface.tsx#L21). ~~The fix was deleting 1 line, deleting the style prop defined in `AnimatedFAB.tsx`: https://github.com/callstack/react-native-paper/commit/3cfdda0c437a55e7c9565773b159e3b9279d9151#diff-975af6e6afe37f64a68ad2557bf0169e45a531c77b0ba2528bbaab8db736db33L99~~
Copied Surface `Props['style']` type due to docs generation [as discussed](#pullrequestreview-1268992712)

### Test plan

Converted `AnimatedFAB.test.js` to TypeScript and wrote a simple test. `yarn typescript` will fail if we revert the changes in `AnimatedFAB.tsx`.
image

### DIFF
--- a/src/components/FAB/AnimatedFAB.tsx
+++ b/src/components/FAB/AnimatedFAB.tsx
@@ -96,7 +96,7 @@ export type Props = $RemoveChildren<typeof Surface> & {
    * Color mappings variant for combinations of container and icon colors.
    */
   variant?: 'primary' | 'secondary' | 'tertiary' | 'surface';
-  style?: StyleProp<ViewStyle>;
+  style?: Animated.WithAnimatedValue<StyleProp<ViewStyle>>;
   /**
    * @optional
    */
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "@types/react-dom": "^18.0.8",
     "@types/react-native": "^0.70.6",
     "@types/react-native-vector-icons": "^6.4.1",
+    "@types/react-test-renderer": "^18.0.0",
     "@typescript-eslint/eslint-plugin": "^5.41.0",
     "@typescript-eslint/parser": "^5.41.0",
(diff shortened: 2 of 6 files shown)


src/components/FAB/AnimatedFAB.tsx
```diff
@@ -96,7 +96,6 @@ export type Props = $RemoveChildren<typeof Surface> & {
    * Color mappings variant for combinations of container and icon colors.
    */
   variant?: 'primary' | 'secondary' | 'tertiary' | 'surface';
-  style?: StyleProp<ViewStyle>;
```
lukewalczak, line 99: Unfortunately, we cannot remove it from there, because it's needed for the purpose of docs generation. You can check on the [docs output](https://output.circle-artifacts.com/output/job/68afc86d-6e58-4bab-8ff7-4fecd5645bd2/artifacts/0/docs/animated-fab.html) there prop `style` is missing which can be confusing. I'm aware it's not ideal, but I guess we need to copy the type from the `Surface`

DimitarNestorov: How about referencing `SurfaceProps['style']` (pushed)?

DimitarNestorov:
Superseded by #3623

PR #3632, opened 2023-01-30
chore: linter rules to detect `forwardRef` usage

DimitarNestorov:
### Summary

I noticed that I missed to remove `React.forwardRef` from some places when working on #3603 so I decided to make a couple of ESLint rules

image
image

|Before|After|
|-|-|
|image|image|
|image|image|

### Test plan

CI passes

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -54,6 +54,15 @@
     "react-native-a11y/has-valid-accessibility-descriptors": "off"
   },
 
+  "overrides": [
+    {
+      "files": ["*.test.js", "*.test.tsx"],
+      "rules": {
+        "local-rules/no-react-forwardref-usage": "off"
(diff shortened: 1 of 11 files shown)


src/components/__tests__/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "local-rules/no-react-forwardref-usage": "off"
```
lukewalczak, line 3: Question: is there an option to ignore files with the extension `*.test.js` in the custom eslint rule? My point being, that file can be unnecessary.

lukewalczak: Sounds good!

lukewalczak (approved):

PR #3638, opened 2023-01-31
refactor: poc button theme

gedu:
### Issue

With the addition of Material Design 3 and making it work with v2 a new variable was created `isV3`. This allowed the load of different UI but added extra responsibility to the components. On the other hand, it makes the component harder to scale, and we ended up with many nested conditions. 

### Summary

Paper_theme_refactor

My idea is to remove any knowledge about versions from components, so it only takes care of behavior and shows the right UI depending on the state.

Reassure
afterRefactor

I compare the current and new, and on average (runner 10 times) the new architecture was faster by 0.35s

Note: I made the new structure into the `ButtonNew` component, so I get thoughts and inputs about it before fixing the tests snapshots 

### Test plan

Run the component in the example app. 
 - Check for UI changes 
 - Ensure that the functionality remains the same
 
Note: You will need to update the index.ts import to use the `ButtonNew` component.


lukewalczak:
Currently closing the PR, however, we will use it as one of the resources for the next Paper team's workshop.

PR #3642, opened 2023-02-01
refactor: fix performance of TextInput

tjaniczek:
Performance optimizations for TextInput component.

### Summary

The most I could squeeze off it without major refactor - just by avoiding unnecessary rerenders and interpolations. From my calculations  we managed to:

- speed up first render by 25%
- speed up subsequent renders by 35%
- speed up rerenders after each key stroke by 40%

### Test plan

Check for visible regressions in the test app

### DIFF
--- a/src/components/TextInput/Label/InputLabel.tsx
+++ b/src/components/TextInput/Label/InputLabel.tsx
@@ -50,27 +50,42 @@ const InputLabel = (props: InputLabelProps) => {
     fontSize,
     lineHeight,
     fontWeight,
+    opacity: hasActiveOutline
+      ? parentState.labeled.interpolate({
+          inputRange: [0, 1],
+          outputRange: [1, 0],
+        })
+      : 0,
     transform: [
       {
         // Wiggle the label when there's an error
-        translateX: parentState.error.interpolate({
-          inputRange: [0, 0.5, 1],
-          outputRange: [0, parentState.value && error ? wiggleOffsetX : 0, 0],
-        }),
+        translateX:
+          parentState.value && error
+            ? parentState.error.interpolate({
+                inputRange: [0, 0.5, 1],
+                outputRange: [0, wiggleOffsetX, 0],
+              })
+            : 0,
       },
       {
         // Move label to top
-        translateY: parentState.labeled.interpolate({
-          inputRange: [0, 1],
-          outputRange: [baseLabelTranslateY, 0],
-        }),
+        translateY:
+          baseLabelTranslateY !== 0
+            ? parentState.labeled.interpolate({
+                inputRange: [0, 1],
+                outputRange: [baseLabelTranslateY, 0],
+              })
+            : 0,
       },
       {
         // Make label smaller
-        scale: parentState.labeled.interpolate({
-          inputRange: [0, 1],
-          outputRange: [labelScale, 1],
-        }),
+        scale:
+          labelScale !== 0
+            ? parentState.labeled.interpolate({
+                inputRange: [0, 1],
+                outputRange: [labelScale, 1],
+              })
+            : parentState.labeled,
       },
     ],
   };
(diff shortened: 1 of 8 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -369,29 +374,37 @@ const TextInput = forwardRef<TextInputHandles, Props>(
...
       }
     }, [focused, value, labeled, scale]);
 
     const onLeftAffixLayoutChange = (event: LayoutChangeEvent) => {
-      setLeftLayout({
-        height: event.nativeEvent.layout.height,
-        width: event.nativeEvent.layout.width,
-      });
+      const height = roundLayoutSize(event.nativeEvent.layout.height);
+      const width = roundLayoutSize(event.nativeEvent.layout.width);
```
tjaniczek, line 388: Sometimes the event would trigger with small change well into the milionths (e.g. 5.000000001 vs 5.5.000000002) which would trigger rerender without visible difference to the user. Rounding should solve the problem.

src/components/TextInput/Label/InputLabel.tsx
```diff
@@ -50,27 +50,42 @@ const InputLabel = (props: InputLabelProps) => {
...
-          inputRange: [0, 0.5, 1],
-          outputRange: [0, parentState.value && error ? wiggleOffsetX : 0, 0],
-        }),
+        translateX:
+          parentState.value && error
+            ? parentState.error.interpolate({
+                inputRange: [0, 0.5, 1],
+                outputRange: [0, wiggleOffsetX, 0],
+              })
+            : 0,
```
tjaniczek, line 68: To clarify all cases with `interpolate` updates:  
- sometimes it would interpolate to `[0, 0]` so there's no point of interpolating at all, we could pass `0` 
- sometimes it would interpolate from `[1, 0]` to `[1, 0]` - once again no point, we could pass the original `Animated.value` instead

src/components/TextInput/TextInputFlat.tsx
```diff
@@ -349,7 +348,9 @@ const TextInputFlat = ({
             ]}
           />
         )}
-        <InputLabel parentState={parentState} labelProps={labelProps} />
+        {label ? (
+          <InputLabel parentState={parentState} labelProps={labelProps} />
+        ) : null}
```
tjaniczek, line 353: Avoids all the calculations in `InputLabel` since it would return `null` in the end anyway...

src/components/TextInput/helpers.tsx
```diff
@@ -245,15 +243,6 @@ export const adjustPaddingFlat = ({
...
 };
 
-export const interpolatePlaceholder = (
-  labeled: Animated.Value,
-  hasActiveOutline: boolean | undefined
-) =>
-  labeled.interpolate({
-    inputRange: [0, 1],
-    outputRange: [hasActiveOutline ? 0 : 1, 1],
-  });
```
tjaniczek, line 255: `hasActiveOutline` was `true` in both cases when this function is used, so no point of having it at all

src/components/TextInput/TextInput.tsx
```diff
@@ -344,17 +345,21 @@ const TextInput = forwardRef<TextInputHandles, Props>(
...
+        }
       }
 
       return () => {
         if (timer.current) {
           clearTimeout(timer.current);
         }
       };
-    }, [focused, rest.label, rest.placeholder]);
+    }, [focused, placeholder, rest.label, rest.placeholder]);
```
DimitarNestorov, line 359: Setting the state with the same value doesn't trigger a rerender. Am I missing something? Made a quick CodeSandbox: https://codesandbox.io/s/react-new?file=/src/App.js

RichardLindhout:
React.memo around the component can also speed up the component

tjaniczek:
> React.memo around the different components can also greatly speed up the component!

Yes, that would solve most of our pain with this component, if only we could safely memo this component. Our users need to pass callbacks to `TextInput` and if they are not memoized - `React.memo` would have a serious negative impact on the performance. 

There may be a valid case to memoize internal components, but imho that would require rather large restructuring of the code. A story for another time...

lukewalczak (approved):

PR #3645, opened 2023-02-01
feat: allow override of custom theme animation

Filippo39:
### Summary

This pull request wants to allow an override of the animation field of the default theme.
Before it was possible now not anymore.

### DIFF
--- a/src/core/Provider.tsx
+++ b/src/core/Provider.tsx
@@ -86,6 +86,7 @@ const Provider = (props: Props) => {
       ...props.theme,
       version: themeVersion,
       animation: {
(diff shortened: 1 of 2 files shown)


src/core/Provider.tsx
```diff
@@ -86,6 +86,7 @@ const Provider = (props: Props) => {
       ...props.theme,
       version: themeVersion,
       animation: {
+        ...props.theme?.animation,
```
lukewalczak, line 89: Could you please provide a unit test for customizing animation to avoid another regression?

lukewalczak (approved):

Issue #3652, opened 2023-02-05
Button flex problem on iOS


szado:
### Current behaviour
There is a problem with buttons with `flex: 1` on iOS - buttons are not visible.

### Expected behaviour
Buttons should have proper width and be visible like on Android or web.

### How to reproduce?
https://snack.expo.dev/@szado/rnp-ios-button-problem?platform=ios
`but1` should occupy the entire available width instead it is invisible.

### Preview
IOS:
![obraz](https://user-images.githubusercontent.com/20907536/216799105-1eb7921b-652e-4cff-afca-c407917488f3.png)
Expected behaviour:
![obraz](https://user-images.githubusercontent.com/20907536/216799117-79481707-abcf-4069-a723-64eb7e8d8608.png)

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x
| android                        | x
| react-native                | newest
| react-native-paper    | newest
| expo sdk                     | newest

lukewalczak:
I will take a look at that

PR #3653, opened 2023-02-05
fix: passing flex to the shared and inner layer Surface styles

lukewalczak:
### Summary

PR refines styles in the `Surface` views for `iOS` platform, to pass the `flex` into all its layers.

#### Related issue

- #3652 

### Test plan

Added unit test covering Surface's layer styles.

### DIFF
--- a/src/components/Surface.tsx
+++ b/src/components/Surface.tsx
@@ -113,7 +113,7 @@ const Surface = forwardRef<View, Props>(
       children,
       theme: overridenTheme,
       style,
-      testID,
+      testID = 'surface',
       ...props
     }: Props,
     ref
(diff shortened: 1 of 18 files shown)


src/components/Surface.tsx
```diff
@@ -227,6 +227,7 @@ const Surface = React.forwardRef<View, Props>(
       bottom,
       start,
       end,
+      flex,
```
gedu, line 230: I would move out of the function of the `iOSShadowOutputRanges` (line 206) Array creation. It has all hardcoded values

lukewalczak: Sure, I will extract all the shadow creation functions to the utils

gedu (approved):
LGTM, just an small comment

src/components/Surface.tsx
lukewalczak: To not block the release process, I will address the Surface refactor in the separate PR.

PR #3659, opened 2023-02-06
fix: allow for custom variants in `Text` (with declaration merging)

DimitarNestorov:
### Summary

Fixes #3658
Closes #3660 

- Updated docs
- Updated example
- Added tests

|iOS|Android|
|-|-|
|image|image|

### Test plan

CI passes. Example works as expected.

PR #3660, opened 2023-02-06
fix: allow for custom variants in `Text` (with factory)

### LINKED ISSUE #3658: Add customVariant example is not working
### Current behaviour
Cant add custom variants with steps from docs.

### Expected behaviour
Add custom font variant properly, would be best with typesafety in variant prop.

### How to reproduce?
I use @lukewalczak snack and add customVariant like said in docs with huge fontSize, but there is no effect.
[snack](https://snack.expo.dev/nqXz0JUGk)

### Your Environment

| react-native                | 0.71.1 (bare rn)
| react-native-paper    | 5.1.3

DimitarNestorov:
### Summary

Fixes #3658
Closes #3659

- Updated docs
- Updated example
- Added tests

|iOS|Android|
|-|-|
|image|image|

### Test plan

CI passes. Example works as expected.

### DIFF
--- a/docs/docs/guides/04-fonts.md
+++ b/docs/docs/guides/04-fonts.md
@@ -393,7 +393,6 @@ Platform.select({
 
 #### Using `configureFonts` helper
 
-<div style={{display: 'none'}}>
 * If there is a need to create a custom font variant, prepare its config object including required all fonts properties. After that, defined `fontConfig` has to be passed under the <b>`variant`</b> name as `config` into the params object:
 
 ```js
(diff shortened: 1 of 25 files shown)


docs/pages/4.fonts.mdx
```diff
@@ -428,7 +427,16 @@ export default function Main() {
   );
 }
 ```
-</div>
+
+If you're using TypeScript you will need to create a custom `Text` component which accepts your custom variants:
+
+```typescript
+import { createText } from 'react-native-paper'
```
lukewalczak, line 434: Shouldn't be `customText`?

Tbh I prefer that approach 👍🏽 imo from the user perspective it's also more _intuitive_

docs/pages/4.fonts.mdx
```diff
@@ -428,7 +427,16 @@ export default function Main() {
...
+
+// Use this instead of importing `Text` from `react-native-paper`
+export const Text = createText<'customVariant'>
```
lukewalczak, line 437:
```suggestion
export const Text = createText<'customVariant'>()
```

docs/pages/4.fonts.mdx
DimitarNestorov: 🤦‍♂️ I should make TypeScript go through the docs

jaworek (approved):

PR #3666, opened 2023-02-09
docs: hide navbar on scroll

yamankatby:


yamankatby:
Closed in favour of #3707

PR #3674, opened 2023-02-12
Text color change on SegmentedButtons

Alenii:
The text in SegmentedButtons is set only by the onSurface color property if the theme is V3, but I think it would be nicer if the text color can either be customised by a new properties for the button prop called labelColor & checkedlabelColor  or at least reverses color like in my pull request, thank you.


lukewalczak:
Closing in favour of: https://github.com/callstack/react-native-paper/pull/3701

PR #3676, opened 2023-02-13
feat: add react-navigation integration for BottomNavigation

satya164:
### Summary

- Added an integration with React Navigation (`createMaterialBottomTabNavigator`) which can be imported from `react-native-paper/react-navigation`
- Move the tab bar from `BottomNavigation` into a new `BottomNavigationBar` component
- Added examples

React Navigation integration example
BottomNavigationBar example

### Test plan

- Tests work
- Examples work


package.json
```diff
@@ -48,6 +49,7 @@
   },
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.8",
+    "@react-navigation/elements": "^1.3.15",
```
satya164, line 52: is this necessary for bottom navigation? if yes then we have to update our code not to need it. we can't have react navigation specific packages in the dependencies.

src/react-navigation/views/MaterialBottomTabView.tsx
```diff
@@ -0,0 +1,200 @@
...
+  state: TabNavigationState<ParamListBase>;
+  navigation: MaterialBottomTabNavigationHelpers;
+  descriptors: MaterialBottomTabDescriptorMap;
+};
+
+type Scene = { route: { key: string } };
+
+// Optionally require vector-icons referenced from react-native-paper:
+// https://github.com/callstack/react-native-paper/blob/4b26429c49053eaa4c3e0fae208639e01093fa87/src/components/MaterialCommunityIcon.tsx#L14
+let MaterialCommunityIcons: React.ComponentType<
```
satya164, line 30: This can probably be imported from existing implementation and removed from here

package.json
DimitarNestorov: Done: https://github.com/callstack/react-native-paper/pull/3676/commits/244d083c86e75dc9dde59a9348dfb8ca67ac4009

src/react-navigation/views/MaterialBottomTabView.tsx
DimitarNestorov: Done: https://github.com/callstack/react-native-paper/pull/3676/commits/488f08e0a61685e8b9bf1d8d9cb576078411d008

src/components/MaterialCommunityIcon.tsx
```diff
@@ -106,3 +106,5 @@ const styles = StyleSheet.create({
 });
 
 export default defaultIcon;
+
+export { MaterialCommunityIcons };
```
satya164, line 110: Why export `MaterialCommunityIcons` instead of using `defaultIcon` which is already exported and used in other places?

react-navigation/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "main": "../lib/commonjs/react-navigation/index",
+  "module": "../lib/module/react-navigation/index",
+  "react-native": "../src/react-navigation/index",
+  "types": "../lib/typescript/react-navigation/index",
+  "peerDependencies": {
+    "@react-navigation/native": "*"
```
satya164, line 7: This isn't a regular package so this won't be picked up by Yarn/npm. Do you think this is necessary to add here? It can be confusing to have dependencies that don't do anything.

DimitarNestorov: > this won't be picked up by Yarn/npm
I agree.

> Do you think this is necessary to add here?
No. The example app Metro config looks at `peerDependencies` from the root `package.json` and since we wouldn't want to add `@react-navigation/native` there decided to add it in this `package.json`. Hardcoded it in the Metro config instead: https://github.com/callstack/react-native-paper/pull/3676/commits/fd3ffa1d2a2a8434b29632d0ee6e6ab95430740b#diff-1500199fd5a82fe7d24127e13ccfe3b2916b105084b7fe9784b9cf71feed245cR18

satya164: imo then hardcode it inside the metro config. Otherwise it's confusing to have a peer dependency that doesn't actually work like peer dependency

example/src/RootNavigator.tsx
```diff
@@ -71,3 +71,14 @@ export default function Root() {
     </Stack.Navigator>
   );
 }
+
+function getHeaderTitle(
```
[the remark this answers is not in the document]
DimitarNestorov: 🤦‍♂️ There was no point in doing that. My bad.

src/components/BottomNavigation/BottomNavigationBar.tsx
```diff
@@ -0,0 +1,948 @@
...
+   * Specifies the largest possible scale a label font can reach.
+   */
+  labelMaxFontSizeMultiplier?: number;
+  /**
+   * Style for the bottom navigation bar.  You can pass a custom background color here:
+   *
+   * ```js
+   * barStyle={{ backgroundColor: '#694fad' }}
+   * ```
+   */
```
DimitarNestorov, line 207:
```suggestion
```

example/src/Examples/BottomNavigationBarExample.tsx
```diff
@@ -0,0 +1,100 @@
...
+            const event = navigation.emit({
+              type: 'tabPress',
+              target: key,
+              canPreventDefault: true,
+            });
+
+            if (event.defaultPrevented) {
+              preventDefault();
+            } else {
+              navigation.navigate({ key, merge: true });
```
satya164, line 46:
```suggestion
              navigation.navigate(route);
```

example/src/Examples/BottomNavigationBarExample.tsx
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            return null;
+          }}
+          getLabelText={({ route }) => {
+            const { options } = props.descriptors[route.key];
+
+            return typeof options.tabBarLabel === 'string'
+              ? options.tabBarLabel
+              : undefined;
```
satya164, line 62:
```suggestion
              : options.title;
```

example/src/Examples/BottomNavigationBarExample.tsx
```diff
@@ -0,0 +1,100 @@
...
+          }}
+          {...props}
```
satya164, line 64:
```suggestion
          state={props.state}
          safeAreaInsets={props.insets}
```

src/components/BottomNavigation/BottomNavigationBar.tsx
```diff
@@ -0,0 +1,1022 @@
...
+   *   routes: [
+   *     { key: 'music', title: 'Favorites', focusedIcon: 'heart', unfocusedIcon: 'heart-outline'},
+   *     { key: 'albums', title: 'Albums', focusedIcon: 'album' },
+   *     { key: 'recents', title: 'Recents', focusedIcon: 'history' },
+   *     { key: 'notifications', title: 'Notifications', focusedIcon: 'bell', unfocusedIcon: 'bell-outline' },
+   *   ]
+   * }
+   * ```
+   *
+   * `BottomNavigation` is a controlled component, which means the `index` needs to be updated via the `onTabPress` callback.
```
satya164, line 116: The references to `BottomNavigation` need to be changed to match the component name

src/components/BottomNavigation/BottomNavigation.tsx
```diff
@@ -114,14 +98,14 @@ export type Props = {
...
+   * `BottomNavigation` is a controlled component, which means the `index` needs to be updated via the `onIndexChange` or `onTabPress` callback.
    */
   navigationState: NavigationState;
   /**
    * Callback which is called on tab change, receives the index of the new tab as argument.
-   * The navigation state needs to be updated when it's called, otherwise the change is dropped.
+   * The navigation state needs to be updated when this or `onTabPress` is called, otherwise the change is dropped.
    */
-  onIndexChange: (index: number) => void;
+  onIndexChange?: (index: number) => void;
```
satya164, line 108: This is not correct. The index may change via other means than a tab press in a `BottomNavigation` component - more specifically via the `jumpTo` function passed to the scenes. Using `onTabPress` to update the state means `jumpTo` won't work.

Consumers should always use `onIndexChange` when using `BottomNavigation` component. It also needs to be required since this is a controlled component. The API of this component shouldn't change at all.

src/components/BottomNavigation/BottomNavigationBar.tsx
```diff
@@ -0,0 +1,1022 @@
...
+ *           getLabelText={({ route }) => {
+ *             const { options } = props.descriptors[route.key];
+ *
+ *             return typeof options.tabBarLabel === 'string'
+ *               ? options.tabBarLabel
+ *               : undefined;
+ *           }}
+ *           {...props}
+ *         />
+ *       )}
```
satya164, line 297:
```suggestion
 *       tabBar={({ navigation, state, descriptors, insets }) => (
 *         <BottomNavigation.Bar
 *           state={state}
 *           onTabPress={({ route, preventDefault }) => {
 *             const event = navigation.emit({
 *               type: 'tabPress',
 *               target: route.key,
 *               canPreventDefault: true,
 *             });
 *
 *             if (event.defaultPrevented) {
 *               preventDefault();
 *             } else {
 *               navigation.navigate(route);
 *             }
 *           }}
 *           renderIcon={({ route, focused, color }) => {
 *             const { options } = descriptors[route.key];
 *             if (options.tabBarIcon) {
 *               return options.tabBarIcon({ focused, color, size: 24 });
 *             }
 *
 *             return null;
 *           }}
 *           getLabelText={({ route }) => {
 *             const { options } = descriptors[route.key];
 *             const label =
 *              options.tabBarLabel !== undefined
 *                ? options.tabBarLabel
 *                : options.title !== undefined
 *                ? options.title
 *                : route.name;
 * 
 *             return label;
 *           }}
 *.          safeAreaInsets={insets}
 *         />
 *       )}
```

DimitarNestorov: Thoughts on renaming `safeAreaInsets` to `insets`?

satya164: `insets` is better but `BottomNavigation` has `safeAreaInsets` so could be confusing

DimitarNestorov: In that case I will revert `state` to `navigationState`.

lukewalczak (approved):
Thanks @DimitarNestorov for addressing all the comments and finishing the work! 🚀

Issue #3665, opened 2023-02-08
TextInput multiline captures vertical drag events on iOS


CodingItWrong:
### Current behaviour
I have a Paper `TextInput` multiline inside a `ScrollView`. Because the text input expands to show all of its contents on Android and iOS, I don't need to scroll vertically within the text input. So I would like vertical drags to scroll the outer ScrollView. This happens successfully on Android.

On iOS, however, after I have typed text input the TextInput, then vertical drags within the TextInput scroll the TextInput's contents up and down. This makes it harder for me as a user to scroll the ScrollView, especially if there are multiple multiline TextInputs.

### Expected behaviour
I would expect iOS to function like Android: vertical drags within a multiline TextInput never move its contents up and down, they just allow the outer ScrollView to scroll.

### How to reproduce?
Placing a Paper TextInput inside a RN ScrollView should be enough to reproduce the issue.

Repro repo: https://github.com/CodingItWrong/paper-sandbox/tree/text-input-issues (note this is not the main branch)

### Preview
![scroll](https://user-images.githubusercontent.com/15832198/217519787-91e08bea-b892-4d2b-b98a-6ed044487d8c.gif)

### What have you tried so far?
I considered trying to disable drag events in the TextInput, but I don't want to disable drags of the text-editing cursor itself, so I did not pursue this idea.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 16.2
| react-native                | 0.70.5
| react-native-paper    | 5.1.3
| node                            | 16.19.0
| npm or yarn                | yarn
| yarn | 1.22.19
| npm | 9.2.0
| expo sdk                     | 47.0.13

CodingItWrong:
@tjaniczek Yup, the line height fixes it great on my end as well 👍 Sounds like it's the right fix. Thank you for investigating this so quickly.

tjaniczek:
Hi @CodingItWrong - we figured out that passing down `scrollEnabled={false}` (https://reactnative.dev/docs/textinput#scrollenabled-ios) in the TextInput is much more reliable on real devices than the `lineHeight` hack. This should do the trick for you as well. Cheers!

PR #3677, opened 2023-02-13
fix: multiline TextInput captures vertical drag on iOS

tjaniczek:
### Summary

Fixes https://github.com/callstack/react-native-paper/issues/3665 

Default multiline TextInput expands to the size of text so no scrolling is needed. Unfortunately on iOS, when font size is specified but the line height is not, operating system still treats it as a scrollable container. This hijacks scroll events when user wants to move page up/down.

This PR adds default line height to all multiline inputs.

### Test plan

Changes should be visible on iOS simulator


tjaniczek:
Closed in favour of passing down `scrollEnabled={false}`

Issue #3593, opened 2023-01-12
React Native V0.71.0 Shadow Warnings


stakez:
### Current behaviour
Since upgrading to React Native 0.71.0, the following warning has been occurring
` WARN  (ADVICE) View #1853 of type RCTView has a shadow set but cannot calculate shadow efficiently. Consider setting a background color to fix this, or apply the shadow to a more specific component.`

After removing several React Native paper Components (especially Surface), the warnings dropped. After reverting to original, the warnings all return. 

### Expected behaviour
No warning should be shown.

### How to reproduce?
Run a React Native App using  V0.71.0 and view the terminal running `react-native start`

### Preview

![Screenshot 2023-01-12 at 1 37 48 PM](https://user-images.githubusercontent.com/99569934/212186665-d86cdcac-a5f3-4dd1-80cf-6306aee1b291.png)

### Your Environment

| software                      | version
| --------------------- | -------
| react-native                |0.71.0
| react-native-paper    | ^5.1.2
| node                            | 18.12.1
| npm or yarn                | yarn
| expo sdk                     | not applicable

lukewalczak:
As you noticed the issue comes form the `Surface` component where we are applying the elevation, but the views don't have a specified background color, that's why the warning appears. I will definitely check what we can do about it.

stakez:
Thanks @lukewalczak. Looking toward to what you find.

Yusk03:
Can agree
Have the same problem, hope you'll resolve it

asurare:
Having the same issue after a recent update. Is it open to PR ?

lukewalczak:
>  Is it open to PR ?

Sure, contributions are welcomed!

visoft:
This is happening with the Portal and Dialog components as well.

Streudal:
Adding on to what @visoft mentioned: this is also happening with the FAB component as well. Basically anything that has a shadow built in it seems like.

lukewalczak:
We are aware of that issue and there is a [PR](https://github.com/callstack/react-native-paper/pull/3738) which should fix the warning.

lukewalczak:
Version `5.5.0` has been released and the issue should be fixed :)

Streudal:
Everything is lookin great! Thanks for the updates!

ericpoulinnz:
@lukewalczak I'm still getting this warning in `5.5.2`. 

Seems to be the `AnimatedFAB`.

spaskhalov:
[...]

PR #3681, opened 2023-02-16
fix: apply background color to surface outer layers

BeeMargarida:
### Summary

Fixes https://github.com/callstack/react-native-paper/issues/3593.

As mentioned in the issue, using version 0.71.0 of RN causes causes warnings to be shown in the console when using the `Surface` component. This has to do with the fact that the component is defining a shadow without a view having a background color, which causes performance issues - see reasoning [here](https://github.com/facebook/react-native/commit/e4c53c28aea7e067e48f5c8c0100c7cafc031b06).

The fix was applying the background color to the outer and inner layer, which meant that it was necessary to modifying the structure of Surface to match the given width, height, margins (and remove them from the inner view) and border radius.

### Test plan

The changes made can be tested in the iOS example app. It should not be possible to see any changes, since it's expected that the behaviour and UI remains the same.

### DIFF
--- a/src/components/Surface.tsx
+++ b/src/components/Surface.tsx
@@ -229,8 +229,26 @@ const Surface = forwardRef<View, Props>(
       start,
       end,
       flex,
+      width,
+      height,
+      transform,
+      opacity,
       ...restStyle
     } = (StyleSheet.flatten(style) || {}) as ViewStyle;
+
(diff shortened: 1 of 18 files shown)


BeeMargarida:
The approach taken is not compatible with the current logic in the Animated FAB component, working on fixing this now

PR #3700, opened 2023-02-21
docs: screen flash on initial docs load

tjaniczek:
### Summary

When opening the new docs, the screen flashes for a brief moment. This is caused by Docusaurus rendering full-screen loader via `react-loadable` (https://github.com/jamiebuilds/react-loadable#avoiding-flash-of-loading-component)

I guess the way to fix it is to defer using `useColorMode` hook until page is fully hydrated.

### Test plan

Go to https://callstack.github.io/react-native-paper/index.html and use Analyze page lode via Chrome's Lighthouse.

### DIFF
--- a/docs/src/components/BannerExample.tsx
+++ b/docs/src/components/BannerExample.tsx
@@ -139,7 +139,30 @@ const BannerExample = () => {
   );
 };
 
-const WithProvider = () => {
+const Shimmer = () => {
+  /* Docusaurus won't call StyleSheet.create() server-side */
+  /* eslint-disable react-native/no-inline-styles, react-native/no-color-literals */
+
+  return (
+    <View
+      style={{
+        display: 'flex',
+        minHeight: 500,
+        borderWidth: 1,
+        borderColor: 'rgba(125, 82, 96, 0.4)',
+        borderStyle: 'solid',
+        borderRadius: 40,
+        alignContent: 'center',
+        justifyContent: 'center',
+        padding: 30,
+        marginTop: 36,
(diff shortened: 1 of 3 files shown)


docs/src/components/BannerExample.tsx
```diff
@@ -139,15 +139,19 @@ const BannerExample = () => {
...
+  return (
+    <Provider theme={isDarkTheme ? DarkTheme : DefaultTheme}>
+      <BannerExample />
+    </Provider>
+  );
+};
+
+export default function ClientSideBannerExample() {
   return (
     <BrowserOnly fallback={<ProgressBar indeterminate />}>
```
yamankatby, line 153: The issue is `react-native-paper` components don't work well outside the `BrowserOnly`

docs/src/components/GetStartedButtons.tsx
```diff
@@ -51,15 +51,15 @@ const GetStartedButton = () => {
...
-      )}
-    </BrowserOnly>
+    <Provider theme={isDarkTheme ? DarkTheme : DefaultTheme}>
+      <GetStartedButton />
+    </Provider>
   );
+};
+
+export default function ClientSideGetStarted() {
+  return <BrowserOnly>{() => <WithProvider />}</BrowserOnly>;
```
yamankatby, line 64: Looks like there is no `fallback` here

tjaniczek:
Update: added shims to avoid layout shifts while hydrating.

image

lukewalczak (approved):

Issue #3671, opened 2023-02-10
Segmented Buttons need an option to select active and inactive background/text+icon colors


Isopodus:
**Is your feature request related to a problem? Please describe.**
In our app we need to set the text and background to be inverse colors, but because of lack of active/inactive text color it is impossible to achieve.
In our case the active button background and text color are the same theme color (`onSurface`), this results in the text being unreadable:

![image](https://user-images.githubusercontent.com/29857052/218118775-3a759c52-ef13-48a7-bd32-3cfc18181eb3.png)

**Describe the solution you'd like**
After searching through the library source I came up with a small patch which adds a third text color state being `checked` color. I've set it to `onSecondaryContainer`. Below is the patch file content I used with patch-package, it basically sets the icon color to be the text color and adds another if block to apply the checked color (it should really be the `else if` of `disabled` block, but you've got the idea):

```patch
diff --git a/node_modules/react-native-paper/src/components/SegmentedButtons/SegmentedButtonItem.tsx b/node_modules/react-native-paper/src/components/SegmentedButtons/SegmentedButtonItem.tsx
index 027fdf5..0ad5f81 100644
--- a/node_modules/react-native-paper/src/components/SegmentedButtons/SegmentedButtonItem.tsx
+++ b/node_modules/react-native-paper/src/components/SegmentedButtons/SegmentedButtonItem.tsx
@@ -176,7 +176,7 @@ const SegmentedButtonItem = ({
               testID={`${testID}-check-icon`}
               style={[iconStyle, { transform: [{ scale: checkScale }] }]}
             >
-              <Icon source={'check'} size={iconSize} />
+              <Icon source={'check'} size={iconSize} color={labelTextStyle.color?.toString()}/>
             </Animated.View>
           ) : null}
           {showIcon ? (
diff --git a/node_modules/react-native-paper/src/components/SegmentedButtons/utils.ts b/node_modules/react-native-paper/src/components/SegmentedButtons/utils.ts
index 701fb90..bf1ce4b 100644
--- a/node_modules/react-native-paper/src/components/SegmentedButtons/utils.ts
+++ b/node_modules/react-native-paper/src/components/SegmentedButtons/utils.ts
@@ -124,11 +124,15 @@ const getSegmentedButtonBorderWidth = ({
 const getSegmentedButtonTextColor = ({
   theme,
   disabled,
-}: Omit<BaseProps, 'checked'>) => {
+  checked,
+}: BaseProps) => {
[...]

PR #3701, opened 2023-02-22
fix: Segment checked state

gedu:
### Summary

`SegmentButton` didn't allow to change the color of the text and icon when it was checked.
I added 2 new props, so you can change the colors through them, but also can get the colors from theme

Only available from v3

#### Related issue

- #3671 

### Test plan

Added unit tests

https://user-images.githubusercontent.com/1676818/220708651-752c7330-f8ae-4d60-a1b1-8c7d4e22dc28.MP4


src/components/__tests__/__snapshots__/SegmentedButton.test.js.snap
```diff
@@ -111,7 +111,7 @@ exports[`renders checked segmented button with selected check 1`] = `
             style={
               Array [
                 Object {
-                  "color": "rgba(28, 27, 31, 1)",
+                  "color": "rgba(29, 25, 43, 1)",
```
lukewalczak, line 114: Please double check why that color has been changed.

gedu: because this line gives true because `value` -> `walk` 
```
const checked =
          multiSelect && Array.isArray(value)
            ? value.includes(item.value)
            : value === item.value;
```
and now because I'm listening to the `checked` state in utils `getSegmentedButtonTextColor`
```
if (checked) {
      return checkedColor ?? theme.colors.onSecondaryContainer;
    }
```
it returns that new color

lukewalczak (approved):

PR #3707, opened 2023-02-24
docs: hide navbar on scroll

yamankatby:


lukewalczak:
I have no strong opinion about that feature, so I've checked a couple of other docs pages to observe the _navbar_ visibility and it appears the navbar is always present, that's why I lean towards closing that PR.

Issue #3703, opened 2023-02-23
Banner border still visible when the banner is hidden


Ross-Landry:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `react-native-paper@5.0.1` for the project I'm working on.

The issue:
I've applied some color to the bottom border of a Banner. When the banner is hidden - the bottom border is still visible. 
( Here's a [snack](https://snack.expo.dev/@realscout/paranoid-churros) demonstrating the issue )

My fix:
I applied a conditional style object to the banner container, that sets the bottom border width to 0 when the banner is not visible.

Here is the diff that solved my problem:

```
diff --git a/node_modules/react-native-paper/src/components/Banner.tsx b/node_modules/react-native-paper/src/components/Banner.tsx
index 4b52832..331f333 100644
--- a/node_modules/react-native-paper/src/components/Banner.tsx
+++ b/node_modules/react-native-paper/src/components/Banner.tsx
@@ -189,7 +189,7 @@ const Banner = ({
   return (
     <Surface
       {...rest}
-      style={[!theme.isV3 && styles.elevation, style]}
+      style={[!theme.isV3 && styles.elevation, style, !visible && styles.hiddenBorder]}
       theme={theme}
       {...(theme.isV3 && { elevation })}
     >
@@ -289,6 +289,9 @@ const styles = StyleSheet.create({
   elevation: {
     elevation: 1,
   },
+  hiddenBorder: {
+    borderBottomWidth:0
+  }
 });
 
 export default withInternalTheme(Banner);

```
<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

lukewalczak:
Hey @Ross-Landry, thanks for the bug report. Could you please attach the [snack](https://snack.expo.dev/) presenting the issue?

PR #3709, opened 2023-02-27
fix: adjust surface wrapper opacity

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3703

### Summary

According to the issue, PR adjusts the `opacity` of the `Banner` component, to set it to `0` once the component is not **visible**.

#### Related issue:

- #3703


gedu (approved):

Ross-Landry:
@lukewalczak Elegant solution! Thank you very much for taking the time

Issue #3704, opened 2023-02-24
AppBar.Content  is incorrectly labeled in the accessibility inspector as a button when no onPress is passed


fwielstra:
### Current behaviour

The AppBar's heading, when inspected via the Accessibility Inspector on iOS, is marked as and read out as an "disiabled button"

### Expected behaviour
If no `onPress` callback is given, it should not be accessible; it should read out the `Text` element of the title, since that already has the correct accessibility properties.

I'm not sure what should happen if the user passes a component as the `title` though, I presume they're on their own then?

Also a disclaimer, I'm not an expert on accessibility, I've just done a quick review of our app in development to see what it looks like. I'm not sure if this is considered an issue.

I suspect the fix is passing something like `accessible={!!onPress}` to the `TouchableWithoutFeedback` element in `src/components/Appbar/AppbarContent.tsx`

### How to reproduce?

Create an AppBar with a title:

```
      <Appbar.Header>
        <Appbar.Content title="hello" />
      </Appbar.Header>
```

- Start the app
- Open Xcode, then go to the Xcode menu -> Open Developer Tool -> Accessibility Inspector
- Select the heading

### Preview
![image](https://user-images.githubusercontent.com/394435/221159248-33fe7fa0-7810-458b-96ab-6f7b7fc42d57.png)

### What have you tried so far?

Set the `accessibilityRole` property on the `AppBar.Content`, but it's not permitted.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 16.1
| android                        | n/a
| react-native                | 0.71.3
| react-native-paper    | 5.0.2
| node                            | 16.16.0
| npm or yarn                | npm 8.11.0
| expo sdk                     | n/a

PR #3719, opened 2023-03-02
fix: changed accessibility label based on onPress callback

barttom:
### Summary
Resolves problem for accessibility label when Appbar.Content has no OnPress action
detailed info: https://github.com/callstack/react-native-paper/issues/3704

### Test plan

- Start the app
- Open Xcode, then go to the Xcode menu -> Open Developer Tool -> Accessibility Inspector
- Select the heading

|   case | expected result  |
|---|---|
|  **when Appbar.Content hasn't onPress callback: content is recognized as a text** |  image |
|  **when Appbar.Content has onPress callback:  component is recognized as a button** |   image
|

### DIFF
--- a/src/components/Appbar/AppbarContent.tsx
+++ b/src/components/Appbar/AppbarContent.tsx
@@ -130,7 +130,7 @@ const AppbarContent = ({
 
   return (
     <TouchableWithoutFeedback
-      accessibilityRole="button"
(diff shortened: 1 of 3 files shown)


lukewalczak (approved):
Thanks @barttom for applying the fix! 🔥

Issue #3717, opened 2023-03-02
Expo SDK48 Source Map Issue


kreshnikalidema:
### Current behaviour
After migrated to Expo SDK 48, more than 150 warnings appeared with the text "Failed to parse source map from sourcemap". The same problem was also in `@react-navigation` which was fixed by upgrading `react-native-builder-bob` from version 0.20.3 to 0.20.4

https://github.com/expo/expo/issues/21276#issuecomment-1451753482

### Your Environment

| software                      | version
| ---------------------     | -------
| react-native                 | 0.71.3
| react-native-paper      | 5.2.0
| node                            | 16.18.1
| npm                             | 8.19.2
| expo sdk                      | 48

kreshnikalidema:
Hey @lukewalczak, I have raised the PR and it is working as expected. Can you check it?

PR #3721, opened 2023-03-03
fix: fix paths in sourcemap files

kreshnikalidema:
Fix paths in sourcemap files.

After migrated to Expo SDK 48, more than 150 warnings appeared with the text "Failed to parse source map from sourcemap". The same problem was also in @react-navigation which was fixed by upgrading react-native-builder-bob from version 0.20.3 to 0.20.4

https://github.com/callstack/react-native-paper/issues/3717

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.8",
     "color": "^3.1.2",
-    "react-native-builder-bob": "^0.20.4",
     "use-event-callback": "^0.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.8",
     "color": "^3.1.2",
+    "react-native-builder-bob": "^0.20.4",
```
lukewalczak, line 52: I think it should stay in `devDependecies`

kreshnikalidema: Fixed. Sorry, I was in a hurry.

lukewalczak (approved):

Issue #3722, opened 2023-03-03
TextInput in a specific configuration shows visual artifact on iOS and Web


CodingItWrong:
### Current behaviour
I get a dark gray bar appearing at the top of a `TextInput` when all the following are true:

- Displaying in light mode
- Multiline
- Disabled
- Flat

### Expected behaviour

I do not expect the dark gray bar to appear; I expect the field background to display consistently as in the other combinations of settings above.

### How to reproduce?
Just render out a `<TextInput multiline disabled />`.

Reproduction is on this branch: [GitHub - CodingItWrong/paper-sandbox at disabled-multiline-text-issue](https://github.com/CodingItWrong/paper-sandbox/tree/disabled-multiline-text-issue)

### Preview

The issue:

![image](https://user-images.githubusercontent.com/15832198/222819344-6ba0948e-b80a-45fa-a57d-5acd6e922eb2.png)

No issue in dark mode:

![image](https://user-images.githubusercontent.com/15832198/222819484-628833e1-c056-450f-a9de-038d06ab18ee.png)

### What have you tried so far?
Narrowed down the bug to only the configuration described above.

Other than that, I'm not totally sure of workarounds I might be able to do.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 16.2
| react-native                | 0.71.3
| react-native-paper    | 5.2.0
| node                            | 18.14.2
| npm or yarn                | yarn
| yarn                             | 1.22.19
| npm                             | 9.6.0
| expo sdk                     | 48.0.5

lukewalczak:
@CodingItWrong thanks for the report, I'm able to reproduce the issue, so will tackle it.

PR #3726, opened 2023-03-06
fix: input's adornment and helper disabled colors

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3722

### Summary

PR correct the following items:

* hiding the `patch-container` when input is disabled

before | after
--- | ---
Zrzut ekranu 2023-03-6 o 10 54 11 | Zrzut ekranu 2023-03-6 o 10 54 49

* disabled text color for `HelperText`
* disabled background color for flat `TextInput`
* disabled icon color for adornments

before | after
--- | ---
Zrzut ekranu 2023-03-6 o 10 46 25 | Zrzut ekranu 2023-03-6 o 10 41 05

### Test plan

Updated test for some colors and patch container. New examples added to the demo app.


src/components/HelperText.tsx
```diff
@@ -122,15 +127,28 @@ const HelperText = ({
     textHeight = e.nativeEvent.layout.height;
   };
 
-  const { colors, dark } = theme;
+  const getTextColor = () => {
```
tjaniczek, line 130: Could we replace that with function from `Adornment/utils.ts` ? They seem oddly similar

Issue #3569, opened 2022-12-31
`Appbar.Header` ignores `borderBottom*Radius` style


fabriziocucci:
### Current behaviour
As per title, it seems that those props don't have any effect:
```js
<Appbar.Header elevated={true} style={{ borderBottomLeftRadius: 16, borderBottomRightRadius: 16 }}>
    <Appbar.Content title="My Title" />
</Appbar.Header>
```

### Expected behaviour
I would expect those styling props to be honoured.

### How to reproduce?
https://snack.expo.dev/@fabrizio.cucci/header-with-borderbottom-radius

### Preview
![WhatsApp Image 2022-12-31 at 19 00 38](https://user-images.githubusercontent.com/8156463/210152024-eab0f1d1-1e65-4a61-8f18-ce35906d8262.jpeg)

### What have you tried so far?
The `Appbar` component seems to naturally work with this (but of course it misses all the juice of `Appbar.Header`):
```js
<Appbar elevated={true} style={{ borderBottomLeftRadius: 16, borderBottomRightRadius: 16 }}>
    <Appbar.Content title="My Title" />
</Appbar>
```

### Your Environment

| software                      | version
| --------------------- | -------
| react-native-paper    | 5.1.2

lukewalczak:
According to the [documentation](https://m3.material.io/components/top-app-bar/guidelines), the `Appbar` should not have curved shapes, that's why it's not supported:

<img width="847" alt="image" src="https://user-images.githubusercontent.com/22746080/210180780-6ace3d70-e011-4faa-81a3-f2a97a0ebe26.png">

I would not consider that behavior as a bug.

fabriziocucci:
Ok, so this is intentionally strictly adhering to the spec and not supporting customisation.

The only thing a bit confusing is the fact that the `Appbar` component supports this but, on the other side, the guideline doesn't seem to be as explicit when it comes to bottom app bar.

Happy to either close this or transform into a feature request, whatever makes sense for you @lukewalczak.

lukewalczak:
According to the MD guidelines, by default `Appbar` cannot have curved shapes. I know it's kinda against the rules, however in the past, there was a similar question about that behavior, and that's why we can transform it into a feature request.

fabriziocucci:
Thanks @lukewalczak, I love the attitude of leaning towards more flexibility vs religious adherence to the spec! 👍

PR #3767, opened 2023-03-20
fix: appbar header ignores border bottom radius

josemak25:
### Summary
This fix Implements: https://github.com/callstack/react-native-paper/issues/3569

- Add fix for app bar header custom styles
   -  This change addresses styles like border bottom radius on the app bar

### Test plan

1. Start the example app 
2. Click on the Appbar
3. Enable custom color switch
4. Enable Elevated
5. And lastly pass your custom styles to the app bar header

    ```javascript
    <Appbar.Header elevated={true} style={{ borderBottomLeftRadius: 16, borderBottomRightRadius: 16 }}>
       <Appbar.Content title="My Title" />
    </Appbar.Header>
    ```

<details>
<summary>screenshot from web</summary>

Screenshot 2023-03-20 at 2 12 14 PM

</details>

<details>
<summary>screenshot from web</summary>

mobile screenshot

</details>


DimitarNestorov (approved):

src/components/Appbar/AppbarHeader.tsx
```diff
@@ -153,10 +155,12 @@ const AppbarHeader = ({
           paddingTop: statusBarHeight ?? top,
           paddingHorizontal: Math.max(left, right),
         },
+        restStyle,
```
lukewalczak, line 158: I'm not sure if it's the correct approach to pass `restStyle` to the `Appbar` root layer (as you named it) which in styles has previously extracted styles properties from destructured style object 🤔 Maybe we should rather just pass the `borderRadius` into that layer?

src/components/Appbar/utils.ts
```diff
@@ -39,6 +39,44 @@ export const getAppbarColor = (
...
+
+  const borders = Object.entries(properties)
+    .filter(([_, value]) => value)
+    .reduce<Record<string, typeof borderRadius>>((acc, [key, value]) => {
+      acc[key] = value;
+      return acc;
+    }, {});
+
+  return borders;
+};
```
lukewalczak, line 78: Maybe we can simplify it as follows:

```
const borders = {};
for (const [key, value] of Object.entries(properties)) {
  if (value) {
    borders[key] = value;
  }
}

return borders;
```

wdyt @josemak25 @DimitarNestorov?

josemak25: I agree this is much simpler to read, thanks for the suggestion 👍

josemak25: Yea, this simplifies it even more

DimitarNestorov (approved):

PR #3771, opened 2023-03-21
fix(snackbar,view): make snackbar width fit to screen

gandol:
in some case the snackbar not fit to device screen because still use `%` instead of use  device width

Screenshot 2023-03-21 at 23 35 51

### Summary

so in this fix i just change to use device width and add padding for the content
Screenshot 2023-03-21 at 23 36 10


lukewalczak:
Hey @gandol, could you please present on the snack, the issue which you are facing and that the PR is fixing?

lukewalczak:
Closing due to missing response.

PR #3772, opened 2023-03-21
fix: data table header title number of lines not working

pac-guerreiro:
### Summary

Data table column titles were stuck on one line, even if numberOfLines > 1.

The problem was that data table header had a fixed height of 48 px.

I updated the data table in the example app to include a column title with two lines.

### Test plan

You can use the data table sample on the example app and tweak the numberOfLines and title text.

Right now, the second column is set to show the column title in 2 lines as you can see below:

Screenshot 2023-03-21 at 16 23 40

### DIFF
--- a/example/src/Examples/DataTableExample.tsx
+++ b/example/src/Examples/DataTableExample.tsx
@@ -83,7 +83,9 @@ const DataTableExample = () => {
             >
               Dessert
             </DataTable.Title>
-            <DataTable.Title numeric>Calories</DataTable.Title>
+            <DataTable.Title numberOfLines={2} numeric>
+              Calories per piece
+            </DataTable.Title>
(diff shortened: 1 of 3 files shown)


example/src/Examples/DataTableExample.tsx
```diff
@@ -83,7 +83,9 @@ const DataTableExample = () => {
             >
               Dessert
             </DataTable.Title>
-            <DataTable.Title numeric>Calories</DataTable.Title>
+            <DataTable.Title numberOfLines={2} numeric>
```
tjaniczek, line 86: It would be nice to add `alignItems: 'center'` to the `styles.container` since we can have multiline headers: 

image

tjaniczek (approved):

Issue #3736, opened 2023-03-08
RTL problem on Expo web 


MMTE:
### Current behaviour

as you can see in the screenshot using expo web, react native paper components aren't RTL

### Expected behaviour

it's expected that the Exclamation mark becomes after RTL text (direction is not RTL). other components aren't RTL too.

### How to reproduce?

here is the source code of GitHub repo:
[rtl-test-react-native-paper](https://github.com/MMTE/react-native-paper-rtl-expo-web)

also steps to reproduce:
```
npx create-expo-app
npx expo install react-native-web@~0.18.10 react-dom@18.2.0
npx expo install @expo/webpack-config@^18.0.1
npx expo start
add a react-paper-component
press w to start expo web
press a to run emulator or use expo-go
```

### Preview

![Screenshot from 2023-03-08 08-18-08](https://user-images.githubusercontent.com/6337624/223622755-2808e669-00db-49ed-98a4-3accfb5cfa08.png)

### What have you tried so far?

I have tried to change styles for each component manually for example for Cards title by changing the properties of the title in the card, but I believe it's not the solution.
also, I don't know whether this problem would be resolved using the [Using on the Web](https://callstack.github.io/react-native-paper/docs/guides/react-native-web) or not. because the expo web approach is more straight forward and the guide on react native paper wasn't clear to me. 

I would appreciate any suggestions. 

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | -
| android                        | 13
| react-native                | 0.71.3
| react-native-paper    | 5.3.1
| node                            | 16.19.0
| npm or yarn                | 8.19.3
| expo sdk                     | 48.0.5

PR #3777, opened 2023-03-23
fix: rtl not being detected correctly on web

pac-guerreiro:
### Summary

RTL detection didn't work on web (#3736) because RN I18nManager only works on native ([See breaking changes from react-native-web@0.18.0](https://github.com/necolas/react-native-web/releases/tag/0.18.0)).

My solution involved the following:

- Adding a prop named `direction` to the `Provider`, so the user can define if the components should be displayed in RTL or LTR.

- Adding an effect that forces RTL on native when `direction === 'rtl'`.

- Wrapping children of the `Provider` with a `View` that injects a prop named `dir` to all it's subtree that will define if components should be displayed in RTL or LTR on web ([See new features from react-native-web@0.18.0](https://github.com/necolas/react-native-web/releases/tag/0.18.0)).

- Change all `paddingLeft` to `paddingStart` ([See RN RTL support article](https://reactnative.dev/blog/2016/08/19/right-to-left-support-for-react-native-apps#overview-changes-in-rn-for-rtl-support)).

- Change all `paddingRight` to `paddingEnd` ([See RN RTL support article](https://reactnative.dev/blog/2016/08/19/right-to-left-support-for-react-native-apps#overview-changes-in-rn-for-rtl-support)).

- Change all `marginLeft` to `marginStart` ([See RN RTL support article](https://reactnative.dev/blog/2016/08/19/right-to-left-support-for-react-native-apps#overview-changes-in-rn-for-rtl-support)).

- Change all `marginRight` to `marginEnd` ([See RN RTL support article](https://reactnative.dev/blog/2016/08/19/right-to-left-support-for-react-native-apps#overview-changes-in-rn-for-rtl-support)).


DimitarNestorov (changes requested):

src/core/Provider.tsx
```diff
@@ -97,18 +109,36 @@ const Provider = (props: Props) => {
...
-        <SettingsProvider value={settings || { icon: MaterialCommunityIcon }}>
-          {/* @ts-expect-error check @callstack/react-theme-provider's children prop */}
-          <ThemeProvider theme={getTheme()}>{children}</ThemeProvider>
-        </SettingsProvider>
+        <LocalizationProvider value={direction}>
+          <SettingsProvider value={settings || { icon: MaterialCommunityIcon }}>
+            {/* @ts-expect-error check @callstack/react-theme-provider's children prop */}
+            <ThemeProvider theme={getTheme()}>
+              {
+                <View style={styles.container} {...{ dir: direction }}>
```
DimitarNestorov, line 126: This container View can be considered a breaking change. I believe we should revert this change and just add a note in the docs and add it in the Example app. It's something specific to react-native-web. Thoughts @lukewalczak @tjaniczek?

lukewalczak: > I believe we should revert this change and just add a note in the docs and add it in the Example app. It's something specific to react-native-web.

Thanks @DimitarNestorov for your input! Could you please elaborate more on your suggestion?

DimitarNestorov: I read from the description of the PR that the `dir` prop of the wrapping `View` is what fixes RTL on web. This wrapping View can be a problem for some. Even conditionally rendering it only for web can be a problem, so we should instead document it well and link people to the docs.

tjaniczek:
Not sure why the pop-up menu is still LTR on web:
image

will pick up this task for the time being. Also, I like @DimitarNestorov approach 👍  removed the `View` already and added docs instead

src/core/Provider.tsx
tjaniczek: `dir` is not a valid prop for RN's `View`.  This wired notation `{...{ dir: direction }}` is there only to avoid TS error. We should get rid of it anyway.

I think the original idea for `react-native-web` was to wrap the whole app in `<html dir="rtl">` - I tested that and it makes much more sense, we'll avoid this additional wrapper, problems with Portals etc. Will rework the `Localization` provider and figure out better API to support this.

tjaniczek:
Ok, since this PR becomes too ridiculous to review I'll try to pinpoint a couple of issues.

First things first, `react-native-web` mocks out the whole `I18nManager` and relies text positioning on `dir` property of parent element (e.g. `<View dir="rtl>`). As far as I know, RN does not support this property so linter throws error trying to use it. Effectively there's no way around it so I added [this convenient wrapper](https://github.com/callstack/react-native-paper/pull/3777/files#diff-696727b69da6b034a61ac170db640d76426194d7e535d6df3873b6765f777fd6R37) to silent the linter. With or without this, there's the key problem:

**Since `react-native-web` parses RTL layouts differently to `RN` we will need to support two ways of defining LTR/RTL layouts** (web & mobile).

Most concerns I have regard [this useEffect](https://github.com/callstack/react-native-paper/pull/3777/files#diff-df7f1205b408a8b779895c080c77457728df4da79a0fa852c0255e54ce897664R83) where I'd assign dir to the root `<html>` element of a page (`<html dir="rtl">`). This handles most cases since the `react-native-way` looks all the way to the top of the tree to find this prop. Without top element with `dir` attribute we'd have to add this in multiple places: 
- below `<Provider>` , 
- in each `<Modal>` 
- and as far as I know inside each `<Surface>` too

Surprisingly, this is not a breaking change, since to apply this `direction` has to be manually passed into `<Provider>` ( [as explained here](https://github.com/callstack/react-native-paper/pull/3777/files#diff-bffc8ea6fc2d7dbf6593758d46277634fb9645efd8ad56d97642f189dd642069R273) ) , and the prop itself is added in this PR.

Other than that, we need a few hacks since we're basically retrofitting the RTL support at this point:
- https://github.com/callstack/react-native-paper/pull/3777/files#diff-46327c8e0610e196d2eec4ef16e3fee3fe0b9f69c0451f9e43e331b7f26ed08bR163
- https://github.com/callstack/react-native-paper/pull/3777/files#diff-40a7ac238bd91da69419785da41b78053017736cab9ece62b22639c8537174e2R61
- https://github.com/callstack/react-native-paper/pull/3777/files#diff-a4dba4502c98f41455802b906263f5842e6a67881b016e03d6d8a9b5718dcbe7R8
- https://github.com/callstack/react-native-paper/pull/3777/files#diff-5146ee7e7b0598cc2982fcdd8c7f9f6cd3eb0d891f87203ebfc5a3255a6d24beR82

And there's also a couple things I couldn't get to work properly:
- `Animated FAB` does not work on web at the moment (both LTR and RTL are broken)
- `Material Bottom Tab Navigator` does not turn the content RTL automatically (need to investigate RTL support in Navigator)
- `TextInput` - absolutely no way of getting this component to work on both web & mobile RTL

satya164:
> Most concerns I have regard [this useEffect](https://github.com/callstack/react-native-paper/pull/3777/files#diff-df7f1205b408a8b779895c080c77457728df4da79a0fa852c0255e54ce897664R83) where I'd assign dir to the root <html> element of a page (<html dir="rtl">). This handles most cases since the react-native-web looks all the way to the top of the tree to find this prop.

This won't work with SSR on Web. What I'm planning for React Navigation is to ask users to pass `dir` to HTML and call `I18nManager.forceRTL` if they need to support `RTL`. The `direction` prop on the `Provider` is simply to inform the library about which direction needs to be used. The library itself shouldn't be trying to call `I18nManager.forceRTL` or mutate the DOM. Doing that affects far more than the library's scope.

Other than that on React Navigation side the changes are essentially replacing usage of `I18nManager` with the `direction` from `Provider`.

pac-guerreiro:
Can I close this PR?

tjaniczek:
No, huge chunk of work left

federicogomezlara:
Hello, do you guys plan to continue to work on this feature? We'd love to be able to use it, we've had to swap out our texts and inputs for others from other library on web

arasrezaei:
lots of stuff has to be done!

satya164:
Closing in favor of https://github.com/callstack/react-native-paper/pull/4921

PR #3782, opened 2023-03-24
docs: integrate appbar with react-navigation doc adjustments

lukewalczak:
### Summary

Adjustments to the documentation related to `Appbar` component integration with `react-navigation`

### Test plan

N/A


satya164 (approved):
Maybe consider removing the 5.x related stuff as well since that version of React Navigation is no longer supported

PR #3791, opened 2023-03-30
feat: support for maxFontSizeMutliplier on Checkbox.Item

dimitrisspacecode:
### Summary

Added max font scaling support for the Checkbox.Item component

### Test plan

Easy way to test it is by running the example on an iOS simulator and start Increasing/Decreasing the Preferred Text Size (on default I set it to 0, meaning it's disabled, set it to 1.1 or more to test it works)

### DIFF
--- a/src/components/Checkbox/CheckboxItem.tsx
+++ b/src/components/Checkbox/CheckboxItem.tsx
@@ -50,6 +50,10 @@ export type Props = {
    */
   style?: StyleProp<ViewStyle>;
   /**
+   * Specifies the largest possible scale a title font can reach.
+   */
+  maxFontSizeMultiplier?: number;
+  /**
    * Style that is passed to Label element.
    */
   labelStyle?: StyleProp<TextStyle>;
@@ -121,6 +125,7 @@ const CheckboxItem = ({
   accessibilityLabel = label,
   disabled,
   labelVariant = 'bodyLarge',
+  maxFontSizeMultiplier = 1.5,
   ...props
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
@@ -168,6 +173,8 @@ const CheckboxItem = ({
         {isLeading && checkbox}
         <Text
           variant={labelVariant}
+          testID={`${testID}-text`}
+          maxFontSizeMultiplier={maxFontSizeMultiplier}
           style={[
             styles.label,
             !theme.isV3 && styles.font,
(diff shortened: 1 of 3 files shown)


src/components/Checkbox/CheckboxItem.tsx
```diff
@@ -168,6 +173,7 @@ const CheckboxItem = ({
         {isLeading && checkbox}
         <Text
           variant={labelVariant}
+          maxFontSizeMultiplier={maxFontSizeMultiplier}
```
lukewalczak, line 176: Please add unit test covering new prop

src/components/Checkbox/CheckboxItem.tsx
```diff
@@ -121,6 +125,7 @@ const CheckboxItem = ({
   accessibilityLabel = label,
   disabled,
   labelVariant = 'bodyLarge',
+  maxFontSizeMultiplier = 0,
```
lukewalczak, line 128: I think we should add at least `1.5` as a default value, like in other places.

src/components/__tests__/Checkbox/CheckboxItem.test.tsx
```diff
@@ -94,3 +94,13 @@ it('disables the row when the prop disabled is true', () => {
...
   });
 });
+
+it('should have maxFontSizeMultiplier set to 1.5 by default', () => {
+  const { getByTestId } = render(
+    <Checkbox.Item label="" testID="checkbox-item" status="unchecked" />
+  );
+  const checkboxItem = getByTestId('checkbox-item');
+  expect(
+    checkboxItem.props.children[0].props.children[1].props.maxFontSizeMultiplier
```
DimitarNestorov, line 104: Add a `testID` to the `Text` in the checkbox.
```diff
         <Text
           variant={labelVariant}
           maxFontSizeMultiplier={maxFontSizeMultiplier}
+          testID={`${testID}-text`}
```

And then update the test

```suggestion
  const checkboxItemText = getByTestId('checkbox-item-text');
  expect(
    checkboxItemText.props.maxFontSizeMultiplier
```

DimitarNestorov (changes requested):

DimitarNestorov (approved):

lukewalczak:
I've created an additional [commit](https://github.com/callstack/react-native-paper/commit/e77119759b152509e367a2e430a5c346997c6063) to rename the `maxFontSizeMultiplier` to `labelMaxFontSizeMultiplier` to keep the consistency with other props related to the same component such as `labelStyle` or `labelVariant`

PR #3794, opened 2023-03-31
chore: remove disabled accessibility trait for `AppbarContent` and `Card`

DimitarNestorov:
### Summary

It's kinda weird to hear "not enabled" for elements of the UI that will never become interactable.

I tried messing around with the `accessibilityState` prop but I couldn't figure out a way without removing the element: https://snack.expo.dev/@dimitarnestorov/touchablewithoutfeedback-can-not-remove-disabled-trait

|Before|After|
|-|-|
|Showcasing traits for AppbarContent before changes in PR|Showcasing traits for AppbarContent after changes in PR|
|Showcasing traits for Card before changes in PR|Showcasing traits for Card after changes in PR|

### Test plan

Tests pass

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -9,6 +9,8 @@ import {
   ViewStyle,
 } from 'react-native';
 
+import useLatestCallback from 'use-latest-callback';
+
 import { useInternalTheme } from '../../core/theming';
 import type { ThemeProp } from '../../types';
 import Surface from '../Surface';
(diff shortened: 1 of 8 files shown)


src/components/Card/Card.tsx
```diff
@@ -269,29 +284,22 @@ const Card = ({
...
+
+      {onPress || onLongPress ? (
+        <TouchableWithoutFeedback
+          delayPressIn={0}
+          disabled={disabled}
+          delayLongPress={delayLongPress}
+          onLongPress={onLongPress}
+          onPress={onPress}
+          onPressIn={() => runElevationAnimation('in')}
+          onPressOut={() => runElevationAnimation('out')}
```
lukewalczak, line 296: Shouldn't we avoid passing inline functions as props to components, as it can lead to performance issues and unnecessary re-renders? 🤔

lukewalczak (approved):
Thanks @DimitarNestorov! 🚀

Issue #3790, opened 2023-03-29
Disable shadows by default


alexdgourlay:
**Is your feature request related to a problem? Please describe.**
The [material documentation on elevation](https://m3.material.io/styles/elevation/overview#:~:text=Shadows%3A%20Instead%20of%20applying%20shadows%20by%20default%20to%20all%20levels%2C%20use%20shadows%20only%20when%20required%20to%20create%20additional%20protection%20against%20a%20background%20or%20to%20encourage%20interaction), for material 3, state that:

> Shadows: Instead of applying shadows by default to all levels, use shadows only when required to create additional protection against a background or to encourage interaction

**Describe the solution you'd like**
Shadows should be disabled by default for react-native-paper components.

lukewalczak:
Hey @alexdgourlay, the `Surface` _(I guess we are discussing that one)_ component was adjusted to the Material You based on the `Elevation` design from MD3 Figma Kit available on their website, which wasn't presenting an option without shadows:

<img width="766" alt="image" src="https://user-images.githubusercontent.com/22746080/228953373-be2a4d8d-207e-4418-b225-73f7b8759fe9.png">

I have a feeling they didn't have the exact specs related to' Elevation' on their site at that point. 

Thank you for bringing this point to my attention! We cannot disable the shadow from the `Surface` since it will be a breaking change, but we can introduce the new prop, which can be then set by the user to make the shadow visible or hidden!

alexdgourlay:
Yes perhaps they updated their specs. 

Introducing a new props sounds good!

PR #3799, opened 2023-04-03
feat: support flat and elevated mode

lukewalczak:
Fixes: #3790 

#### Related issue

- #3790 

Introducing prop `mode` to the `Surface` component with two possible options `elevated` (default) and `flat`.

The `flat` mode indicates the `Surface` component will have the background color appropriate to the set level, however, it will not have a `shadow`.

ios | android | web
-- | --- | ---
Zrzut ekranu 2023-04-3 o 12 58 38 | Zrzut ekranu 2023-04-3 o 12 57 27 | Zrzut ekranu 2023-04-3 o 12 57 43

### Test plan

New prop `mode` covered by unit tests.

### DIFF
--- a/src/components/__tests__/Surface.test.tsx
+++ b/src/components/__tests__/Surface.test.tsx
@@ -4,6 +4,7 @@ import { Platform } from 'react-native';
 
 import { render } from '@testing-library/react-native';
 
+import { getTheme } from '../../core/theming';
 import Surface from '../Surface';
 
 describe('Surface', () => {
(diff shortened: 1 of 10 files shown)


src/components/__tests__/Surface.test.tsx
```diff
@@ -39,6 +40,35 @@ describe('Surface', () => {
       },
     });
 
+    it('should render Surface with appropriate bg color but without shadow, if mode is set to "flat"', () => {
```
tjaniczek, line 43: It would be nice to have a test with `mode="elevated"` too

PR #3803, opened 2023-04-05
fix: correct the way how disabling touchable is handled

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3303

### Summary

Introducing hook `useIsTouchableEvent` to check whether any touchable event was passed to the `TouchableRipple` to properly set `disabled` prop.

Context: 

I guess touchable was initially disabled when `onPress` callback is not defined, in order to prevent running custom functionality handled by onPressIne.g. in Button [component](https://github.com/callstack/react-native-paper/blob/main/src/components/Button/Button.tsx?rgh-link-date=2022-09-23T12%3A45%3A14Z#L208). Probably the same applies to the Chip component.

#### Related issue:

- #3303 

### Test plan

- [X] - covered new hook with unit test
- [x] - write unit test in `Chip` and `Button` component to confirm that `disabled` is properly set


DimitarNestorov:
Why a hook?

```ts
const hasTouchHandler = (...events: (event: GestureResponderEvent) => void) =>
  Boolean(events.filter(Boolean).length);
```

Usage:
```ts
const hasPassedTouchHandler = hasTouchHandler(
  onPress,
  onLongPress,
  onPressIn,
  onPressOut,
);

// disabled={disabled || !hasPassedTouchHandler}
```

lukewalczak:
> Why a hook?

Tbh there is no specific reason, just creating a hook came to my mind. Creating simple util helper makes sense as well and most likely will be faster and less complicated 🙌🏽 
I will refactor it, thanks!

src/utils/hasTouchHandler.tsx
```diff
@@ -0,0 +1,25 @@
...
+type TouchableEventObject = {
+  onPress?: ((event: GestureResponderEvent) => void) | undefined;
+  onPressIn?: ((event: GestureResponderEvent) => void) | undefined;
+  onPressOut?: ((event: GestureResponderEvent) => void) | undefined;
+  onLongPress?: ((event: GestureResponderEvent) => void) | undefined;
+};
+
+const hasTouchHandler = (
+  touchableEventObject: TouchableEventObject
+): boolean => {
```
DimitarNestorov, line 12:
```suggestion
export default function hasTouchHandler(
  touchableEventObject: TouchableEventObject
) {
```

src/utils/hasTouchHandler.tsx
```diff
@@ -0,0 +1,25 @@
+import type { GestureResponderEvent } from 'react-native';
+
+type TouchableEventObject = {
+  onPress?: ((event: GestureResponderEvent) => void) | undefined;
+  onPressIn?: ((event: GestureResponderEvent) => void) | undefined;
+  onPressOut?: ((event: GestureResponderEvent) => void) | undefined;
+  onLongPress?: ((event: GestureResponderEvent) => void) | undefined;
+};
```
DimitarNestorov, line 8: Move the `touchableEvents` array out of the function and mark it as const and then update the type:
```suggestion
type TouchableEventObject = Partial<Record<keyof touchableEvents[number], (event: GestureResponderEvent) => void>>;
```
image

DimitarNestorov (changes requested):

DimitarNestorov (changes requested):

DimitarNestorov (approved):

PR #3837, opened 2023-04-22
docs: Fixed images linking to incorrect path

D3faIt:
Path to images didn't link to anything, now it should link correctly, at least according to the docs at https://callstack.github.io/react-native-paper/docs/guides/react-navigation/

### Summary

Fixed images linking to incorrect path

### Test plan

Tested directly on https://callstack.github.io, changing the links from `../../` to `/react-native-paper/` makes them show up

PR #3847, opened 2023-04-26
Fix broken showcase paths

### LINKED ISSUE #3846: Showcase images 404
### Current behaviour
https://callstack.github.io/react-native-paper/docs/showcase/

All images are 404

chmac:
I haven't tested this, but I believe it should solve the issue. I don't know what your build process is for the docs site, so I haven't tried it out.

fix #3846

### DIFF
--- a/docs/src/components/Showcase.tsx
+++ b/docs/src/components/Showcase.tsx
@@ -19,7 +19,7 @@ const data: Data[] = [
   {
     color: '#6200EE',
     name: 'Paper Example App',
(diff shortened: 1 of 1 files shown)


chmac:
The preview from @callstack-bot doesn't seem to work. I'm not sure why, maybe the paths are still wrong and my PR doesn't fix the issue.

chmac:
I didn't see #3837, which was just deployed, which fixed the issue!

PR #3851, opened 2023-04-27
docs: add section for component theme colors

lukewalczak:
### Summary

We're adding a "Theme colors" section to each component's doc to simplify the process of identifying theme color names and their corresponding roles without diving into the source code

* Colors are manually collected within the file `docs/src/data/themeColors` for each component
* Colors are defined for possible component's:
	* states such as `active` or `selected`
	* modes such as `flat` or `outlined`
	

### Test plan

- [X] - added tests for utils
- [X] - tested manually on the build

### DIFF
--- a/docs/component-docs-plugin/generatePageMDX.js
+++ b/docs/component-docs-plugin/generatePageMDX.js
@@ -37,12 +37,19 @@ function generatePageMDX(doc, link) {
     .replace(/src="screenshots/g, `src="${baseUrl}screenshots`)
     .replace(/@extends.+$/, '');
 
+  const data = JSON.stringify(customFields.themeColors[doc.title]);
+
   const mdx = `
 ---
 title: ${doc.title}
 ---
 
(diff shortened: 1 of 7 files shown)


docs/component-docs-plugin/generatePageMDX.js
```diff
@@ -37,12 +37,19 @@ function generatePageMDX(doc, link) {
...
 ---
 title: ${doc.title}
 ---
 
 import PropTable from '@site/src/components/PropTable.tsx';
+import ThemeColorsTable from '@site/src/components/ThemeColorsTable.tsx';
+
+## Theme colors
+
+<ThemeColorsTable data={${data}} componentName="${doc.title}" />
```
DimitarNestorov, line 52: Let's move it below examples? Maybe below props?

lukewalczak: Sure, nice catch, I totally forgot to relocate it, after recording the the video sneak peek.

PR #3858, opened 2023-04-28
fix: added style to component used for text measurement on Android

Y0lk:
### Summary
On Android, the text on AnimatedFAB is calculated using a 2nd component, but this component doesn't take the text style into account. This leads to incorrect text measurement.

### Test plan
This issue is easy to see on Android if you're using the MD2 theme. Depending on the length of your label, the spacing between the icon and the text will look different because the width of the label is wrong.

### DIFF
--- a/src/components/FAB/AnimatedFAB.tsx
+++ b/src/components/FAB/AnimatedFAB.tsx
@@ -13,7 +13,6 @@ import {
   ScrollView,
   StyleProp,
   StyleSheet,
-  Text,
   View,
   ViewStyle,
 } from 'react-native';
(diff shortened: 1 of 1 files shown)


src/components/FAB/AnimatedFAB.tsx
```diff
@@ -496,7 +496,12 @@ const AnimatedFAB = ({
         // proper text measurements there is a need to additionaly render that text, but
         // wrapped in absolutely positioned `ScrollView` which height is 0.
         <ScrollView style={styles.textPlaceholderContainer}>
-          <Text onTextLayout={onTextLayout}>{label}</Text>
+          <Text
+            onTextLayout={onTextLayout}
+            style={[uppercase && styles.uppercaseLabel, textStyle]}
```
lukewalczak, line 501: Shouldn't we copy all the styles from the "main" text? Maybe we can even use the same `Animated.Text` here for the measurements purposed, did you try that?

Y0lk: I suppose we could! I don't think it would change anything in this case but it might be a bit more future-proof. I updated the PR.

The only other style that I added was `styles.label` (which currently has only `position: 'absolute'`, but could have more in the future) since the rest of the styles either use the calculated text width, or have no effect on the text width so I don't think it makes sense to include them.

lukewalczak (approved):

PR #3866, opened 2023-05-02
feat: add possibility to override touchable ripple color

lukewalczak:
### Summary

PR is introducing the possibility to override the `rippleColor` in each component as well as customizing `Provider` to disable the ripple effect in all Paper's components.

### Test plan

- [X] - tested manually
- [X] - added docs related to `Ripple effect`
- [x] - write unit tests for `TouchableRipple` which is a base for other components


src/components/Chip/helpers.tsx
```diff
@@ -239,6 +244,10 @@ const getUnderlayColor = ({
     isOutlined,
   });
 
+  if (customRippleColor) {
+    return customRippleColor;
+  }
```
DimitarNestorov, line 249: Since it's an instant return let's move it to the very top of the function

src/core/aaa.md
DimitarNestorov, line 1: I'm guessing this is a temporary file

src/core/settings.tsx
```diff
@@ -5,9 +5,11 @@ import MaterialCommunityIcon, {
 } from '../components/MaterialCommunityIcon';
 
 export type Settings = {
-  icon: ({ name, color, size, direction }: IconProps) => React.ReactNode;
+  icon?: ({ name, color, size, direction }: IconProps) => React.ReactNode;
```
DimitarNestorov, line 8: I don't understand why you made this optional

docs/docs/guides/11-ripple-effect.md
```diff
@@ -0,0 +1,37 @@
...
+
+```
+<Button rippleColor="transparent" icon="camera" mode="contained" onPress={() => console.log('Pressed')}>
+  Press me
+</Button>
+```
+
+## Disable ripple effect in all components
+
+To disable the ripple effect in *all* of Paper's components, you need to customize the `Provider`  that is likely wrapping your root component. To do this, simply pass `rippleEffectEnabled: false` to the `settings` object prop within the mentioned `Provider`.
```
DimitarNestorov, line 21: Let's rename `Provider` to `PaperProvider` and add `export const Provider = PaperProvider` and update the docs everywhere to not have `Provider as PaperProvider` and in the next major version we can remove `Provider`.
```suggestion
To disable the ripple effect in **all** of Paper's components set `rippleEffectEnabled: false` on the `settings` prop of `PaperProvider`.
```

docs/docs/guides/11-ripple-effect.md
```diff
@@ -0,0 +1,37 @@
...
+The ripple effect is a visual feedback that occurs when a user interacts with a clickable UI element, such as a button. This response takes the form of a circular ripple expanding from the point of contact, much like a drop falling into water and creating ripples.
+
+The ripple effect is an essential aspect of Material Design, and Paper's pressable components have it enabled by default. Nonetheless, it can be tailored to suit specific needs.
+
+## Customize ripple effect color in component
+
+The `rippleColor` prop is available for every pressable component which allows you to set the color of the ripple effect. For the instance, to make the `Button` component's ripple effect transparent, simply pass the desired color value to the prop:
+
+```
+<Button rippleColor="transparent" icon="camera" mode="contained" onPress={() => console.log('Pressed')}>
```
DimitarNestorov, line 14: `transparent`? Wouldn't that be the same as disabling the effect? Let's have a solid color like `#FF0000`.

docs/docs/guides/11-ripple-effect.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Ripple effect
+---
+
+The ripple effect is a visual feedback that occurs when a user interacts with a clickable UI element, such as a button. This response takes the form of a circular ripple expanding from the point of contact, much like a drop falling into water and creating ripples.
```
DimitarNestorov, line 5:
```diff
- clickable
+ pressable
```

docs/docs/guides/11-ripple-effect.md
```diff
@@ -0,0 +1,37 @@
...
+
+The ripple effect is an essential aspect of Material Design, and Paper's pressable components have it enabled by default. Nonetheless, it can be tailored to suit specific needs.
```
DimitarNestorov, line 7: iOS doesn't have a ripple effect so I think it's worth noting.

example/src/DrawerItems.tsx
```diff
@@ -222,6 +226,17 @@ const DrawerItems = ({
                 </View>
               </TouchableRipple>
             )}
+
+            {isV3 && (
+              <TouchableRipple onPress={toggleRippleEffect}>
+                <View style={[styles.preference, isV3 && styles.v3Preference]}>
+                  <Text variant="labelLarge">Ripple effect *</Text>
```
DimitarNestorov, line 233: On iOS it should be called highlight effect?

DimitarNestorov (changes requested):

src/components/TouchableRipple/TouchableRipple.tsx
```diff
@@ -111,132 +113,136 @@ const TouchableRipple = ({
   const theme = useInternalTheme(themeOverrides);
   const { onPress, onLongPress, onPressIn, onPressOut } = rest;
 
-  const handlePressIn = (e: any) => {
-    const { centered } = rest;
-
+  const handlePressIn = (e: any, rippleEffectEnabled?: boolean) => {
```
tjaniczek, line 116: Rather large function here. It may benefit from `useCallback` in case there's a lot of buttons on the screen

src/components/Drawer/DrawerItem.tsx
```diff
@@ -123,7 +126,7 @@ const DrawerItem = ({
         accessibilityRole="button"
         accessibilityState={{ selected: active }}
         accessibilityLabel={accessibilityLabel}
-        underlayColor={underlayColor}
+        rippleColor={customRippleColor || rippleColor}
```
tjaniczek, line 129: This boils down to:
```
customRippleColor || (isV3 ? color(contentColor).alpha(0.12).rgb().string())
```
Asking just to be on the safe side - do we want to pass `customRippleColor` in V2?

docs/docs/guides/11-ripple-effect.md
lukewalczak: I've just used the `transparent` color, since based on the issues, users often ask how to customize the color to disable the effect, however, I inclined to your comment and changed the color.

docs/docs/guides/11-ripple-effect.md
lukewalczak: > Let's rename Provider to PaperProvider and add export const Provider = PaperProvider and update the docs everywhere to not have Provider as PaperProvider and in the next major version we can remove Provider.

Agree, I will create a separated PR for it.

src/components/Drawer/DrawerItem.tsx
lukewalczak: Yes, overriding the rippleColor should be applicable for two MD generations (_I've just correct the switch in the example app which was available only for v3)_

src/core/settings.tsx
lukewalczak: Before introducing the `rippleEffectEnabled` into the `Settings` there was defined only one property - the `icon`, so the user passing the `settings` prop into the `PaperProvider` could configure only the source for the `icon`. 
At the moment, when there are two options, `icon` should be optional in the same way as `rippleEffectEnabled`, to avoid TS error, when the user is overriding only one of the options e.g. `rippleEffectEnabled`.

src/components/TouchableRipple/TouchableRipple.native.tsx
```diff
@@ -84,39 +86,52 @@ const TouchableRipple = ({
...
     return (
-      <Pressable
-        {...rest}
-        disabled={disabled}
-        style={[borderless && styles.overflowHidden, style]}
-        android_ripple={
-          background != null
-            ? background
-            : {
+      <SettingsConsumer>
```
DimitarNestorov, line 89: Let's also use hooks here?

docs/docs/guides/11-ripple-effect.md
lukewalczak: https://github.com/callstack/react-native-paper/pull/3889

src/core/settings.tsx
DimitarNestorov: Makes complete sense! Forgot that this is a user prop.

src/components/TouchableRipple/TouchableRipple.tsx
```diff
@@ -249,15 +261,21 @@ const TouchableRipple = ({
...
-    <Pressable
-      {...rest}
-      onPressIn={handlePressIn}
-      onPressOut={handlePressOut}
-      disabled={disabled}
-      style={[styles.touchable, borderless && styles.borderless, style]}
-    >
-      {React.Children.only(children)}
-    </Pressable>
+    <SettingsConsumer>
```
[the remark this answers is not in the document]
lukewalczak: Done.

DimitarNestorov (approved):

tjaniczek (approved):

Issue #3871, opened 2023-05-03
TouchableRipple underlayColor ignored


jahirfiquitiva:
Hi

I have set a custom underlayColor for TouchableRipple on iOS, but it seems to not display anything at all.

I checked the source code and it looks like the color is ignored and not used?

https://github.com/callstack/react-native-paper/blob/896ab233398ace5d12f7892e31888d57ba298a0a/src/components/TouchableRipple/TouchableRipple.tsx#LL106C3-L106C16

What the correct way to set the underlayColor and have it show on iOS?

lukewalczak:
Hey @jahirfiquitiva could you please present your issue on the snack? I was not able to observe the problem, please check the `TouchableRipple` [example](https://snack.expo.dev/@react-native-paper/underlaycolor-in-touchableripple) with set `underlayColor`.

> I checked the source code and it looks like the color is ignored and not used?
> [896ab23/src/components/TouchableRipple/TouchableRipple.tsx#LL106C3-L106C16](https://github.com/callstack/react-native-paper/blob/896ab233398ace5d12f7892e31888d57ba298a0a/src/components/TouchableRipple/TouchableRipple.tsx#LL106C3-L106C16)

The file is related to the `web` platform, for the mobile please check the file with `.native` extension.

jahirfiquitiva:
@lukewalczak  I don't know how to do the snack thing, but here's what I've found:


```typescript
<TouchableRipple
  testID={`meetingList_${formattedDate}`}
  onPress={() => setExpanded(!expanded)}
  rippleColor={paperTheme.colors.accentRipple}
  underlayColor={paperTheme.colors.accentRipple}
>
  <View
    style={{
      ...styles.header,
      backgroundColor: paperTheme.colors.accentLight,
    }}
  >
	...
  </View>
</TouchableRipple>
```

This does **not** work. The ripple is **not** shown in android and the underlay is **not** shown in iOS. Here's a video:

https://user-images.githubusercontent.com/10360816/236030823-d695b4f3-a8ea-49d6-b91c-9cc69bad26c1.mp4

-----

```typescript
<TouchableRipple
  testID={`meetingList_${formattedDate}`}
  onPress={() => setExpanded(!expanded)}
  rippleColor={paperTheme.colors.accentRipple}
  underlayColor={paperTheme.colors.accentRipple}
  style={{
    backgroundColor: paperTheme.colors.accentLight,
  }}
>
  <View style={styles.header}>
	...
  </View>
[...]

PR #3875, opened 2023-05-08
fix: touchable ripple overlay

lukewalczak:
### Summary

Correct the `TouchableRipple` underlay color support, when the ripple is not supported, so the additional layer is displayed, to have a similar effect to `TouchableHighlight`.

#### Related issue

- #3871 

### Test plan

- [X] - added unit tests


gedu (approved):
LGTM

Issue #3180, opened 2022-05-21
Add support for onLongPress on FAB.Group component


fabriziocucci:
The experience I'd like to achieve with the `FAB.Group` component is the following:
* `onLongPress` => show the underlying FABs
* `onPress` => do something else

I'm wondering, was the callback left out intentionally (e.g. to comply with the Material Design spec) or just an oversight (considering that the FAB supports `onLongPress` already)?

Happy to put together a PR if this is something desirable.

lukewalczak:
Hey @fabriziocucci, I guess there is no `onLongPress` functionality since MD didn't indicate that or apps following the MD don't have that functionality as well.
Generally, I think we can add support for that, however opening the stack of FABs should be still handled on `onPress`.

fabriziocucci:
> however opening the stack of FABs should be still handled on onPress

I guess it makes sense as default behaviour:
`onPress` => open stack of FABs
`onLongPress` => do nothing or open stack of FABs

Although opening the stack of FABs even when `onPress` is explicitly overridden IMHO would be somewhat inflexible.

To avoid breaking changes maybe something like:
```
```

fabriziocucci:
> however opening the stack of FABs should be still handled on onPress

I guess it makes sense as default behaviour:
`onPress` => open stack of FABs
`onLongPress` => do nothing or open stack of FABs

Although opening the stack of FABs even when `onPress` is explicitly overridden IMHO would still be somewhat inflexible.

I understand this is the current behaviour so maybe to avoid breaking changes something like this could work?
```
<FAB
    onPress={() => {
        // the onPress specified by the user can return true/false to indicate whether the stack of FABs should open
        const shouldToggle = onPress?.() || true;
        shouldToggle && toggle();
    }}
    ...
```

lukewalczak:
I guess the main purpose of `onPress` on the `FABGroup` is to handle the callback on the main `FAB` when stack is opened.  
Imo`onLongPress` should be an auxiliary option for handling pressing the main `FAB` when stack is not opened.

fabriziocucci:
I see your point @lukewalczak, I'm wondering though what's the downside of keeping the component more flexible.

From a user point of view, a `FAB` and a `FABGroup` are indistinguishable, there is no expectation in terms of interaction.

[...]

PR #3881, opened 2023-05-10
feat: add long press support for fab group

kwereszczynski:
### Summary

Added `onLongPress` support. Allows to revert the logic of displaying actions stack from `onPress` to `onLongPress`.

## Related issue:
https://github.com/callstack/react-native-paper/issues/3180

### Test plan
- [ ] Check if long press triggers action, and single press toggles stack.
- [ ] Open actions stack by press.
- [ ] Click "Toggle on Long Press" action.
- [ ] Check if logic got reverted, and now single press triggers action, and long press toggles stack.

### DIFF
--- a/src/components/FAB/FABGroup.tsx
+++ b/src/components/FAB/FABGroup.tsx
@@ -83,6 +83,14 @@ export type Props = {
    */
   toggleStackOnLongPress?: boolean;
   /**
+   * Changes the delay for long press reaction.
+   */
+  delayLongPress?: number;
+  /**
+   * Allows for onLongPress when stack is opened.
+   */
+  enableLongPressWhenStackOpened?: boolean;
+  /**
    * Whether the speed dial is open.
    */
   open: boolean;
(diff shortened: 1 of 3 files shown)


src/components/FAB/FABGroup.tsx
```diff
@@ -416,8 +428,19 @@ const FABGroup = ({
         <FAB
           onPress={(e) => {
             onPress?.(e);
-            toggle();
+            if (!toggleStackOnLongPress || open) {
+              toggle();
+            }
+          }}
+          onLongPress={() => {
+            if (!open) {
```
DimitarNestorov, line 436: Is this necessary? Maybe someone would like to have their `onLongPress` callback work even when the group is open.

src/components/FAB/FABGroup.tsx
```diff
@@ -416,8 +428,19 @@ const FABGroup = ({
...
+              onLongPress?.();
+              if (toggleStackOnLongPress) {
+                toggle();
+              }
+            }
           }}
+          delayLongPress={200}
```
lukewalczak, line 443: Add a prop for that so user can customize their values.

src/components/FAB/FABGroup.tsx
kwereszczynski: As i was testing it, this felt weird. I agree that we could allow for it, but make it optional?

DimitarNestorov (approved):

PR #3889, opened 2023-05-15
refactor: export PaperProvider to simplify its usage

lukewalczak:
### Summary

Usually react apps in the root contain many `Provider`s, so to simplify using `Provider` from the Paper, it's going to be exported as `PaperProvider` to avoid renaming it during the import.

Note: It's not a breaking change, there are both `Provider` and `PaperProvider` available atm.

### Test plan

CI should pass.


DimitarNestorov:
For VSCode autocomplete it will be worth to also rename it in `core/Provider`: https://github.com/callstack/react-native-paper/blob/05321cfb70bd1e4f7cac68e1ff26debf4731f792/src/core/Provider.tsx#L23

DimitarNestorov (approved):

DimitarNestorov (approved):

Issue #3776, opened 2023-03-23
TextInput Label disappears when the component is unfocused on mobile


ChrisGadek1:
### Current behaviour
[Screencast from 23.03.2023 01:28:16.webm](https://user-images.githubusercontent.com/44200187/227068378-4f224862-46b3-411f-9105-1086a332e382.webm)
If TextInput is unfocused, then the label disappears. It was tested only on android.

### Expected behaviour
[Screencast from 23.03.2023 01:31:55.webm](https://user-images.githubusercontent.com/44200187/227068824-d101b02d-a35a-4f99-82e5-575c3a7af778.webm)
The label should stay and not disappear

### How to reproduce?
`<TextInput
       label={i18n.t('login.login')}
       value={username}
       onChangeText={setUsername}
       />`

### What have you tried so far?
I think the problem is with TextInput label opacity, so I made a little change there:
https://github.com/callstack/react-native-paper/blob/d88ff2051c3540fab5d7980efe3423f4ddfc98bd/src/components/TextInput/Label/InputLabel.tsx#L65
I replaced this line with this line:
`      outputRange: [1, 1],`
And everything began to work correctly (That's how I made preview videos in sections _Current behaviour_ and _Expected behaviour_)

### Your Environment

| software                      | version
| --------------------- | -------
| android                       |13
| react-native                |0.71.4
| react-native-paper      |5.5.0
| node                           |18.15.0
| npm                            |9.5.0
| expo sdk                    |(no expo)

lukewalczak:
Hey @ChrisGadek1, are you able to reproduce your issue on the expo snack? – please attach the link to it then.

ChrisGadek1:
> Hey @ChrisGadek1, are you able to reproduce your issue on the expo snack? – please attach the link to it then.

I was not able to reproduce, on expo everythink seems to work as expected. The problem is only in pure React Native project

lukewalczak:
> I was not able to reproduce, on expo everythink seems to work as expected. The problem is only in pure React Native project created by create-react-native-app

Could you please create minimal github repo presenting the issue, which we can clone to investigate the bug?

ChrisGadek1:
> > I was not able to reproduce, on expo everythink seems to work as expected. The problem is only in pure React Native project created by create-react-native-app
> 
> Could you please create minimal github repo presenting the issue, which we can clone to investigate the bug?

[...]

PR #3899, opened 2023-05-21
fix: unfocused TextInput label disappears on new architecture (#3776)

osman-turan:
### Summary

The `TextInput` label disappears when the input is unfocused on React Native's New Architecture (on both Android and iOS). After some digging, I noticed the root cause is switching between an `Animated.Value` and some `number`s depending on some booleans:

```ts
const placeholderOpacity = hasActiveOutline
  ? parentState.labeled // <-- `Animated.Value`
  : parentState.labelLayout.measured
  ? 1
  : 0;
```

If we switch between `Animated.Value` instances, the problem disappears:

```ts
const { current: placeholderOpacityAnims } = React.useRef([
  new Animated.Value(0),
  new Animated.Value(1),
]);

const placeholderOpacity = hasActiveOutline
  ? parentState.labeled
  : placeholderOpacityAnims[parentState.labelLayout.measured ? 1 : 0];
```

#3776 has more details about the issue and contains an example repo to reproduce it.

### Test plan

https://github.com/ChrisGadek1/react-native-paper-textinput-bug-reproduction repo from https://github.com/callstack/react-native-paper/issues/3776#issuecomment-1481136624 can be used for testing this fix.

### DIFF
--- a/src/components/TextInput/TextInputFlat.tsx
+++ b/src/components/TextInput/TextInputFlat.tsx
@@ -216,11 +216,14 @@ const TextInputFlat = ({
   const baseLabelTranslateY =
     -labelHalfHeight - (topPosition + MINIMIZED_LABEL_Y_OFFSET);
 
+  const { current: placeholderOpacityAnims } = React.useRef([
+    new Animated.Value(0),
+    new Animated.Value(1),
+  ]);
+
   const placeholderOpacity = hasActiveOutline
     ? parentState.labeled
-    : parentState.labelLayout.measured
-    ? 1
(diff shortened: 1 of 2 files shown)


archieWu:
This bug was fixed one month but has not been released in production

lukewalczak (approved):
Thanks for help! 🙌🏽

osman-turan:
@lukewalczak Thanks! 🙌 

@archieWu Thanks for pointing out. But I tried some of the later releases after my PR, and none of them worked for me.

Issue #3911, opened 2023-06-01
`overlayAccessibilityLabel` prop for Dialog component


meatnordrink:
**Is your feature request related to a problem? Please describe.**
The accessibility label for modals is editable via `overlayAccessibilityLabel`, which is necessary given the accessibility difficulties surrounding modals; but there is no corresponding prop on dialogs, leaving us unable to edit the default "close modal" text announced to screen readers.

**Describe the solution you'd like**
I would like to see an `overlayAccessibilityLabel` prop on `Dialog` components.

**Describe alternatives you've considered**
If some of the accessibility props of RN were exposed through `Dialog` (like `accessibilityLabel`, etc.), that might be an alternative.

**Additional context**
n/a

PR #3915, opened 2023-06-05
feat: add overlayAccessibilityLabel prop to Dialog

jeferson-sb:
### Summary
Forward `overlayAccessibilityLabel` on Dialog component to be able to change the accessible label on the modal.

Addresses #3911

### DIFF
--- a/src/components/Dialog/Dialog.tsx
+++ b/src/components/Dialog/Dialog.tsx
@@ -31,6 +31,10 @@ export type Props = {
    */
   onDismiss?: () => void;
   /**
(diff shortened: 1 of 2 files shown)


ahammam-ddpa:
Bumping this PR - any particular reason this isn't merged in?

MikitasK (commented):
LGTM 👌  
could you just add a regression test in `Dialog.test.tsx` confirming that `overlayAccessibilityLabel` is applied to the backdrop? this will help prevent future regressions

PR #3922, opened 2023-06-08
docs: fix Appbar.Header default height

alex-fournier:
Default height for Appbar.Header component was not matching the actual height (56 vs. 64).


lukewalczak (approved):
@alex-fournier thanks for correction!

PR #3934, opened 2023-06-15
docs: Update Chip.tsx

meatnordrink:
Update description

Small expansion of the Chip documentation.

### Summary

The current Chip documentation doesn't give much info as to how the Chip element is intended to be used in Material UI. While the user can certainly look it up, most Material UI libraries give at least a brief summary of [the usage guidelines](https://m2.material.io/components/chips) in their own documentation, and I think it's helpful. 

So this PR adds some brief guidance as to how the Chip component is used in the Material UI system. (I thought a link to the [official guidelines](https://m2.material.io/components/chips) might also be nice, but wasn't sure if that might be against policy, since it's not already being done...? If not, and you think it's a good idea, I can include one.)

### Test plan

Check the description for the Chip component, see if it makes sense and feels sufficiently informative.

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -125,7 +125,15 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
 };
 
 /**
- * Chips can be used to display entities in small blocks.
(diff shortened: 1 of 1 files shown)


meatnordrink:
Forgot about the title, commit format; re-opening with corrections.

Issue #3902, opened 2023-05-22
Can't use number `0` as an `id` prop on `List.Accordion`


JPStrydom:
### Current behaviour
When passing an `id` prop of the number `0` to a `List.Accordion`, the following incorrect error get's thrown:
```
'List.Accordion is used inside a List.AccordionGroup without specifying an id prop.'
```

### Expected behaviour
The number `0` should be allowed as an acceptable `id` prop value.

### How to reproduce?
Pass `id` `0` to a `List.Accordion` component - say from a `.map` index.

### Preview
![image](https://github.com/callstack/react-native-paper/assets/25905330/387ef6c4-1b43-47de-95a8-4986b5a03601)

### What have you tried so far?
It looks like the following code in the `List` component is the culprit:
```
  if (groupContext !== null && !id) {
    throw new Error(
      'List.Accordion is used inside a List.AccordionGroup without specifying an id prop.'
    );
  }
```
which should probably be something like:
```
  if (groupContext !== null && (id || id === 0)) {
    throw new Error(
      'List.Accordion is used inside a List.AccordionGroup without specifying an id prop.'
    );
  }
```
but I'm not sure how that would effect the underlying implimentation.

### Your Environment

| software                      | version
| --------------------- | -------
| react-native                | 0.71.8
| react-native-paper    | 5.7.2
| node                            | 18.16.0
| npm                            | 9.5.1
| expo                         | 48.0.16

JPStrydom:
> Hey! Thanks for opening the issue. Can you provide a minimal repro which demonstrates the issue? Posting a snippet of your code in the issue is useful, but it's not usually straightforward to run. A repro will help us debug the issue faster. Please try to keep the repro as small as possible. The easiest way to provide a repro is on [snack.expo.dev](https://snack.expo.dev). If it's not possible to repro it on [snack.expo.dev](https://snack.expo.dev), then you can also provide the repro in a GitHub repository.

I'm sure the issue is simple enough to not require a reproducible demo. Simply pass an `id` of `0` to the `List.Accordion` as such:
```
<List.AccordionGroup>
  <List.Accordion id={0}>This is a test</List.Accordion>
</List.AccordionGroup>
```
and you'll see a prop type warning saying:
```
'List.Accordion is used inside a List.AccordionGroup without specifying an id prop.'
```

PR #3936, opened 2023-06-15
fix: correct condition to accept id equal zero

lukewalczak:
Fixes: #3902 

### Summary

Corrects the condition to not throw an error when `id={0}` since it's typed to accept the string or number.

#### Related issue

- #3902 

### Test plan

Covered with unit tests.


gedu (approved):
LGTM

PR #3937, opened 2023-06-15
fix: MD2 Checkbox.Android disabled color

jennifersalas:
### Summary

This fixes an issue where on MD2 themes, the Android Checkbox component did not use the correct color when disabled.

### Test plan
The issue can be reproduced in this expo snack: https://snack.expo.dev/@jvsalas/restless-bagel

Screenshot 2023-06-15 at 4 19 07 PM


lukewalczak (approved):
Thanks!

Issue #3811, opened 2023-04-10
TextInput outlined strange inside a view


Angelk90:
### Current behaviour

As you see from start to weird behavior you don't see placeholder text.
Which instead of in the same input field as above but not found in the view, is seen correctly.

### Expected behaviour

Same behavior as the input field above.

### How to reproduce?

You can also try it with android expo emulator:
https://snack.expo.dev/exQQvm0ve

### Preview

https://user-images.githubusercontent.com/20476002/230880067-fe8d8e37-1edf-41ae-9463-43d68709da86.mov

### Your Environment

| software                      | version
| --------------------- | -------
| android                        | 10
| react-native                | 0.71.4
| react-native-paper    | 5.6.0
| node                            | 16.14.2
| npm                             | 9.6.2

lukewalczak:
> It seems to work, but you have to understand why the component breaks.

Could you please explain what is the issue on the attached snack?

lukewalczak:
Sorry, but I feel confused atm, has your problem been solved or do you need more help with that issue?

lukewalczak:
I see, I will check what's the problem with the placeholder

PR #3941, opened 2023-06-16
fix: set input min width

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3811 https://github.com/callstack/react-native-paper/issues/4039

### Summary

Setting input min width, to be properly displayed for shorter label and placeholers

### Test plan

Updated snapshots

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -6,6 +6,8 @@ import {
   TextInput as NativeTextInput,
   TextStyle,
   ViewStyle,
+  NativeSyntheticEvent,
+  TextLayoutEventData,
 } from 'react-native';
 
 import TextInputAffix, {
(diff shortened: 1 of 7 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -458,6 +464,13 @@ const TextInput = forwardRef<TextInputHandles, Props>(
       [labelLayout.height, labelLayout.width]
     );
 
+    const handleLabelTextLayout = React.useCallback(
+      ({ nativeEvent }: NativeSyntheticEvent<TextLayoutEventData>) => {
+        setLabelTextLayout({ width: Math.ceil(nativeEvent.lines[0].width) });
```
tjaniczek, line 469: I guess
```
Math.ceil(Math.max(33, ...nativeEvent.lines.map(line => line.width)))
```
if we want to be precise

tjaniczek (approved):
Omg, how good that Accordion is 🎉

Issue #3952, opened 2023-06-26
export 'useLinkTools' (imported as 'useLinkTools') was not found in '@react-navigation/native'


nth-zik:
Hi all, I currently using react-native-paper for web with typescript and receive this error

```
ERROR in ./node_modules/react-native-paper/lib/module/react-navigation/adapter.js 7:16-28
export 'useLinkTools' (imported as 'useLinkTools') was not found in '@react-navigation/native' 
```
Work fine on android, ios but on web throw this error
Seem my webpack not ignore error from node_modules
Here is my CRA script
![image](https://github.com/callstack/react-native-paper/assets/22862747/dc686fce-7503-4199-9640-ab145bddfb14)

Thanks all

PR #3954, opened 2023-06-26
Draft: fix: conditionally import react navigation

tjaniczek:
Fixes https://github.com/callstack/react-native-paper/issues/3952

### Summary

Unfortunately Webpack won't allows us to import missing export via ES6. Dynamic imports are a no-go as well. They're async and since we want to import hook we end up with `Promise<useLinkTools>`. This would break rules of hooks since we'd have a conditional hook rendered (before and after async import).

The only solution I found is to import the whole package as `CommonJS`. This won't break the rules since it's sync. It will import once and then only one hook will be returned. 

### Test plan

Example app does not throw an error anymore on either v6 or v7

### DIFF
--- a/src/react-navigation/adapter.ts
+++ b/src/react-navigation/adapter.ts
@@ -1,19 +1,14 @@
-import {
-  // @ts-ignore: this hook is not available in React Navigation v7
-  useLinkBuilder,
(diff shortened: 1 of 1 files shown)


DimitarNestorov:
Using CommonJS also isn't ideal because bundlers do not do tree shaking

PR #3969, opened 2023-07-10
feat: option for hiding check icon on selected chip

Big-Cheung:
### Summary
Added optional prop for enabling/disabling default checkmark icon when selecting a chip.

### Test plan
Testable using a Chip component with a selected item.  The check mark next to the selected chip should be hidden if the prop option "showSelectedIcon" is set to false, and shown if the option is set to true. It is set to true by default.

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -66,7 +66,7 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
    */
   showSelectedOverlay?: boolean;
   /**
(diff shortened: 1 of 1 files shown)


lukewalczak:
Hey @Big-Cheung, you have to adjust spacings when the selected icon is hidden since those are not equal then:

image

Big-Cheung:
Good catch! Adjusted the spacing to use normal spacing. I also renamed the prop to be consistent with SegmentedButton.

lukewalczak (approved):

Issue #3988, opened 2023-07-20
Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?


rzfzr:
### Current behaviour

When using a Button, I get the following error.

```
import React from 'react';
import {Link} from 'expo-router';
import {Button} from 'react-native-paper';

export default function Page() {
  return (
    <Link href="/onboarding" asChild>
      <Button icon="camera" mode="contained">
        Go About
      </Button>
    </Link>
  );
}

```
```
 ERROR  Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `SlotClone`.
    in Button (created by SlotClone)
    in SlotClone (created by Slot)
    in Slot
    in ExpoRouterLink (created by Page)
    in Page
    in Unknown (created by Route(index))
    in Route (created by Route(index))
    in Route(index) (created by SceneView)
    in StaticContainer
    in EnsureSingleNavigator (created by SceneView)
    in SceneView (created by QualifiedSlot)
    in QualifiedSlot (created by DefaultNavigator)
    in PreventRemoveProvider (created by NavigationContent)
    in NavigationContent
    in Unknown (created by QualifiedNavigator)
    in QualifiedNavigator (created by Navigator)
    in Navigator (created by DefaultNavigator)
    in RNCSafeAreaView
    in Unknown (created by DefaultNavigator)
    in DefaultNavigator
    in Unknown (created by Route())
    in Route (created by Route())
    in Route() (created by ContextNavigator)
    in RNCSafeAreaProvider (created by SafeAreaProvider)
    in SafeAreaProvider (created by wrapper)
    in RNGestureHandlerRootView (created by GestureHandlerRootView)
    in GestureHandlerRootView (created by GestureHandlerRootView)
    in GestureHandlerRootView (created by wrapper)
    in wrapper (created by ContextNavigator)
    in EnsureSingleNavigator
    in BaseNavigationContainer
    in ThemeProvider
    in NavigationContainerInner (created by ContextNavigator)
    in ContextNavigator (created by ExpoRoot)
    in ExpoRoot (created by App)
    in ThemeProvider (created by PaperProvider)
    in RCTView (created by View)
    in View (created by Portal.Host)
    in Portal.Host (created by PaperProvider)
    in RNCSafeAreaProvider (created by SafeAreaProvider)
    in SafeAreaProvider (created by SafeAreaInsetsContext)
[...]

PR #3992, opened 2023-07-22
fix: added ref support to button and touchable ripple component.

sangameshsomawar:
Fixes: https://github.com/callstack/react-native-paper/issues/3988

### Summary

Added ref support for Button and TouchableRipple component.

### Test plan

I tried creating ref for button and touchable ripples, it worked.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -12,6 +12,7 @@ import {
 
 import color from 'color';
 
+import { forwardRef } from '../..//utils/forwardRef';
 import { useInternalTheme } from '../../core/theming';
 import type { $Omit, ThemeProp } from '../../types';
 import hasTouchHandler from '../../utils/hasTouchHandler';
@@ -125,6 +126,7 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
    * testID to be used on tests.
    */
   testID?: string;
+  ref?: React.Ref<ButtonRef>;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


sangamesh1439:
@lukewalczak Let me know if I can do something more to help reviewing/testing code?

src/components/Button/Button.tsx
```diff
@@ -12,6 +12,7 @@ import {
 
 import color from 'color';
 
+import { forwardRef } from '../..//utils/forwardRef';
```
DimitarNestorov, line 15:
```suggestion
import { forwardRef } from '../../utils/forwardRef';
```

src/components/Button/Button.tsx
```diff
@@ -144,33 +146,39 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
...
-  contentStyle,
-  labelStyle,
-  testID = 'button',
-  accessible,
-  ...rest
-}: Props) => {
+
+type ButtonRef = React.ElementRef<typeof TouchableRipple>;
+
+const Button: React.ForwardRefRenderFunction<ButtonRef, Props> = (
```
DimitarNestorov, line 152: Have you tried something like this?
```suggestion
const Button = forwardRef({
```

sangameshsomawar: @DimitarNestorov I could not understand your point.

`forwardRef` is already added in the last line `export default forwardRef(Button);`

sangamesh1439: @DimitarNestorov We cannot use `forwardRef` from `src/utils/forwardRef.tsx` because it does not support mentioning component level attribute,  In our case `TouchableRipple.supported = true;` is a component level attribute.

Issue #3924, opened 2023-06-09
How to remove bottom border radius on Card.Cover


verbowersock:
Currently, it looks like Card.Cover has border bottom radius that is impossible to remove. Setting border radius to zero on style in Card.Cover does nothing. Attached is the screenshot from react-native-paper-examples snack.
![Screenshot 2023-06-09 at 1 24 41 PM](https://github.com/callstack/react-native-paper/assets/24816633/465b9252-7ff0-44af-9515-12fafba15832)

How do i remove this border radius? I need the line to be straight, more like in v4. I am using 5.7.2 at the moment 
![Screenshot 2023-06-09 at 1 29 44 PM](https://github.com/callstack/react-native-paper/assets/24816633/7dfec2b9-bdd4-4717-b060-13761a668bb5)

PR #3996, opened 2023-07-24
fix: correct overriding border radius

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3924

### Summary

Allow customizing `borderRadius` value in `Card.Cover`.

### Test plan

Adding new example and unit test case.

### DIFF
--- a/example/src/Examples/CardExample.tsx
+++ b/example/src/Examples/CardExample.tsx
@@ -118,6 +118,19 @@ const CardExample = () => {
             </Button>
           </Card.Actions>
         </Card>
+        <Card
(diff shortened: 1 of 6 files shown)


tjaniczek (approved):

meatnordrink:
I am noticing what might be an unintended side-effect to this improvement; now, if you pass in rounding to a single corner (as we were already doing in a number of places), it applies that value to _all_ corners of the `<Card.cover>`. For example, this is what we're seeing, after upgrading and pulling in this work: 

```css
  cardCover: {
    borderTopRightRadius: 8, // All four borders have 8px rounding
    }
```

PR #3998, opened 2023-07-25
fix: forward ref to snackbar

### LINKED ISSUE #3976: Snackbar ref is null
### Current behaviour
The ref is null 

### Expected behaviour
ref should provide a reference to the Snackbar

### How to reproduce?
https://snack.expo.dev/@ad9889/e90742

### Preview
See snack

Nothing else

### Your Environment

| ios                                | 16.5.1
| android                        | 11
| react-native                | 0.69.9
| react-native-paper    |  5.9.1
| node                            | 18.16.0
| yarn                | 1.22.19
| expo sdk                     | 46

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3976

### Summary

Provide `ref` to the `Snackbar`.

### DIFF
--- a/src/components/Snackbar.tsx
+++ b/src/components/Snackbar.tsx
@@ -14,12 +14,20 @@ import useLatestCallback from 'use-latest-callback';
 
 import { useInternalTheme } from '../core/theming';
 import type { $Omit, $RemoveChildren, ThemeProp } from '../types';
+import { forwardRef } from '../utils/forwardRef';
 import Button from './Button/Button';
 import type { IconSource } from './Icon';
 import IconButton from './IconButton/IconButton';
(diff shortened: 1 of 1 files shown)


src/components/Snackbar.tsx
```diff
@@ -128,218 +136,224 @@ const DURATION_LONG = 10000;
...
+                        inputRange: [0, 1],
+                        outputRange: [0.9, 1],
+                      })
+                    : 1,
+                },
+              ],
+            },
+            style,
+          ]}
+          ref={ref}
```
tjaniczek, line 301: What would be the use case of having ref on `Surface`? Maybe we should pass ref from `renderChildrenWithWrapper` so the user can check if the Snackbar is currently open? WDYT?

PR #4000, opened 2023-07-25
chore: update dependencies in lib and example

lukewalczak:
#### Library updates

* updating `react-native` to `0.71.6`
* updating `react` to `18.0.2`
* updating `jest` to `29.2.1`

#### Example updates

* removing `entryPoint` from `app.json`
* updating `expo` to `48.0.0`
* updating other deps to be compatible with expo@48 using `expo doctor`

#### Other

* adding resolution on `expo-assets` to get rid of warning:

> Constants.platform.ios.model has been deprecated in favor of expo-device's Device.modelName property. This API will be removed in SDK 45.

* ignore the warning in webpack, coming from `reanimated`:

> export 'default'.'interpolate' (imported as 'Animated') was not found in 'react-native-reanimated' ...

* linting files

* update snapshots

### Test plan

✅ CI pass
✅ iOS app work
✅ Android app work
✅ Android native app work
✅ web app work


src/components/MaterialCommunityIcon.tsx
```diff
@@ -64,7 +64,7 @@ try {
 export const accessibilityProps =
   Platform.OS === 'web'
     ? {
-        role: 'img',
+        role: 'img' as Role,
```
satya164, line 67: avoid `as`other than `as const` since it's not typesafe. you could do `somethingrandom` as Role and there won't be a warning.

Issue #3467, opened 2022-11-14
Android elevation/shadow issue


iM-GeeKy:
### Current behaviour

Navigating around the example app below on Android results in an odd elevation/shadow rendering issue. Note, this behavior isn't present on iOS. This problem seems to be intermittent  and I'd suspect this might be an upstream issue, but I want to post it here as it indirectly affects this repository and to bring visibility to the matter.

### Expected behaviour

The shadow should behave in a consistent manor without the shadow flickering.

### How to reproduce?

https://github.com/iM-GeeKy/shadows-bug

### Preview

This video was captured in Expo Go on a Pixel 5a running Android 13.
https://user-images.githubusercontent.com/7604441/201551315-70020fb4-ee8f-4516-b6ba-aafe6a1efcf5.mov

### What have you tried so far?

I haven't quite been able to pin down a solution, but I did find a few threads that relate to the issue.
- https://github.com/facebook/react-native/issues/23090
- https://github.com/callstack/react-native-paper/issues/1349

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x
| android                        | 13
| react-native                | 0.70.5
| react-native-paper    | 5.0.0-rc.9
| node                            | v16.16.0
| npm or yarn                | yarn
| expo sdk                     | 47.0.3

iM-GeeKy:
@lukewalczak Do you have any ideas on this? I'm still seeing this behavior and am not sure if it's just a react-native issue or if it's a direct result of this library.

lukewalczak:
Hey @iM-GeeKy, unfortunately didn't check that, however I will try to address the issue.

dalenaval:
Currently facing the same issue

iM-GeeKy:
Hi @lukewalczak and @tjaniczek, I'm reopening this issue with the release of Expo 49. I've reproduced the example in the same repo under the sdk-49 branch [here](https://github.com/iM-GeeKy/shadows-bug/tree/sdk-49).

iM-GeeKy:
> Hi @lukewalczak and @tjaniczek, I think this issue should be reopened with the release of Expo 49. I've reproduced the example in the same repo under the sdk-49 branch [here](https://github.com/iM-GeeKy/shadows-bug/tree/sdk-49).

Should I reopen this issue again or will it be tracked in this thread?

PR #4008, opened 2023-07-31
fix: adjust offset target value for translation on axis y

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3467

### Summary

Adjust the value for offset, depending on the scene activity. 

### Test plan

Tested manually on the example app, based on the expo SDK 49 attached in the issue.
Tested also in the Paper example app.

### DIFF
--- a/src/components/BottomNavigation/BottomNavigation.tsx
+++ b/src/components/BottomNavigation/BottomNavigation.tsx
@@ -509,14 +509,14 @@ const BottomNavigation = <Route extends BaseRoute>({
             ? 1
             : 0;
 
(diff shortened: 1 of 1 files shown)


iM-GeeKy:
> Hey @iM-GeeKy, would you mind applying changes from the PR in your app and checking if there is no issue with the shadow?

Yeah man sure, I'll get back today once I get off work.

iM-GeeKy:
@lukewalczak Actually I tested it really quickly and it does seem to resolve the issue!

Issue #3975, opened 2023-07-12
Dynamic theme generator colors is not working properly


SoyDiego:
### Current behaviour

The theme generator builder is not OK when i'm trying to customize my colors.

### Expected behaviour

Show and generate my colors

### How to reproduce?

When I'm going to the theme builder of material design 3. I can choose my colors (left side of the screenshot) but when I'm trying to add to the theme generator of your website, the inputs are OK but the preview is ALL BLUE like the primary color.

- If I paste my #hex detect the color in the input but continue like the primary in the preview.
- If I write my hex number by number, only I can write 4 numbers and then the color is assigned and close but is not my color.

### Preview

![image](https://github.com/callstack/react-native-paper/assets/19764334/6d5578f9-a4c9-4e0d-9fa7-e4f5008c56b6)

![problem](https://github.com/callstack/react-native-paper/assets/19764334/5c4b4600-2504-4d33-b34e-80e7a926f116)

SoyDiego:
any news about this?

tjaniczek:
Hi @SoyDiego , finally had time to come back to this. Three things:
- As for the primary/secondary/tertiary colors - believe or not, all blues are the resulting colors from MD generator. The reason they look different on https://m3.material.io/theme-builder#/custom is that they have a bug in their own color picker. Play around with it a little bit and you'll see it yourself. 
- For more context why they are all blue, please refer to my comment here: https://github.com/callstack/react-native-paper/discussions/4034
- Adding custom colors in our generator was indeed buggy previously, there's a PR to address that and will be merged soon. Thanks for the feedback!

Hope that helps!

SoyDiego:
> Hi @SoyDiego , finally had time to come back to this. Three things:
> 
>     * As for the primary/secondary/tertiary colors - believe or not, all blues are the resulting colors from MD generator. The reason they look different on https://m3.material.io/theme-builder#/custom is that they have a bug in their own color picker. Play around with it a little bit and you'll see it yourself.
> 
>     * For more context why they are all blue, please refer to my comment here: [Trying coolors.co with dynamic theme colors ... any advice ? #4034](https://github.com/callstack/react-native-paper/discussions/4034)
> 
>     * Adding custom colors in our generator was indeed buggy previously, there's a PR to address that and will be merged soon. Thanks for the feedback!
> 
> 
> Hope that helps!

Thanks for your reply and all the information @tjaniczek

lukewalczak:
[...]

PR #4033, opened 2023-08-17
fix: custom colors list in docs theme builder

tjaniczek:
### Summary

Fixes: https://github.com/callstack/react-native-paper/issues/3975

The issue was that keys in custom colors list were not unique

### Test plan

Described in the original issue: go to docs, make a theme, play around with custom colors


lukewalczak (approved):
Tested and works and expected 🎉

image

PR #4036, opened 2023-08-18
fix: custom theme property for RadioButtonItem component

alexzel:
This PR contains two fixes for RadioButtonItem when passing custom colors through `theme` property:
```
theme={{
  colors: {
     onSurface: ....,
     onSurfaceDisabled: ...,
  },
}}
```

1. Fix for type error:
```
error TS2740: Type '{ onSurface: string; onSurfaceDisabled: string; }' is missing the following properties from type 'MD3Colors': primary, primaryContainer, secondary, secondaryContainer, and 27 more.
```
Solution: Change `theme` type to `ThemeProp`.

2. Radio button icon does not apply disabled color from custom `onSurfaceDisabled` theme color.

Solution: pass `theme` down to the components handling icon colors.

### DIFF
--- a/src/components/RadioButton/RadioButtonItem.tsx
+++ b/src/components/RadioButton/RadioButtonItem.tsx
@@ -135,7 +135,14 @@ const RadioButtonItem = ({
   labelVariant = 'bodyLarge',
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
-  const radioButtonProps = { value, disabled, status, color, uncheckedColor };
+  const radioButtonProps = {
+    value,
+    disabled,
+    status,
+    color,
+    theme,
+    uncheckedColor,
+  };
   const isLeading = position === 'leading';
   let radioButton: any;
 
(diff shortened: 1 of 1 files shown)


src/components/RadioButton/RadioButtonItem.tsx
```diff
@@ -79,7 +79,7 @@ export type Props = {
   /**
    * @optional
    */
-  theme?: InternalTheme;
+  theme?: ThemeProp;
```
lukewalczak, line 82: Could you please correct the `theme` type also in `RadioButtonIOS.tsx`?

lukewalczak (approved):

src/components/RadioButton/RadioButtonItem.tsx
alexzel: Sorry for the delay, didn't have time for that on weekend. I see you have merged it and updated `RadioButtonIOS.tsx`. Shouldn't we also update `theme` type in `RadioButtonAndroid.tsx`? Or maybe it is intentionally stays `InternalTheme` in there?

lukewalczak: Had to overlook it, will update android file as well

PR #4049, opened 2023-08-30
fix: Update Card.tsx

meatnordrink:
Fix `accessible` prop not being passed to `TouchableWithoutFeedback`.

### Summary

In the update to 5.6.0, the Card component was refactored, and in the refactor, the `accessible` property, passed in as a prop, was assigned to the component's inner `View`, rather than the `TouchableWithoutFeedback`, as it had been previously - and as it needs to be to have effect. 

This meant that passing `accessible={false}` had no effect, and thus there was no way to make content within a card's content area accessible to VoiceOver on iOS (which, by default, aggregates the accessibility tree if it encounters an element with an `onPress` handler).

I removed the ineffectual `accessible` prop on the inner content, and added it back to `TouchableWithoutFeedback`. This fixes the issue in my tests.

### Test plan

I have tested locally in our repo, but this would be a sufficient test that the fix has worked: 

```ts
<Card 
  accessible={false}
>
    <Card.Content>
            <Button onPress={() => console.log('button pressed!'}>Can you click me via VoiceOver?</Button>
    </Card.Content>
</Card>
```

This is my first contribution outside the docs, so please let me know if I can be more helpful as far as testing. There was nothing specific beyond linting in the contributing guidelines on testing (like, an existing test repo with a Maestro suite you can run against it, or something), but if something like that exists I'm happy to contribute to testing.

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -261,7 +261,6 @@ const Card = ({
     <View
       style={[styles.innerContainer, contentStyle]}
       testID={testID}
-      accessible={accessible}
     >
       {React.Children.map(children, (child, index) =>
         React.isValidElement(child)
@@ -311,6 +310,7 @@ const Card = ({
 
       {hasPassedTouchHandler ? (
         <TouchableWithoutFeedback
+          accessible={accessible}
           delayPressIn={0}
           disabled={disabled}
           delayLongPress={delayLongPress}


meatnordrink:
CI failed because I had `Fix` instead of `fix` in the title; correcting it didn't trigger CI to run again, and doesn't look like I can request a re-run myself. Sorry. :disappointed:

lukewalczak:
Hey @meatnordrink, please also correct the `fix` in the commit message and check the `lint`

meatnordrink:
@lukewalczak , I did this via the GitHub interface, and I don't know of a way to change a commit message on GitHub itself. Instead, I just re-did the PR, here: 

https://github.com/callstack/react-native-paper/pull/4051

It's passed the semantic PR and lint steps.

PR #4076, opened 2023-09-13
fix: using direct dark prop in appbar

lukewalczak:
### Summary

Unblock using `dark` prop in the `Appbar`

### DIFF
--- a/src/components/Appbar/utils.ts
+++ b/src/components/Appbar/utils.ts
@@ -56,19 +56,15 @@ export const getAppbarColor = ({
     return color;
   }
 
-  if (isV3) {
-    if (isDark) {
-      return white;
-    } else {
-      return undefined;
-    }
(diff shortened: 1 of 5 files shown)


src/components/Appbar/utils.ts
```diff
@@ -47,6 +47,30 @@ export const getAppbarColor = (
...
+export const getAppbarColor = ({
+  color,
+  isDark,
+  isV3,
+}: BaseProps & { color: string }) => {
+  if (typeof color !== 'undefined') {
+    return color;
+  }
+
+  if (isV3) {
```
tjaniczek, line 59:
```
if (isDark) {
   return white;
}

return isV3 ? undefined : black;
```

seems easier to read

tjaniczek (approved):

PR #4077, opened 2023-09-15
feat: add `elevation` property for Menu component

alexzel:
Adds `elevation` property for Menu component so we can customize Surface shadow in MD3.

### Summary

This PR makes it possible to redefine Surface elevation in the Menu component, so shadow and default background color will be picked based on this value.

### Test plan

Make sure when the property is omitted the component looks the same as before the change. Provide Menu component with a custom elevation value.


lukewalczak (approved):

alexzel:
@lukewalczak thanks!

Issue #3669, opened 2023-02-10
Unknown margin on top of TextInput


somasekharkakarla:
### Current behaviour
The unknown margin on top

### Expected behaviour
There should be no margin

### How to reproduce?
```
 <View style={{flex:1, borderWidth:2, borderRadius:8, borderColor:'grey', justifyContent:'center'}}>
        <RNTextInput mode={'outlined'} outlineStyle={{borderRadius:0, borderWidth:0, backgroundColor:'red', margin:0, padding:0,flex:1, alignSelf:'center'}}/>
      </View>
```

### Preview
<img width="392" alt="Screenshot 2023-02-10 at 4 44 46 PM" src="https://user-images.githubusercontent.com/65530762/218095826-a24614a3-1c61-420a-92c9-51705223d911.png">

### Your Environment

| software                      | version
| --------------------- | -------
| android                        | 12
| react-native                | 0.70.5
| react-native-paper    | 5.0.0-rc.9

lukewalczak:
Please try to install the latest stable version `5.2.0` and if the issue still occurs please attach the [snack](https://snack.expo.dev/)

somasekharkakarla:
Hi @lukewalczak,

As you suggested I upgraded my react-native-paper lib to v5.2.0 but still the issue exists. Here is the snack link https://snack.expo.dev/qiAeK7t3h

lukewalczak:
Thanks, we'll check the issue.

tjaniczek:
Hi @somasekharkakarla !
<img width="403" alt="image" src="https://user-images.githubusercontent.com/7429558/220075158-4febee8d-1fd7-4644-a33a-8cf4b06aa636.png">

Please note this padding is needed so that the floating label won't get obscured while input field is focused. Input fields without the label are not recommended by Material due to accessibility issues: https://m3.material.io/components/text-fields/guidelines#d65e62c9-97a4-437c-a099-23acea26e68a

PR #4086, opened 2023-09-18
fix: adjust top padding for outlined input without label

lukewalczak:
Fixes: https://github.com/callstack/react-native-paper/issues/3669

### Summary

Sets the `paddingTop` value according to the `label` visibility.

### Test plan

Updated snapshot.

### DIFF
--- a/example/src/Examples/TextInputExample.tsx
+++ b/example/src/Examples/TextInputExample.tsx
@@ -656,6 +656,9 @@ const TextInputExample = () => {
               />
             </View>
             <View style={styles.inputContainerStyle}>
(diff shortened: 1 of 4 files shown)


davismj:
This is the greatest and best commit.

davismj:
It seems that this change may have affected the positioning of adornments, as their position seems to be computed assuming the presence of a label.

image

lukewalczak:
Thanks @davismj, I had to overlooked it. Will check it

lukewalczak:
Fixed and released in `5.10.6`

PR #4092, opened 2023-09-21
fix: allow pass-through of accessibilityRole prop on Button.tsx

meatnordrink:
### Summary

In cases where a button is used because it is stylistically desired, but the `onPress` event opens a link, we should be able to set the `accessibilityRole` of the button as `link` for screen readers. Otherwise, it should remain `button` by default.

The inclusion of an icon or styling can indicate link behavior to a sighted user, but is not detectable by a screen reader.

### Test plan

* Pass `accessibilityRole='link'` as a prop to a button, and navigate to the button via screen reader. It should announce as a link.
* Do not pass any `accessibilityRole` prop, and navigate to the button via screen reader; it should announce as a button.

I have tested this locally, and it works as expected.


src/components/Button/Button.tsx
```diff
@@ -88,6 +89,10 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
    * Accessibility hint for the button. This is read by the screen reader when the user taps the button.
    */
   accessibilityHint?: string;
+  /**
+   * Accessibility role for the button. Button by default.
```
lukewalczak, line 93:
```suggestion
   * Accessibility role for the button. The "button" role is set by default.
```

lukewalczak:
Hey @meatnordrink, could you please resolve the conflict?

src/components/Button/Button.tsx
```diff
@@ -321,7 +327,7 @@ const Button = (
         delayLongPress={delayLongPress}
         accessibilityLabel={accessibilityLabel}
         accessibilityHint={accessibilityHint}
-        accessibilityRole="button"
+        accessibilityRole={accessibilityRole || 'button'}
```
lukewalczak, line 330: Can we move the "button" role to the line #172 as a default value for `accessibilityRole` prop?

lukewalczak (approved):

PR #4095, opened 2023-09-22
fix: configureFonts should be deterministic

### LINKED ISSUE #4094: configureFonts is mutating the base typescale object
### Current behaviour
Every call to `configureFonts` mutates the base typescale object, which renders this function not deterministic.

### Expected behaviour
Every call to `configureFonts` should return a clean fonts config based on a fixed default config plus a custom config passed as parameter.

### How to reproduce?
In the following Snack, the text should not be colored. `themeA` is not even used.

https://snack.expo.dev/@alexfournier/configurefonts

### Your Environment

| react-native-paper    | 5.10.6

alex-fournier:
Fixes #4094 

Every call to `configureFonts` should always return the same result for a given configuration.

### Summary

Previous calls to `configureFonts` mutated the base typescale object, resulting in a corrupted fonts object.
Each call to `configureFonts` should be the result of a fixed default config plus a custom config passed as argument.

### Test plan

If you call `configureFonts` multiple times with different configs, you should not find custom config from previous calls in your resulting object.

### DIFF
--- a/src/styles/__tests__/fonts.test.js
+++ b/src/styles/__tests__/fonts.test.js
@@ -271,4 +271,20 @@ describe('configureFonts', () => {
       },
     });
   });
+
+  it('should be deterministic', () => {
+    // first call that should not mutate the original config
+    configureFonts({
+      config: {
+        labelMedium: {
+          color: 'coral',
+        },
+      },
+    });
+
+    // second call that should return the original config without modification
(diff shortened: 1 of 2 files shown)


lukewalczak (approved):

lukewalczak:
Thanks @alex-fournier!

Issue #4114, opened 2023-10-06
Textinput's outline appears behind the label if Text component is passed to label


martenjurgens:
### Current behaviour
If Text component is passed to the label of a TextInput, the outline isn't hidden properly and appears behind the label.

### Expected behaviour
Acts the same as if a string is passed to the label.

### How to reproduce?

```javascript
const [text, setText] = React.useState("");

<TextInput
   // label={"Last name"}
  label={<Text style={{ color: "black" }}>Last name</Text>}
  value={text}
  onChangeText={text => setText(text)}
   />
``` 

### Preview

How it looks like when Text component is passed to label
![image](https://github.com/callstack/react-native-paper/assets/55079581/30f1f55f-b0a6-4835-9f8e-dd587ec0ea73)

How it looks like when string is passed to label
![image](https://github.com/callstack/react-native-paper/assets/55079581/6ea807f5-4178-4cff-913b-1edc7b0cfcbc)

### Your Environment

| software                      | version
| --------------------- | -------
| android                        | 13
| react-native                | 0.72.5
| react-native-paper    | 5.10.6

Lev-Shapiro:
I'll start working on this today soon, please assign this issue to me

lukewalczak:
Thanks @Lev-Shapiro!

WanYingT:
How do I remove the label background just like this?
<img width="313" alt="Screenshot 2023-11-03 at 10 19 12 PM" src="https://github.com/callstack/react-native-paper/assets/45457137/95501d23-030b-4085-99dc-a3fe5a6a0f7e">

Lev-Shapiro:
I just checked my code that was merged and it doesn't seem to have any problems with background labeling. Could you share why you decided to comment on this issue? Anyway, I'll be happy to fix this

WanYingT:
I found that passing RNP Text component  to the label props will still have this problem occurred, while using RN Text component is totally fine. Also, the problem I mentioned happens when I  add { borderRadius: 30 } to the TextInput component to make it close to design, label background seems to be overflow. I understand I can remove the white background by using colors: { background: 'transparent' }, but it will not leave a field for that. Is there any possible approach to fix?

Lev-Shapiro:
Yes, creating a new issue and giving code to reproduce the problem

PR #4115, opened 2023-10-06
fix(text-input): remove text duplication of label as component

Lev-Shapiro:
### Summary

Fixed issue #4114.

Removed duplication of label in TextInput component when using label as a component ( and not string )

### Test plan

Added example, so that it can be checked and used in future

### DIFF
--- a/example/src/Examples/TextInputExample.tsx
+++ b/example/src/Examples/TextInputExample.tsx
@@ -295,12 +295,7 @@ const TextInputExample = () => {
             <TextInput
               mode="outlined"
               style={[styles.inputContainerStyle, styles.fontSize]}
-              label={
-                // eslint-disable-next-line react-native/no-inline-styles, react-native/no-color-literals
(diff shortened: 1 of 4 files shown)


example/src/Examples/TextInputExample.tsx
```diff
@@ -290,6 +292,21 @@ const TextInputExample = () => {
                 />
               }
             />
+            <TextInput
+              mode="outlined"
+              style={[styles.inputContainerStyle, styles.fontSize]}
+              label={
+                // eslint-disable-next-line react-native/no-inline-styles, react-native/no-color-literals
+                <Text selectionColor="blue" style={{ color: 'orange' }}>
```
lukewalczak, line 300: Move it to the `StyleSheet` and use the color from `MD3Colors`

Lev-Shapiro:
@lukewalczak I'm new to contributing, so idk if it notifies you when I make a new commit, so I'm just writing to notify you that I've updated the code and waiting for review

PR #4125, opened 2023-10-11
Update Button.tsx

td-tomasz:


lukewalczak:
Hello @td-tomasz-joniec, could you kindly provide details on the problem that the PR is intended to address?

td-tomasz-joniec:
> Hello @td-tomasz-joniec, could you kindly provide details on the problem that the PR is intended to address?

Button uses internally Text which has its own prop adjustsFontSizeToFit, this change will allow to use this prop in order to scale text when there is little space available.

PR #4166, opened 2023-11-05
docs: Add Bluebirding Application to Showcase

crussell122:
Adding the Bluebirding Application to the Showcase.

Looks like there's a few instances of showcase images that are being streched/squished due to the 9:16 aspect ratio of the `showcase-image` CSS class. Is there any reason why images are required to have a 9:16 aspect ratio?


iM-GeeKy:
@lukewalczak I helped @crussell122 get this started, and it looks like there are some missing dependencies to get the docusaurus site up and running, hence the updated `package.json` and `yarn.lock`.

As for the aspect ratio issue I provided @crussell122 with a screenshot directly from the iOS simulator of our application. I had to fix this in [react-native-paper-dates](https://github.com/web-ridge/react-native-paper-dates/pull/345), the library I co-author, when I copied the showcase code you all have.

PR #4191, opened 2023-11-20
feat: pressable card title

mtch99:
### Motivation
I needed the option to press the card title

### Test plan
Two test cases on the Card.Title component
      - Card.Title without onPressTitle props {...props, onPressTitle:undefined}
            make sure the card functions normally
      - Card.Title with onPressTitle props = () => {Alert.alert("Pressed")}

Both are tested in the Card example👇
Capture d’écran, le 2023-11-19 à 23 18 44

*After press event on card title*

Capture d’écran, le 2023-11-19 à 23 20 32

### DIFF
--- a/src/components/Card/CardTitle.tsx
+++ b/src/components/Card/CardTitle.tsx
@@ -234,7 +234,6 @@ const styles = StyleSheet.create({
     flex: 1,
     flexDirection: 'column',
     justifyContent: 'center',
-    alignSelf: 'flex-start',
   },
 
   title: {
--- a/src/components/Card/CardTitle.tsx
+++ b/src/components/Card/CardTitle.tsx
@@ -159,11 +159,22 @@ const CardTitle = ({
   onPressTitle,
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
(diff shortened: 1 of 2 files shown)


src/components/Card/CardTitle.tsx
```diff
@@ -169,7 +176,18 @@ const CardTitle = ({
       ) : null}
 
       <View style={[styles.titles]}>
-        {title && (
+        {title && onPressTitle ? (
+          <TouchableOpacity onPress={onPressTitle} style={[styles.title]}>
```
lukewalczak, line 180: Avoid code repetition and move the title outside the render then used wrapped in pressable and without it.

lukewalczak, line 180: Use `Pressable` instead of `TouchableOpacity`

mtch99: After moving the Title component from the render, I did the same for subtitle to keep preserve coding style consistency, also had to rename some variables

mtch99: Are the changes okay? Or Is there something I have to correct? @lukewalczak

Issue #4214, opened 2023-12-05
RadioButton.Item does not take onLongPress


ondrejnedoma:
### Current behaviour
When an "onLongPress" prop is passed to RadioButton.Item, it is ignored and the function inside the prop does not run on a long press.

### Expected behaviour
The onLongPress event would be passed to the TouchableRipple component under the hood.

### How to reproduce?
Create a RadioButton.Item component and pass it the onLongPress prop and verify that the function does not fire.

### What have you tried so far?
I tried wrapping the RadioButton.Item with a TouchableWithoutFeedback element, but that did not do anything.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x
| android                        | 13
| react-native                | 0.72.7
| react-native-paper    | 5.11.1
| node                            | 18.17.1
| npm or yarn                | npm 9.6.7
| expo sdk                     | x.x.x

PR #4215, opened 2023-12-05
feat: added onLongPress to `RadioButton.Item` and `Checkbox.Item`

ondrejnedoma:
### Motivation

RadioButton.Item currently does not pass the "onLongPress" prop to the TouchableRipple component. This would be useful to implement to make RadioButton.Item elements be able to handle a secondary action.

### Related issue

#4214 

### Test plan

Create a RadioButton.Item and pass it an onLongPress prop with a function that logs something into the console.

### DIFF
--- a/src/components/RadioButton/RadioButtonItem.tsx
+++ b/src/components/RadioButton/RadioButtonItem.tsx
@@ -37,6 +37,10 @@ export type Props = {
    */
   onPress?: (e: GestureResponderEvent) => void;
   /**
+   * Function to execute on long press.
+   */
+  onLongPress?: (e: GestureResponderEvent) => void;
+  /**
    * Accessibility label for the touchable. This is read by the screen reader when the user taps the touchable.
    */
   accessibilityLabel?: string;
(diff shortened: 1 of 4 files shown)


src/components/RadioButton/RadioButtonItem.tsx
```diff
@@ -195,6 +200,7 @@ const RadioButtonItem = ({
                 event,
               })
             }
+            onLongPress={onLongPress}
```
kaushik1094, line 203: This is good, Thanks for making this change.

kaushik1094 (approved):
Thanks for making necessary changes.

lukewalczak:
Hey @ondrejnedoma, please rename the commit message to follow the library [convention](https://github.com/callstack/react-native-paper/blob/main/CONTRIBUTING.md#commit-message-convention)

ondrejnedoma:
Fixed the name, also took me a while to figure out how fork syncing works (or does not work) so sorry for the spam :D

Issue #4213, opened 2023-12-05
navigationState is showing deprecated


ananduremanan:
navigationState is showing deprecated with BottomNavigation.Bar. How do I fix this?

My Dependencies

```
"dependencies": {
    "@react-navigation/bottom-tabs": "^6.5.11",
    "expo": "~49.0.15",
    "expo-status-bar": "~1.6.0",
    "react": "18.2.0",
    "react-native": "0.72.6",
    "react-native-paper": "^5.11.3",
    "react-native-vector-icons": "^10.0.2"
  },
```
#### Screenshot

![image](https://github.com/callstack/react-native-paper/assets/48993261/ee3897af-c8cf-4511-b9d2-63a627491e4b)

Any other workaround for this?

KristampsWong:
same issues here, anyone fixed it?

lukewalczak:
Please install library version `5.11.4`

TobyEalden:
I'm still seeing this with `5.11.4`

![Screenshot 2023-12-21 at 08 30 32](https://github.com/callstack/react-native-paper/assets/1628365/22ce325c-c8ff-4b05-85db-40e70202f6a7)

![Screenshot 2023-12-21 at 08 31 16](https://github.com/callstack/react-native-paper/assets/1628365/3b90ae1e-b08a-4dc7-b793-5c63ef298b85)

amirreza-km:
I have the same issue now 
How to fix?

ananduremanan:
@amirreza-km it was fixed with the update for me, delete the package from npm module and delete the package lock json as well, then reinstall the package the issue will be gone..

amirreza-km:
> @amirreza-km it was fixed with the update for me, delete the package from npm module and delete the package lock json as well, then reinstall the package the issue will be gone..

I uninstalled and installed again but not fixed!
What version have you installed?

algorhythmic92:
This is not fixed. Even after a `rm -rf node_modules/ && rm package-lock.json && npm install` on 5.12.1, this issue persists:

package.json:
![image](https://github.com/callstack/react-native-paper/assets/149957752/77b54986-5d41-4048-bd30-68b9ccab2787)

Lock file resolved version:
![image](https://github.com/callstack/react-native-paper/assets/149957752/768c7cc8-9653-4963-8229-c2ca9f652824)

Your module:
[...]

PR #4223, opened 2023-12-11
fix: bottom navigation navigation state type correction

lukewalczak:
### Motivation

Correction jsdoc `deprecated` annotation, for documentation purposes, which makes `navigationState` as deprecated, which actually is required.

### Related issue

- #4213 

### Test plan

`navigationState` is still required and IDE doesn't mark it as `deprecated`. Documentation work as expected.

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -222,6 +222,7 @@ a .badge .badge-text {
   border-color: rgba(98, 0, 238, 1);
 }
 
+.badge-deprecatedProperty,
 .badge-deprecated {
   color: rgba(176, 0, 32, 1);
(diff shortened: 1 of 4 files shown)


zohaib-ahmad1020:
Found this issue on latest react native version 0.79.2 
React native paper version: 5.11.3

localhost:8081/index…pp.testing.uat:1170 Warning: TypeError: 0, _useLatestCallback.default is not a function (it is undefined)

This error is located at:
    at Chip (http://localhost:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&excludeSource=true&sourcePaths=url-server&app=com.adgm.app.testing.uat:229907:25)

Issue #4224, opened 2023-12-12
Error: Style property 'shadowOffset' is not supported by native animated module


jslok:
### Current behaviour
On rn 0.73.0 when using a rn-paper button with mode='contained', pressing on the button causes a crash with the listed error. Any other mode works fine. Only tested on Android.

Worked fine before I upgraded to rn 0.73.0.

### Expected behaviour
No error

### How to reproduce?
Tap a button with mode='contained'. 

### What have you tried so far?
rn-paper v4 and v5
disabling useNativeDriver in node_modules/react-native-paper/src/components/Button/Button.tsx makes it work but obviously not ideal.

### Your Environment

| software                      | version
| --------------------- | -------
| android                        | 13
| react-native                | 0.73.0
| react-native-paper    | v4 and v5 latest

jslok:
Same error with Chip with mode='outlined'

MatthieuKephyre:
``` diff --git a/node_modules/react-native-paper/src/components/Chip/Chip.tsx b/node_modules/react-native-paper/src/components/Chip/Chip.tsx
index 229f4e5..1c5bbdf 100644
--- a/node_modules/react-native-paper/src/components/Chip/Chip.tsx
+++ b/node_modules/react-native-paper/src/components/Chip/Chip.tsx
@@ -215,20 +215,22 @@ const Chip = ({
   const handlePressIn = useLatestCallback((e: GestureResponderEvent) => {
     const { scale } = theme.animation;
     onPressIn?.(e);
+
     Animated.timing(elevation, {
       toValue: isV3 ? (elevated ? 2 : 0) : 4,
       duration: 200 * scale,
-      useNativeDriver: true,
+      useNativeDriver: false,
     }).start();
   });
 
   const handlePressOut = useLatestCallback((e: GestureResponderEvent) => {
     const { scale } = theme.animation;
     onPressOut?.(e);
+
     Animated.timing(elevation, {
       toValue: isV3 && elevated ? 1 : 0,
       duration: 150 * scale,
-      useNativeDriver: true,
+      useNativeDriver: false,
     }).start();
   });```

This patch package fixed the problem for me.

The "elevation" animated value here is not used for ios, it's only user for android, so you can use `useNativeDriver: false` or before Animated.timing return void if Platform.OS === "ios"

tgermain-impirica:
Same issue here. Upgrading from RN 0.71.x to 0.73.0.

lucaswitch:
[...]

PR #4246, opened 2023-12-27
fix: adjust native driver value based on the rn version

lukewalczak:
### Motivation

Adjust the `useNativeDriver` value in shadow animation to be based on the `react-native` version.

### Related issue

- #4224 

### Test plan

Tested on react-native version 0.71, 0.72 and 0.73


mnahkies:
@lukewalczak I think this created a regression for those targeting web using `react-native-web` - I've raised a PR with more details / proposed fix here #4257

PR #4249, opened 2023-12-31
Update 02-theming.mdx replace "PaperProvider"

henrique-ft:
### Motivation

When i was coding with this library, i found that the "PaperProvider" is not exported anymore and now we have a "ThemeProvider" instead 🤔


PaitoAnderson:
You're probably using version `4.12.6` instead of `5.11.5`.

lukewalczak:
Everything is correct and there is no need to introduce any changes - we are exporting from the library both `ThemeProvider` and [`PaperProvider`](https://github.com/callstack/react-native-paper/blob/main/src/index.tsx#L14)

Issue #2581, opened 2021-03-03
TextInput: long text results in linebreak


Hektar90:
### Current behaviour

Typing a long text in TextInput and unfocusing it breaks the line

### Expected behaviour

Text is cut off at the end and is only visible by scrolling while editing

### Code sample

<img width="920" alt="Bildschirmfoto 2021-03-03 um 18 39 29" src="https://user-images.githubusercontent.com/15243732/109847750-db64a980-7c4f-11eb-8856-d8743d42f279.png">

### Screenshots (if applicable)

<img width="397" alt="Bildschirmfoto 2021-03-03 um 18 47 24" src="https://user-images.githubusercontent.com/15243732/109849068-3ba81b00-7c51-11eb-8629-04cde5baaf8a.png">

### What have you tried

set numberOfLines to 1 and multiline to false

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | ios
| react-native          | 0.63.4
| react-native-paper    | 4.7.2
| node                  | 12.14.0
| npm           | 6.13.4
| expo             | -

dingziqi:
[It seems to be caused by the `textAlign`](https://github.com/callstack/react-native-paper/blob/975cfa3be4956b629aa8255e226ff9ca068448eb/src/components/TextInput/TextInputOutlined.tsx#L336), only when `textAlign: default`, the ellipsis will work corectly。

Hektar90:
> [It seems to be caused by the `textAlign`](https://github.com/callstack/react-native-paper/blob/975cfa3be4956b629aa8255e226ff9ca068448eb/src/components/TextInput/TextInputOutlined.tsx#L336), only when `textAlign: default`, the ellipsis will work corectly。This is the bug of react-native, see the [issues](https://github.com/facebook/react-native/issues/29068#issue-631523357).
> 
> You can passdown `textAlign: 'auto'` to fixed it.

Unfortunately this is not working. Either adding textAlign:'auto' | 'left' | 'right' nor leaving the property out to take the default has no effect.
The latter leads to an type error saying textAlign needs to be set.

shllg:
I my case `textAlign: 'auto'` does not work.

Additionally, I recognized the following behavior in development:
- changing the vertical padding will "adjust" the component correctly. The ellipsis is still missing but the content is in one line
- entering any text and remove it again results again in line breaks

Initially, it renders correctly (except of missing ellipsis) and changes result in line breaks.

[...]

PR #4261, opened 2024-01-08
fix: textinput long text causes wrapping (ios)

john-trieu-nguyen:
### Motivation

Release 5.10.5 introduced #4084 which created a bug in TextInput (iOS) when there is a long text value.

iOS has a specific style requirement ([possibly related](https://github.com/facebook/react-native/issues/29068)) for TextInput to show the ellipsis and not wrap when there is a long text value.
```
RequiredStyle: {
    textAlign: 'auto',
    lineHeight: undefined,
}
```

#4084 prevents lineHeight having "undefined" as a possible value. Given that #4084 was specifically only addressing web requirements, this pull request introduces a web platform check.

### Related issue

Fixes: #2581 

### Test plan
Have written an example TextInput component to test for expected iOS behaviour.

Requesting some assistance in writing the Jest test.


john-trieu-nguyen:
Not sure how to write test to expect different behaviours between web and native.

lukewalczak (approved):

Issue #4266, opened 2024-01-11
IconButton TouchableRipple borderRadius is calculated from the icon size and not the button borderRadius


jahirfiquitiva:
### Current behaviour

The `IconButton` component `TouchableRipple` `borderRadius` is calculated based on the icon size and the button padding.

https://github.com/callstack/react-native-paper/blob/8e5d12da0685fc5073fd930d4043131b1c472f8b/src/components/IconButton/IconButton.tsx#L150-L155

https://github.com/callstack/react-native-paper/blob/8e5d12da0685fc5073fd930d4043131b1c472f8b/src/components/IconButton/IconButton.tsx#L186

### Expected behaviour

The `borderRadius` for the `TouchableRipple` should match the `borderRadius` set for the button in `style`s and fill the button surface

### How to reproduce?

https://snack.expo.dev/@jahirfiquitiva/rnp-iconbutton-ripple

### Preview

![Shot 2024-01-11 at 12 10 10@2x](https://github.com/callstack/react-native-paper/assets/10360816/edbe7e00-be5a-4f70-9b61-d1e7049e9b64)

### Alternative solution

Add a prop to customize the `TouchableRipple` `borderRadius`.

jahirfiquitiva:
I have updated the expo example to show the same issue with the `Button` component.

I have set a bigger padding and the `TouchableRipple` does not fill the `Button` surface

![Shot 2024-01-11 at 12 28 45@2x](https://github.com/callstack/react-native-paper/assets/10360816/29b7fc28-eabe-45c8-aeb4-ac2d35b8c798)

lukewalczak:
Hey @jahirfiquitiva, would you like to investigate the problem and propose the solution?

jahirfiquitiva:
@lukewalczak  I have opened the PR #4278 which should fix the issues and provide alternatives for the things that cannot be easily fixed

PR #4278, opened 2024-01-16
fix: Button and IconButton TouchableRipple borderRadius

jahirfiquitiva:
### Motivation

Resolve issue #4266 

#### Removes the `borderRadius` from the `IconButton` inner `TouchableRipple`.

Using `overflow: hidden` is enough for the inner views to fit the parent view, even taking care of the parent view `borderRadius`.

| Before | After |
| ------- | ------- |
| Shot 2024-01-15 at 20 26 40 | Shot 2024-01-15 at 20 27 22 |

After the changes, we can even set custom border radius for each corner of `IconButton` and the `TouchableRipple` will fill the button properly

https://github.com/callstack/react-native-paper/assets/10360816/55de1430-d59f-4844-a7bf-fc43b69aa731

#### Updates the `Button` inner `TouchableRipple` props to make the ripple fill the `Button` surface.

The `overflow: hidden` trick could not be applied to `Button` as its `Surface` can be elevated, and when using `overflow: hidden` the elevation shadow will not be displayed correctly.

> It's slightly hard to notice, but before, in the rounded corners it would be a space of ~1px due to the `borderWidth`

| Before | After |
| ------- | ------- |
| Shot 2024-01-15 at 20 23 54 | Shot 2024-01-15 at 20 22 40 |
| Shot 2024-01-15 at 20 39 19@2x | Shot 2024-01-15 at 20 38 37@2x |

#### Adds `contentStyle` prop to `IconButton`, similar to the one in `Button`

Previously there was no way to customize the `IconButton` padding, as if we set it to the `style` prop, it would cause the inner `TouchableRipple` to not fill the `IconButton`. Now, `contentStyle` allows doing so by setting the styles to the inner `TouchableRipple` view.

### Test plan

Run `yarn example web` to run the example project and view the updated examples for `Button` and `IconButton`

### DIFF
--- a/src/components/Button/utils.tsx
+++ b/src/components/Button/utils.tsx
@@ -4,6 +4,7 @@ import color from 'color';
 
 import { black, white } from '../../styles/themes/v2/colors';
 import type { InternalTheme } from '../../types';
+import { splitStyles } from '../../utils/splitStyles';
 
 export type ButtonMode =
   | 'text'
@@ -251,10 +252,14 @@ export const getButtonTouchableRippleStyle = (
 ): ViewStyleBorderRadiusStyles => {
   if (!style) return {};
   const touchableRippleStyle: ViewStyleBorderRadiusStyles = {};
+
+  const [, borderRadiusStyles] = splitStyles(
+    style,
+    (style) => style.startsWith('border') && style.endsWith('Radius')
+  );
+
   (
-    Object.keys(style).filter(
-      (key) => key.startsWith('border') && key.endsWith('Radius')
-    ) as Array<keyof ViewStyleBorderRadiusStyles>
+    Object.keys(borderRadiusStyles) as Array<keyof ViewStyleBorderRadiusStyles>
   ).forEach((key) => {
     const value = style[key as keyof ViewStyleBorderRadiusStyles];
     if (typeof value === 'number') {
--- a/src/components/Button/utils.tsx
+++ b/src/components/Button/utils.tsx
@@ -257,9 +257,11 @@ export const getButtonTouchableRippleStyle = (
     ) as Array<keyof ViewStyleBorderRadiusStyles>
   ).forEach((key) => {
     const value = style[key as keyof ViewStyleBorderRadiusStyles];
-    if (typeof value !== 'undefined' && typeof value === 'number')
-      touchableRippleStyle[key as keyof ViewStyleBorderRadiusStyles] =
-        value - borderWidth;
+    if (typeof value !== 'undefined' && typeof value === 'number') {
+      // Only subtract borderWidth if value is greater than 0
+      const radius = value > 0 ? value - borderWidth : 0;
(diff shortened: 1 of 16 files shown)


src/components/Button/utils.tsx
```diff
@@ -230,3 +230,36 @@ export const getButtonColors = ({
...
+    | 'borderBottomRightRadius'
+    | 'borderBottomStartRadius'
+    | 'borderTopEndRadius'
+    | 'borderTopLeftRadius'
+    | 'borderTopRightRadius'
+    | 'borderTopStartRadius'
+    | 'borderRadius'
+  >
+>;
+export const getButtonTouchableRippleStyle = (
```
lukewalczak, line 248: Could you please add a unit test covering that functionality?

jahirfiquitiva: @lukewalczak for this specific function? or for the `Button` component? Because the previous tests were still passing, meaning the functionality did not affect their styles. The only test that changed was for the outlined button, where the result inner border radius is 19 (20 - 1 (border))

lukewalczak: @jahirfiquitiva Perfectly for both cases

jahirfiquitiva:
@lukewalczak I have added a test for an outlined button with custom radius. There were already tests for non-outlined buttons with custom radius

jahirfiquitiva:
@lukewalczak would you mind taking another look, please?

src/components/Button/utils.tsx
```diff
@@ -230,3 +230,38 @@ export const getButtonColors = ({
...
+): ViewStyleBorderRadiusStyles => {
+  if (!style) return {};
+  const touchableRippleStyle: ViewStyleBorderRadiusStyles = {};
+  (
+    Object.keys(style).filter(
+      (key) => key.startsWith('border') && key.endsWith('Radius')
+    ) as Array<keyof ViewStyleBorderRadiusStyles>
+  ).forEach((key) => {
+    const value = style[key as keyof ViewStyleBorderRadiusStyles];
+    if (typeof value !== 'undefined' && typeof value === 'number') {
```
seb-zabielski, line 260: I think that `typeof value !== 'undefined'` is not necessary here, as condition is met only for `typeof value === 'number'`. `Value` with type `number` will not be `undefined`

src/components/Button/utils.tsx
```diff
@@ -230,3 +230,38 @@ export const getButtonColors = ({
...
+  >
+>;
+export const getButtonTouchableRippleStyle = (
+  style?: ViewStyle,
+  borderWidth: number = 0
+): ViewStyleBorderRadiusStyles => {
+  if (!style) return {};
+  const touchableRippleStyle: ViewStyleBorderRadiusStyles = {};
+  (
+    Object.keys(style).filter(
```
seb-zabielski, line 255: You can use `splitStyles` method from `src/utils/splitStyles.ts`, to keep code clean and consistent

lukewalczak (approved):

Issue #4282, opened 2024-01-18
disabled property on IconButton is not working


atheck:
### Current behaviour

The `onPress` handler is called even when the `IconButton` is disabled.

### Expected behaviour

The `onPress` handler should not be called when the `IconButton` is disabled. 

The docs say:

"A disabled button is greyed out and onPress is not called on touch."

### How to reproduce?

```tsx
<IconButton
  icon={"some-icon"}
  onPress={onPress}
  disabled={true}
/>
```

The icon is grayed out, but the button can still be pressed on `onPress` is called.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x
| android                        | 14
| react-native                | 0.72.9
| react-native-paper    | 5.12.1
| node                            | 20.10.0
| npm or yarn                | npm 10.2.4
| expo sdk                     | x.x.x

lukewalczak:
Hey @atheck, based on the code you've provided, I created the [snack](https://snack.expo.dev/@react-native-paper/icon-button---disabled) where you can observe the `onPress` callback is not triggered.

Closing the ticket, however feel free to reopen it with your issue reproduction.

atheck:
Hi, thank you for trying it. I figured out the reason for the behavior. I use a `Tooltip` around the `IconButton` that seems to cause the issue:

```tsx
<Tooltip title={"Title"}>
  <IconButton
    icon={'camera'}
    onPress={() => console.log("pressed")}
    disabled={true}
  />
</Tooltip>
```

lukewalczak:
@atheck will check it.

atheck:
Thank you @lukewalczak

PR #4291, opened 2024-01-22
fix: prevent pressing disabled children within tooltip

lukewalczak:
### Motivation

Adding a fix preventing triggering children's `onPress` callback wrapped within a `Tooltip` component.

### Related issue

- #4282 

### Test plan

Tested on app, updated example.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -122,6 +122,7 @@ const Tooltip = ({
   };
 
   const handleTouchStart = () => {
+    console.log('start');
     if (hideTooltipTimer.current.length) {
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -122,6 +122,7 @@ const Tooltip = ({
   };
 
   const handleTouchStart = () => {
+    console.log('start');
```
brunohkbx, line 125: @lukewalczak you forgot this console.log here

lukewalczak: Damn, thanks. Will fix later today

Issue #4294, opened 2024-01-22
TextInput right operand of 'in' is not an object


nimeshmaharjan1:
### Current behaviour

Getting this, "Right operand of 'in' is not and object" error after when using the TextInput component from react native paper. If I remove the TextInput component the app works fine. The TextInput from react-native is working though

### How to reproduce?

Just a newly created expo project with react native paper and rendering the TextInput component

### Preview

![image](https://github.com/callstack/react-native-paper/assets/86953478/9919620d-2d48-47b0-a7c3-7496f06dca09)

### Your Environment

| software                      | version
| --------------------- | -------

| react-native                | 0.73.2
| react-native-paper    | 5.12.1
| expo sdk                     | 50.0.1
| react                     | 18.2.0

AnMiZa:
Hey @nimeshmaharjan1 can You share Your `TextInput` code that causes that error?

harbouli:
The same Issue TextInput from react native paper not working

Aztaro97:
I'm facing the same issues with Searchbar components using Expo SDK 50.0.2

Aztaro97:
@nimeshmaharjan1 ,  I was able to fix the problem after following this [solution #756](https://github.com/marklawlor/nativewind/issues/756#issue-2092873397)

TAQ2:
@Aztaro97 sorry I'm a react native noob - how would you apply the fix? don't we need to wait until react-native-paper update their nativewind dependency?

gyanbhartip:
Could someone please explain why does this fix work? Even if short-circuiting is used, it would return a false instead of undefined (after the solution). But Does react-native not accept false as a value in style array? Why would it fail?

PR #4305, opened 2024-02-03
fix: :bug: Remove outline on web from TextInputFlat component

Aztaro97:
### Motivation
[Comment #4294](https://github.com/callstack/react-native-paper/issues/4294#issue-2094449471)

### Related issue
Isuue [ #4294](https://github.com/callstack/react-native-paper/issues/4294#issue-2094449471)

### Test plan

CI passes ✅


gyanbhartip:
Could someone please explain why does this fix work? Even if short-circuiting is used, it would return a false instead of undefined (after the solution). But Does react-native not accept false as a value in style array? Why would it fail?

Issue #4324, opened 2024-02-26
Broken button child width on Android


zvonimirr:
### Current behaviour
I have a button with a child that is a View with the following style.
```js
{
    width: "100%",
    flexDirection: "row",
    alignItems: "center",
    justifyContent: "space-between",
}
```

On iOS it works without any issues, however on Android it seems to be having a bit of trouble.

### Expected behaviour
I expect both platform to honor the width: "100%", and fill the entire button.

### How to reproduce?
```jsx
<Button>
  <View style={{
    width: "100%",
    flexDirection: "row",
    alignItems: "center",
    justifyContent: "space-between",
  }}>
    <Icon1 />
    <Text>
      View All Transactions
    </Text>
    <Icon2 />
  </View>
</Button>
```

### Preview
<img width="674" alt="Screenshot 2024-02-26 at 16 57 44" src="https://github.com/callstack/react-native-paper/assets/18758022/0d8eb7df-32ea-4a3e-a363-8cdcb18538dd">

Added the extra background color for clarity.

### What have you tried so far?
I tried experimenting with onLayout prop, but in the end just ended up hardcoding the padding on the Text element inside the View.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 17.0.1
| android                        | 13
| react-native                | 0.73.4
| react-native-paper    | 5.12.3
| node                            | 20.5.0
| npm or yarn                | 10.2.5
| expo sdk                     | 50.0.7

lukewalczak:
Tested on the latest version - `5.13.4`, didn't spot the issue:

ios | android
--- | ---
<img width="517" alt="Image" src="https://github.com/user-attachments/assets/4f52907b-7d6a-4b17-b818-a44f09199c2f" /> | <img width="437" alt="Image" src="https://github.com/user-attachments/assets/76ec257e-1688-4a9e-be14-bcc598312b42" />

PR #4325, opened 2024-02-26
feat(button): add trailing icon support

zvonimirr:
This PR adds a support for a trailing icon on the button component.

### Motivation

The motivation mostly stems from the issue I encountered in #4324.
I suppose I'm not the only one that needed an extra icon on their button so this PR adds exactly that.

### Related issue

Not necessarily related, but motivated me to open this PR. #4324 

### Test plan

Import the Button component and add a `trailingIcon` prop.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -75,7 +75,7 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
   /**
    * Trailing icon to display for the `Button`.
    */
-  trailingIcon? IconSource;
+  trailingIcon?: IconSource;
   /**
    * Whether the button is disabled. A disabled button is greyed out and `onPress` is not called on touch.
    */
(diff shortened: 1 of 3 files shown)


lukaevet:
+1

sztobar:
Wouldn't it be better to rename existing `icon` prop as `startIcon` and change `trailingIcon` to `endIcon`, just like material-ui does? https://mui.com/material-ui/react-button/#buttons-with-icons-and-label

sztobar:
> Wouldn't it be better to rename existing `icon` prop as `startIcon` and change `trailingIcon` to `endIcon`, just like material-ui does? https://mui.com/material-ui/react-button/#buttons-with-icons-and-label

Ok, my bad. It took me time to notice, that there's an easy way to make icon appear on the right side with `contentStyle={{ flexDirection: 'row-reverse' }}`: https://callstack.github.io/react-native-paper/docs/components/Button/#contentstyle

zvonimirr:
@sztobar correct. Just adding an icon and making it reverse is not a problem for me. The problem arises when I want to have two icons.

See #4324 to see what my original motivation was.

gedu:
Hey, we follow material design guidelines and 2 icons in a button is a Don't. I can open an internal discussion to see if we want to give better support to it: https://m3.material.io/components/buttons/guidelines#89ba49e3-31b5-47ae-9665-3e7f33452d0f

zvonimirr:
Makes sense @gedu. I think this can be closed then? I doubt it's a common use case to require dual icons in a button anyway

PR #4327, opened 2024-02-27
fix: allow pass-through of accessibilityRole on Chip

meatnordrink:
### Motivation

The `Chip` component is frequently used for display purposes, and in these cases is often never clickable. However, the `Chip` element still announces as a button because it has `accessibilityRole="button"` hardcoded.

When the `Chip` element is not being used as a button, a more sensible `accessibilityRole` should be able to be passed through. This PR keeps the default `"button"` `accessibilityRole`, but adds the ability to pass in a different `accessibilityRole` if `"button"` doesn't make sense.

### Related issue

n/a

### Test plan

* Pass `accessibilityRole='text'` as a prop to a Chip, and navigate to the Chip via screen reader. It should *not* announce as a button.
* Do not pass any `accessibilityRole` prop, and navigate to the chip via screen reader; it should announce as a button.

I have tested this locally, and it works as expected.

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -181,6 +181,7 @@ const Chip = ({
   disabled = false,
   background,
   accessibilityLabel,
+  accessibilityRole = 'button',
   closeIconAccessibilityLabel = 'Close',
   onPress,
   onLongPress,
@@ -323,7 +324,7 @@ const Chip = ({
         rippleColor={rippleColor}
         disabled={disabled}
         accessibilityLabel={accessibilityLabel}
-        accessibilityRole="button"
+        accessibilityRole={accessibilityRole}
         accessibilityState={accessibilityState}
         testID={testID}
         theme={theme}


billomore:
very simple, straightforward and important PR

gedu:
LGTM, this PR will be on version 5.12.4

jcubic:
@gedu can you explain why the PR was not merged, only is waiting for release?
Kind of odd release procedure.

gedu:
> @gedu can you explain why the PR was not merged, only is waiting for release? How you can release something if the PR was not merged?
> 
> I'm asking because I want to contribute, and no one even bother to merge a valid PR that was accepted. A lot of unanswered issues. Doesn't look like a well maintained project.

Hey, the main maintainer is on vacation, and I was hopping to chat with him to know more about the process, that's why I didn't merge it. On the other hand, the last weeks were really complicated on my side. I tried to replay Issues and on Discord
We really appreciate your contribution, hope you can keep doing it.

gedu (approved):

Issue #4351, opened 2024-03-21
Nested Text components do not inherit style


joebochill:
### Current behaviour
When nesting `Text` components, the child component does not inherit the style of the parent element.

### Expected behaviour
I would expect the styles to be inherited and overridden by the child component instead of simply overridden. The Text component from react-native exhibits this behavior.

### How to reproduce?
Nest a Text component inside another Text component and apply styles to both (e.g., color on one and fontWeight on the other). Note the nested component ignores the styles set on the parent component.

https://snack.expo.dev/lfm3PI-XWFR_YP56qwbn7

### Preview
![image](https://github.com/callstack/react-native-paper/assets/29152776/bc1d88c0-69de-4651-9014-112f03ef1f9a)

### What have you tried so far?
Replacing the nested Text element with the react-native version (instead of RNP) fixes the inheritance, but you lose the default styles from RNP that you may wish to preserve (e.g., if the nested Text is a different variant from the parent, you need to manually apply all of the variant styles to the nested element).

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 17.2
| android                        | 30
| react-native                | 0.73.2
| react-native-paper    | 4.9.2
| node                            | 18.17.1
| yarn                             | 1.22.19
| expo sdk                     | 50.0.0

rayan1810:
As per my understanding, this is caused due to the default color applied from the theme to a NativeText internally. Ideally, it should be overridden by the parent text if that has any conflicting style passed in style prop.

The solution I'm proposing is to provide the Text component with its own context that can be read by the child Text component and merged with the default styling from the theme. Here's the [PR](https://github.com/callstack/react-native-paper/pull/4356) with the suggested changes, It's still a WIP, hoping for a little review from the maintainers if this is aligned with them or not before completing it.

PR #4356, opened 2024-03-23
fix: nested text styling not inheriting from parent

rayan1810:
This addresses the issue #4351 by adding a React Context with Text Component and checks + reads the parent Text's context to get it's style and merges it with the current Text's style, just like how React Native behaves.

### Related issue
#4351

### Test plan
Testing is pending, the PR is still a Work in progress!


src/components/Typography/Text.tsx
```diff
@@ -149,20 +153,35 @@ const Text = (
...
   } else {
     const font = theme.isV3 ? theme.fonts.default : theme.fonts?.regular;
     const textStyle = {
       ...font,
       color: theme.isV3 ? theme.colors?.onSurface : theme.colors.text,
     };
+
+    if (parentTextContext) {
+      Object.assign(textStyle, StyleSheet.flatten(parentTextContext.style));
+    }
```
scarletcs, line 169: It seems to me as though this will overwrite any element-specific styling with the parent styling, such that this would output "Foo" in blue, not red:

```jsx
<Text style={{color: "blue"}}>
  <Text style={{color: "red"}}>
    Foo
  </Text>
</Text>
```

So the following code would be preferable instead:

```js
if (parentTextContext) {
  textStyle = Object.assign(StyleSheet.flatten(parentTextContext.style), textStyle);
}
```

rayan1810: Hey @scarletcs , Sorry for the late response but I don't think this would override the specific style, even the example you shared, I have tried it out and it seems to provide the expected results.

rayan1810:
Ready for Review ✅

PR #4392, opened 2024-05-01
fix: increase react version

gedu:
### Motivation

Increasing React-Native version to keep it updated

### Related issue

-

### Test plan

- Go through the app and check that nothing is broken

https://github.com/callstack/react-native-paper/assets/1676818/3a630b3a-7a0c-40ed-9d1c-c1548c76cef6

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,51 +54,58 @@
   },
   "devDependencies": {
     "@babel/core": "^7.20.7",
-    "@babel/plugin-proposal-export-namespace-from": "^7.18.9",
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -112,9 +112,6 @@
       "pre-commit": "yarn lint-no-fix && yarn typescript && yarn test"
     }
   },
-  "engines": {
```
raajnadar: Probably you can revert this change?

gedu: Cool, Yes I can add it back.

Issue #4407, opened 2024-05-14
TextInput.Affix - Right affix is not properly aligned when using onPress


megharora97:
### Current behaviour
In the latest package, when we are using TextInput.Affix with onPress then it is not visible.

### Expected behaviour
It should be aligned as Left Affix.

### How to reproduce?
To reproduce or give it a try you can check my snack where I added both Left & Right Affix and it's clearly visible that Left Affix is visible as expected but Right is not visible

### Preview
<img width="1290" alt="Screenshot 2024-05-14 at 9 09 55 PM" src="https://github.com/callstack/react-native-paper/assets/36987098/80877b47-bbea-4270-a75c-f8926f3b5ef3">

lukewalczak:
Closing via: https://github.com/callstack/react-native-paper/pull/4415

PR #4415, opened 2024-05-22
fix(text-input-affix): fix right TextInput.Affix alignment with onPress

seb-zabielski:
### Motivation

The position of the right TextInput.Affix is not calculated correctly when the onPress method is provided. 

### Related issue
#4407 

### Test plan

Run Example and add a right affix to TextInput with the onPress property. Affix should render the same as without onPress. You can also try the example from the related issue.

**The first recording shows how the component behaves before the fix (without and with the onPress property). The second is the result after the fix.**

https://github.com/callstack/react-native-paper/assets/165782545/91013325-ef40-458f-817e-7b6207961945

https://github.com/callstack/react-native-paper/assets/165782545/1b7e2a75-49bd-471a-8542-dc073c3f57da

### DIFF
--- a/src/components/TextInput/Adornment/TextInputAffix.tsx
+++ b/src/components/TextInput/Adornment/TextInputAffix.tsx
@@ -153,7 +153,18 @@ const TextInputAffix = ({
 
   const textColor = getTextColor({ theme, disabled });
 
-  const affix = (
+  const content = (
+    <Text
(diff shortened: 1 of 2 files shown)


gedu:
@seb-zabielski Cool thanks for the PR, can you provide some video showing the fix?

seb-zabielski:
@gedu I have updated the description and added the recordings there

gedu (approved):

Issue #4441, opened 2024-06-21
Shadow Issue with Colored React Native Elements in TextInput Label


IsmailParacha:
### Current behaviour
I'm using react-native-paper for TextInput components. When I pass React Native elements as the label and apply color to them, I see a shadow/replica of the elements behind them. However, if I don't apply any color, the elements display properly.

My requirement is to show an asterisk (*) in the field label, where the asterisk should be red and the label text should be black.

### Preview
https://github.com/callstack/react-native-paper/assets/84890110/21813ff8-f38e-4779-8d05-4d5f2d9d6c45

### code
```
import React, {useState} from 'react';
import {TextInput} from 'react-native-paper';
import {StyleSheet, View, Button, Text} from 'react-native';
const Home = () => {
const [surname, setSurname] = useState('');
   <View style={styles.container}>
      <TextInput
        label={
          <>
            <Text style={{fontSize: 20, color: 'red'}}>{'Surname'}</Text>
            <Text style={{fontSize: 20, color: 'red'}}> *</Text>
          </>
        }
        mode="outlined"
        onChangeText={text => {
          setSurname(text);
        }}
        value={surname}
        placeholder="Enter Surname"
        style={{width: '80%', height: 50}}
      />
      <TextInput
        label={
          <>
            <Text style={{fontSize: 20}}>{'Forename'}</Text>
            <Text style={{fontSize: 20}}> *</Text>
          </>
        }
        mode="outlined"
        onChangeText={text => {
          setSurname(text);
        }}
        value={surname}
        placeholder="Enter Forename"
        style={{width: '80%', height: 50}}
      />
 </View>
  );
};
const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
});

export default Home;
```

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 17
| android                        | 11
| react-native                | 0.73.4
| react-native-paper    | ^5.12.3

seb-zabielski:
Hey,
Can you try to use the label in a different way, without the Fragment?

For example something like this:
```js
label={
  <Text style={{ fontSize: 20, color: 'red' }}>
    Surname <Text style={{ fontSize: 20, color: 'red' }}>*</Text>
  </Text>
}
```

IsmailParacha:
i have tried to put a text component in the label instead of using fragment but got the same issue as you can see
[...]

PR #4442, opened 2024-06-24
fix(text-input-label-background): custom component background height

seb-zabielski:
### Motivation
Label with nested text components are displayed twice.

Previously, to calculate the background of a label, the label itself was rendered and the color of its text was set to `transparent`. Everything worked for the label as a string or a single <Text> element. For nested <Text> components, setting `color: "transparent"` didn't work and the text was duplicated (nested text color was not set as transparent). This pull request changes the way to get the right shape, as we already store the height of the label. Instead of rendering the label, we simply use the height value. 

### Related issue
#4441 

### Test plan
Run example app and use label from the related issue or example below

```js
<TextInput
  mode="outlined"
  label={
    <Text style={{ color: 'red' }}>
      First <Text style={{ color: 'blue' }}>Second</Text>
    </Text>
  }
  placeholder="Type something"
  value={outlinedText}
  onChangeText={(outlinedText) =>
    inputActionHandler('outlinedText', outlinedText)
  }
/>
```
 
| **BEFORE**                                                                                                       | **AFTER**                                                                                                       |
|------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
|  image | image |
| Red "First" is not duplicated, but blue "Second" is displayed twice. Second one is a tiny one                    | The label is not duplicated. The background has not changed                                                     |

### DIFF
--- a/src/components/TextInput/Label/InputLabel.tsx
+++ b/src/components/TextInput/Label/InputLabel.tsx
@@ -20,6 +20,7 @@ const InputLabel = (props: InputLabelProps) => {
     focused,
     opacity,
     labelLayoutWidth,
+    labelLayoutHeight,
     labelBackground,
     label,
     labelError,
@@ -157,6 +158,7 @@ const InputLabel = (props: InputLabelProps) => {
           {labelBackground?.({
             labeled,
             labelLayoutWidth,
+            labelLayoutHeight,
             labelStyle,
             placeholderStyle,
             baseLabelTranslateX,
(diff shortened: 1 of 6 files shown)


gedu:
Does this work even if I wrap the Text with View or Fragment

seb-zabielski:
`Fragment` yes. 
For `View` background also yes, but in this case label itself isn't positioned correctly. Position of the label is not related to this PR and it looks the same without this fix.

| **Fragment**                                                                                                    | **View**                                                                                                        |
|-----------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
| image | image |

gedu:
Awesome, just wanna double, we have at least 2 workarounds for now, thanks

lukewalczak (approved):

Issue #4445, opened 2024-06-27
[Android-Only] Snackbar visibility issue with RN New Architecture


bpeltonc:
Note: this is ONLY reproducible on Android with RN New Architecture enabled

### Current behaviour
When the Snackbar `visible` prop is set to `true`, the Snackbar is rendered onto the screen, but it's opacity remains as `0`, making it invisible. When the `visible` prop is set back to `false` the Snackbar "flashes" before disappearing again.

### Expected behaviour
When the Snackbar `visible` prop is `true`, the Snackbar should be rendered on the screen and visible.

### How to reproduce?
Please visit this [repo](https://github.com/bpeltonc/snackbar-bug-mre) and follow the readme.

### Preview
https://github.com/callstack/react-native-paper/assets/87504652/6cb3c56d-610b-4332-8104-90ea88894ca3

### What have you tried so far?
I have spend a considerable amount of time modifying the show/hide logic and the animation logic for the Snackbar component. I have isolated the issue to either the RN Animated API itself, or your library's use of it. By adding `react-native-reanimated` to the project and migrating all animations to their APIs, the component behaves as expected.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | N/A
| android                        | 14
| react-native                | 0.74.2
| react-native-paper    | 5.12.3
| node                            | 20.13.0
| npm or yarn                | 10.5.2 (npm)
| expo sdk                     | 51.0.14

seb-zabielski:
Hey @bpeltonc I've created pull request with the fix ;)

lukewalczak:
Closing – it should be fixed from `5.13.1`

PR #4447, opened 2024-07-01
fix(snackbar): fix show animation on new architecture

seb-zabielski:
### Motivation
This bug is only available on android with new architecture enabled and is well described (with and example) in the related issue #4445 
From what I've been able to find, the problem occurs when the animation is started before the component is rendered.
In our scenario, the component returns null as long as the `hidden` state is set to true. In the `handleOnVisible` function we have the main steps `setHidden(false)` and Animation start. To fix the bug, I moved the start of the animation to a useEffect that depends on the `hidden` value, so that the animation will start after the state is changed (and component re-rendered).

btw, since this bug only occurs with the new architecture, it is either also a react-native bug or an expected change. I have not found an answer yet. 

### Related issue
#4445 

### Test plan
Run android app with New Architecture enabled and try to show/hide snackbar.

| **Before**                                                                                            | **After**                                                                                             |
|-------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
| <video src="https://github.com/callstack/react-native-paper/assets/165782545/973ecf72-c749-499c-99fd-1d6b7d6f287e"> |<video src="https://github.com/callstack/react-native-paper/assets/165782545/eb498a83-700c-4607-85bb-5afeee816272"> |

### DIFF
--- a/src/components/Snackbar.tsx
+++ b/src/components/Snackbar.tsx
@@ -172,10 +172,9 @@ const Snackbar = ({
 
   const { scale } = theme.animation;
 
-  const handleOnVisible = useLatestCallback(() => {
-    // show
+  const animateShow = useLatestCallback(() => {
     if (hideTimeout.current) clearTimeout(hideTimeout.current);
-    setHidden(false);
+
     Animated.timing(opacity, {
       toValue: 1,
       duration: 200 * scale,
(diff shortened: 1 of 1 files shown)


iM-GeeKy:
@seb-zabielski @gedu Is there a tentative timeline for [5.12.4](https://github.com/callstack/react-native-paper/milestone/14). Also, random question, but is @lukewalczak no longer maintaining this?

lukewalczak (approved):

PR #4464, opened 2024-07-27
Typo behaviour => behavior

SantiagoSiordia:
Just a small typo in the issues template


BogiKay:
Hey, thank you very much for your contribution. I think both are correct and it depends on regional spelling 🇺🇸 vs 🇬🇧 , so there no need to introduce such a change, especially the PR doesn't change all occurrences in docs, but only one.

PR #4477, opened 2024-08-13
feat: bumped React Native and Expo version

kamilkedzierski:
### Motivation

We needed to bump the Android SDK version (see the attached screenshot for details). To accommodate this, we also had to update React Native and Expo, as the versions we were using were outdated and no longer supported.

#### Version Updates:

- React Native was upgraded using the [React Native Upgrade Helper](https://react-native-community.github.io/upgrade-helper/).
- Expo and related libraries were updated with the help of [Expo Doctor](https://www.npmjs.com/package/expo-doctor).

### Issues Encountered & Resolved:

- Icons were not displaying on the web.
- Some test cases failed.
- Assets from the react-native-paper library were not displaying in the example project.

All issues were resolved after the updates.

### Related issue

image

### Test plan

I would appreciate it if the reviewer could double-check the example app to ensure that no bugs were missed during the update. 🙏


gedu:
This looks good, I only found 2 issues.
FAB on web doesn't look good, but I think it is a known issue.
floatingActionButton_web

What I think it is from this PR, is on iOS when switching to RTL, the app restarts infinitely.


https://github.com/user-attachments/assets/35aec213-8267-44a8-b30e-608aaddf1874

Can we fix that?

example/package.json
```diff
@@ -9,44 +9,47 @@
     "start": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start",
```
gedu, line 9:
```suggestion
    "start": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --go",
```

iM-GeeKy:
> This looks good, I only found 2 issues. FAB on web doesn't look good, but I think it is a known issue. floatingActionButton_web
> 
> What I think it is from this PR, is on iOS when switching to RTL, the app restarts infinitely.
> 
>  Screen.Recording.2024-09-28.at.19.52.15.mp4 
> Can we fix that?

Here is the [FAB issue](https://github.com/callstack/react-native-paper/issues/4478) for future reference.

example/package.json
kamilkedzierski: added!

iM-GeeKy: @gedu Can we get this merged? It's an important PR toward gaining support for the new arch.

gedu:
RTL update LGTM

Issue #4206, opened 2023-11-29
Touch to display tooltip (Not long touch)


Steven-MKN:
**Is your feature request related to a problem? Please describe.**
I have an info icon like ℹ️ or similar. I would like my users to touch the icon, without long touch, as this would be more intuitive

**Describe the solution you'd like**
Can we add an option to touch without to long touch for the tooltip text to be triggered? I'd be happy to pick this up

**Describe alternatives you've considered**
I have tried setting the `enterTouchDelay` prop to something like `0` or `1` so that it might trigger much sooner. Seems like this delay is only applicable after a long touch is registered

**Additional context**
<img width="342" alt="Screenshot 2023-11-29 at 04 57 31" src="https://github.com/callstack/react-native-paper/assets/42501398/60761953-db93-4399-87f0-689fbfa9c96a">
an example where I'd prefer a touch instead of a long touch

Jhh910:
> **Is your feature request related to a problem? Please describe.** I have an info icon like ℹ️ or similar. I would like my users to touch the icon, without long touch, as this would be more intuitive
> 
> **Describe the solution you'd like** Can we add an option to touch without to long touch for the tooltip text to be triggered? I'd be happy to pick this up
> 
> **Describe alternatives you've considered** I have tried setting the `enterTouchDelay` prop to something like `0` or `1` so that it might trigger much sooner. Seems like this delay is only applicable after a long touch is registered
> 
[...]

PR #4479, opened 2024-08-19
feat: touch to display tooltip

Steven-MKN:
### Motivation

I would like my users to touch the info icon (not long touch), as this would be more intuitive when it's on a TextInput or similar.
286460785-60761953-db93-4399-87f0-689fbfa9c96a
an example where I'd prefer a touch instead of a long touch

### Related issue

https://github.com/callstack/react-native-paper/issues/4206

### Test plan

To test this I have created an example tooltip in a TextInput, the tooltip is set to display on touch. Here is an example:

[Screen_recording_20240819_155635.webm](https://github.com/user-attachments/assets/9321ae1c-da64-4c9f-9026-e2f236170dc4)


mnikolaus:
Why not just using `enterTouchDelay={0}` ?

quan-nh2:
It is a nice prop, but seems the repo dying 😢

Steven-MKN:
> Why not just using `enterTouchDelay={0}` ?

Hey, I explained here https://github.com/callstack/react-native-paper/issues/4206

Steven-MKN:
> It is a nice prop, but seems the repo dying 😢

Sad. In favour of what? I can't imagine a react native app without this lib tbh

Issue #2769, opened 2021-06-11
Menu does not avoid intersecting keyboard on Android


WesleyAC:
### Current behaviour

When a `Menu` is created near the bottom of the app, while the keyboard is expanded, the `Menu` opens behind the keyboard.

### Expected behaviour

I expect the menu to open upwards instead.

### Code sample

I don't have one that's easy to run, unfortunately :(

### Screenshots (if applicable)

![Screenshot_20210611-141306](https://user-images.githubusercontent.com/5001092/121749370-68c16080-cabf-11eb-8a72-2b4f3fb4fe8e.png)
![Screenshot_20210611-141336](https://user-images.githubusercontent.com/5001092/121749376-69f28d80-cabf-11eb-9ba7-a2651738c9e8.png)

### What have you tried

I looked at the documented props, and did not find any that controlled the direction that the `Menu` attempts to open in.

### Your Environment

| software              | version
| --------------------- | -------
| ios or android        | Android 11
| react-native          | 0.63.4
| react-native-paper    | ^4.8.1
| node                  | N/A
| npm or yarn           | yarn
| expo sdk              | N/A

WesleyAC:
[Here's a repro](https://snack.expo.io/vN5XcHE4R)

Tested on react-native-paper 4.9.1, same issue.

brunohkbx:
@WesleyAC how about wrapping it with [KeyboardAvoidingView](https://reactnative.dev/docs/keyboardavoidingview) ?

bombillazo:
Any solution to this issue?

PR #4488, opened 2024-08-23
fix(menu): refresh window layout when virtual keyboard is present

sneas:
### Motivation

Hi, my use case of the `Menu` element is somewhat peculiar: I have a text input in the menu. When the input is focused, and the virtual keyboard shows up - it covers the menu element.

**Before input is focused**

Simulator Screenshot - iPhone 15 Pro Max - 2024-08-22 at 17 06 15

**After input is focused**

Simulator Screenshot - iPhone 15 Pro Max - 2024-08-22 at 17 06 24

### Related issue

#2769

### Test plan

Do you think I should update examples to provide the use case?

I was also hesitant about providing a unit test because it would be bulky, rely on external mocks, and be fragile.

**When the fix is applied, the window layout gets recalculated, and the menu moves above the keyboard**

Simulator Screenshot - iPhone 15 Pro Max - 2024-08-22 at 17 06 54

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -326,7 +326,7 @@ class Menu extends React.Component<Props, State> {
   };
 
   private show = async () => {
-    const windowLayout = Dimensions.get('window');
+    const windowLayout = this.getWindowLayout();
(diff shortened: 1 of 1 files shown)


sneas:
sorry

Issue #4493, opened 2024-09-10
[Android] Wrong menu position in Menu component


alex-fournier:
### Current behaviour
In Menu component, the menu is not positioned as expected when visible. For example with `anchorPosition="bottom"`, the menu comes over the anchor.

### Expected behaviour
The menu should appear under anchor.

### How to reproduce?
https://snack.expo.dev/@alexfournier/ludicrous-green-churros

Tested with two physical devices: Google Pixel 7 and Samsung Galaxy S24.

| actual | expected
| -- | --
| ![actual](https://github.com/user-attachments/assets/f4c13c1e-6b82-430a-9ff3-80267e9e2fea) | ![expected](https://github.com/user-attachments/assets/9fe6629f-345f-45e0-bf02-eb597d6e45a0)

### What have you tried so far?
When passing the correct status bar height using `statusBarHeight`, it works fine. I think this piece of code is not working as expected as it returns 0 on my two devices:

```tsx
// src/constants.tsx
const estimatedStatusBarHeight =
  NativeModules.NativeUnimoduleProxy?.modulesConstants?.ExponentConstants
    ?.statusBarHeight ?? 0;
```

### Your Environment

| software                      | version
| --------------------- | -------
| android                        | 14
| react-native                | 0.74.5
| react-native-paper    | 5.12.5
| node                            | 20.10.0
| npm or yarn                | 1.22.19
| expo sdk                     | 51.0.32

alex-fournier:
I think the regression comes from this commit in Expo: https://github.com/expo/expo/pull/28014/files

Constants from expo-constants are not exported anymore so `NativeModules.NativeUnimoduleProxy.modulesConstants` is now empty.

We could use [`StatusBar.currentHeight`](https://reactnative.dev/docs/0.74/statusbar#currentheight-android) and apply this value only when status bar is set as translucent but I didn't find a way to know this without native code yet.

gedu:
Yes, there is a comment that we should replace it with `react-native-safe-area-context` probably with
```
import { useSafeAreaInsets } from 'react-native-safe-area-context';
```

alex-fournier:
I am working on a PR to fix that.

Geoffrey63:
Hello, I've also noticed this bug on Android when the menu is displayed from an Appbar. I've tried to compensate by adding some padding, but it doesn't work. Have you managed to fix it?

alex-fournier:
Hello, I have this fix in my code:

```jsx
[...]

PR #4495, opened 2024-09-13
fix(menu): dropdown position with translucent status bar

alex-fournier:
This PR replaces the way the status bar height is calculated on Android. Now I use `react-native-safe-area-context` to get the top inset. The old way using `NativeModules.NativeUnimoduleProxy?.modulesConstants` was no longer working since this commit in Expo: https://github.com/expo/expo/pull/28014/files

I removed the `expo-constants` dependency as it was only used for the old way of calculating status bar height that has been removed.

### Motivation

The initial problem was an incorrect position of the menu under the anchor when using `anchorPosition=bottom` with a translucent status bar in Android.

### Related issue

#4493

### Test plan

I added a test case in the example app. A menu with anchor position set to bottom. It is displayed the same, either with a status bar translucent or not.


alex-fournier:
In 66b8bfd1a14e253bdf419214a12c2b97eb9dcf9e, I fixed the component name of `Menu` and `Card` for the documentation website. (Card vs. CardComponent)

Capture d’écran 2024-09-13 à 15 03 05

PR #4498, opened 2024-09-17
fix(fab.group): subitems are accessibility focusable when collapsed

personalizedrefrigerator:
### Motivation

**Goal**: Improve the accessibility of the `FAB.Group` component by preventing hidden items from being focused.

### Related issue

Previously, FAB.Group items were accessibility focusable even when the menu was collapsed:

[recording1.webm](https://github.com/user-attachments/assets/f239db6e-8d22-4de7-8d14-0c89f6cab775)

**Note**: This only fixes the issue on Android and iOS. A similar issue persists on web.

### Test plan

**iOS**:
1. Open the example app in the simulator.
2. Open the "Floating Action Button" example.
3. Enable VoiceOver system-wide.
4. Move VoiceOver focus to the simulator.
5. Move VoiceOver focus to the next item. Verify that the item is not a hidden member of the `FAB.Group`.
6. Repeat step 5 until focus is on the "+" FAB.
7. Activate the current item.
8. Move VoiceOver focus to the next item until the floating calendar FAB has focus. Verify that the menu items ("unlabeled", "star", "email", "remind", toggle on long press") each are focusable.
   - Note: VoiceOver will focus items not in the FAB.Group menu during this step. This is a separate issue.

**Screen recording**:

[recording2.webm](https://github.com/user-attachments/assets/39674380-51c2-48c4-9342-d22260b4608d)

**Note**: The above screen recording also uses Expo SDK 51 (I was having trouble deploying the example with SDK 48).

### DIFF
--- a/example/src/Examples/FABExample.tsx
+++ b/example/src/Examples/FABExample.tsx
@@ -145,6 +145,7 @@ const FABExample = () => {
           <FAB.Group
             open={open}
             icon={open ? 'calendar-today' : 'plus'}
+            accessibilityLabel='Calendar FAB'
             toggleStackOnLongPress={toggleStackOnLongPress}
             actions={[
               { icon: 'plus', onPress: () => {} },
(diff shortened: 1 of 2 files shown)


lukewalczak (approved):
Thanks @personalizedrefrigerator for the contribution!

Issue #4209, opened 2023-12-01
How to create Material 3: Determinate Circular Progress Indicator


benjaminkomen:
### Ask your Question
I am looking to implement the determinate circular progress indicator as described [here](https://m3.material.io/components/progress-indicators/specs#16fa008b-87cd-4398-b026-b28bc2d14ad9). That way I can show the download progress between 0 and 100% as it slowly increases.

It seems that React Native Paper has:
- [An indeterminate circular progress indicator](https://callstack.github.io/react-native-paper/docs/components/ActivityIndicator)
- [A linear progress indicator](https://callstack.github.io/react-native-paper/docs/components/ProgressBar)

But not the specific thing I want, but maybe I'm missing something? Does it exist in react native paper, and if not, what are alternatives which work well? I saw [this](https://www.npmjs.com/package/react-native-circular-progress-indicator) library already. Or is this on the roadmap to be added to react native paper?

lukewalczak:
Hey @benjaminkomen, the determinate circular progress indicator is not implemented in the library, however, I believe it can be integrated. If you would like to create it, I will be more than happy to accept your PR

benjaminkomen:
Thanks @lukewalczak , that might be a fun Christmas holiday project, I'll see what I can do.

mfulton26:
Can we reopen this until support is added? I see there's a PR open but not yet merged. I too would like a circular determinate progress indicator.

craiganderson-iotv:
I would also love to have a determinate loading indicator, especially if the `<Button />` could have a determinate loading mode!

PR #4509, opened 2024-09-24
feat: add a determinate circular progress bar

xmtrc:
### Motivation

There is no determinant circular progress bar in react native paper yet.

### Related issue

#4209 

### Test plan

I've added some unit tests and added an example to the example app.


mfulton26:
This has been opened now for a bit. Any ideas on if it is ready to be merged and released?

Issue #4478, opened 2024-08-17
AnimatedFab Label Styling (Web)


iM-GeeKy:
### Current behaviour

The `AnimatedFab` component `label` should appears to wrap and ellipse when running on web, but works as expected on mobile.

### Expected behaviour

The `AnimatedFab` component `label` should span horizontally similar to mobile

### How to reproduce?

https://snack.expo.dev/@fitzwabs/thankful-violet-cheese

### Preview

<img width="300" alt="Screenshot 2024-08-17 at 4 36 36 PM" src="https://github.com/user-attachments/assets/b94f8479-0648-432e-ab50-6beb30116a70">

<img width="300" alt="Screenshot 2024-08-17 at 4 37 16 PM" src="https://github.com/user-attachments/assets/dec92623-a1c7-4a0c-9b61-37e26e86c9b0">

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 17.5
| android                        | 14.0
| react-native                | 0.74.5
| react-native-paper    | 5.12.5
| node                            | 20.16.0
| npm or yarn                | yarn
| expo sdk                     | 51.0.28

lukewalczak:
Tested provided snack and works as expected on the latest version - `5.13.2`.

PR #4567, opened 2024-11-28
fix(animated-fab): label styling (web)

BogiKay:
### Motivation

Currently, whole styling related logic in AnimatedFAB component relies on `onTextLayout`, which is mobile specific, so FAB's label is not styled well in web.  

### Related issue

https://github.com/callstack/react-native-paper/issues/4478

### Test plan

* Animated FAB is looking good when example app is running in web browser and there's no regression for mobile devices.

### DIFF
--- a/src/components/FAB/utils.ts
+++ b/src/components/FAB/utils.ts
@@ -436,13 +436,17 @@ export const getExtendedFabStyle = ({
   return isV3 ? v3Extended : extended;
 };
 
+let cachedContext: CanvasRenderingContext2D | null = null;
+
 const getCanvasContext = () => {
-  const fragment = document.createDocumentFragment();
+  if (cachedContext) {
+    return cachedContext;
+  }
+
   const canvas = document.createElement('canvas');
+  cachedContext = canvas.getContext('2d');
 
-  fragment.appendChild(canvas);
-
-  return canvas.getContext('2d');
+  return cachedContext;
 };
 
 export const getLabelSizeWeb = (ref: MutableRefObject<HTMLElement | null>) => {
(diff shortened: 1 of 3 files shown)


ericpoulinnz:
@BogiKay is this able to be merged in soon? I'd like to use this component on our web app

BogiKay:
> @BogiKay is this able to be merged in soon? I'd like to use this component on our web app

I'll take care of it. I'll do my best to have it merged by the end of next week.

src/components/FAB/utils.ts
```diff
@@ -428,3 +435,36 @@ export const getExtendedFabStyle = ({
 
   return isV3 ? v3Extended : extended;
 };
+
+const getCanvasContext = () => {
```
gedu, line 439: I'm not expert on web, won't impact the performance creating everytime a canvas? will work cache it? something like
```
let cachedCanvasContext: CanvasRenderingContext2D | null = null;

const getCanvasContext = () => {
  if (!cachedCanvasContext) {
    const canvas = document.createElement('canvas');
    cachedCanvasContext = canvas.getContext('2d');
  }
  return cachedCanvasContext;
};
```

BogiKay: Valid suggestion

I did some performance testing for 100 FAB component instances rendered and TBH I didn't notice significant difference between variant with caching and the one without it except little memory consumption improvement, but still it's improvement, so caching added.  💪 

Screenshot 2025-01-24 at 21 11 01

ericpoulinnz:
This looks good to me! Is it able to be merged? It's one of a couple of issues holding up a new web launch for my team.

gedu (approved):
LGTM

Issue #4544, opened 2024-11-07
ProgressBar crashes app in new arch on android


pnthach95:
### Current behaviour
In new arch android, `ProgressBar` crashes app when long length float value is used in `progress`. It works with 1/2, 1/4 but not 1/3.

Error log: `Warning: Error: Exception in HostFunction: Loss of precision during arithmetic conversion: (long) 33.33333333333333`

It works on iOS but you have this error in xcode log, `Error while converting prop 'accessibilityValue': Loss of precision during arithmetic conversion: (long long) 33.33333333333333`

### Expected behaviour
`ProgressBar` should work with long length float value in new arch on android

### How to reproduce?

- Create new project and enable new arch
- Run this code and press buttons

```jsx
const [progress, setProgress] = useState(1);

<ProgressBar progress={progress} />
<Button
  onPress={() => {
     setProgress(1 / 2);
  }}>
  1/2 work
</Button>
<Button
  onPress={() => {
     setProgress(1 / 3);
  }}>
  1/3 not work
</Button>
```

### Preview

https://github.com/user-attachments/assets/70128034-f190-4e9a-8c81-0efd6336fdcc

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 18.1
| android                        | 14
| react-native                | 0.76.1
| react-native-paper    | 5.12.5
| node                            | 20.12.1
| yarn                | 3.6.4
| expo sdk                     | x.x.x

lovegaoshi:
same problem as https://github.com/Sharcoux/slider/issues/102. u can just patch package the accessibilityvalur with a math. round

duanykruger:
Uppp, same problem! I tried the solution above, but still have the same problem.

AlenMartinez:
You must make sure that it is always an integer and not a float 33.33 is not an integer.

netdown:
@AlenMartinez ProgressBar expects progress value between 0 and 1.

ThomasLaSalmonie:
Unfortunately I don't have a lot of time to investigate before next week but I face the same issue when I wanted to upgrade to expo 51 to 52. I'm using my iOS device (Iphone 14 Pro) iOS Beta 18.2

But I manage to make thing work using the props `animatedValue` instead of `progress` keep the same float (the fix using Math.round was not working for me either)
 I know it's not exactly the same feature but At least the component work.

BogiKay:
Current, I'm working on a new arch support and this issue going to be fixed as part of the work

[...]

PR #4568, opened 2024-11-29
Fix - ProgressBar crashes app in new arch on android

mmuminovic:
### Motivation
In new arch android, ProgressBar crashes app when long length float value is used in `progress`

Error log: Warning: Error: Exception in HostFunction: Loss of precision during arithmetic conversion: (long) 33.33333333333333

So I did a fix and now we can't have long decimal numbers in `accessibilityValue`

### Related issue
https://github.com/callstack/react-native-paper/issues/4544

### Test plan
To reproduce a bug put in ProgressBar prop `progress={0.07}`

0.07*100 in JS is equal to 7.000000000000001

and that was my case.

I tested using dependencies:
"expo": "^52.0.0",
"react-native": "0.76.3",
"react-native-paper": "^5.12.5"

Everything works fine now

### DIFF
--- a/src/components/ProgressBar.tsx
+++ b/src/components/ProgressBar.tsx
@@ -203,7 +203,9 @@ const ProgressBar = ({
       accessibilityRole="progressbar"
       accessibilityState={{ busy: visible }}
       accessibilityValue={
-        indeterminate ? {} : { min: 0, max: 100, now: progress * 100 }
+        indeterminate
+          ? {}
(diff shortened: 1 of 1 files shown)


BogiKay:
@mmuminovic Thank you very much for your effort! 

Expected type of `now` is integer, so your solution won't work, because the result is float. 

The fix for this issue is already implemented as part of new-arch support work, so there's no need to make any changes on your end.

PR #4591, opened 2025-01-01
feat: allow usage of custom border radius for animated fab

babucarr32:
### Motivation
I was using the `AnimatedFAB` component, But I wanted it's `borderRadius` to be completely round. But, when I applied the `borderRadius` property on the style, it does not work. So, to fix that issue and allow the usage of custom `borderRadius`, I came up with this feature.

### Test plan
Assign any number to `borderRadius`, and it should work.
Screenshot 2025-01-01 at 10 32 40 PM
Screenshot 2025-01-01 at 10 32 55 PM

### DIFF
--- a/src/components/FAB/AnimatedFAB.tsx
+++ b/src/components/FAB/AnimatedFAB.tsx
@@ -419,7 +419,10 @@ const AnimatedFAB = ({
         </View>
         <Animated.View
           pointerEvents="box-none"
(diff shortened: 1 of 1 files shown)


adorum:
Hi, it would be nice to enhance animated fab with option to change border radius. Can you merge this PR?

babucarr32:
Alternative https://react-native-paper-abstracted.vercel.app

PR #4592, opened 2025-01-02
[BugFix - Severity HIGH] Hide modal through onDismiss callback

mdaamir151:
**Current behavior**: When user presses outside of modal, it gets dismissed internally regardless of whether app logic wants to close the modal and then user provided callback is invoked. This results in flicker - internal close and then immediate open due to `visible` state still `true` on modal if the app logic doesn't dismiss the modal immediately. Flicker is also caused if there are frame delays between user callback being invoked and `visible` state being set

**New Behavior**: Initiate dismiss modal only when app logic sets `visible` state to false

**Steps to reproduce bug**: Pass empty function as `onDismiss` callback and press outside of modal - modal flickers

### Motivation
Modal flickering on Android on close

https://github.com/user-attachments/assets/2e8f7294-b2f0-4855-9846-1a25014f4dd7

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -206,14 +206,14 @@ function Modal({
       accessibilityViewIsModal
       accessibilityLiveRegion="polite"
       style={StyleSheet.absoluteFill}
-      onAccessibilityEscape={hideModal}
+      onAccessibilityEscape={onDismissCallback}
       testID={testID}
     >
       <AnimatedPressable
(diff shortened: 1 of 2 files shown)


BogiKay:
Hey @mdaamir151, while working on new-arch support this bug was taken into consideration. Could you verify if the bug still occurs in the latest release 5.13.1?

mdaamir151:
It's fixed in 5.13.1
Thanks!

mdaamir151:
Resolved in 5.13.1

PR #4600, opened 2025-01-13
fix: issue when uncontrolled input value is set programmatically i.e …

shubhamdeol:
…without typing

### Motivation

We have a use case which I feel is very common. We have remark suggestions for an entry. Once suggestion is chosen from list of suggestions and value is set using setNativeProps method, uncontrolledValue is not updated this is resulting input label and palceholder to get overlapped.

This can also happen in cases where pastes the content into input without typing it.  

### Screenshots
- Before

Screenshot_20250113-170518

- After
Screenshot_20250113-170721

### Test plan
- I don't this any major testing is required. I have kept the changes to minimum. It should not effect the other functionality.


shubhamdeol:
@callstack-bot @lukewalczak why this PR is not addressed?

Issue #4579, opened 2024-12-18
Outdated `use-latest-callback` dependency causes type error


rsmidt:
### Current behaviour
When trying to use a simple `FAB.Group`, a type error is emitted because of the "use-latest-callback" dependency:

```
Warning: TypeError: 0, _useLatestCallback.default is not a function (it is undefined)

This error is located at:
    in CardComponent (created by FAB.Group)
    in RCTView (created by View)
    in View (created by FAB.Group)
    in RCTView (created by View)
    in View (created by FAB.Group)
    in RCTView (created by View)
    in View (created by FAB.Group)
    in RCTView (created by View)
    in View (created by FAB.Group)
    in RCTView (created by View)
    in View (created by FAB.Group)
    in FAB.Group (created by TeamActionsFab)
....
```

Seems like a simple dep 

### Expected behaviour
It should not throw a type error.

### How to reproduce?
The component definition:
```tsx
function TeamActionsFab({ teamId }: TeamActionsFabProps) {
  const [fabState, setFabState] = useState({ open: false });

  const onStateChange = ({ open }: { open: boolean }) => setFabState({ open });

  return (
    <Portal>
      <FAB.Group
        visible
        open={fabState.open}
        icon="plus"
        actions={[
          {
            icon: "calendar",
            label: i18n.t("app.teams.actions.add_event"),
            onPress: () => console.log("Pressed add event"),
          },
        ]}
        onStateChange={onStateChange}
      />
    </Portal>
  );
}
```

### What have you tried so far?
There's a similar issue in react-navigation: https://github.com/react-navigation/react-navigation/issues/12056.
I've tried the steps laid out there (changing resolution, manually installing a newer version, ...), but nothing worked out.

**What actually worked** is manually overriding the dependency in the package.json:
```json
  "overrides": {
    "use-latest-callback": "^0.2.3"
  }
```

If it's just a matter of bumping the version, I can prepare a PR.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x (not used yet)
| android                        | 35
| react-native                | 0.76.5
| react-native-paper    | 5.12.5
| node                            | 22.11.0
| npm or yarn                | npm  (10.9.0)
| expo sdk                     | 52.0.19

BogiKay:
Hey @rsmidt, this issue is not related to the paper package itself, but it may happening because npm resolves packages in different way if major version is 0, so I'll update dep to `0.2` to prevent using broken version.

[...]

PR #4623, opened 2025-02-08
chore: update use-latest-callback to 0.2.x

BogiKay:
### Motivation

There's one version 0.1.x of `use-latest-callback` that may cause a type error mentioned in #4579. 
Using caret with major version eq 0 doesn't allow to use higher minor version by default, so I decided to update this dep to 0.2.3 to reduce the risk of usage the broken version which is one of 0.1.x. 

### Related issue

#4579

### Test plan

Example app should work as before and no error should be thrown.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.9",
     "color": "^3.1.2",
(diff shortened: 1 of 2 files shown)


andrew-stupchuk:
Could you please reopen this PR? @gedu Reproducing the issue on the latest version of React Native for the Chip component

PR #4634, opened 2025-02-19
fix: fix Surface styling when flex set

### LINKED ISSUE #4632: iOS surface with `flex` style render weird.
### Current behaviour

In the source code, I noticed that iOS surface has outerLayer and inner view. However, outerLayerStyleProperties filter `flex`, than the inner will never has `flex` style. Therefore, it will never fill the remaining height. In android and web, it renders nomal. 

```
const [filteredStyles, outerLayerStyles, borderRadiusStyles] =
  splitStyles(
    flattenedStyles,
    (style) =>
      outerLayerStyleProperties.includes(style) ||
      style.startsWith('margin'),
    (style) => style.startsWith('border') && style.endsWith('Radius')
  );
```

### Your Environment

| react-native                | x.x.x
| react-native-paper    | 5.13.1

BogiKay:
### Motivation

PR fixes the issue when flex is applied to the Surface component and flex style is re-applied only to outer component, not the nested one, so styles can be broken if e.g `flex` and `justifyContent` are applied together.

Exact repro is provided in #4632

### Related issue

Fixes #4632

### DIFF
--- a/src/components/Surface.tsx
+++ b/src/components/Surface.tsx
@@ -202,7 +202,7 @@ const SurfaceIOS = forwardRef<
         ...(isElevated && getStyleForShadowLayer(elevation, 1)),
         ...filteredStyles,
         ...borderRadiusStyles,
-        flex: flattenedStyles.height ? 1 : undefined,
+        flex: flattenedStyles.height || flattenedStyles.flex ? 1 : undefined,
         backgroundColor: bgColor,
       };
 


dribble-njr:
I have a question, will the flex-related properties (outerLayerStyleProperties) take effect in the inner box? On andriod and web it will be applied to the only box.

Issue #4490, opened 2024-08-29
Text gets cut in half (vertically) for DataTable.Title 


jektvik:
### Current behaviour
Text gets cut in half (vertically) for DataTable.Title 

### Expected behaviour
Text should not get cut in half

### How to reproduce?
1. In accessibility settings of android, magnify font and icon size to maximum like you're 90 years old.
2. Create a DataTable and display on android f.ex.:
<DataTable>
        <DataTable.Header
          style={{
            backgroundColor: 'inherit',
          }}>
          DataTable.Title
            style={{
...
            }}>
            <Text style={{fontWeight: 'bold'}}>Temp &#127777;</Text>
          </DataTable.Title>

### Preview
![image](https://github.com/user-attachments/assets/0d84556c-8ca5-4430-992e-042e3851e95c)

### What have you tried so far?
A variety of css props. They didn't work because it seems that maxHeight is hardcoded inside the Title component.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | n/a
| android                        | 15
| react-native                | 0.74.3
| react-native-paper    | 5.12.5
| node                            | x20.11.0
| npm or yarn                | 10.2.4
| expo sdk                     | n/a

kacperzolkiewski:
Hi @jektvik,

I just created a PR with a fix for this issue. When you have a moment, could you check if it works as expected? Let me know if any adjustments are needed!

Thanks!

jektvik:
I appreciate very much.
Please don't wait on me - I am way out of context at this moment to reproduce and verify unfortunately.

PR #4640, opened 2025-02-28
fix: ensure DataTable.Title respects font scaling to prevent clipping

kacperzolkiewski:
### Motivation
Currently, text inside DataTable.Title gets cut off vertically when users increase font size in accessibility settings. This happens because the maxHeight does not account for scaled font sizes.

This PR ensures that the component properly respects PixelRatio.getFontScale(), preventing text clipping when font size is increased.

### Related issue
Fixes: [#4490](https://github.com/callstack/react-native-paper/issues/4490)

### Test plan

Steps to Reproduce the Issue (Before Fix)

- Open a screen with DataTable.Title.
- Go to Device Settings > Accessibility > Font Size and increase the font size.
- Observe that the text gets cut off vertically.

Steps to Verify the Fix

- Apply this PR’s changes.
- Increase font size in accessibility settings.
- Verify that the text is now fully visible and scales correctly without clipping.

### DIFF
--- a/src/components/DataTable/DataTableTitle.tsx
+++ b/src/components/DataTable/DataTableTitle.tsx
@@ -3,6 +3,7 @@ import {
   Animated,
   GestureResponderEvent,
   I18nManager,
+  PixelRatio,
   Pressable,
   StyleProp,
   StyleSheet,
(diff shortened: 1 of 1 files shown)


lukewalczak (approved):
Works as expected, thanks @kacperzolkiewski!

PR #1435, opened 2019-10-30
feat: add ability to change padding between FAB group items

mmarkusik:
### Motivation

This pull requests introduces a feature proposed in [Add ability to change padding between fab group children #1274](https://github.com/callstack/react-native-paper/issues/1274)

### Test plan

- Go to `FABExample` component
- Add `itemMargin` with numeric value to the rendered `FAB.Group` component
- Run the app and in the example the spaces between the items should change

PR #4657, opened 2025-03-17
feat: add ability to add styles to the action-wrapper in FABGroup

ChenBri:
Added an `actionWrapper` prop to the action object, enabling the ability to add styles of the action-item wrapper.

### Motivation
Currently, there is no way to access the wrapper component of a FABGroup action item. It prevents modifying aspects like spacing between action items. To address this, I added an `actionWrapper` prop within the action prop, allowing the ability to add styles to the wrapper styles.

### Related issue
https://github.com/callstack/react-native-paper/pull/1435


src/components/FAB/FABGroup.tsx
```diff
@@ -376,6 +378,7 @@ const FABGroup = ({
               <View
                 key={i} // eslint-disable-line react/no-array-index-key
                 style={[
+                  it.wrapperStyle,
```
lukewalczak, line 381: Please place it after `marginHorizontal`

Issue #4648, opened 2025-03-07
Card.Cover padding


Rankarusu:
### Current behaviour
I upgraded from react native version 75.4 to 76.7 today. Since that a card component with a cover has weird padding around the cover image.

### Expected behaviour
no padding around like before

### How to reproduce?
My component looks like this currently
```ts
<Card
            testID={`${tile.id}_Card`}
            disabled={disabled}
            key={tile.id}
            style={styles.card}
            onPress={() => setCurrentApp(app)}
          >
            <>
              <Card.Cover source={{ uri: tile.image }} src={tile.image} />
              {disabled && <Card.Cover style={styles.tint} />}
              {maintenanceUntil !== null && (
                <View style={styles.float}>
                  <Text variant="bodyLarge">
                    {t("applicationSelection.maintenance")}
                  </Text>
                </View>
              )}
            </>
            <Card.Title
              title={tile.title}
              subtitle={tile.description}
              titleStyle={disabled ? styles.disabledText : undefined}
              subtitleStyle={disabled ? styles.disabledText : undefined}
            />
          </Card>
```

### Preview

<img width="403" alt="Image" src="https://github.com/user-attachments/assets/2e481dba-2fa0-4aa4-8620-fb585dafd44e" />

### What have you tried so far?
Adding padding:0 to styles does nothing as it affects the View around the image and not the image itself. I looked into papers source code and found, that a padding of 16 is set on the image inside Card.Cover. I have no clue why this was not an issue before now.

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 18
| android                        | 15
| react-native                | 0.76.7
| react-native-paper    | 5.13.1
| node                            | 18.20.4
| yarn                             | 1.22.19

haydencrain:
+1 Can reproduce.

This is quite odd since Git blame shows this has been in the source code for 9 years https://github.com/callstack/react-native-paper/blame/366e0d63fd5ae0d5dc7799dbd97eee8e4062134c/src/components/Card/CardCover.tsx#L89

haydencrain:
For now I've just copied across from the source code and removed the padding and it works for my limited use cases

gcgoncalves:
Just to report I'm getting the same behaviour, but only on iOS, not on android.


software | version
-- | --
ios | 16.2
android | 16
react-native | 0.76.9
react-native-paper | 5.13.1
node | v23.11.0
yarn | 1.22.22

Rankarusu:
[...]

PR #4683, opened 2025-04-17
fix: remove cover padding

lukewalczak:
### Motivation

Removing padding around the image within `Card.Cover`

### Related issue

Fixes:  #4648 

### Test plan

Tested on iOS, Android and web, with all three modes `elevated`, `outlined` and `contained`.


gedu (approved):
LGTM

PR #4725, opened 2025-05-09
refactor: remove MD2 theme

lukewalczak:
### Motivation

Removing all the code related to `isV3` and MD2 in the `example` directory.

### Test plan

CI is failing due to ts and tests issues - it's ok, we are in the process of removing it also from the components.

Tested app on ios, android and web.


.gitignore
```diff
@@ -71,4 +71,6 @@ lib/
 !.yarn/plugins
 !.yarn/releases
 !.yarn/sdks
-!.yarn/versions
\ No newline at end of file
+!.yarn/versions
+
+example/ios/*
```
lukewalczak, line 76: TODO: Revert it

PR #4742, opened 2025-05-15
fix: add container prop and improve flex/layout handling

lukewalczak:
### Motivation

This PR improves the behavior of the `Surface` component regarding flexbox layout and style inheritance, especially on iOS, and introduces a new **internal** `container` prop for more explicit control. 
This prop allows internal components (like Button, Card, etc.) to explicitly specify whether the `Surface` should behave as a main layout container or as a simple wrapper.

### Related issue

#3652 
#4632 

### Test plan

Tested against two mentioned issues:

Zrzut ekranu 2025-05-15 o 13 01 37

### DIFF
--- a/src/components/Appbar/Appbar.tsx
+++ b/src/components/Appbar/Appbar.tsx
@@ -254,6 +254,7 @@ const Appbar = ({
         !theme.isV3 && { elevation },
       ]}
       elevation={elevation as MD3Elevation}
+      container
       {...rest}
     >
       {shouldAddLeftSpacing ? <View style={spacingStyle} /> : null}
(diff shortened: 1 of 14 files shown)


lukewalczak:
@durandt could you please whether it fixes the issue?

durandt:
> @durandt could you please whether it fixes the issue?

Sure @lukewalczak , I would be glad to help. I have only been working with RN for a year or so, what would be the best way to test your PR? Apply-patch? Or do you have the build artifacts somewhere? (I didn't find them in the PR build)

durandt:
Thank you very much @lukewalczak 🙏, that now works as intended with the provided patch. Wishing you a nice day 🌤️

PR #4743, opened 2025-05-15
docs: Add paragraph explaining how to test a specific PR/commit in CO…

durandt:
### Motivation

Make it easier for developers using `react-native-paper` to find information about how to target a specific PR/commit of `react-native-paper` to test Pull-Requests
I would have found this information helpful, but feel free to turn down the PR if you find it too trivial.

### Related issue

-

### Test plan

-


lukewalczak (approved):
I think we can keep it, so in case of similar questions in the future, we can easily redirect users to this section.

PR #4746, opened 2025-05-16
Add custom icon for checkbox component

tux2nicolae:
We use MaterialIcons in our app instead of MaterialComunityIcons and we have a problem with the checkbox not having a property for specifying a custom icon


tux2nicolae:
@lukewalczak can you check please?

PR #4818, opened 2025-10-10
fix: switch state sync when used inside a Portal on iOS

### LINKED ISSUE #4789: Switch component doesn't update visually on first press when used inside Portal/Modal
### Current behaviour
When using React Native Paper's Switch component inside a Portal/Modal, the switch doesn't update its visual state on the first press, even though the state change is properly triggered and logged. The switch only updates visually on the second press. This issue doesn't occur when the same Switch is used outside of Portal/Modal contexts.

### Expected behaviour
It should re render the switch as the state has changed.

### How to reproduce?
```
// Minimal reproduction case
import { useState } from 'react';
import { Modal, Portal, Switch } from 'react-native-paper';

export default function BugDemo() {
  const [visible, setVisible] = useState(true);
  const [checked, setChecked] = useState(false);

  return (
    <Portal>
      <Modal visible={visible}>
        <Switch 
          value={checked} 
          onValueChange={setChecked} // State updates but UI doesn't on first press
        />
      </Modal>
    </Portal>
  );
}
```

it works If I apply a dynamic key to the Switch like this 

`<Switch 
  key={`switch-${checked}`} // This fixes the issue
  value={checked} 
  onValueChange={setChecked}
/>`

### Your Environment

| ios                                | 18.5
| react-native                | 0.79.5
| node                            | 22.17.1
| expo sdk                     | 53.0.20

PaitoAnderson:
### Motivation

When a `<Switch />` is inside a `<Portal />` you must tap the switch component twice to visually toggle the state on iOS.

### Related issue

Fixes #4789

### Test plan

I updated the example app to Expo 54, and then added this code in the `SwitchExample.tsx` file and tried toggling with a single tap.

```tsx
<Portal>
  <View style={[styles.row, { top: 500 }]}>
    <TextComponent>Portal {checked ? 'on' : 'off'}</TextComponent>
    <Switch value={checked} onValueChange={setChecked} />
  </View>
</Portal>
```

### DIFF
--- a/src/components/Switch/Switch.tsx
+++ b/src/components/Switch/Switch.tsx
@@ -67,12 +67,32 @@ const Switch = ({
   ...rest
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
+
+  // Internal state is needed to fix #4789 where Switch is inside a Portal
+  const [internalValue, setInternalValue] = React.useState<boolean>(() =>
+    typeof value === 'boolean' ? value : false
+  );
+
+  React.useEffect(() => {
+    if (typeof value === 'boolean' && value !== internalValue) {
+      setInternalValue(value);
+    }
+  }, [value, internalValue]);
+
   const { checkedColor, onTintColor, thumbTintColor } = getSwitchColor({
     theme,
     disabled,
-    value,
+    value: internalValue,
     color,
   });
+
+  const handleValueChange = React.useCallback(
+    (newValue: boolean) => {
+      setInternalValue(newValue);
+      onValueChange?.(newValue);
+    },
+    [onValueChange]
+  );
 
   const props =
     version && version.major === 0 && version.minor <= 56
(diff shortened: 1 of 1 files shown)


src/components/Switch/Switch.tsx
```diff
@@ -67,13 +67,33 @@ const Switch = ({
...
+  // Internal state is needed to fix #4789 where Switch is inside a Portal
+  const [internalValue, setInternalValue] = React.useState<boolean>(() =>
+    typeof value === 'boolean' ? value : false
+  );
+
+  React.useEffect(() => {
+    if (typeof value === 'boolean' && value !== internalValue) {
+      setInternalValue(value);
+    }
+  }, [value, internalValue]);
```
gkueny:
```suggestion
  React.useEffect(() => {
    setInternalValue(oldValue => {
      if (typeof value === 'boolean' && value !== oldValue) {
        return value;
      }
      return oldValue;
    });
  }, [value, internalValue]);
```

Siimone:
any update on this?

Issue #4807, opened 2025-09-17
Menu can only be opened once on expo 54 / RN 0.81


LiamMorrow:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `react-native-paper@5.14.5` for the project I'm working on.
 I noticed that the `Menu` component in expo 54 (RN 0.81) can only be opened once, then it stays closed. This affects both iOS and Android. 
For some reason I noticed that the `finished` flag was always false on the `hide` callback, I'm unsure why this would be. This patch isn't perfect, it now seems like animations only work the first time, however it at least opens and closes fine. 

### Video of bug

https://github.com/user-attachments/assets/1272df15-bc65-4034-992e-bd07a7c5f186

### Video of patched version (no animations after first, but at least opens)

https://github.com/user-attachments/assets/b4419fe7-ef02-4c93-a2db-2dc0c326803b

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/react-native-paper/src/components/BottomNavigation/BottomNavigationBar.tsx b/node_modules/react-native-paper/src/components/BottomNavigation/BottomNavigationBar.tsx
index 05264c6..e2ca0db 100644
--- a/node_modules/react-native-paper/src/components/BottomNavigation/BottomNavigationBar.tsx
+++ b/node_modules/react-native-paper/src/components/BottomNavigation/BottomNavigationBar.tsx
@@ -975,6 +975,10 @@ const styles = StyleSheet.create({
     fontSize: 12,
     height: BAR_HEIGHT,
     textAlign: 'center',
+    // Unsure why - rendering issue the text is not being centered properly in ios
+    ...(Platform.OS==='ios' ?{
+      paddingLeft:6
+    }:null),
     backgroundColor: 'transparent',
     ...(Platform.OS === 'web'
       ? {
diff --git a/node_modules/react-native-paper/src/components/Menu/Menu.tsx b/node_modules/react-native-paper/src/components/Menu/Menu.tsx
index 55922c1..77f3400 100644
--- a/node_modules/react-native-paper/src/components/Menu/Menu.tsx
+++ b/node_modules/react-native-paper/src/components/Menu/Menu.tsx
@@ -299,10 +299,10 @@ const Menu = ({
           });
         }
       }),
-    [anchor]
-  );
+      [anchor]
+    );
 
-  const show = React.useCallback(async () => {
[...]

Issue #4763, opened 2025-06-23
Menu component closes and fails to open again on re-render


julwil:
### Current behaviour
- Open the Menu component
- The parent component re-renders
- Menu component closes and can't be opened again

### Expected behaviour
- Open the Menu component
- The parent component re-renders
- Menu component stays open

### How to reproduce?
```TSX
import React, {useState} from 'react';
import {View} from 'react-native';
import {Button, Menu, PaperProvider} from 'react-native-paper';

const MyComponent = () => {
  const [randomVarialbeToForceRerender, setRandomVarialbeToForceRerender] =
    useState(0);
  const [visible, setVisible] = useState(false);

  const openMenu = () => {
    console.debug('Opening menu');
    setVisible(true);

    /**
     * We update a state variable to force a re-render of the component.
     * A real-world example could be a network request or some other
     * asynchronous operation that updates the state.
     */
    setTimeout(() => setRandomVarialbeToForceRerender(Math.random()), 1000);
  };

  const closeMenu = () => {
    setVisible(false);
    console.debug('Closing menu');
  };

  console.debug('visible', visible);
  return (
    <PaperProvider>
      <View
        style={{
          paddingTop: 50,
          flexDirection: 'row',
          justifyContent: 'center',
        }}>
        <Menu
          visible={visible}
          onDismiss={closeMenu}
          anchor={<Button onPress={openMenu}>Show menu</Button>}>
          <Menu.Item
            onPress={() => {}}
            title={`Random  ${randomVarialbeToForceRerender}`}
          />
        </Menu>
      </View>
    </PaperProvider>
  );
};

export default MyComponent;
```

### Preview
https://github.com/user-attachments/assets/3d995026-ab4e-4fd3-bbf3-5544164a63ec

### What have you tried so far?
Memoization to avoid re-rendering the menu. However, this approach is limited and breaks
when the menu items are non-static (c.f. example)

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 18.5
| android                        | x
| react-native                | 0.77.2
| react-native-paper    | 5.13.1
| node                            | 18.20.8
| npm or yarn                | 3.8.7
| expo sdk                     | x.x.x

bjeld:
We ran into the same issue after upgrading to React Native 0.80  ("react-native-paper": "5.14.5") 
Only solution for now is to downgrade to React Native 0.79.4

bjeld:
[...]

PR #4842, opened 2026-01-08
chore: Update react-native to 0.82

ruben-rebelo:
Updates the react-native version to 0.82 and Example app expo to 54.
There were several updates to unit tests and typescript.

### Motivation

Just to keep everything updated

### Test plan

- All the components render and work as before
- Example app works as usual

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   default:
     docker:
-      - image: cimg/node:18.20.2
+      - image: cimg/node:20.19.0
     working_directory: ~/react-native-paper
 
 commands:
@@ -106,7 +106,6 @@ jobs:
           name: Build docs
           command: |
             export DOCUSAURUS_BASE_URL="/output/job/$CIRCLE_WORKFLOW_JOB_ID/artifacts/$CIRCLE_NODE_INDEX/docs/"
-            export NODE_OPTIONS="--require $(pwd)/docs/scripts/node-shims.cjs"
             cd docs && yarn build && cd ..
             bash .circleci/comment-artifacts.sh
       - store_artifacts:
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
   },
   "dependencies": {
     "@callstack/react-theme-provider": "^3.0.9",
+    "@react-native/babel-preset": "^0.82.1",
```
jbroma: this should be at most `devDependency` and never a `dependency` - this will get installed with the paper itself in the projects and the projects have their own babel config.

but...

I would avoid the preset altogether since every project will use that preset with it's own babel config. In theory, running the preset in the lib should reduce the transpilation time of the library in the projects, but every file still is going to be parsed by babel and then (most likely, needs verifying) transformations will try to run again but will result in `noop` since it was already transformed here.

src/components/Typography/AnimatedText.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { JSX } from 'react';
```
jbroma: shouldn't be needed to import this, probably a configuration issue somewhere

package.json
satya164: iirc it's needed for tests etc.

the lib isn't actually compiled with this since bob has an internal Babel config unless overridden. the main benefit of precompiling is to make it seamlessly work with anything other than metro (jest, vite, webpack) since they don't compile `node_modules` by default.

fracico-tech:
Please double check because with React Native > 0.81 the menu component is not working well.
See #4807, #4763

mosoakinyemi:
> Please double check because with React Native > 0.81 the menu component is not working well. See #4807, #4763

Thanks for bringing these up @fracico-tech !
@ruben-rebelo can we take a look at those issues related to >RN 0.81? 👀
- https://github.com/callstack/react-native-paper/issues/4807
- https://github.com/callstack/react-native-paper/issues/4763

salememd:
When this upgrade expected to be released? Thanks.

PR #4880, opened 2026-03-20
fix(TextInput): fall back to theme fontWeight to prevent SIGSEGV on Fabric

isaacrowntree:
### Motivation

`TextInputOutlined` and `TextInputFlat` destructure `fontWeight` from the user's `style` prop and spread it **after** the theme font object:

```ts
const { fontWeight } = (StyleSheet.flatten(style) || {}) as TextStyle;
// ...
textStyle: { ...font, fontSize, lineHeight, fontWeight }
```

When the user doesn't provide `fontWeight` in their style (the common case — most `<TextInput>` usage omits it), the destructured value is `undefined`. Spreading `{ ...font, fontWeight: undefined }` then overrides `font.fontWeight` (e.g. `'400'` from the theme) with `undefined`.

On React Native's **Fabric renderer** (New Architecture), this `undefined` flows as `nil` to the native `RCTGetFontWeight` function, which passes it to `CFStringFind` → `CFStringGetLength` on a NULL pointer → **EXC_BAD_ACCESS (SIGSEGV)**.

This is a **fatal crash** that kills the app on any screen with a `<TextInput>` that doesn't explicitly set `fontWeight` in its style prop. Reproduced on iOS with React Native 0.78 + react-native-paper 5.15.0 + Fabric enabled.

### Related issue

No existing issue — discovered via Crashlytics crash report in production. Native crash stack:

```
Thread 4 Crashed:
0  CoreFoundation    CF_IS_OBJC + 12
1  CoreFoundation    CFStringGetLength + 232
2  CoreFoundation    CFStringFind + 60
3  React             RCTGetFontWeight + 212
4  React             RCTFontWithFontProperties + 3024
5  React             RCTNSTextAttributesFromTextAttributes + 316
6  React             RCTNSAttributedStringFromAttributedString + 364
7  React             -[RCTTextLayoutManager measureAttributedString:...] + 76
```

### Test plan

1. Render `<TextInput mode="outlined" />` without a `fontWeight` in the style prop → previously crashes on Fabric, now uses theme font's fontWeight
2. Render `<TextInput mode="outlined" style={{ fontWeight: '700' }} />` → still applies the user-provided weight (nullish coalescing only kicks in for `undefined`/`null`)
3. Both `TextInputOutlined` and `TextInputFlat` are patched with the same fix
4. Lint, TypeScript, and tests should pass — the change is a single-line nullish coalescing fallback per file


satya164 (approved):
Thanks for the PR

satya164:
Can you update the snapshots?

isaacrowntree:
Snapshots updated in 4c70d4a — CI is green. Ready for another look @satya164, thanks!

PR #4882, opened 2026-03-23
fix: Add aria-hidden to interior radio button in RadioButtonItem

meatnordrink:
### Motivation

tl;dr - `RadioButton.Item` re-presents the interior `RadioButton` when using Android TalkBack.

Android Talkback is Android's screen-reader, and it behaves somewhat differently than iOS's VoiceOver (which is more commonly used for screen-reader auditing). Talkback allows access to buttons-within-buttons (iOS does not), which can be a confusing experience for Talkback users if both the outer and inner buttons on an element present themselves as distinct elements, when in fact it's just a button with an `onPress` on an outer container as well as the inner button.

In the case of `Radiobutton.Item`, which groups a radio button and a label, Talkback first presents the user with the pressable radio button + label, as intended. However, it then presents just the radio button, which is within the `RadioButton.Item`, as a separate element. This is confusing, and is never the intended behavior. 

This PR adds `aria-hidden`, which Talkback recognizes, to the inner `RadioButton` in `RadioButtonItem`, ensuring that Talkback does not read out the interior radio button as a separate element. It is still read out as part of the radio button + label group, as intended. 

This change has no effect on iOS, as iOS does not reveal child pressables to VoiceOver (precisely for this reason). 

It has no effect on the non screen-reader experience on any platform. 

### Related issue

https://github.com/callstack/react-native-paper/issues/4881

### Test plan

Attempt to navigate a `RadioButton.Item` with Android Talkback, iOS VoiceOver, and no screen reader. (I completed these tests locally.)

### DIFF
--- a/src/components/RadioButton/RadioButtonItem.tsx
+++ b/src/components/RadioButton/RadioButtonItem.tsx
@@ -177,7 +177,7 @@ const RadioButtonItem = ({
   let radioButton: any;
 
   if (mode === 'android') {
-    radioButton = <RadioButtonAndroid {...radioButtonProps} />;
+    radioButton = <RadioButtonAndroid {...radioButtonProps} aria-hidden />;
   } else if (mode === 'ios') {
     radioButton = <RadioButtonIOS {...radioButtonProps} />;
   } else {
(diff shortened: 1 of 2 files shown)


satya164 (changes requested):
Thanks for the PR. Since the snapshot tests are affected by this change, please run the tests locally to update the snapshots.

It'd also be great to add a comment above the prop describing why it's needed for future readers.

meatnordrink:
@satya164, I've added the comment, and updated the snapshot. 

My apologies for the long wait, I initially had some trouble building the repo locally to generate the snapshots (yarn and npm fighting on my machine), and hadn't been able to make the time to debug until now.

PR #4895, opened 2026-04-13
chore: remove md2

xdimota7x:
### Motivation

This PR removes Material Design 2 support and makes the library MD3-only.
The goal is to simplify the codebase by removing MD2/MD3 branching across components, theming, and utilities, reducing maintenance overhead and inconsistency risk.

### Related issue

Paper currently supports both MD2 and MD3, with many conditional code paths (isV3, version-specific theme logic) spread across components and core theming.
This PR addresses that by removing MD2 themes and all associated MD2 behavior/exports.

### Test plan

- All test suites should pass
- MD3 components should behave as they currently do
- Components previously branching on MD2/MD3 still render correctly (e.g. Button, Chip, Searchbar, Snackbar, TextInput, SegmentedButtons)


xdimota7x:
Closed in favor of #4898

Issue #4810, opened 2025-09-24
TouchableRipple ripple effect not working in RN 0.81


mo-elnady:
### Current behaviour
After upgrading to RN 0.81 all ripple effect in TouchableRipple component doesn't show anymore.

### Expected behaviour
rippleColour prop is not effective anymore, it should normally apply the selected colour.

### How to reproduce?
`    
          <TouchableRipple
              rippleColor={'blue'}
              onPress={
                () => console.log("Pressed");
              }
            >
              <Image />
            </TouchableRipple>
`

### Your Environment
| software                      | version
| --------------------- | -------
| android                        | 15
| react-native                | 0.81
| react-native-paper    | 5.14.5

farynaio:
I found it has some problems with theme styling with expo 54

ryonakae:
I was experiencing the same issue, but I was able to get the Ripple Effect to display on Android by modifying the code as follows:

```ts
import { TouchableRipple } from 'react-native-paper'
import { type PressableAndroidRippleConfig } from 'react-native'

<TouchableRipple
  // rippleColor={tapHighlightColor}
  underlayColor={tapHighlightColor}
  background={
    {
      color: tapHighlightColor,
      foreground: true,
    } as PressableAndroidRippleConfig
  }
  style={styles.content}
>
...
</TouchableRipple>
```

### Environment
- Android: 16
- react-native: 0.81.4
- expo: 54.0.12
- react-native-paper: 5.14.5

CatLover01:
Adding `borderless={true}` made it work for me

RenanSantos7:
> I was experiencing the same issue, but I was able to get the Ripple Effect to display on Android by modifying the code as follows:
> 
> import { TouchableRipple } from 'react-native-paper'
> import { type PressableAndroidRippleConfig } from 'react-native'
> 
> <TouchableRipple
>   // rippleColor={tapHighlightColor}
>   underlayColor={tapHighlightColor}
>   background={
>     {
>       color: tapHighlightColor,
>       foreground: true,
>     } as PressableAndroidRippleConfig
>   }
>   style={styles.content}
> >
> ...
> </TouchableRipple>
> 
> ### Environment
> 
>     * Android: 16
> 
>     * react-native: 0.81.4
> 
>     * expo: 54.0.12
> 
>     * react-native-paper: 5.14.5

This doesn't work for me on Android 16, expo 54.0.33 and react-native 0.81.5.

[...]

PR #4897, opened 2026-04-15
fix: ripple not showing

adrcotfas:
### Motivation

Ripple was not showing on several components like list items.

#### Order of merging:
- #4898
- #4897
- #4899
- #4900
- #4901

### Related issue
#4810
### Test plan

Open the example app with and without this fix and observe the fact that the ripple now works for the main screen list items and the Touchable example.


ruben-rebelo:
LGTM!

PR #4898, opened 2026-04-15
chore: remove md2

xdimota7x:
### Motivation

This PR removes Material Design 2 support and makes the library MD3-only.
The goal is to simplify the codebase by removing MD2/MD3 branching across components, theming, and utilities, reducing maintenance overhead and inconsistency risk.

### Related issue

Paper currently supports both MD2 and MD3, with many conditional code paths (isV3, version-specific theme logic) spread across components and core theming.
This PR addresses that by removing MD2 themes and all associated MD2 behavior/exports.

### Test plan

- All test suites should pass
- MD3 components should behave as they currently do
- Components previously branching on MD2/MD3 still render correctly (e.g., Button, Chip, Searchbar, Snackbar, TextInput, SegmentedButtons)


adrcotfas:
I think we should remove `src/components/Typography/v2` too together with any `isV3` usage which is always true.

xdimota7x:
Thanks for the suggestion @adrcotfas. Addressed the `Typography/v2`, as well as the version specific references from types and config in the latest commit.

src/components/Appbar/AppbarContent.tsx
```diff
@@ -102,8 +104,6 @@ export type Props = $RemoveChildren<typeof View> & {
  */
 const AppbarContent = ({
   color: titleColor,
-  subtitle,
-  subtitleStyle,
```
ruben-rebelo, line 106: Since we have removed the usage of subtitle and subtitleStyle we can remove it from the type.

src/components/BottomNavigation/BottomNavigationBar.tsx
```diff
@@ -321,7 +317,7 @@ const BottomNavigationBar = <Route extends BaseRoute>({
   ),
   getLabelText = ({ route }: { route: Route }) => route.title,
   getBadge = ({ route }: { route: Route }) => route.badge,
-  getColor = ({ route }: { route: Route }) => route.color,
+  getColor: _getColor = ({ route }: { route: Route }) => route.color,
```
ruben-rebelo: Since we are removing the usage of getColor please remove it from the Props<Route> or add a dev warning on first call so consumers migrating from 5.x know to drop it.

ruben-rebelo:
Thank you so much for you work @xdimota7x! 
I added really small comment just for small clean up of types that I could find.
Overall it looks really good and I'm happy to merge it!

xdimota7x:
Thank you for taking your time reviewing this large PR and for the suggestions @ruben-rebelo!
I addressed and cleaned up both suggestions in the latest commits, having removed `getColor` and `subtitle/subtitleStyle`.

PR #4899, opened 2026-04-15
feat: remove customRippleColor from components

### LINKED ISSUE #4894: refactor: remove customRippleColor prop
`customRippleColor` allows overriding the ripple effect color, but MD3 guidelines specify that ripple color should be derived from the content color, making this an API surface that works against spec. 

Removing it simplifies the API without meaningful loss of expressiveness.
Components that use it:
  - Button
  - Chip
  - Searchbar
  - IconButton
  - Menu.Item
  - SegmentedButton (item)
  - FAB
  - AnimatedFAB
  - List.Accordion
  - Drawer.Item

adrcotfas:
### Motivation

- We're modernizing Paper to support `PlatformColor` for Dynamic Themes without native modules which does not support applying an `alpha` to it, something required for the current implementation of ripple color.
- Dynamic Themes using `PlatformColor` and custom ripple color is not compatible.
- MD3 guidelines specify that ripple color is derived from the content color, not a custom color so the question "why do we need to support that" comes to mind.
- The current ripple implementation in components like `Button`, `IconButton`, `SearchBar`, `Chip` etc does not follow the MD3 guidelines.
- The native Jetpack Compose library does not support adding a custom ripple color without some special `CompositionLocalProvider` config so custom ripples are not encouraged.
- In MD3, the default ripple color is using a `0.1` alpha applied to a neutral color. Per specs, some components like `Button` need a `0.1` alpha applied to a primary/tertiary color. At this opacity it's hard to distinguish the neutral from the primary ripple and we don't have a choice when using `PlatformColor` anyway.
- I've proposed extending `android_ripple` to support `PlatformColor` and an alpha that will allow Paper to follow the MD3 ripple specs for all components: https://github.com/facebook/react-native/pull/56395.
- These are the reasons I'm proposing we drop the custom ripple color and make our life easier for the next steps.

#### Order of merging:
- #4898
- #4897
- #4899
- #4900
- #4901

### Related issue

- Closes #4894
- https://github.com/facebook/react-native/pull/56395

### Test plan

Run the example app and observe the ripple color for components; `Button` for example will have a slightly different tint.

### DIFF
--- a/src/components/Appbar/AppbarAction.tsx
+++ b/src/components/Appbar/AppbarAction.tsx
@@ -1,11 +1,5 @@
 import * as React from 'react';
-import type {
-  Animated,
-  ColorValue,
-  StyleProp,
-  View,
-  ViewStyle,
-} from 'react-native';
+import type { Animated, StyleProp, View, ViewStyle } from 'react-native';
 
 import type { ThemeProp } from 'src/types';
 
@@ -20,10 +14,6 @@ export type Props = React.ComponentPropsWithoutRef<typeof IconButton> & {
    *  Custom color for action icon.
    */
   color?: string;
-  /**
-   * Color of the ripple effect.
-   */
-  rippleColor?: ColorValue;
   /**
    * Name of the icon to show.
    */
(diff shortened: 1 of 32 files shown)


ruben-rebelo:
LGTM

PR #4900, opened 2026-04-15
fix(MD3): update color schema and tokens 

adrcotfas:
### Motivation
This PR updates the color schema and tokens for the MD3 themes. See:  
- https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/color/res/values/tokens.xml
- https://m3.material.io/styles/color/roles

#### Order of merging:
- #4898
- #4897
- #4899
- #4900
- #4901

### Test plan

Run the example app and observe the slightly different colors used for components, according to the latest MD3 specs.


thymikee (approved):

src/types.tsx
```diff
@@ -58,6 +65,18 @@ export type MD3Colors = {
...
+  onPrimaryFixed: string;
+  onPrimaryFixedVariant: string;
+  secondaryFixed: string;
+  secondaryFixedDim: string;
+  onSecondaryFixed: string;
+  onSecondaryFixedVariant: string;
+  tertiaryFixed: string;
+  tertiaryFixedDim: string;
+  onTertiaryFixed: string;
+  onTertiaryFixedVariant: string;
```
ruben-rebelo, line 79: The 19 new properties are required. Any consumer who built a custom MD3Colors literal pre-upgrade will now fail TS. Is this intended?

adrcotfas: Yes, it's a breaking change in that case but MD3 theme generators like https://material-foundation.github.io/material-theme-builder/ include these properties so it should not be a big deal to regenerate.  
Shipping a generateThemeFromSeed(hex, { variant? }) via `@material/material-color-utilities` removes the need for anyone to hand-author a palette and it's part of my plan on v6 branch. We're already using the `material-color-utilities` but only for the docs/website.

Issue #4878, opened 2026-03-12
Outlined TextInput label background visible through Modal backdrop, transparent background causes outline strikethrough


samdaw13:
### Current behaviour

Outlined `TextInput` renders correctly under normal conditions, the label background matches the page and everything looks fine. The issue only occurs when the TextInput is visible behind a semi-transparent overlay, such as a `Modal` backdrop.

When a `Modal` opens over a screen containing outlined TextInputs, the label's opaque background rectangle is visible through the darkened backdrop, since the label background color does not update to account for the overlay. This makes the label look like it was struck through. 

Attempting to fix this by setting the background to `'transparent'` does not work:
- Setting `backgroundColor: 'transparent'` in the `style` prop causes the label to fall back to `theme.colors.background` (see code reference below), so the opaque rectangle remains — the outline border also strikes through the label text
- Setting `theme.colors.background` to `'transparent'` removes the opaque rectangle, but the outline border still renders through the label text as a strikethrough, since there is no surface behind the label to mask the outline gap

### Expected behaviour

When the label background is set to `'transparent'`, the outline gap behind the label text should still be clipped/masked so the border does not strike through the label. This would allow the TextInput to look correct both normally and behind overlays.

Alternatively, the label background color could be independently configurable from the input background, so it can be changed without affecting the outline gap behavior.

### How to reproduce?

1. Render an outlined `TextInput` with a value (so the label is floating)
2. Open a `Modal` over it — the label background rectangle is visible through the semi-transparent backdrop
3. Try to fix it by setting `theme={{ colors: { background: 'transparent' } }}`
4. The opaque rectangle is gone, but the outline border now strikes through the label text

```tsx
<Modal visible={modalVisible} onDismiss={dismiss}>
  {/* modal content */}
</Modal>

<TextInput
  mode='outlined'
  label='Example'
  value='Some text'
  theme={{ colors: { background: 'transparent' } }}
/>
```

**Note:** Without the `transparent` override, the TextInput looks perfect on its own — this is specifically about how the label background interacts with overlays.

### Preview

**1. Modal closed — label looks correct:**

![Image](https://github.com/user-attachments/assets/238b9dc4-cc49-4ff0-85ed-2c2ba71618d6)

**2. Modal open — label background rectangle visible through backdrop:**

![Image](https://github.com/user-attachments/assets/14032523-fdf1-4acb-bab6-da14c2721e88)

[...]

Issue #4329, opened 2024-03-01
Label clips when  calling render function prop of TextInput component in outline mode 


jamesmogambi:
### Current behaviour

The label of TextInput component clips in outline mode when I call the render function prop. I am calling the render prop so as to render my Custom input, in this case a dropdown select input. Space is sufficient so it should not clip the label. Is the their a way to work around it ?

### Expected behaviour

Should not clip the label
### What have you tried so far?

I have tried applying different styles but still not working
### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x
| android                        | 14
| react-native                | 0.73.4
| react-native-paper    | 5.12.3
| node                            | 18.17.1
| npm                 | 9.6.7
| expo sdk                     | 50.0.7
![Screenshot_1709322988](https://github.com/callstack/react-native-paper/assets/31744209/b04c7461-7c44-41e9-b179-e0d8947becd9)

gedu:
Hey, do you have a snack, or can show some code? That way would be easier to replicate and see what it is going on. Thanks

ayloncarrijo:
Same here. It was working fine on version 5.10.6.

jamesmogambi:
> Same here. It was working fine on version 5.10.6.

Seems it breaks in later versions

pentolbakso:
still not working

<img width="375" alt="Screenshot 2024-09-23 at 00 48 01" src="https://github.com/user-attachments/assets/edf9d292-405b-4eb5-80ab-e338d1ecedc1">


```javascript
<TextInput
  label="Timezone"
  value="something value"
  mode="outlined"
  style={{marginTop: 10}}
  error={!!errors.timezone}
  onBlur={onBlur}
  render={() => <View />}
/>

dccarmo:
For what is worth, the issue is the `labelWidth` calculation here: https://github.com/callstack/react-native-paper/blob/main/src/components/TextInput/Label/InputLabel.tsx#L118

jamesmogambi:
Fine let me check it

bviala:
bump

on my end label is fine when at the top (focus || hasValue state) but cut off when inside the input (!focus && !hasValue state)

Issue #4235, opened 2023-12-18
Input component labels in outlined mode need some work


StefanWallin:
### Current behaviour

1. When background and primary are close to each other we are not able to style text-color of labels on input-components separately from primary color in theme. It's weird that primary is used both in backgrounds and texts without escape hatches.
2. When placing a **rounded** `<TextInput>` with `mode="outlined"` and `outlineStyle={{borderColor:'transparent'}}` on a elevated `<Surface>`, then the backgroundColor of the label gets a weird shape.

### Expected behaviour

I expect:
- The rounding of the input to hide the label background
- The label background to be styleable without creating a label element
- Each of the inputs parts to have to have escapeHatches for styling, such as these props, or a style function that takes it's state(focused, normal, disabled) `labelStyle`, `activeLabelStyle`, `inputStyle`, `activeInputStyle`, 

### How to reproduce?

https://snack.expo.dev/t8WLJwSQ6

### **Background is dark to show low contrast between background and primary text**
![IMG_2656](https://github.com/callstack/react-native-paper/assets/457653/25b938a3-03e3-4468-a8b6-a8d9c6035d76)
### **Background is set to red to highlight the weird shape of the label**
![IMG_2655](https://github.com/callstack/react-native-paper/assets/457653/9780026f-bb2e-45e4-b97d-7bcb872d1402)

### What have you tried so far?

Creating custom label components without getting it to work consistently across platforms (android & iOS)
### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | 17.1.2
| android                        | 13
| react-native                | 0.72.1
| react-native-paper    | 5.11.4
| node                            | 18.13.0
| npm or yarn                | 9.6.3
| expo sdk                     | any/not-applicable/47 in snack

ramakula:
@StefanWallin If we pass a Text component with a color as a label to a text input in outlined mode, it looks like this:

![Simulator Screenshot - iPhone 14 Pro Max - 2023-12-19 at 14 51 19](https://github.com/callstack/react-native-paper/assets/2745769/aa897aee-1c85-4f0a-a81d-f78d41494d2d)

<TextInput....
.....
label: <Sext style={{ color: '#333333' }}</Text>

HamidAliSE:
[...]

PR #4904, opened 2026-04-22
feat: TextField component

wonderlul:
### Motivation

**File structure**
The component is split by variant (`filled/ `and `outlined/`) and a root that wires shared behavior. Each area keeps logic, styles, utils, and constants in separate files. That follows patterns already used elsewhere in the library, but goes one step further so responsibilities stay obvious: variant-specific layout and theming do not drown in shared code, and the public API file stays focused on behavior and types.

**`LeftAccessory` / `RightAccessory` vs `TextInput` adornments**
`TextInput` composes leading and trailing content through `left` and `right`, which are built around icons and affixes (`TextInput.Icon`, `TextInput.Affix`) and internal adornment types. `TextField` instead exposes `LeftAccessory` and `RightAccessory` as render props (component types). The field passes curated layout and state—notably the merged style for alignment with the field, plus status, multiline, and editable—so accessories stay aligned with the input without re-implementing field internals. That supports arbitrary leading/trailing UI (clear actions, custom buttons, non-icon content) while still inheriting the important structural styles from the component.

**`filled` / `outlined` instead of `flat` / `outlined`**
Material Design 3 describes text fields in terms of `filled` and `outlined` styles. The existing TextInput API uses mode: `'flat' | 'outlined'`, where “flat” corresponds to the filled look. The new component names variants filled and outlined so the public API matches MD3 language and is easier to map from the spec and design tools.

**Style overrides**
TextField is built as a small stack of clear layers, and each layer can be adjusted without fighting the rest. The outer pressable wrapper, the field shell (border, background, row that includes accessories), and the inner content wrapper (label + TextInput) each accept dedicated style props (pressableStyle, fieldStyle, containerStyle). The underlying TextInput still uses the normal style prop (and the rest of TextInputProps) for typography, padding, and input-specific layout. Label and helper text can be customized through labelProps and helperProps (including style). Leading/trailing UI uses LeftAccessory / RightAccessory, which receive a prepared style from the field so custom content stays aligned while remaining fully under your control. Together, this gives predictable “override the part you mean” behavior instead of a single opaque style that’s hard to reason about.

**TextField instead of TextInput**
Material Design 3 uses the term text field for this control. Exporting a second “Paper” input named `TextInput` would also blur the built-in React Native `TextInput` in imports and documentation. A dedicated `TextField` name keeps the design-system component clearly namespaced and aligned with MD3, while the underlying control remains React Native’s `TextInput` where appropriate.

**Positioning relative to TextInput**
`TextField` is intended as the modern replacement path for form text entry in react-native-paper: implementation is structured for clarity and maintainability, and it adopts MD3-oriented theming (including use of the `PlatformColor API` where it fits platform tokens). Compared to the legacy `TextInput` stack, this design aims to be easier to follow, less ad hoc in how variants and layout are split, and more efficient in how styles and state are applied—giving teams a refreshed, spec-aligned building block for new work without forcing an immediate break for existing `TextInput` users.

### Order of merging
https://github.com/callstack/react-native-paper/pull/4901

### Related issue
https://github.com/callstack/react-native-paper/issues/4878
https://github.com/callstack/react-native-paper/issues/4329
https://github.com/callstack/react-native-paper/issues/4235

### Test plan

Run the Example app.

`filled`
text-field-filled

`outlined`
text-field-outlined


src/components/TextField/outlined/logic.ts
```diff
@@ -0,0 +1,184 @@
...
+  const $fieldStyles = [$fieldStyle, $fieldStyleOverride];
+
+  const $outlineStyles = [
+    $outlineStyle,
+    {
+      borderWidth: hasActiveOutline ? 2 : 1,
+      borderColor: hasActiveOutline ? activeOutlineColor : inactiveOutlineColor,
+    },
+    disabled && $disabledStyle,
+    $fieldStyleOverride,
```
adrcotfas, line 113: Is this intentional? It's already part of `fieldStyles` above.

src/components/TextField/styles.ts
```diff
@@ -0,0 +1,45 @@
...
+  paddingVertical: 0,
+  paddingHorizontal: 0,
+};
+
+export const $helperStyle: TextStyle = {
+  marginTop: HELPER_MARGIN_TOP,
+  paddingHorizontal: TEXT_FIELD_INPUT_WRAPPER_PADDING_HORIZONTAL,
+  fontSize: HELPER_FONT_SIZE,
+  fontWeight: '400',
+  textAlign: 'left',
```
adrcotfas, line 26: I think we should consider RTL instead of hardcoding `left`.

src/components/TextField/filled/constants.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  ACCESORY_SIZE,
```
adrcotfas: Typo for `ACCESSORY_SIZE` here and in styles.ts:, filled/constants.ts, outlined/constants.ts.

src/components/TextField/filled/logic.ts
```diff
@@ -0,0 +1,167 @@
+import { Animated, StyleProp, TextStyle, ViewStyle } from 'react-native';
```
adrcotfas: The filled variant does not handle RTL as the outlined variant does. Also, some RTL tests would be useful.

src/components/TextField/outlined/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+}: {
+  theme: InternalTheme;
+  disabled: boolean;
+  hasError: boolean;
+}) => {
+  const { colors } = theme;
+
+  let inactiveOutlineColor = colors.outline;
+
+  if (disabled) {
```
adrcotfas: According to [specs](https://m3.material.io/components/text-fields/specs):  
```
Outlined text field disabled outline color -> md.sys.color.on-surface
Outlined text field disabled outline opacity -> 0.12
```

src/components/TextField/filled/logic.ts
```diff
@@ -0,0 +1,167 @@
...
+    $animatedLabelTextStyle,
+    labelProps?.style,
+  ];
+
+  const $fieldStyles = [
+    $fieldStyle,
+    {
+      backgroundColor: disabled ? surfaceContainerHighest : surfaceVariant,
+    },
+    disabled && $disabledStyle,
```
adrcotfas: The outlined variant applies `disabled && $disabledStyle` per element (input, label, accessories). The filled variant applies it only to the field parent.  

`$animatedLabelTextStyles` has no disabled dimming.

According to [specs](https://m3.material.io/components/text-fields/specs):

```
Filled text field disabled container opacity -> 0.04
```

example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,210 @@
...
+          helper="Enter a valid email address."
+          placeholder="name@example.com"
+          status="error"
+          value={errorField}
+          onChangeText={setErrorField}
+          keyboardType="email-address"
+          autoCapitalize="none"
+          pressableStyle={styles.field}
+        />
+        <TextField
```
adrcotfas, line 166: No outline is visible here.

src/components/TextField/styles.ts
```diff
@@ -0,0 +1,45 @@
+import { StyleProp, TextStyle, ViewStyle } from 'react-native';
+
+import {
+  ACCESORY_SIZE,
+  HELPER_FONT_SIZE,
```
adrcotfas, line 5: For consistency with the specs, I would rename all occurrences of `helper` to `supportingText` and `accessory` to `icon`.

adrcotfas (commented):
Below are some components described in the specs I see were not covered. Do we want them?  

- in all error examples from the specs there's a trailing error icon
- the option to have leading and trailing text (called prefix and suffix) for currency, units of measurement etc; for suffix, the text field is aligned to the end instead of start;
- character counter (aligned to the end, same style as supporting text)
- The native filled `TextField` has an animation for the outline when selecting it: center to edges reveal 
- see https://m3.material.io/components/text-fields/guidelines

Some other notes:
- Please double check the tokens from the specs. I added some comments about opacity/color for certain elements that was missing; maybe there are other inconsistencies.
- A more complex example would be nice to cover more use cases and these elements, if we add them.

Here's a recording from this app showing the features mentioned above: https://github.com/material-components/material-components-android. I recommend installing this app to double check how `TextField` looks and behaves like in native. 

https://github.com/user-attachments/assets/5211c39c-b333-46e3-b752-5620b94e34db

src/components/TextField/filled/logic.ts
```diff
@@ -0,0 +1,167 @@
+import { Animated, StyleProp, TextStyle, ViewStyle } from 'react-native';
+
```
adrcotfas, line 8: In the specs placeholders are hidden until the TextField is selected.

src/components/TextField/outlined/logic.ts
wonderlul: Yes, this is intentional. `outlineStyles` are kind of additional outline to handle the focus state that has to be overridden as well.

src/components/TextField/styles.ts
wonderlul: Considering RTL using `left` is actually handling both cases depending on the settings. Do you have something else in mind?

adrcotfas: By considering RTL you mean `outlined/logic.ts` `writingDirection`? Doesn't the text still anchor to the left side regardless of `writingDirection`?  
Shouldn't this be `auto` instead of `left` or just remove it?

wonderlul: With `auto` it aligns text to the left and doesn't respect the RTL settings. I guess with `textAlign: left` it considers `left` as the start no matter if it's left or right :-). Don't even ask me how long it took me to figure that out.

src/components/TextField/styles.ts
wonderlul: MD3 describes the common case as leading/trailing icons, but the slots are intentionally generic: users may pass text, badges, custom controls, or non-icon content. Naming those props *Icon would misdescribe valid usage and nudge people toward icons only. “Accessory”-style naming is a deliberate pattern for flexible slots (similar to how many form libraries avoid over-specific names).

Agreed on supportingText — that matches the MD3 wording for the area under the field and keeps the API aligned with the spec without changing behavior. I’ll rename helper / related names to supportingText (and adjust docs/tests accordingly). Thanks for the suggestion.

adrcotfas: Sure, that makes sense but do we want to encourage users to add customization beyond what the guidelines describe like icons(icon buttons)?  
The only place I found a non-icon accessory described was for the credit card example here, but we could argue that icon/image are equivalent: https://m3.material.io/components/text-fields/guidelines. 

> An image can help contextualize the required input text such as a credit card number.

For text there are different props we don't have yet in `TextField`: `prefixText` and `suffixText` which should be aligned with the actual text inside the text field, not centered vertically like the current accessories. See screen-20260423-165915-1776952725646.mp4.

In my opinion users of the library would appreciate a react-native-paper API as close as possible in naming with the MD3 specs (`leadingIcon`, `trailingIcon`) making it easier to follow the specs.  
What do you think?

src/components/TextField/filled/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+// PLACEHOLDER & MULTILINE POSITIONING
+// ==================
+
+export const PADDING_TOP = ACTIVE_LABEL_FONT_SIZE + TEXT_FIELD_PADDING_VERTICAL;
+
+// ==================
+// OPACITY
+// ==================
+
+export const FILLED_CONTAINER_ALPHA = 0.04;
```
adrcotfas, line 33: Let's go with `DISABLED_CONTAINER_OPACITY` so we align with the specs token `filled-text-field.disabled.container.opacity`.

src/components/TextField/outlined/constants.ts
```diff
@@ -0,0 +1,55 @@
...
+// ==================
+
+export const PLACEHOLDER_TOP_POSITION =
+  TEXT_FIELD_PADDING_VERTICAL + LINE_HEIGHT_DELTA;
+
+// ============
+// OPACITY
+// ============
+
+export const OUTLINE_ALPHA = 0.12;
```
adrcotfas, line 55: Let's go with `DISABLED_OUTLINE_OPACITY` to align with the specs `outlined-text-field.disabled.outline.opacity`

src/components/TextField/filled/styles.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const $labelTextStyle: TextStyle = {
+  fontWeight: '400',
+  includeFontPadding: false,
+  paddingVertical: 0,
+  paddingHorizontal: 0,
+};
+
+/* Disabled tint (`onSurface @ 0.04`) is rendered as a childless overlay so its
```
adrcotfas, line 43: This comment is duplicated in `logic.ts` and contains a magic number that might change. Perhaps keep the comment only here and mention the constant?

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,293 @@
...
+  disabled: boolean;
+  hasAccessory: boolean;
+  hasError: boolean;
+  $animatedLabelWrapperStyle: Animated.WithAnimatedObject<ViewStyle>;
+  $animatedLabelTextStyle: Animated.WithAnimatedObject<TextStyle>;
+  $animatedPlaceholderStyle: Animated.WithAnimatedObject<TextStyle>;
+  $animatedActiveOutlineStyle?: Animated.WithAnimatedObject<ViewStyle>;
+};
+
+export interface TextFieldProps extends TextInputProps {
```
adrcotfas, line 40: It would be nice to support `prefixText` and `suffixText` aligned to the actual text field like in the specs instead of `accessory`/`icon` which is centered vertically. See screen-20260423-165915-1776952725646.mp4.  For `suffixText` there should be a way to align the text in the text field at the end - something visible in the video example.

They use the same text style as the text in the textfield with `onSurfaceVariant` color for all states (`isError`, `enabled`, `focused`, `unfocused`) except for `disabled` where it's `onSurfaceVariant` at 0.38 opacity.

src/components/TextField/styles.ts
wonderlul: I don’t have a strong opinion on this one — I think it’s really a product direction call: whether the public API should track MD3 as closely as possible (easier mental mapping from the spec) or stay a bit more generic so non–MD3 patterns stay natural in React Native.

Splitting things into leading/trailing icon plus prefix/suffix text (or similar) would line up better with the spec and can feel more “complete” but it also multiplies props and concepts for both authors and consumers, and we’d need clear rules for composition and order when several adornments exist. So it’s a tradeoff between spec fidelity and API surface / noise, not a purely technical right-or-wrong.

I’m fine either way the team wants to go; happy to adjust the implementation once there’s a preferred direction.

src/components/TextField/filled/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+// ==================
+// LABEL POSITIONING
+// ==================
+
+export const LABEL_LEFT_OFFSET_WITH_ACCESSORY =
+  ACCESSORY_SIZE +
+  TEXT_FIELD_ACCESSORY_MARGIN_HORIZONTAL +
+  TEXT_FIELD_INPUT_WRAPPER_PADDING_HORIZONTAL;
+
+export const LABEL_LEFT_OFFSET_WITHOUT_ACCESSORY =
```
adrcotfas, line 18: It would be nice to refactor all constants containing `left`/`right` to `start`/`end` to comply with RTL and MD naming.

src/components/TextField/styles.ts
wonderlul: I’m going to consider an alternative that accounts for all the possibilities and attempts to blend an MD3-first design with flexibility. I'll let you know soon.

src/components/TextField/TextField.tsx
adrcotfas, line 40: It would be nice to support the `counterEnabled` + `counterMaxLength` which lives together with `supportingText` aligned to the end.  
BTW, counter is different from suffix but could serve the same purpose.

In the MD examples I see `supportingText` together with the counter in the same TextField. Exceeding the limit would not modify the `supportingText` but apply the error style to the counter:

https://github.com/user-attachments/assets/eab22ff0-9310-416a-9e86-98492341475a

example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,213 @@
+import * as React from 'react';
```
adrcotfas, line 1: It would be nice to:
- improve the examples containing a trailing icon button to show a ripple when pressing it: now we don't have a visual feedback like in the `TextInput` examples
- add other examples with counter, suffix/prefix text 
- add any other examples you might find valuable equivalent to what we showcase in `TextInput` since we're replacing that component with `TextField`

adrcotfas (changes requested):

PR #4906, opened 2026-04-28
Fix react native 0.85 absolutefillobject

fracico-tech:
### Motivation

This PR fixes a layout regression where Dialogs, Modals, and Portals appear at the bottom of the screen instead of being centered or properly positioned.

The issue was introduced with React Native 0.85, which removed StyleSheet.absoluteFillObject. Since React Native Paper uses the spread operator (...StyleSheet.absoluteFillObject) in several components, these styles were returning undefined, causing absolute positioning to fail.

This change introduces a robust fallback to ensure the library remains compatible with both React Native 0.85+ and older versions.

### Related issue
Fixes an issue where UI components using Portal or Modal lose their fullscreen overlay and centering on RN 0.85.

### Test plan
Environment: React Native 0.85.0, Android/iOS.

Steps:

Open a Dialog or Modal component.

Verify the component is centered on the screen and the underlay covers the entire background.

Backward Compatibility:

Verified that the manual object fallback { position: 'absolute', ... } behaves identically to the previous absoluteFillObject on older versions of React Native.

Before:

Dialogs stuck at the bottom of the screen.

Underlay/Overlay has 0 height/width.

After:

Dialogs correctly centered.

Underlay correctly covers the full screen.

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -236,7 +236,7 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   wrapper: {
-    ...(StyleSheet.absoluteFill),
+    ...StyleSheet.absoluteFill,
     justifyContent: 'center',
   },
   // eslint-disable-next-line react-native/no-color-literals
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -236,13 +236,7 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   wrapper: {
-    ...(StyleSheet.absoluteFillObject || {
-      position: 'absolute',
-      left: 0,
-      right: 0,
-      top: 0,
-      bottom: 0,
-    }),
+    ...(StyleSheet.absoluteFill),
     justifyContent: 'center',
   },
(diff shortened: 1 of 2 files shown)


OkancanCosar:
hey, this pr need to merge for rn 0.85.x please consider to merge urgently

satya164 (changes requested):
Thanks for the PR. However it doesn't need condition like this. Replacing `StyleSheet.absoluteFillObject` with `StyleSheet.absoluteFill` is safe and enough.

fracico-tech:
I can't simply swap it for ...StyleSheet.absoluteFill here. Since absoluteFill can return a style ID (number) rather than an object, spreading it (...) would result in an empty object and break the positioning.
I think mine is the safer approach

satya164:
> I can't simply swap it for ...StyleSheet.absoluteFill here. Since absoluteFill can return a style ID

It hasn't returned a number since React Native 0.60. It's essentially the same as `absoluteFillObject` https://github.com/facebook/react-native/commit/a8e3c7f5780516eb0297830632862484ad032c10

satya164:
Thank you

PR #4909, opened 2026-04-30
feat: TextInput component

### LINKED ISSUE #4878: Outlined TextInput label background visible through Modal backdrop, transparent background causes outline strikethrough

wonderlul:
BREAKING CHANGE: `TextInput` has been rewritten for Material Design 3. The public API, adornments, variant naming, and styling overrides have changed. See the migration section below for required updates.

## Summary

Replaces the Paper 5.x `TextInput` with a new MD3-aligned implementation. The component name stays **`TextInput`**, but most props and composition patterns have changed.

The new `TextInput` supports **`filled`** and **`outlined`** variants, floating labels, **`supportingText`**, a built-in character **`counter`**, **`prefix`** / **`suffix`**, leading/trailing **`startAccessory`** / **`endAccessory`** render props, a custom **`render`** prop for masked inputs, and improved accessibility. Colors and layout follow the theme; per-prop color overrides for outlines and underlines were removed.

## What changed

| Before (v5) | After (v6) |
| --- | --- |
| `mode="flat"` | `variant="filled"` (default) |
| `mode="outlined"` | `variant="outlined"` |
| `left` / `right` element props | `startAccessory` / `endAccessory` render props |
| `TextInput.Affix` | `prefix` / `suffix` strings, or custom accessories |
| `HelperText` below the field | `supportingText` on the field |
| `label` as `string` or `ReactNode` | `label` as `string` only |
| `dense`, `contentStyle`, `outlineStyle`, `underlineStyle` | removed — use `style` and `theme` |
| `textColor`, `underlineColor`, `activeUnderlineColor`, `outlineColor`, `activeOutlineColor` | removed — use `style` and `theme` |
| `disabled` blocked interaction | `disabled` blocks interaction; use `editable={false}` for read-only fields that can still be focused |

### Still supported

- **`TextInput.Icon`** — use inside `startAccessory` / `endAccessory`
- **`render`** — swap the native input (e.g. masked inputs)
- **`selectionColor`** and **`cursorColor`** — inherited from React Native `TextInput` props
- **`ref`** — exposes `TextInputHandles` (`focus`, `blur`, `clear`, `isFocused`, `setNativeProps`, `setSelection`); `clear()` also syncs internal state and label animation

## What was removed

- **`TextInput.Affix`**
- **`HelperText`** as the recommended companion for text fields (use `supportingText` instead)
- **`mode`**, **`left`**, **`right`**
- **`dense`**, **`contentStyle`**, **`outlineStyle`**, **`underlineStyle`**
- **`textColor`**, **`underlineColor`**, **`activeUnderlineColor`**, **`outlineColor`**, **`activeOutlineColor`**

## Migration

Import stays the same:

```tsx
import { TextInput, type TextInputProps } from 'react-native-paper';
```

### Variant

```tsx
// Before (v5)
<TextInput mode="flat" label="Email" />
<TextInput mode="outlined" label="Password" />

// After (v6)
<TextInput variant="filled" label="Email" />
<TextInput variant="outlined" label="Password" />
```

### Icons and adornments

`left` / `right` element props become **`startAccessory`** / **`endAccessory`** render functions. Spread the props TextInput passes in — they include layout `style`, plus `error`, `disabled`, and `multiline`.

```tsx
// Before (v5)
<TextInput
  label="Search"
  value={text}
  onChangeText={setText}
  left={<TextInput.Icon icon="magnify" />}
  right={<TextInput.Affix text={`${text.length}/80`} />}
/>

// After (v6)
<TextInput
  label="Search"
  value={text}
  onChangeText={setText}
  startAccessory={(props) => <TextInput.Icon {...props} icon="magnify" />}
  endAccessory={(props) => (
    <TextInput.Icon {...props} icon="close" onPress={() => setText('')} />
  )}
/>
```

Replace **`TextInput.Affix`** with **`prefix`** / **`suffix`** for inline text, or a custom **`endAccessory`**:

```tsx
// Before (v5) — affix as trailing adornment
<TextInput
  right={<TextInput.Affix text="/100" />}
  maxLength={100}
/>

// After (v6) — prefix/suffix and/or counter
<TextInput
  prefix="$"
  suffix="/100"
  maxLength={100}
  counter
/>
```

> **Note:** `prefix` and `suffix` are shown only when the label is floated (field focused or has a value).

### Supporting text and errors

```tsx
// Before (v5)
<>
  <TextInput label="Email" error={hasError} />
  <HelperText type="error" visible={hasError}>
    Enter a valid email
  </HelperText>
</>

// After (v6)
<TextInput
  label="Email"
  error={hasError}
  supportingText={hasError ? 'Enter a valid email' : 'We will never share your email'}
/>
```

When `error` is `true` and no `endAccessory` is provided, a default error icon is shown on the trailing side.

### Disabled vs read-only

```tsx
// Disabled — not editable, not focusable, disabled styling
<TextInput disabled label="Email" value={email} />

// Read-only — can focus and select text, but not edit
<TextInput editable={false} label="Email" value={email} />
```

### Styling and colors

Outline, underline, label, and disabled colors now come from the **theme**. Override input text with the standard **`style`** prop; override accent colors via **`theme`** or, where applicable, inherited props:

```tsx
// Before (v5)
<TextInput
  dense
  contentStyle={{ paddingHorizontal: 12 }}
  outlineStyle={{ borderRadius: 12 }}
  outlineColor="#79747E"
  activeOutlineColor="#6750A4"
  textColor="#1C1B1F"
  style={{ fontSize: 16 }}
/>

// After (v6)
<TextInput
  theme={{
    ...MD3LightTheme,
    colors: {
      ...MD3LightTheme.colors,
      outline: '#79747E',
      primary: '#6750A4',
    },
  }}
  style={{ fontSize: 16, color: '#1C1B1F' }}
  selectionColor="#6750A4"
  cursorColor="#6750A4"
/>
```

### Custom input (`render` prop)

```tsx
<TextInput
  label="Phone"
  render={(props) => <TextInputMask {...props} mask="+[00] [000] [000] [000]" />}
/>
```

### Imperative handle

```tsx
const ref = React.useRef<TextInputHandles>(null);

<TextInput ref={ref} label="Name" defaultValue="Jane" onChangeText={setName} />;

ref.current?.clear(); // clears text, syncs counter/label animation, calls onChangeText('')
```

## New props reference

| Prop | Description |
| --- | --- |
| `variant` | `'filled'` (default) or `'outlined'` |
| `supportingText` | Helper/error text below the field |
| `counter` | Shows `currentLength/maxLength`; requires `maxLength` |
| `prefix` / `suffix` | Short inline strings inside the field |
| `startAccessory` / `endAccessory` | Render props for leading/trailing content |
| `disabled` | Disables the field (separate from `editable={false}`) |
| `render` | Custom native input renderer |

## Related

- Depends on: https://github.com/callstack/react-native-paper/pull/4912
- Closes: https://github.com/callstack/react-native-paper/issues/4878, https://github.com/callstack/react-native-paper/issues/4329, https://github.com/callstack/react-native-paper/issues/4235
- Migration guide: `docs/docs/guides/12-migration.md`

## Test plan

Run the Example app and verify:

- [ ] **`filled`** variant — label animation, underline indicator, disabled overlay, error state
- [ ] **`outlined`** variant — outline, label notch behavior, error state
- [ ] **`startAccessory`** / **`endAccessory`** with `TextInput.Icon` and custom content
- [ ] **`prefix`** / **`suffix`** visibility when focused vs blurred
- [ ] **`counter`** with `maxLength`
- [ ] **`supportingText`** with and without `error`
- [ ] **`disabled`** vs **`editable={false}`**
- [ ] **`render`** prop with a custom input
- [ ] **`ref.clear()`** on controlled and uncontrolled fields
- [ ] RTL layout (accessories swap sides)
- [ ] Multiline fields

**Screenshots**

`filled`

text-field-filled

`outlined`

text-field-outlined


example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,300 @@
...
+        />
+        <TextField
+          variant="filled"
+          label="TextField.Icon"
+          value={filledIconQuery}
+          onChangeText={setFilledIconQuery}
+          StartAccessory={(props: TextFieldAccessoryProps) => (
+            <TextField.Icon {...props} icon="magnify" color={iconMuted} />
+          )}
+          EndAccessory={(props: TextFieldAccessoryProps) => (
```
adrcotfas: - Please move this `EndAccessory` to the `TextField` above and remove this one.  
The two are identical except for the missing ripple from the first. I think it does not make sense to exemplify a button without feedback.  
- Same for the first two `TextField`s from "Outlined" below.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+import { useTextField } from './logic';
+import type { InternalTheme, ThemeProp } from '../../types';
+
+export type TextFieldVariant = 'filled' | 'outlined';
+
+export interface TextFieldAccessoryProps {
+  style: StyleProp<ViewStyle>;
+  status?: 'error' | 'disabled';
+  multiline: boolean;
```
adrcotfas: Is `multiline` needed here?

src/components/TextField/TextFieldIcon.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const iconSize = size ?? ACCESSORY_SIZE;
+
+  const iconColor = getIconColor({
+    theme,
+    color,
+    status,
+  });
+
+  const onPressHandler = editable ? onPress : undefined;
```
adrcotfas: The press guard works fine visually since `TouchableRipple` already handles missing `onPress`. The problem is accessibility because `IconButton` sets `accessibilityState` from its own `disabled` prop, which is never passed, so screen readers won't know it's disabled. Should pass `disabled={!editable}` to `IconButton`.

src/components/TextField/filled/logic.ts
```diff
@@ -0,0 +1,267 @@
...
+    },
+    disabled && $disabledStyle,
+    $inputStyleOverride,
+  ];
+
+  const $prefixStyles: StyleProp<TextStyle> = [
+    $inputStyle,
+    {
+      fontSize: INPUT_FONT_SIZE,
+      color: onSurface,
```
adrcotfas: Looking at the native implementation, it should be `color: onSurfaceVariant`.
Same for suffixStyles, same for outlined/logic.

src/components/TextField/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  if (status === 'error') {
+    return error;
+  }
+  return onSurfaceVariant;
+};
+
+/**
+ * Returns the solid background color for the filled field container, or
+ * `undefined` when disabled. The disabled tint (`onSurface @ 0.04`) is rendered
```
adrcotfas: Please don't use magic numbers because `DISABLED_CONTAINER_OPACITY` might change.

example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,300 @@
...
+  TextField,
+  type TextFieldAccessoryProps,
+} from 'react-native-paper';
+
+import { useExampleTheme } from '../hooks/useExampleTheme';
+import ScreenWrapper from '../ScreenWrapper';
+
+const TextFieldExample = () => {
+  const { colors } = useExampleTheme();
+  const iconMuted = colors.onSurfaceVariant;
```
adrcotfas: This looks obsolete. I think we should leave it to `getIconColor` to do its job and not override it with `iconMuted` because it affects the color according to states.

example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,300 @@
...
+        />
+        <TextField
+          variant="filled"
+          label="Without accessories"
+          value={email}
+          onChangeText={setEmail}
+          keyboardType="email-address"
+          autoCapitalize="none"
+          autoCorrect={false}
+          pressableStyle={styles.field}
```
adrcotfas: `pressableStyle={styles.field}` is used everywhere but it's empty.

example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,300 @@
+import * as React from 'react';
```
adrcotfas: I used this example to check against the specs and native implementation but had to extend it locally to check all states.  

It would be nice to have an example for disabled fields that contain accessories and suffix/prefix or even some buttons that toggles states for a more compact example page.

example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,300 @@
...
+  const [outlinedText, setOutlinedText] = React.useState('');
+  const [outlinedPassword, setOutlinedPassword] = React.useState('');
+  const [outlinedNotes, setOutlinedNotes] = React.useState('');
+  const [outlinedPrefix, setOutlinedPrefix] = React.useState('');
+  const [outlinedSuffix, setOutlinedSuffix] = React.useState('');
+  const [errorField, setErrorField] = React.useState('invalid@');
+  const [filledIconQuery, setFilledIconQuery] = React.useState('');
+  const [outlinedIconQuery, setOutlinedIconQuery] = React.useState('');
+
+  const ClearFilledSearchAccessory = ({
```
adrcotfas: `ClearFilledSearchAccessory`, `ClearOutlinedSearchAccessory`, `SearchLeadingAccessory`  look obsolete  - we already have `IconButton/TouchableRipple` built in `TextField.Icon`.

src/components/TextField/filled/logic.ts
```diff
@@ -0,0 +1,267 @@
...
+    $supportingTextStyle,
+    {
+      color: supportingTextColor,
+      writingDirection: isRTL ? 'rtl' : 'ltr',
+    },
+    disabled && $disabledStyle,
+    supportingTextProps?.style,
+  ];
+
+  const $inputStyles: StyleProp<TextStyle> = [
```
adrcotfas: At least on Android there's a visual bug with the opacity of the input text.  
When disabled, the opacity on the input text appears more than it should be (near invisible).  
The fix I found was to extract a `$inputWrapperStyles` and move the `{ flex: 1 },
    disabled && $disabledStyle` to it then place the RN `TextInput` from TextField.tsx inside of a `<View style={$inputWrapperStyles}>`.

src/components/TextField/filled/logic.ts
```diff
@@ -0,0 +1,267 @@
+import {
```
adrcotfas: I checked the native implementation:
-  prefix should have an end padding of 2
- suffix should have a start padding of 2

src/components/TextField/outlined/logic.ts
```diff
@@ -0,0 +1,233 @@
+import {
```
adrcotfas: Multiline issues:

- The padding breaks when adding a new line to an outlined TextField.  
I see `MULTILINE_PADDING_TOP` is not used here.

- There's a visual artifact/clipping/redraw when adding new lines for both filled and outlined: if this is hard to fix, we can ignore for now

adrcotfas (changes requested):

example/src/Examples/TextFieldExample.tsx
wonderlul: This was intentional. The example is meant to show that accessories are not limited to our built-in helpers (for example TextField.Icon): developers can wire in their own components, including ones with intentionally different interaction patterns — here, a control that doesn’t use a ripple to contrast with the standard affordance.

src/components/TextField/TextField.tsx
wonderlul: It’s hard to say upfront whether every consumer will need multiline, but exposing it doesn’t really cost us anything and keeps the API honest: accessories are rendered in a context that does depend on whether the field is single- or multi-line. If someone wants to override default accessory layout or behavior (alignment, padding, placement), having multiline on the accessory props avoids workarounds and guessing from inside a custom StartAccessory / EndAccessory.

src/components/TextField/filled/logic.ts
wonderlul: Per M3 tokens (InputPrefixColor / InputSuffixColor = OnSurfaceVariant; InputColor = OnSurface), updated prefix/suffix only—left input text on onSurface.

example/src/Examples/TextFieldExample.tsx
wonderlul: This was intentional. The example is meant to show that accessories are not limited to our built-in helpers (for example TextField.Icon): developers can wire in their own components, including ones with intentionally different interaction patterns — here, a control that doesn’t use a ripple to contrast with the standard affordance.

adrcotfas: In that case why not put some examples of different accessories, not something that is confused with icon buttons, because they are clickable, and should provide a visual feedback; IMHO, keeping extra elements here just for the sake of it confuses users that look at the examples.

What I'm trying to say: why showcase (and implement in the first place) things that you cannot find in the native specs/ guidelines?

src/components/TextField/filled/logic.ts
adrcotfas: Sure, my bad for pointing the wrong line number.

src/components/TextField/TextField.tsx
adrcotfas: I'm extremely curious about what use case you have in mind that needs multiline for accessories. YAGNI.

example/src/Examples/TextFieldExample.tsx
adrcotfas: In that case why not put some examples of different accessories, not something that is confused with icon buttons, because they are clickable, and should provide a visual feedback; IMHO, keeping extra elements here just for the sake of it confuses users that look at the examples.

What I'm trying to say: why showcase (and implement in the first place) things that you cannot find in the native specs/ guidelines?

src/components/TextField/TextField.tsx
wonderlul: I'd say accessories sometimes need different layout when the field is multiline (e.g. top-align an icon with the first line instead of vertically centering it in a tall input, or move an end action to the bottom).

adrcotfas:
Here's a comment from my friend Claude about duplicate code:

## `logic.ts` -- highest duplication (~28%)

  The most significant overlap. Both files share these sections **verbatim**:                                                                                                                                                                                                              
   
  | Block | Lines each |                                                                                                                                                                                                                                                                   
  |---|---|                                              
  | Props destructuring | ~13 |                                                                                                                                                                                                                                                            
  | `isRTL` extraction | 1 |                                                                                                                                                                                                                                                               
  | `getLabelColor(...)` call | ~5 |
  | `getSupportingTextColor(...)` call | ~5 |                                                                                                                                                                                                                                              
  | `$animatedLabelTextStyles` array | ~9 |              
  | `$containerStyles` array | ~4 |                                                                                                                                                                                                                                                        
  | `$supportingTextStyles` array | ~8 |                                                                                                                                                                                                                                                   
  | `$counterStyles` array | ~8 |                                                                                                                                                                                                                                                          
  | `$prefixStyles` array | ~6 |                                                                                                                                                                                                                                                           
  | `$suffixStyles` array | ~6 |                         
  | `$leadingAccessoryStyles` | ~3 |                                                                                                                                                                                                                                                       
  | `$trailingAccessoryStyles` | ~3 |
  | **Total** | **~71 identical lines** |                                                                                                                                                                                                                                                  
                                                         
  Outlined has 233 lines, filled has 280 -- so **~71/233 (30%) and ~71/280 (25%)** respectively are shared logic.                                                                                                                                                                          
   
  What's genuinely variant: the `$fieldStyles` (filled adds `backgroundColor`), `$outlineStyles` (border vs bottom-bar), `$animatedActiveOutlineStyles` (filled-only), `$disabledBackgroundStyles` (undefined in outlined, overlay in filled), and the `labelBackgroundColor` extraction   
  (outlined only).                                       
                                                                                                                                                                                                                                                                                           
  ---                                                    

  ## `styles.ts` -- low duplication (~10%)

  `$labelTextStyle` is **byte-for-byte identical** in both files (5 lines). It belongs in the shared `../styles.ts`.

  Everything else differs meaningfully:                                                                                                                                                                                                                                                    
  - `$fieldStyle`: same keys, `borderRadius` (outlined) vs `borderTopStart/EndRadius` (filled)
  - `$outlineStyle`: full-perimeter absolute (outlined) vs bottom-only absolute (filled)                                                                                                                                                                                                   
  - `$containerStyle`: `alignItems: 'center'` (outlined) vs `'flex-end'` (filled)                                                                                                                                                                                                          
  - `$labelWrapperStyle`: outlined adds `paddingHorizontal`, filled is empty
  - `$disabledBackgroundStyle`: filled-only                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                           
  ---
                                                                                                                                                                                                                                                                                           
  ## `constants.ts` -- minimal duplication (~5%)         

  `LABEL_START_OFFSET_WITHOUT_ACCESSORY` resolves to `TEXT_FIELD_INPUT_WRAPPER_PADDING_HORIZONTAL` in both -- could live in shared `../constants.ts`.                                                                                                                                      
   
  Everything else is variant: `LABEL_PADDING_HORIZONTAL` and the RTL translate constants are outlined-only; `MULTILINE_PADDING_TOP` is filled-only; `ACTIVE_LABEL_TOP_POSITION` uses different formulas; the opacity values differ (0.12 vs 0.04).                                         
                                                         
  ---                                                                                                                                                                                                                                                                                      
                                                         
  ## `utils.ts` -- 0% unifiable

  Both export `getOutlineColor` but with incompatible signatures: outlined takes `hasError: boolean`, filled takes `status?: 'error' | 'disabled'`. Same name, different contract -- unifying them would require a signature change that ripples into both `logic.ts` files.               
   
  ---                                                                                                                                                                                                                                                                                      
                                                         
  ## Overall

  | File | Outlined lines | Filled lines | Duplicated | % |
  |---|---|---|---|---|
  | `logic.ts` | 233 | 280 | ~71 | ~28% |
  | `styles.ts` | 45 | 54 | ~5 | ~10% |                                                                                                                                                                                                                                                    
  | `constants.ts` | 49 | 35 | ~2 | ~5% |
  | `utils.ts` | 39 | 34 | 0 | 0% |                                                                                                                                                                                                                                                        
  | **Total** | **366** | **403** | **~78** | **~20%** |

src/components/TextField/outlined/logic.ts
wonderlul: outlined-multiline

https://m3.material.io/components/text-fields/specs


- As per screenshot from documentation, top padding for multi-line variant is visibly smaller than for a regular one. I assume it's native behaviour. 
- I'll look into it

adrcotfas: Too bad they don't have consistency between these examples and implementation. I checked the Jetpack Compose and MDC native implementations and the padding remains the same for multi-line.

wonderlul: OK, done it.

example/src/Examples/TextFieldExample.tsx
adrcotfas: The new examples are awesome! Good job!

example/src/Examples/TextFieldExample.tsx
adrcotfas: Love the new example!

adrcotfas:
Some more issues I found:
- Outlined + Multiline in combination with any leading icon/ trailing icon/ error/suffix/ prefix has a bug; give it a try as now it's much easier to manually test combinations
- We should not allow prefix + suffix in the same field 

Edit: the filled version lost the top rounded corners when in error mode and after toggling the "disabled" state.

src/components/TextInput/TextInputIcon.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
```
adrcotfas, line 1: There's a better way to get the native error icon in the correct size as this one is larger than expected.

Remove `export const ERROR_ICON_SIZE = 16;`

```
import React from 'react';
import { StyleProp, View, ViewStyle } from 'react-native';

import { ACCESSORY_SIZE } from './constants';
import { useInternalTheme } from '../../core/theming';
import type { ThemeProp } from '../../types';
import Icon from '../Icon';

interface TextFieldErrorIconProps {
  style?: StyleProp<ViewStyle>;
  theme?: ThemeProp;
}

const TextFieldErrorIcon = ({
  style: $wrapperStyle,
  theme: themeOverride,
}: TextFieldErrorIconProps) => {
  const theme = useInternalTheme(themeOverride);

  return (
    <View style={$wrapperStyle}>
      <Icon
        source="alert-circle"
        size={ACCESSORY_SIZE}
        color={theme.colors.error}
      />
    </View>
  );
};

export default TextFieldErrorIcon;
```

adrcotfas: Also, I think we should have this error icon clickable as on native.

adrcotfas:
When the helper text is empty, the counter gets aligned to start. It should stay aligned to end.

adrcotfas:
Error seems to take precedence over Disabled but I would expect that a disabled field is not editable even if it has "error".

wonderlul:
> Some more issues I found:
> 
> * ~The filled version lost the top rounded corners~ Or not - it was a weird visual bug occuring once
> * Outlined + Multiline in combination with any leading icon/ trailing icon/ error/suffix/ prefix has a bug; give it a try as now it's much easier to manually test combinations
> * We should not allow prefix + suffix in the same field

Multi-line fixed.

I disagree with disallowing prefix + suffix. I think it's up to the developer to decide whether to use both, one or neither, but we shouldn't disable the option for developers.

wonderlul:
> Error seems to take precedence over Disabled but I would expect that a disabled field is not editable even if it has "error".

Yeah, this is on purpose — we’re using a clear order for those states.

My thinking was: if there’s an error, the field shouldn’t really be blocked. You usually need to change the value to fix it, so disabling it felt backwards.

If we’re okay with that idea, we can stick to this gradation and it keeps the code a lot simpler — less special cases.

I know some developers expect disabled to always win even when error is set; if that’s what we want product-wise, we can always revisit it. Let me know.

src/components/TextInput/TextInputIcon.tsx
wonderlul: I’ve been weighing this one. The error icon is mostly a fallback indicator when there isn’t a custom trailing control: people who need an actionable control already have Accessory / TextInput.Icon, which is tappable and fully customizable.

If we made the default error icon behave like native by default, we’d need a clear action (or we’d be implying one that doesn’t exist). I’m wary of adding something like onErrorIconPress just for that case — it feels like API clutter for a path that’s often “show error only,” not “do something on press.”

That said, I’m not married to that take — if you’re confident native parity here is worth the extra surface (or you’d rather we align with platform behavior in another way), I’m happy to follow your call. Let me know what you prefer.

adrcotfas: Here's the native behavior:
- If a trailing icon exists, the error icon will override it and have its own action when clicked:
https://github.com/user-attachments/assets/2d93d3c5-08b2-4591-a581-de0bb3414066
A clear action is a different story.

I'm fine with it this way if you feel that it's overkill to cover that.

adrcotfas (approved):

satya164 (commented):
Let's keep the code style consistent with other parts of the codebase.

Avoid too many small helpers. The code is hard for me to follow because it's split between so many parts. Prefer normal hooks and components for composition instead of helper functions.

We're not using `$` prefix for styles. Use `StyleSheet.create` API so it uses atomic styles on React Native Web. Avoid unnecessary inline styles.

Also, avoid comments like these:

```
// ============
// OPACITY
// ============
```

Use normal multiline or single-line comments when necessary and don't comment on the obvious.

wonderlul:
> Let's keep the code style consistent with other parts of the codebase.
> 
> Avoid too many small helpers. The code is hard for me to follow because it's split between so many parts. Prefer normal hooks and components for composition instead of helper functions.
> 
> We're not using `$` prefix for styles. Use `StyleSheet.create` API so it uses atomic styles on React Native Web. Avoid unnecessary inline styles.
> 
> Also, avoid comments like these:
> 
> ```
> // ============
> // OPACITY
> // ============
> ```
> 
> Use normal multiline or single-line comments when necessary and don't comment on the obvious.

In the context of hooks vs. helpers, I see it more as a matter of convention and preference. If reusable logic does not manage application state, I usually prefer helper/util terminology over hooks. The helpers I created contain reusable logic, which is why I decided to extract them. Similarly, when a piece of logic becomes relatively large (e.g. getFilledTextFieldData), I tend to move it into a helper as well.

I split the logic into separate places because, from my perspective, it improves readability through clearer separation of responsibilities. Of course, this can be somewhat subjective. I think the current solution is a reasonable middle ground. I removed the filled and outlined folders and moved their logic into individual files within the main folder. Considering the complexity of the component, the current file structure feels like a good balance between readability and organization.

I also added the StyleSheet API for static styles, although most of the styles in this component are dynamic. I removed the $ notation as well.

As for comments, I only added them where I felt additional context improved readability — mainly around more complex functions, variables, styles, or to separate logical sections within files. Following your suggestion, I updated them to regular multiline or single-line comments.

example/src/Examples/TextFieldExample.tsx
```diff
@@ -0,0 +1,244 @@
...
+  type TextFieldAccessoryProps,
+  type TextFieldVariant,
+} from 'react-native-paper';
+
+import { useExampleTheme } from '../hooks/useExampleTheme';
+import ScreenWrapper from '../ScreenWrapper';
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
```
satya164: please remove comments like this

src/components/TextField/hooks.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Hooks
+   */
+
+  const input = useRef<TextInput>(null);
+
+  const theme = useInternalTheme(themeOverride);
+
+  const [isFocused, setIsFocused] = useState<boolean>(false);
```
satya164: avoid re-renders unless absolutely necessary. if the focus state can be stored in a reanimated shared value, then it'd be better for performance.

src/components/TextField/hooks.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  useImperativeHandle(ref, () => input.current as TextInput);
+
+  /**
+   * Constants
+   */
+
+  const { isRTL } = I18nManager.getConstants();
```
satya164: use the `useLocale` hook to get writing direction.

src/components/TextField/hooks.ts
```diff
@@ -0,0 +1,157 @@
...
+  const disabled = props.editable === false;
```
satya164: imo `editable={false}` has a different meaning than disabled. `editable={false}` can be used to show a read-only text input while still signaling to the user that it's interactive (e.g., focusing and selecting the state). On web, you can see that `input` supports both `disabled` and `readonly` attributes with different behaviors.

we should add a separate `disabled`.

src/components/TextField/hooks.ts
```diff
@@ -0,0 +1,157 @@
...
+  const isFloating = isFocused || !!props.value;
```
satya164: relying on `props.value` means this component relies on controlled usage. controlled inputs in react native perform very poorly, and are not reliable enough. they also re-render the whole component on each keystroke.

given that one of the goal of the refactor is to improve performance, it needs to eliminate all unnecessary re-renders.

it should also be uncontrolled by default. controlled api should be supported, but not required.

utilize reanimated shared values to reduce re-render and store state such as this that'll affect styling.

src/components/TextField/hooks.ts
```diff
@@ -0,0 +1,157 @@
...
+    animatedLabelTextStyle,
+    animatedActiveOutlineStyle,
+  };
+
+  /**
+   * Components
+   */
+
+  const LeadingAccessory = isRTL ? props.EndAccessory : props.StartAccessory;
+  const TrailingAccessory = isRTL ? props.StartAccessory : props.EndAccessory;
```
satya164: looks like `LeadingAccessory` and `TrailingAccessory` props are components. In such cases, it's easy to cause performance issues if user provides inline functions for these that re-mount each render (especially bad during controlled usage).

if they don't accept props, accept react elements instead. otherwise accept render functions that you call with arguments to get the react element.

the prop names should also be changed to camel case to accomodate this.

src/components/TextField/hooks.ts
```diff
@@ -0,0 +1,157 @@
...
+  // https://github.com/facebook/react-native/issues/31573
+  const placeholder = isFocused ? props.placeholder : ' ';
+  const counterText = `${props.value?.length ?? 0}/${props.maxLength}`;
```
satya164: this is another thing that'll need re-render on each keystroke to work. it's trickier to move it to reanimated, but a `Text` element and calling `setNativeProps` on it to update its content should work as an alternative

src/components/TextField/utils.ts
```diff
@@ -0,0 +1,636 @@
...
+      textAlign: hasSuffix === shared.isRTL ? 'left' : 'right',
+      writingDirection: shared.isRTL ? 'rtl' : 'ltr',
+    },
+    textInputProps.multiline && {
+      height: 'auto',
+      paddingTop: FILLED_MULTILINE_PADDING_TOP,
+    },
+    isWeb && {
+      outlineStyle: 'none' as TextStyle['outlineStyle'],
+    },
```
satya164: Don't use `as type`. In such cases, often `as const` will work and is typesafe.

src/components/TextField/utils.ts
```diff
@@ -0,0 +1,636 @@
...
+      textAlign: hasSuffix === shared.isRTL ? 'left' : 'right',
+      writingDirection: shared.isRTL ? 'rtl' : 'ltr',
+    },
+    textInputProps.multiline && {
+      height: 'auto',
+      textAlignVertical: 'top',
+      paddingTop: OUTLINED_MULTILINE_PADDING_TOP,
+    },
+    isWeb && {
+      outlineStyle: 'none' as TextStyle['outlineStyle'],
```
satya164: Don't use `as type`. In such cases, often `as const` will work and is typesafe.

jest/testSetup.js
```diff
@@ -89,6 +86,8 @@ jest.mock('react-native', () => {
   RN.Animated.loop = loop;
   RN.Animated.parallel = parallel;
 
+  jest.spyOn(RN.PixelRatio, 'getFontScale').mockReturnValue(1);
```
satya164: Why mock this?

src/components/TextField/constants.ts
```diff
@@ -0,0 +1,112 @@
+import { I18nManager, PixelRatio, Platform } from 'react-native';
+
+import { tokens } from '../../theme/tokens';
+import { motionDuration } from '../../theme/tokens/sys/motion';
+import { defaultShapes } from '../../theme/tokens/sys/shape';
+
+export const isWeb = Platform.OS === 'web';
```
satya164: Use `Platform.OS === 'web'` checks inline instead of `isWeb`. This way a metro can drop code for other platforms, which can't be done if it's exported like this.

src/components/TextField/constants.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const SUPPORTING_TEXT_MARGIN_TOP = 4;
+
+export const ANIMATION_DURATION_MS = motionDuration.short3;
+
+export const ACTIVE_INDICATOR_SIZE = 2;
+export const INACTIVE_INDICATOR_SIZE = 1;
+
+const isRTL = I18nManager.getConstants().isRTL;
+const layoutSupportMultiplier = isRTL ? -1 : 1;
```
satya164: Need to use `useLocale` inside the component.

src/components/TextField/TextFieldIcon.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { useInternalTheme } from '../../core/theming';
+import type { ThemeProp } from '../../types';
+import type { IconSource } from '../Icon';
+import { ACCESSORY_SIZE } from './constants';
+import { styles } from './styles';
+import type { TextFieldAccessoryProps } from './TextField';
+import { getIconColor } from './utils';
+import IconButton from '../IconButton/IconButton';
+
+export interface TextFieldIconProps extends TextFieldAccessoryProps {
```
satya164: rest of the codebase uses type aliased and not interfaces. please keep it consistent

src/components/TextField/TextFieldIcon.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <View style={[styles.iconWrapper, style]}>
+      <IconButton
+        icon={icon}
+        iconColor={iconColor}
+        size={iconSize}
+        style={styles.icon}
+        onPress={onPressHandler}
+        {...accessibility}
```
satya164: the `accessibility` name doesn't make much sense since it's spread normally, so any props can be passed. the pattern consistent with rest of the code base would be if `TextFieldIconProps` included `IconButtonProps` and forwarded `...rest`.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,422 @@
...
+            {...textInputProps}
+            placeholder={placeholder}
+            style={inputStyles}
+          />
+
+          {hasSuffix && (
+            <Text aria-hidden {...suffixProps} style={suffixStyles}>
+              {suffix}
+            </Text>
+          )}
```
satya164: should prefix and suffix be hidden from screen readers? normally it'd be incorrect to have info for non-screen reader users that screen reader users can't access

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,422 @@
...
+          <Animated.View
+            aria-hidden
+            pointerEvents="none"
+            style={animatedLabelWrapperStyles}
+          >
+            <Animated.Text {...labelProps} style={animatedLabelTextStyles}>
+              {label}
+            </Animated.Text>
+          </Animated.View>
+        )}
```
satya164: I believe react-native-web warns if you pass `pointerEvents` as a prop. pass them in `style` instead.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,422 @@
...
+   */
+  prefixProps?: TextProps;
+  /**
+   * A short text string displayed at the end of the input (e.g. `"/100"`).
+   */
+  suffix?: string;
+  /**
+   * Pass any additional props directly to the suffix `Text` component.
+   */
+  suffixProps?: TextProps;
```
satya164: not sure we need `labelProps`, `supportingTextProps`, `counterProps`, `prefixProps`, `suffixProps`. normally, we avoid exposing internals, otherwise it's easy to make breaking changes.

let's remove them. Adding them in the future if needed is easier than removing. make sure to keep the API surface small.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,422 @@
...
+  /**
+   * Style overrides for the field container (the bordered row that includes
+   * StartAccessory, input content, and EndAccessory).
+   */
+  fieldStyle?: StyleProp<ViewStyle>;
+  /**
+   * Style overrides for the input content wrapper (the area containing
+   * the label and TextInput, excluding accessories).
+   */
+  containerStyle?: StyleProp<ViewStyle>;
```
satya164: same for `pressableStyle`, `fieldStyle` and `containerStyle`. do we need these? these expose internal implementation details.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,422 @@
...
+  label?: string;
+  /**
+   * Pass any additional props directly to the label Text component.
+   */
+  labelProps?: TextProps;
+  /**
+   * Supporting text to display below the input (Material Design 3). When
+   * `error` is `true`, this text is styled as an error message.
+   */
+  supportingText?: string;
```
satya164: wouldn't `errorText` be more intuitive if it's an error message?

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,422 @@
...
+  placeholder: string | undefined;
+  counterText: string;
+  LeadingAccessory: ComponentType<TextFieldAccessoryProps> | undefined;
+  TrailingAccessory: ComponentType<TextFieldAccessoryProps> | undefined;
+  onFocusHandler: (e: FocusEvent) => void;
+  onBlurHandler: (e: BlurEvent) => void;
+  focusInput: () => void;
+};
+
+export interface TextFieldProps extends TextInputProps {
```
satya164: we use type aliases in rest of the codebase. keep it consistent.

satya164 (changes requested):
Thanks for the PR. I left few comments.

And found few issues:

- Are inputs supposed to look this big on the web?
- Alignment seems off for outlined input on the web
  Image
- Tapping on leading icon in the input seems to unfocus in the input. Since it looks like part of the input, I'd expect it to act like tapping on the input. Ideally it should be positioned absolutely on top and text input should have padding to offset it 
- Measurements seems a bit off on iOS (the app doesn't build on Android, so haven't checked). The placeholder text seems larger and as a result has less space below it compared to the official example.
  Image
- The input supports multiline but leading and trailing icons stay vertically centered. I'd expect them to be aligned at the top when there is multiline text.

Also I think while we should make breaking changes to align with MD3, it'd be valuable to keep the API same where we don't need to deviate.

- `TextField` matches MD3 naming, but I think `TextInput` is better here since users already know it, and it'll be easier migration. So we should keep `TextInput` as the component name.
- Missing props: `textColor`, `underlineColor`, `activeUnderlineColor`, `outlineColor`, `activeOutlineColor`, `selectionColor`, `cursorColor` - we should at leasy support selection and cursor color.
- Missing `render` prop to render a custom input. This is necessary to use an input implementation that supports masking etc.

src/components/TextField/TextField.tsx
wonderlul: I think we should stick with MD3 nomenclature.

satya164: from MD documentation, it doesn't seem like supporting text is specific to errors, even tho error text is displayed as supporting text

CleanShot 2026-05-15 at 11 10 59@2x

otherwise it doesn't make much sense to say supporting text if it's only for error

src/components/TextField/TextField.tsx
wonderlul: It depends. The current TextInput allows developers to override some styles or colors. I’d say we should either allow overriding everything or nothing at all — nothing in between. In my opinion, a component library should provide default styling with the ability to override it. Users can always do this by overriding the theme (since we use useInternalTheme), but in that case we’re only talking about colors, not behavior. Ultimately, I can remove the mentioned attributes assuming that will be sufficient. If they turn out to be needed, the community will definitely let us know.

src/components/TextField/TextField.tsx
wonderlul: It depends. The current TextInput allows developers to override some styles or colors. I’d say we should either allow overriding everything or nothing at all — nothing in between. In my opinion, a component library should provide default styling with the ability to override it. Users can always do this by overriding the theme (since we use useInternalTheme), but in that case we’re only talking about colors, not behavior. Ultimately, I can remove the mentioned attributes assuming that will be sufficient. If they turn out to be needed, the community will definitely let us know.

src/components/TextField/TextField.tsx
satya164: > I’d say we should either allow overriding everything or nothing at all — nothing in between

what we expose should depend on actual use cases. exposing internal implementation detail means we can't change implementation without breaking user's code. which is fine if you're building a internal component library for a client. but it's better to be conservative for public libraries.

src/components/TextField/TextField.tsx
wonderlul: Oh it's not just for error. Maybe the description is misleading, I'll remove the part regarding error styling.

src/components/TextField/TextField.tsx
wonderlul: I’ve gone back and forth on this a lot.

I leaned toward aria-hidden on prefix/suffix mainly for two reasons: 
- to avoid fragmented or duplicated announcements where the reader might walk prefix → field → suffix separately in a clunky way, or repeat something that’s already in the label;
- we document these as short visual affordances (e.g. $, /100) and assume anything meaningful for AT is carried by label / supportingText / accessibilityLabel on the TextInput.

I’m not 100% sold that’s always right if prefix/suffix carry info that isn’t duplicated elsewhere. If you’d rather expose them (or bake them into the accessible name), I’m happy to change it.

src/components/TextField/utils.ts
wonderlul: `Type '"none"' is not assignable to type '"solid" | "dotted" | "dashed" | undefined'.ts(2322)`

Our stack does not provide “better” types that include 'none'. RN’s defs are narrower than CSS and RNW does not ship TS extensions that fix that.

src/components/TextField/utils.ts
wonderlul: `Type '"none"' is not assignable to type '"solid" | "dotted" | "dashed" | undefined'.ts(2322)`

Our stack does not provide “better” types that include 'none'. RN’s defs are narrower than CSS and RNW does not ship TS extensions that fix that.

adrcotfas:
> * Measurements seems a bit off on iOS (the app doesn't build on Android, so haven't checked). The placeholder text seems larger and as a result has less space below it compared to the official example.
>   Image

FYI, in another round of review I found small inconsistencies between the visual examples and tokens, but also between the two native implementations: Jetpack Compose and MDC.

> * The input supports multiline but leading and trailing icons stay vertically centered. I'd expect them to be aligned at the top when there is multiline text.

FYI, both native implementations keep the icons centered by default but there are ways to force a top alignment.

> * `TextField` matches MD3 naming, but I think `TextInput` is better here since users already know it, and it'll be easier migration. So we should keep `TextInput` as the component name.

I disagree. IMHO this is a library that states that it follows the MD guidelines and we should use the same naming for components, tokens etc.

satya164:
> I disagree. IMHO this is a library that states that it follows the MD guidelines and we should use the same naming for components, tokens etc.

It's also a library for React Native, which uses the `TextInput` naming, and Paper has been using this name for almost a decade. Point is it'd be a big breaking change. We already have a lot of breaking changes and we should try to make the upgrade smoother instead of harder.

src/components/TextField/TextField.tsx
satya164: @wonderlul we should check how native material text field works with screen readers and replicate it.

src/components/TextField/utils.ts
satya164: @wonderlul in that case add a @ts-expect-error so it's explicit

adrcotfas:
> > I disagree. IMHO this is a library that states that it follows the MD guidelines and we should use the same naming for components, tokens etc.
> 
> It's also a library for React Native, which uses the `TextInput` naming, and Paper has been using this name for almost a decade. Point is it'd be a big breaking change. We already have a lot of breaking changes and we should try to make the upgrade smoother instead of harder.

Ok, fair point but I think we have a solution for that: type alias / adapter with `deprecated` annotation for the `TextInput`.

src/components/TextField/hooks.ts
wonderlul: In this TextField, isFocused isn’t only feeding animated styles. It’s part of the same render-time field state as isFloating (isFocused || !!value), which drives hasPrefix / hasSuffix and therefore what actually gets mounted in the tree. It also drives the placeholder workaround that has to be applied as a normal TextInput prop from the JS render path.

So we can’t drop JS knowledge of focus without either reworking that behavior or keeping a second source of truth synced from a shared value (which tends to defeat the purpose or add complexity).

On cost: focus/blur typically adds at most one render per transition, while a controlled field usually re-renders on every keystroke via value / onChangeText. For that reason I’d treat “avoid focus-triggered renders” as lower leverage here unless we measure it as a real hotspot.

wonderlul:
> Thanks for the PR. I left few comments.
> 
> And found few issues:
> 
> * Are inputs supposed to look this big on the web?
> * Alignment seems off for outlined input on the web
>   Image
> * Tapping on leading icon in the input seems to unfocus in the input. Since it looks like part of the input, I'd expect it to act like tapping on the input. Ideally it should be positioned absolutely on top and text input should have padding to offset it
> * Measurements seems a bit off on iOS (the app doesn't build on Android, so haven't checked). The placeholder text seems larger and as a result has less space below it compared to the official example.
>   Image
> * The input supports multiline but leading and trailing icons stay vertically centered. I'd expect them to be aligned at the top when there is multiline text.
> 
> Also I think while we should make breaking changes to align with MD3, it'd be valuable to keep the API same where we don't need to deviate.
> 
> * `TextField` matches MD3 naming, but I think `TextInput` is better here since users already know it, and it'll be easier migration. So we should keep `TextInput` as the component name.
> * Missing props: `textColor`, `underlineColor`, `activeUnderlineColor`, `outlineColor`, `activeOutlineColor`, `selectionColor`, `cursorColor` - we should at leasy support selection and cursor color.
> * Missing `render` prop to render a custom input. This is necessary to use an input implementation that supports masking etc.

1. **Are inputs supposed to look this big on the web?** It's hard to say how they should look like on web. They are `flex:1`, so you can wrap them in container to reduce the size.
2. **Alignment seems off for outlined input on the web**. Would you please share the browser details? Google Chrome and Safari works well for me.
3. **Measurements seems a bit off on iOS**. We can't mirror 1:1 every behaviour of MD native app due to differences in styling. Some of the implementations are hacky unfortunately. I adjusted this one.
4. **The input supports multiline but leading and trailing icons stay vertically centered. I'd expect them to be aligned at the top when there is multiline text**. According to MD specs they are centered.
5.  **`TextField` matches MD3 naming, but I think `TextInput` is better here since users already know it, and it'll be easier migration. So we should keep `TextInput` as the component name.** It's up to you. I'm fine with either TextField or TextInput. I'd still ship with TextField for this PR and possibly edit it in migration PR.
6. **Missing props**. That was exactly my point here: https://github.com/callstack/react-native-paper/pull/4909#discussion_r3247150732. Either we allow user to override styles or we don't. In my opinion there shouldn't be in-betweens that we allow to edit **some**.  Text color can be update with styles prop. selectionColor and cursorColor are acceptable props for TextField since it extends TextInputProps so they can override the defaults. Other colors can be overridden with custom theming and I'd stick with it.
7. **Missing `render` prop**. Added.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,423 @@
...
+
+export type TextFieldRenderProps = React.ComponentPropsWithRef<
+  typeof TextInput
+>;
+
+export type TextFieldProps = TextInputProps & {
+  /**
+   * Ref forwarded to the underlying TextInput.
+   */
+  ref?: React.Ref<TextInput>;
```
satya164: The exposed ref should be our custom ref, not `TextInput` ref directly, so we can have proper control and information. e.g. calling `clear()` doesn't trigger `onTextChange` but we still need to handle such cases to adjust styles/animations.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,423 @@
...
+  /**
+   * - `filled` text fields are often used in dialogs and short forms where their style draws more attention.
+   * - `outlined` text fields are often used in long forms where their reduced emphasis helps simplify the layout.
+   */
+  variant?: TextFieldVariant;
```
satya164: Include a description of what the prop is for and what the types are before saying how to use them. Something like:

```suggestion
   * Controls how the text field is styled
   * 
   * - `filled` - uses a background and colored underline that draws more attention
   * - `outline` - uses only an outline for reduced emphasis
   *  
   * `filled` variant is often used in dialogs and short forms where their style draws more attention.
   * `outlined` variant is often used in long forms where their reduced emphasis helps simplify the layout.
   */
  variant?: TextFieldVariant;
```

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,423 @@
...
+  endAccessory?: (props: TextFieldAccessoryProps) => React.ReactNode;
+  /**
+   * Callback to render a custom input component in place of the native `TextInput`.
+   * Receives all props that would be passed to `TextInput`, allowing integration
+   * with third-party inputs such as masked inputs.
+   */
+  render?: (props: TextFieldRenderProps) => React.ReactNode;
+};
+
+const DefaultRenderer = (props: TextFieldRenderProps) => (
```
satya164: This isn't a component, so shouldn't be named like a component

```suggestion
const defaultRenderer = (props: TextFieldRenderProps) => (
```

src/components/TextField/hooks.ts
```diff
@@ -0,0 +1,291 @@
...
+  const onChangeText = (text: string) => {
+    if (!isControlled) {
+      setUncontrolledValue(text);
+    }
+
+    props.onChangeText?.(text);
+  };
+
+  return { value, onChangeText };
+};
```
satya164: We already talked about this several times. The component should not be tracking `value` in state. Animations shouldn't be driven by state, but events. It's key to having responsive animations.

If tracking state is necessary (which it shouldn't be for most things), it should track a boolean, such as whether the field is empty or not, instead of tracking every keystroke. And even then, animations shouldn't be driven by this tracked state, only what's absolutely necessary.

satya164 (changes requested):
Please remove existing `TextInput` and rename the component to `TextInput` in this PR.

And please address what we already talked about regarding not tracking `value` and focus state in render.

satya164:
> Either we allow user to override styles or we don't

I don't follow this line of thought. We allow users to override what we deem safe and not an implementation detail. We don't allow users to override implementation details that can change just because of a refactor. Things like selection color, cursor color aren't going to change because of a refactor. It's a text input - these things will never become invalid no matter how you write this component.

src/components/TextField/hooks.ts
satya164: > which drives hasPrefix / hasSuffix and therefore what actually gets mounted in the tree

These are all implementation details. There is no reason it can't be a style toggle hiding/showing instead of unmounting/mounting.

> truth synced from a shared value (which tends to defeat the purpose or add complexity)

It doesn't really defeat the purpose. Not every input shows counter, and even then counter can be its own component that gets the value from shared value and keeps it localized instead of re-rendering the whole input. It's just a single component that needs to actually know the full text length.

wonderlul:
> > Either we allow user to override styles or we don't
> 
> I don't follow this line of thought. We allow users to override what we deem safe and not an implementation detail. We don't allow users to override implementation details that can change just because of a refactor. Things like selection color, cursor color aren't going to change because of a refactor. It's a text input - these things will never become invalid no matter how you write this component.

I agree. That's what I meant by `Text color can be update with styles prop. selectionColor and cursorColor are acceptable props for TextField since it extends TextInputProps so they can override the defaults. Other colors can be overridden with custom theming`. Apart from the inherited ones, I wouldn’t choose more or less important styles, because that’s relative. One developer may want to override the border on an input, while another may want to change the opacity for the disabled variant. Nevertheless, if you want to do it, I’m open to making changes. I’d just ask you to point out specifically which ones you’d like to make overridable.

wonderlul:
@satya164 I’d also ask you to take a look at the newly added `12-migration.md` document. I’m not sure whether we want to present it as a separate document or somehow merge/edit the `Getting Started` guide instead. Let me know how you’d like to approach it.

src/components/TextInput/hooks.ts
```diff
@@ -0,0 +1,463 @@
...
+
+      return layout;
+    },
+    /**
+     * `input` is a stable ref. `props` is omitted — only `multiline` affects layout.
+     * `style` is omitted — assumed stable; dynamic `style` changes won't invalidate layout.
+     * Animated styles are stable `useAnimatedStyle` objects and are omitted from deps.
+     * `isFocused` drives static focus styles (label color, outline border).
+     */
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- see comment
```
satya164: > `style` is omitted — assumed stable

why is it assumed stable? style changes should not be ignored. avoid ignoring the eslint rule unless you're actually doing some side effect that should depend on specific scenario and you are confident about the value. you can't make such assumptions in library code that'll be used in various use cases.

tho memoization is unnecessary in this scenario anyway. creating and accessing objects are not expensive and having stable identity isn't important in this case. so the memoization doesn't add any benefits, and causes bugs.

src/components/TextInput/hooks.ts
```diff
@@ -0,0 +1,463 @@
...
+
+  const renderLeadingAccessory = flags.isRTL
+    ? props.endAccessory
+    : props.startAccessory;
+  const renderTrailingAccessory = flags.isRTL
+    ? props.startAccessory
+    : props.endAccessory;
+
+  // https://github.com/facebook/react-native/issues/31573
+  const placeholder = isFocused ? props.placeholder : ' ';
```
satya164: since `label` is also optional, this should only hide the placeholder if a label is given and the text is empty.

otherwise there's no placholder when no label is given even if input is empty.

satya164 (changes requested):

src/components/TextInput/hooks.ts
wonderlul: According to the [Material Design 3 text field specs](https://m3.material.io/components/text-fields/specs), the placeholder text should only appear when the field is focused, regardless of whether a label is present or not. This is the current behavior in the code.

What you're suggesting is a deviation from MD3 specs - that when there's no label, the placeholder should remain visible even when unfocused. I agree this would be a UX enhancement to ensure users understand what the field is for when there's no label present. If you accept that behavior then I'll edit.

satya164: @wonderlul where does it say this? i don't see any mention of placeholder on the linked page. can you quote/screenshot the section that says placeholder shouldn't appear for empty text field regardless of label?

satya164: also in the official material web components, removing the label and setting a placeholder displays the placeholder regardless of whether the input is focused https://material-web.dev/components/text-field/stories/

wonderlul: My bad, sorry. I was thinking about codebase. https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/textfield/TextInputLayout.java#L2530-L2536

wonderlul: I guess those are the inconsistencies @adrcotfas mentioned between different sources.

wonderlul: OK I'll update this.

src/components/TextInput/TextInput.tsx
```diff
@@ -220,358 +289,152 @@ const DefaultRenderer = (props: RenderProps) => <NativeTextInput {...props} />;
...
+    theme,
+    prefix,
+    suffix,
+    counter,
+    disabled,
+    startAccessory,
+    endAccessory,
+    render = defaultRenderer,
+    ...textInputProps
+  } = props;
```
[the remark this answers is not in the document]
wonderlul: `ref` is explicitly typed in TextInputProps — it's declared as `ref?: React.Ref<TextInputHandles>` at line 183, overriding the inherited ref type from `NativeTextInputProps`. So it arrives as a proper prop, not undefined.

src/components/TextInput/hooks.ts
```diff
@@ -0,0 +1,463 @@
...
+    setIsFocused(false);
+
+    runBlurAnimation(getHasText());
+  };
+
+  const focusInput = () => {
+    if (isDisabled) return;
+
+    input.current?.focus();
+  };
```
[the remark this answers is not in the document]
wonderlul: The distinction between disabled and editable props is clearly intentional and correctly modeled. Looking at useTextInputFlags:

`hooks.ts`
```
  return {
    isRTL,
    isFloating,
    isDisabled: !!props.disabled,
    isEditable: props.disabled ? false : props.editable,
    hasError: !!props.error,
    hasCounter: !!(props.counter && props.maxLength),
    hasAccessory,
    hasPrefix: !!props.prefix && isFloating,
    hasSuffix: !!props.suffix && isFloating,
  };
```
`disabled=true` → fully non-interactive (blocks focusInput, forces editable: false, applies disabled styling)
`editable=false` → read-only, can still be focused and text selected but not typed into

Copilot is collapsing these two into one, but the design is deliberate: a read-only field may legitimately receive focus (e.g. to let users copy text), so the onFocus animation and state running for editable=false is correct behavior.

The only narrow edge case Copilot's comment hints at is that a caller could programmatically call ref.focus() on a disabled field via the imperative handle, since that path doesn't have an isDisabled guard. I added it.

satya164 (approved):

PR #4912, opened 2026-05-04
feat: react-native-reanimated as peer dependency

wonderlul:
### Motivation
react-native-reanimated (v4) is the intended path for animated Paper components. Today the library still uses React Native’s built-in Animated API everywhere; this PR does not migrate any UI yet. It only declares the new peer dependencies so consumers install compatible versions up front, and so maintainers can add Reanimated-based animations in follow-up PRs without changing the public dependency contract again.

Reanimated 4 depends on react-native-worklets, so both packages are added as peers (and as dev dependencies for local development, type-checking, and tests).


package.json
```diff
@@ -105,7 +107,9 @@
   "peerDependencies": {
     "react": "*",
     "react-native": "*",
-    "react-native-safe-area-context": "*"
+    "react-native-reanimated": ">=4.3.0",
+    "react-native-safe-area-context": "*",
+    "react-native-worklets": ">=0.8.1"
```
adrcotfas, line 112: Just a comment, not necessarily requesting changes as I'm not sure about best practices for `devDependencies` vs `peerDependencies` but `reanimated 4.3.0`'s own peer dep is `react-native-worklets: 0.8.x`. This would allow `react-native-worklets: 0.9.0` even though reanimated would reject it.

testSetup.js
```diff
@@ -4,6 +4,10 @@ jest.useFakeTimers();
 
 jest.mock('react-native-safe-area-context', () => mockSafeAreaContext);
 
+jest.mock('react-native-worklets', () =>
+  require('react-native-worklets/src/mock')
```
adrcotfas: I suggest `react-native-worklets/lib/module/mock` instead like reanimated uses too in their source code. It's the same artifact runtime code uses, so it doesn't depend on a working TS transform pipeline at test time.

jestSetupAfterEnv.js
```diff
@@ -0,0 +1 @@
+require('react-native-reanimated').setUpTests();
```
adrcotfas: I suggest converting to ESM with a named import so this file matches `testSetup.js` (the sibling jest setup file already uses import):                                                               
```                                                                                                                                                                   
 import { setUpTests } from 'react-native-reanimated';                                                                                                                                            
                                                       
 setUpTests();
```

docs/docs/guides/01-getting-started.md
```diff
@@ -18,6 +18,16 @@ npm install react-native-paper
...
 ```
 
+- You also need to install [react-native-reanimated](https://docs.swmansion.com/react-native-reanimated/) and [react-native-worklets](https://docs.swmansion.com/react-native-worklets/) for animations.
+
+```bash npm2yarn
+npm install react-native-reanimated react-native-worklets
+```
+
+:::note
+If you're using a bare React Native project (not Expo), you need to add `react-native-worklets/plugin` to your `babel.config.js` plugins array. See the [Reanimated installation guide](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/getting-started/) for details.
```
adrcotfas: A comment from Claude: 

> Two problems with the original wording.
> 
> Structural: the plugin instruction was inside a :::note block, which reads as "supplemental info you can skip." But this plugin is mandatory; without it, worklets fails at runtime with confusing errors. Required setup belongs in the install steps, not in a sidebar.
> 
> Factual: the original implied Expo users skip this. Reanimated v4 / worklets v0.8 require the babel plugin regardless of toolchain. The safe wording is "required for both Expo and bare React Native" rather than guessing per setup.

adrcotfas (commented):
`example/package.json` uses reanimated and worklets entries but older versions. We should bump those too or use `*` to avoid maintenance.

package.json
wonderlul: I'm not quite sure here as well. I think Paper should only states minimums; Reanimated’s own peerDependencies and the app’s lockfile are where compatible worklets / Reanimated pairings are enforced.

adrcotfas (approved):

package.json
ruben-rebelo: I'm fine having the minimums for peer dependencies, we just need to make sure this is described on documentation for user troubleshooting purposes.

jest/jestSetupAfterEnv.js
```diff
@@ -0,0 +1,3 @@
+import { setUpTests } from 'react-native-reanimated';
```
satya164, line 1: lets create a `jest` folder and move the file to there, so other future jest specific stuff can live there.

docs/docs/guides/01-getting-started.md
```diff
@@ -18,6 +18,14 @@ npm install react-native-paper
 npm install react-native-safe-area-context
 ```
 
+- You also need to install [react-native-reanimated](https://docs.swmansion.com/react-native-reanimated/) and [react-native-worklets](https://docs.swmansion.com/react-native-worklets/) for animations.
+
+```bash npm2yarn
+npm install react-native-reanimated react-native-worklets
+```
```
satya164: No need to make installation multiple steps. There should be a single installation step that installs both safe area context and reanimated + worklets.

We should also mention the installation steps for Expo. Example:

```
React Native Paper uses [react-native-safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context) for handling safe area, [react-native-reanimated](https://docs.swmansion.com/react-native-reanimated/) and [react-native-worklets](https://docs.swmansion.com/react-native-worklets/) for animations.
```

Then under the description we have 2 tabs, Expo and Community CLI and have commands that shows how to install (Expo willl use `expo install`).

The previous "From v5" wording also seems unncessary, this documentation refers to the latest version so there is no need to mention this.

docs/docs/guides/01-getting-started.md
```diff
@@ -18,6 +18,14 @@ npm install react-native-paper
...
+
+- Follow the [React Native Reanimated installation guide](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/getting-started/) for your toolchain (Expo vs React Native Community CLI). **Community CLI** setups must add `'react-native-worklets/plugin'` **last** in your `babel.config.js` `plugins` array. **Expo** setups follow that guide’s Expo steps (install packages and rebuild native code); starter templates [since Expo SDK 50](https://expo.dev/changelog/2024/01-18-sdk-50) usually include the Worklets Babel plugin already—add it only if your `babel.config.js` does not. Paper declares **minimum** supported versions for `react-native-reanimated` and `react-native-worklets` under `peerDependencies` in [`package.json`](https://github.com/callstack/react-native-paper/blob/main/package.json) (`>=` ranges, not pinned versions). When troubleshooting animation or native build problems, confirm your installed packages meet those minimums.
```
satya164: Avoid wall of text, make this bullet points so it's readable. After adding tabs as per previous comment, move relevant part to it's own tab so there is less to read.

> Paper declares **minimum** supported versions for `react-native-reanimated` and `react-native-worklets` under `peerDependencies` in [`package.json`](https://github.com/callstack/react-native-paper/blob/main/package.json) (`>=` ranges, not pinned versions). When troubleshooting animation or native build problems, confirm your installed packages meet those minimums.

We don't need to say we declare peer deps. User will get the warning/error if they don't match it from package manager. "`>=` ranges, not pinned versions" is also an unnecessary detail to mention. What we should do is document something as follows:

```js
Minimum requirements:

- `react-native` [version]
- `expo` [version]
- `react-native-safe-area-context` [version]
- etc.
```

satya164 (changes requested):

docs/docs/guides/01-getting-started.mdx
```diff
@@ -2,29 +2,60 @@
...
+  <TabItem value="community-cli" label="Community CLI">
 
 ```bash npm2yarn
-npm install react-native-safe-area-context
+npm install react-native-paper
 ```
 
-Additionaly for `iOS` platform there is a requirement to link the native parts of the library:
+  </TabItem>
+</Tabs>
```
satya164: You don't need `expo install` for `react-native-paper`. only for native  dependencies such as safe area context and reanimated. It should use npm install.

docs/docs/guides/01-getting-started.mdx
```diff
@@ -40,7 +71,7 @@ If you use Expo, you don't need to install vector icons - those are the part of
```
satya164: Move this to existing tabs for Expo/Community CLI. Otherwise we have tabs for them but still have these instructions which are not relevant to one of them.

satya164 (changes requested):

satya164 (approved):

PR #4917, opened 2026-05-05
feat: add theme.state opacity tokens and split theme types

adrcotfas:
### Motivation                                                                                                                                                                                   
                                                         
  **1. `theme.state` was missing.**                                                                                                                                                                
  Interaction-state opacities (hover, focus, press, drag) lived in the raw token layer but were never on the theme object. Components had to reach into the token layer directly, bypassing theme
  overrides. `theme.state.opacity` fixes that.                                                                                                                                                     
                                                         
  **2. `src/theme/types.ts` was getting too big.**                                                                                                                                                 
  All theme types lived in one 179-line file. Split into `src/theme/types/` one file per concern so new types (like `ThemeState`) have a clear place to go.
                                                                                                                                                                                                   
  No behavior changes for existing consumers: `Theme` gains one new required field (`state`), all built-in schemes populate it, and the barrel re-exports everything from the same path.           
                                                                                                                                                                                                   
  ### Related issue                                                                                                                                                                                
                                                         
    #### Merge order:

  ### Test plan

  - `yarn test` -- all tests pass; updated `ListSection` snapshot reflects the new `state` field


src/theme/types/typography.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export type Fonts = {
+  regular: Font;
+  medium: Font;
+  light: Font;
+  thin: Font;
+};
+
+export enum TypescaleKey {
```
satya164: Don't use enums. Use a union type.

I see it was already there, but since we're touching this code, let's change this.

src/theme/types/elevation.ts
```diff
@@ -0,0 +1,14 @@
+export type Elevation = 0 | 1 | 2 | 3 | 4 | 5;
+
+export enum ElevationLevels {
```
satya164: Don't use enums. Use union types.

Treat TypeScript as a linter and avoid TypeScript-specific constructs. This has been the direction TypeScript is also going where the syntax is aligned with JS syntax.

Enums also increase bundle size, more verbose requiring an import where just typing a string will suffice with union types, and don't really provide any advantages.

We should set `erasableSyntaxOnly` in TS config to enforce this in the future.

src/theme/types/elevation.ts
```diff
@@ -0,0 +1,14 @@
...
+  'level1',
+  'level2',
+  'level3',
+  'level4',
+  'level5',
+}
+
+export type ElevationColors = {
+  [key in keyof typeof ElevationLevels]: string;
+};
```
satya164: Simpler code would be to do this

```suggestion
export type ElevationColors = Record<`level${Elevation}`, string>
```

src/theme/types/theme.ts
```diff
@@ -0,0 +1,27 @@
...
+  animation: {
+    scale: number;
+    defaultAnimationDuration?: number;
+  };
+};
+
+export type Theme = ThemeBase & {
+  colors: ThemeColors;
+  fonts: Typescale;
+  state: ThemeState;
```
satya164: I'm not sure about the name `state`. The term "state" already means something else in React, and it's better to avoid such overloaded terms.

src/theme/tokens/sys/state.ts
```diff
@@ -0,0 +1,15 @@
...
+const { stateOpacity } = tokens.md.ref;
+
+export const defaultState: ThemeState = {
+  opacity: {
+    hovered: stateOpacity.hovered,
+    focused: stateOpacity.focused,
+    pressed: stateOpacity.pressed,
+    dragged: stateOpacity.dragged,
+    disabled: stateOpacity.disabled,
+    enabled: stateOpacity.enabled,
```
satya164: Don't think this makes sense to expose `enabled` as a public API to customize. There shouldn't be a different opacity for enabled components other than 1.

satya164 (changes requested):
Do we really need to add this to the theme? Is this something that the user should be customizing according to MD guidelines? We don't need to expose every material design token in the theme, only what's relevant for customization. I'd lean towards keeping it internal.

Follow the philosophy of not adding public APIs unless actually needed. We can always add things easily in the future if people need them. Removing things is a breaking change and much harder.

The theme adds new tokens, but components don't use them. I assume it'll be a follow up PR

src/theme/types/elevation.ts
adrcotfas: Done

src/theme/tokens/sys/state.ts
adrcotfas: I removed it from the theme but it's still in `ref`. I think it's valuable in places where we have conditionals like `disabled ? stateOpacity.disabled : stateOpacity.enabled`

adrcotfas:
> Do we really need to add this to the theme? Is this something that the user should be customizing according to MD guidelines? We don't need to expose every material design token in the theme, only what's relevant for customization. I'd lean towards keeping it internal.
> 
> Follow the philosophy of not adding public APIs unless actually needed. We can always add things easily in the future if people need them. Removing things is a breaking change and much harder.
> 

Agree. It's in `ref` now.

> The theme adds new tokens, but components don't use them. I assume it'll be a follow up PR
Yes, we'll need to go through each component, modernize it and plug in the tokens.

satya164 (approved):

PR #4919, opened 2026-05-05
feat: add theme.shapes and cornersToStyle

adrcotfas:
### Motivation

Adds the 10 MD3 shape scale tokens to theme.shapes.*, ShapeCorners for directional variants, and cornersToStyle() for mapping to RN style props. Deprecates theme.roundness in favour of theme.shapes.

### Related issue

  #### Merge order:

### Test plan

### DIFF
--- a/src/theme/tokens/sys/shape.ts
+++ b/src/theme/tokens/sys/shape.ts
@@ -1,0 +1,17 @@
+import type { ThemeShapes } from '../../types';
+
+export const cornerNone = 0;
+export const cornerFull = 9999;
+
+export const defaultShapes: ThemeShapes = {
+  corner: {
+    extraSmall: 4,
+    small: 8,
+    medium: 12,
+    large: 16,
+    largeIncreased: 20,
+    extraLarge: 28,
+    extraLargeIncreased: 32,
+    extraExtraLarge: 48,
+  },
+};
--- a/src/theme/types/shape.ts
+++ b/src/theme/types/shape.ts
@@ -1,0 +1,14 @@
+export type ThemeShapeCorners = {
+  extraSmall: number;
+  small: number;
+  medium: number;
+  large: number;
+  largeIncreased: number;
+  extraLarge: number;
+  extraLargeIncreased: number;
+  extraExtraLarge: number;
+};
+
+export type ThemeShapes = {
+  corner: ThemeShapeCorners;
+};
--- a/src/theme/types/theme.ts
+++ b/src/theme/types/theme.ts
@@ -1,6 +1,7 @@
 import type { $DeepPartial } from '@callstack/react-theme-provider';
 
 import type { ThemeColors } from './color';
+import type { ThemeShapes } from './shape';
 import type { Typescale } from './typography';
 
 type Mode = 'adaptive' | 'exact';
@@ -8,7 +9,6 @@ type Mode = 'adaptive' | 'exact';
 export type ThemeBase = {
   dark: boolean;
   mode?: Mode;
-  roundness: number;
   animation: {
     scale: number;
     defaultAnimationDuration?: number;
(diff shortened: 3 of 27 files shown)


src/theme/types/theme.ts
```diff
@@ -9,6 +10,7 @@ type Mode = 'adaptive' | 'exact';
 export type ThemeBase = {
   dark: boolean;
   mode?: Mode;
+  /** @deprecated Use `theme.shapes.*` instead. Will be removed in a future version. */
   roundness: number;
```
satya164: Don't keep deprecated values in the theme. It makes things more complicated for components. If there is a replacement, drop the old one.

src/index.tsx
```diff
@@ -14,6 +14,7 @@ export { default as Provider } from './core/PaperProvider';
 export { default as PaperProvider } from './core/PaperProvider';
 export { default as shadow } from './theme/shadow';
 export { default as configureFonts } from './theme/fonts';
+export { cornersToStyle } from './theme/tokens/sys/shape';
```
satya164: Don't expose such tiny utilities. Keep the library public API surface small. It's very straightforward for users to write such utility themselves, should they need it.

src/theme/tokens/sys/shape.ts
```diff
@@ -0,0 +1,33 @@
+import type { ViewStyle } from 'react-native';
+
+import type { ShapeCorners, ThemeShapes } from '../../types';
+
+export const defaultShapes: ThemeShapes = {
+  none: 0,
```
satya164: Don't think it makes sense to expose `none` as a customizable theme value. It'll always be `0`.

src/theme/tokens/sys/shape.ts
```diff
@@ -0,0 +1,33 @@
...
+  extraSmall: 4,
+  small: 8,
+  medium: 12,
+  large: 16,
+  largeIncreased: 20,
+  extraLarge: 28,
+  extraLargeIncreased: 32,
+  extraExtraLarge: 48,
+  full: 9999,
```
satya164: Same as `none`. Don't think it makes sense to expose it as a customizable theme value.

src/theme/types/theme.ts
```diff
@@ -20,6 +22,7 @@ export type Theme = ThemeBase & {
   colors: ThemeColors;
   fonts: Typescale;
   state: ThemeState;
+  shapes: ThemeShapes;
```
satya164, line 21: I'm not sure about the name "shapes". It says "shapes" but it exposes roundness/border radius values, not actual shapes, so the name is confusing.

src/theme/types/shape.ts
```diff
@@ -0,0 +1,19 @@
+export type ShapeCorners = {
+  topStart: number;
+  topEnd: number;
+  bottomStart: number;
+  bottomEnd: number;
+};
```
satya164: Not sure we need this. Isn't this just style properties?

satya164 (changes requested):

src/theme/types/theme.ts
adrcotfas: At this point in time I wasn't sure we want backwards compatibility. I'll port the appropriate changes from #4926.

src/theme/types/theme.ts
adrcotfas: I added an extra layer of `corner` and we can extend it to include something like `primitives` for the 35 predefined shapes I didn't add yet(but are used in the loading indicator component which needs modernization).  
I'm trying to use a similar token hierarchy as in the specs.  

```
export type ThemeShapeCorners = {
  extraSmall: number;
  small: number;
  medium: number;
  large: number;
  largeIncreased: number;
  extraLarge: number;
  extraLargeIncreased: number;
  extraExtraLarge: number;
};

export type ThemeShapes = {
  corner: ThemeShapeCorners;
};
```
If you have other suggestions, please let me know.

src/theme/tokens/sys/shape.ts
adrcotfas: I extracted `none` and `full` as constants. Please let me know if you prefer a different approach.

satya164 (approved):

PR #4921, opened 2026-05-05
feat: add direction prop to PaperProvider and useLocale hook for RTL support

hristototov:
### Motivation

Components in react-native-paper read RTL state directly from
`I18nManager.getConstants().isRTL`. This has two issues:

- `I18nManager` is a no-op on React Native Web, so RTL breaks entirely on Web
- The value is read at module load or render time with no way to override it reactively

## Solution

New: `useLocale()` hook

```ts
import { useLocale } from 'react-native-paper';

const { direction } = useLocale(); // 'ltr' | 'rtl'

<PaperProvider direction="rtl">
  {children}
</PaperProvider>
```

Defaults to I18nManager.getConstants().isRTL when not provided, so existing apps are unaffected. Setting the prop does not call I18nManager.forceRTL or modify any native state; it only tells Paper which direction to use for rendering.

### Related issue

N/A 
[Notion link](https://www.notion.so/callstack/Improve-RTL-handling-in-paper-32e5d027c0f8807caf67e6cdbc6d4d74)

### Test plan

 - Add direction="rtl" to PaperProvider in the example app and verify RTL layout renders correctly across components
 Remove the prop and verify the app falls back to I18nManager (default behaviour unchanged)
- On React Native Web, set direction="rtl" and verify RTL layout works
 - All existing tests pass (yarn test)


tjaniczek (approved):

src/core/PaperProvider.tsx
```diff
@@ -101,7 +105,9 @@ const PaperProvider = (props: Props) => {
     <SafeAreaProviderCompat>
       <PortalHost>
         <SettingsProvider value={settingsValue}>
-          <ThemeProvider theme={theme}>{children}</ThemeProvider>
+          <LocaleProvider value={{ direction }}>
```
adrcotfas: We get a new object on every render and this affects FPS and re-render counts on screens that use Paper components.  
We already memoizes `theme` and `settingsValue`. Do the same for this too.

adrcotfas (changes requested):

src/core/PaperProvider.tsx
hristototov: updated

src/core/locale.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const getDefaultDirection = (): Direction =>
+  I18nManager.getConstants?.().isRTL ? 'rtl' : 'ltr';
+
+export const LocaleContext = React.createContext<LocaleContextValue>({
+  direction: getDefaultDirection(),
+});
+
+export const { Provider: LocaleProvider } = LocaleContext;
+
+export const useLocale = () => React.useContext(LocaleContext);
```
satya164: It shouldn't provide a default locale in the context itself. It must be provided when rendering the provider so it stays consistent, and components outside the tree properly throw.

The `useLocale` hook should throw an error when used outside the provider.


```suggestion
export const LocaleContext = React.createContext<LocaleContextValue | undefined>();

export const { Provider: LocaleProvider } = LocaleContext;

export function useLocale() {
  const locale = React.useContext(LocaleContext);
  
  if (locale === undefined) {
    throw new Error("Couldn't find a locale. Is your component inside PaperProvider?");
  }
  
  return locale;
}
```

satya164 (changes requested):
Thanks for the PR. I left some comments. In addition, lets do the following:

- Export a `LocaleProvider` like done for `ThemeProvider`, so user can wrap a specific subtree in a different locale if needed. The defaults etc., can live in this component directly instead of a `getDefaultDirection` export
- Inherit the nearest locale in `Portal.tsx` as done for the theme
- Update documentation. Here is an example https://reactnavigation.org/docs/navigation-container/#direction

src/core/locale.tsx
hristototov: I've addressed the three bullet points above. 

However, the original task in Notion asked for a default value: https://www.notion.so/callstack/Improve-RTL-handling-in-paper-32e5d027c0f8807caf67e6cdbc6d4d74?v=3285d027c0f880a68ccb000c1ce3c56e

In your comment, you suggest we don't have one in order to be consistent. However, `theme` does have a default value `MD3LightTheme`, which means components using `useTheme` work without any wrapper in tests. 

If we were to not have a default value for `locale`, then we'd have to update all (~46) test suites to wrap renders with a `LocaleProvider` (since useLocale would throw outside one), which I think is a bit too much for this PR. LMK your thoughts and if I got something wrong.

satya164: > the original task in Notion asked for a default value

Default value in the provider, not in the context itself.

> then we'd have to update all (~46) test suites to wrap renders with a LocaleProvider

That's fine. Tests should be testing a real scenario - where users will have `PaperProvider` - so tests should use that.

hristototov: updated

- null default, `useLocale` throws outside a provider
- added `test-utils.tsx` - minimal wrapper (SettingsProvider + LocaleProvider + ThemeProvider, no SafeAreaProviderCompat) so Modal/Snackbar tests with their own safe area mocks aren't affected
- 45 test files render now imported from test-utils
- added `"types": ["jest"]` to tsconfig file as I was having IDE issues with the test files

docs/docs/guides/10-rtl.md
```diff
@@ -0,0 +1,72 @@
...
+
+# RTL Support
+
+React Native Paper supports right-to-left (RTL) layouts for languages such as Arabic and Hebrew.
+
+## How it works
+
+On React Native, the writing direction is normally controlled by `I18nManager.forceRTL`. React Native Paper reads this automatically i.e. no configuration is needed for native apps that already set up RTL via `I18nManager`.
+
+However, `I18nManager` is a no-op on **React Native Web**, which means RTL layouts break silently in web apps. The `direction` prop on `PaperProvider` (and the `LocaleProvider` component) lets you explicitly control the writing direction so Paper behaves correctly on all platforms.
```
satya164: this is not full story, since calling `forceRTL` isn't enough. a re-render with the new value is necessary. also as I remember, with Expo it was needed to specify it in `app.json` in certain contexts instead of using `forceRTL`. so recommending to link to their docs instead.

it's also not right to say "lets you explicitly control the writing direction" since below we say it's only for informing, not controlling. we need different code on web to enable it (setting `dir` attribute).

i'd rephrase it to:


```suggestion
By default, React Native Paper reads the writing direction from `I18nManager.getConstants().isRTL` on native platforms. So it will use your existing RTL setup on initial render.

See [I18nManager](http://reactnative.dev/docs/i18nmanager) docs and [Enabling RTL support in Expo](https://docs.expo.dev/guides/localization/#enabling-rtl-support) to configure your app properly.

On the Web, the RTL value is not set globally, unlike native platforms. `I18nManager.getConstants().isRTL` is a no-op on [React Native Web](https://necolas.github.io/react-native-web/). To enable RTL globally, you can specify `dir` attribute on the `html` element:

<html dir="rtl">
  <!-- App content -->
</html>

Then, let `react-native-paper` know about it by using the `direction` prop on `PaperProvider` or the `LocaleProvider` component to match the writing direction in your app.
```

tsconfig.json
```diff
@@ -30,7 +30,8 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "esnext"
+    "target": "esnext",
+    "types": ["jest", "node"]
```
satya164, line 34: tests should import from `@jest/global` instead of making the types global but that can be a follow-up.

satya164 (changes requested):

tsconfig.json
hristototov: Sure, if you end up making a ticket for that, feel free to assign it to me.

satya164 (approved):

tsconfig.json
satya164: i can't assign it to you but here is the ticket https://github.com/orgs/callstack/projects/14/views/2?pane=issue&itemId=187197137

cc @ruben-rebelo

ruben-rebelo (approved):

PR #4922, opened 2026-05-05
feat: add motion tokens (spring, easing, duration)

adrcotfas:
### Motivation

Adds theme.motion with two preset schemes (expressiveMotion default, standardMotion), M3 easing curves, and duration constants verified against the material.io spec. Includes toRawSpring() helper to convert the spec damping ratio to the raw coefficient expected by Animated.spring and Reanimated. Deprecates animation.defaultAnimationDuration in favour of theme.motion.duration.*. No component consumes the tokens yet.

### Related issue

See https://www.notion.so/callstack/React-Native-Paper-Foundation-for-MD3-Expressive-34c5d027c0f880edba3df107cd35946f?source=copy_link

  #### Merge order:
- https://github.com/callstack/react-native-paper/pull/4910
- https://github.com/callstack/react-native-paper/pull/4915
- https://github.com/callstack/react-native-paper/pull/4916
- https://github.com/callstack/react-native-paper/pull/4917
- https://github.com/callstack/react-native-paper/pull/4919
- https://github.com/callstack/react-native-paper/pull/4920
- https://github.com/callstack/react-native-paper/pull/4922
- https://github.com/callstack/react-native-paper/pull/4923
- https://github.com/callstack/react-native-paper/pull/4924
- https://github.com/callstack/react-native-paper/pull/4925
- https://github.com/callstack/react-native-paper/pull/4926
- https://github.com/callstack/react-native-paper/pull/4927   

### Test plan

  - `yarn typescript` -- no new type errors
  - `yarn test` -- all tests pass

PR #4920, opened 2026-05-05
feat: add emphasized typescale and fix letter-spacing spec bugs

adrcotfas:
### Motivation

Adds 15 *Emphasized entries per M3 Design Kit: Display/Headline/TitleLarge/Body use Medium (500), TitleMedium/TitleSmall/Label use Bold (700). Adds weightBold to ref.typeface. Fixes displayLarge letterSpacing (0 -> -0.25) and bodyLarge letterSpacing (0.15 -> 0.5).

### Related issue

See https://www.notion.so/callstack/React-Native-Paper-Foundation-for-MD3-Expressive-34c5d027c0f880edba3df107cd35946f?source=copy_link

  #### Merge order:
- https://github.com/callstack/react-native-paper/pull/4910
- https://github.com/callstack/react-native-paper/pull/4915
- https://github.com/callstack/react-native-paper/pull/4916
- https://github.com/callstack/react-native-paper/pull/4917
- https://github.com/callstack/react-native-paper/pull/4919
- https://github.com/callstack/react-native-paper/pull/4920
- https://github.com/callstack/react-native-paper/pull/4922
- https://github.com/callstack/react-native-paper/pull/4923
- https://github.com/callstack/react-native-paper/pull/4924
- https://github.com/callstack/react-native-paper/pull/4925
- https://github.com/callstack/react-native-paper/pull/4926
- https://github.com/callstack/react-native-paper/pull/4927

### Test plan

  - `yarn typescript` -- no new type errors
  - `yarn test` -- all tests pass

PR #4923, opened 2026-05-05
feat: add elevation tokens and fix Android dp spec bug

adrcotfas:

### Motivation

Part of the foundation of adding tokens and structure. 

Add theme.elevation sys token and clean up elevation types                                                                                                                                                                                                                                                                           
                                
  Replace the `ElevationLevels` numeric enum with an `ElevationLevel` string union, add `ThemeElevation` and `theme.elevation.level0-5` as a proper sys token (matching `md.sys.elevation.*` spec naming), export `elevationInputRange` as the single source of truth for the [0-5] interpolation range, fix iOS shadow color to read from         
  `theme.colors.shadow` instead of a hardcoded #000, and remove `ELEVATION_LEVELS_MAP` from `Menu` in favor of the template literal pattern already used in `Surface`.
  
### Related issue                                                                                                                                                                                
                                                         
#### Merge order:

  ### Test plan


src/theme/shadow.tsx
```diff
@@ -1,44 +1,10 @@
...
-        height: shadowHeight[elevation],
-      },
-      shadowRadius: shadowRadius[elevation],
-    };
-  }
+import { shadow as _shadow } from './tokens/sys/elevation';
+
+/**
+ * @deprecated Pass `theme.colors.shadow` as the second argument and import
+ * `shadow` from `react-native-paper/theme/tokens/sys/elevation`. Will be removed in a future version.
```
satya164: Users shouldn't be importing things using deep imports. Either make a clean breaking change or keep the API backward-compatible.

src/components/Surface.tsx
```diff
@@ -75,54 +80,41 @@ const outerLayerStyleProperties: (keyof ViewStyle)[] = [
...
-          outputRange: iOSShadowOutputRanges[layer].height,
+          inputRange: [...elevationInputRange],
+          outputRange: [...shadowLayers[layer].height],
         }),
       },
       shadowRadius: elevation.interpolate({
-        inputRange,
-        outputRange: iOSShadowOutputRanges[layer].shadowRadius,
+        inputRange: [...elevationInputRange],
+        outputRange: [...shadowLayers[layer].shadowRadius],
```
satya164: why are these spread instead of using directly?

src/components/Menu/Menu.tsx
```diff
@@ -675,7 +670,9 @@ const Menu = ({
                   shadowMenuContainerStyle,
                   {
                     backgroundColor:
-                      md3Colors.elevation[ELEVATION_LEVELS_MAP[elevation]],
+                      md3Colors.elevation[
+                        `level${elevation}` as ElevationLevel
```
satya164: don't use `as`

src/components/Surface.tsx
```diff
@@ -255,13 +249,11 @@ const Surface = forwardRef<View, Props>(
...
 
-    const inputRange = [0, 1, 2, 3, 4, 5];
-
     const backgroundColor = (() => {
       if (isAnimatedValue(elevation)) {
         return elevation.interpolate({
-          inputRange,
-          outputRange: inputRange.map((elevation) => {
+          inputRange: [...elevationInputRange],
+          outputRange: [...elevationInputRange].map((elevation) => {
```
satya164: why are these spreading?

src/theme/tokens/sys/elevation.ts
```diff
@@ -0,0 +1,69 @@
...
+  level1: 1,
+  level2: 2,
+  level3: 3,
+  level4: 4,
+  level5: 5,
+};
+
+export const elevationInputRange = [0, 1, 2, 3, 4, 5] as const;
+
+export const androidElevationLevels = [0, 1, 3, 6, 8, 12] as const;
```
satya164: if these are the actual values, why not expose directly instead of having 0-5 range and translating to a different format? avoiding the indirection would keep things simpler.

src/theme/tokens/sys/elevation.ts
```diff
@@ -0,0 +1,69 @@
...
+        }),
+      },
+      shadowOpacity: elevation.interpolate({
+        inputRange: [0, 1],
+        outputRange: [0, 0.3],
+        extrapolate: 'clamp',
+      }),
+      shadowRadius: elevation.interpolate({
+        inputRange: [...elevationInputRange],
+        outputRange: [...shadowLayers[0].shadowRadius],
```
satya164: the spreading seems unnecessary

src/theme/types/elevation.ts
```diff
@@ -1,14 +1,13 @@
...
-  'level4',
-  'level5',
-}
+export type ElevationLevel =
+  | 'level0'
+  | 'level1'
+  | 'level2'
+  | 'level3'
+  | 'level4'
+  | 'level5';
```
satya164: There is an opportunity to reduce repetition and simplify these. The same things don't need to be define as enum, union, a numeric union, then later as object in src/theme/tokens/sys/elevation.ts, and a separate range - it could be derived from the map with `Object.values`, types could be derived from other types, redundant enum could be removed, etc.

src/theme/tokens/sys/elevation.ts
```diff
@@ -0,0 +1,69 @@
...
+    height: [0, 1, 1, 1, 2, 4],
+    shadowRadius: [0, 2, 2, 3, 3, 4],
+  },
+] as const;
+
+export function shadow(
+  elevation: number | Animated.Value = 0,
+  shadowColor: string
+) {
+  if (elevation instanceof Animated.Value) {
```
satya164: does this work for animated interpolations etc.? i also saw there was a `isAnimatedValue` helper in another PR

src/components/Surface.tsx
```diff
@@ -292,12 +286,12 @@ const Surface = forwardRef<View, Props>(
...
 
     if (Platform.OS === 'android') {
-      const elevationLevel = [0, 3, 6, 9, 12, 15];
+      const elevationLevel = [...androidElevationLevels];
 
       const getElevationAndroid = () => {
         if (isAnimatedValue(elevation)) {
           return elevation.interpolate({
-            inputRange,
+            inputRange: [...elevationInputRange],
```
satya164: is spreading necessary?

satya164 (changes requested):

src/components/Surface.tsx
adrcotfas: My bad.

src/theme/tokens/sys/elevation.ts
adrcotfas: The 0-5 levels are MD3 specific and the dp values are a render detail; these are used in `Surface`'s elevatin prop.  
Would you prefer to have this different?  

Edit: I added a separate commit for the review findings for easier re-review.

satya164 (approved):

PR #4924, opened 2026-05-06
refactor(PaperProvider): add reduceMotion prop and useReduceMotion hook

adrcotfas:

### Motivation

The library's existing reduce-motion support set `animation.scale = 0` -- a v2-era flag that components had to explicitly check, with no connection to the MD3 `theme.motion` spring and duration tokens.
                                                                                                                                                                                                                                                                                     
This PR wires OS accessibility into the theme properly. `useAccessibleTheme` replaces the inline `PaperProvider` logic, subscribes to `reduceMotionChanged`, and when active collapses `theme.motion` to instant springs and zero durations via the new `reducedMotion` preset. A clean `theme.motion.prefersReducedMotion: boolean` flag gives components a single readable signal instead of the magic scale === 0 check.
                                                                                                                                                                                                                                                                                     
  `animation.scale` is kept and still set to 0 for backward compatibility with components that haven't migrated yet. It is now formally deprecated in favour of `theme.motion.prefersReducedMotion`.                                                                                     
   
Two further deprecations are added while here: `theme.mode` ('adaptive' | 'exact'), which is an MD2 elevation-overlay concept superseded by tonal surface colors in `theme.colors.elevation.*`; and `animation.defaultAnimationDuration`, which was never read at runtime.               
No component behavior changes. The reduce-motion path is functionally equivalent to before for all existing components. theme.motion.prefersReducedMotion and the reducedMotion preset are foundation plumbing -- they will be consumed when components migrate to `theme.motion` tokens in the per-component PRs that follow.
  
### Related issue

#### Merge order: 

- https://github.com/callstack/react-native-paper/pull/4945

### Test plan

### DIFF
--- a/src/core/PaperProvider.tsx
+++ b/src/core/PaperProvider.tsx
@@ -1,92 +1,47 @@
 import * as React from 'react';
-import {
-  AccessibilityInfo,
-  Appearance,
-  ColorSchemeName,
-  NativeEventSubscription,
-} from 'react-native';
 
 import { getDefaultDirection, LocaleProvider, type Direction } from './locale';
 import SafeAreaProviderCompat from './SafeAreaProviderCompat';
 import { Provider as SettingsProvider, Settings } from './settings';
 import { defaultThemes, ThemeProvider } from './theming';
+import {
+  useResolvedReduceMotion,
+  type ReduceMotionPreference,
+} from './useResolvedReduceMotion';
+import { useSystemColorScheme } from './useSystemColorScheme';
 import MaterialCommunityIcon from '../components/MaterialCommunityIcon';
 import PortalHost from '../components/Portal/PortalHost';
+import { ReduceMotionContext } from '../theme/accessibility/ReduceMotionContext';
 import type { ThemeProp } from '../types';
-import { addEventListener } from '../utils/addEventListener';
 
 export type Props = {
   children: React.ReactNode;
   theme?: ThemeProp;
   settings?: Settings;
   direction?: Direction;
+  reduceMotion?: ReduceMotionPreference;
 };
 
 const PaperProvider = (props: Props) => {
-  const colorSchemeName =
-    (!props.theme && Appearance?.getColorScheme()) || 'light';
+  const { reduceMotion = 'auto' } = props;
 
-  const [reduceMotionEnabled, setReduceMotionEnabled] =
-    React.useState<boolean>(false);
-  const [colorScheme, setColorScheme] =
-    React.useState<ColorSchemeName>(colorSchemeName);
-
-  const handleAppearanceChange = (
-    preferences: Appearance.AppearancePreferences
-  ) => {
-    const { colorScheme } = preferences;
-    setColorScheme(colorScheme);
-  };
-
-  React.useEffect(() => {
-    let subscription: NativeEventSubscription | undefined;
-
-    if (!props.theme) {
-      subscription = addEventListener(
-        AccessibilityInfo,
-        'reduceMotionChanged',
-        setReduceMotionEnabled
-      );
-    }
-    return () => {
-      if (!props.theme) {
-        subscription?.remove();
-      }
-    };
-  }, [props.theme]);
-
-  React.useEffect(() => {
-    let appearanceSubscription: NativeEventSubscription | undefined;
-    if (!props.theme) {
-      appearanceSubscription = Appearance?.addChangeListener(
-        handleAppearanceChange
-      ) as NativeEventSubscription | undefined;
-    }
-    return () => {
-      if (!props.theme) {
-        if (appearanceSubscription) {
-          appearanceSubscription.remove();
-        } else {
-          // @ts-expect-error: We keep deprecated listener remove method for backwards compat with old RN versions
-          Appearance?.removeChangeListener(handleAppearanceChange);
-        }
-      }
-    };
-  }, [props.theme]);
+  const colorScheme = useSystemColorScheme(!props.theme);
+  const resolvedReduceMotion = useResolvedReduceMotion(reduceMotion);
 
   const theme = React.useMemo(() => {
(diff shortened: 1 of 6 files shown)


src/index.tsx
```diff
@@ -18,9 +18,12 @@ export { cornersToStyle } from './theme/tokens/sys/shape';
 export {
   expressiveMotion,
   standardMotion,
+  reducedMotion,
   toRawSpring,
 } from './theme/tokens/sys/motion';
 
+export { useAccessibleTheme } from './theme/accessibility';
```
satya164: Don't think this belongs as a public API. We should only export what's absolutely necessary. Avoid expanding the API surface unless really needed.

src/theme/types/theme.ts
```diff
@@ -7,16 +7,20 @@ import type { ThemeShapes } from './shape';
...
   mode?: Mode;
   /** @deprecated Use `theme.shapes.*` instead. Will be removed in a future version. */
   roundness: number;
+  /** @deprecated Use `theme.motion.*` instead. Will be removed in a future version. */
   animation: {
+    /** @deprecated Use `theme.motion.prefersReducedMotion` instead. Will be removed in a future version. */
     scale: number;
-    /** @deprecated Use `theme.motion.duration.*` instead. Will be removed in a future version. */
+    /** @deprecated No-op. Use `theme.motion.duration.*` instead. Will be removed in a future version. */
     defaultAnimationDuration?: number;
```
satya164: Remove deprecated theme properties. We don't want another situation like V2/v3 themes.

If necessary, the migration guide can show a helper that converts old theme to new theme for easier migration.

src/index.tsx
```diff
@@ -18,9 +18,12 @@ export { cornersToStyle } from './theme/tokens/sys/shape';
 export {
   expressiveMotion,
   standardMotion,
+  reducedMotion,
```
satya164: I'm not sure it's necessary to expose this.

src/theme/tokens/sys/motion.ts
```diff
@@ -78,12 +78,44 @@ export const expressiveMotion: MotionConfig = {
...
+    long1: 0,
+    long2: 0,
+    long3: 0,
+    long4: 0,
+    extraLong1: 0,
+    extraLong2: 0,
+    extraLong3: 0,
+    extraLong4: 0,
+  },
 };
```
satya164: I'm not sure setting all of these to 0/instant animation makes sense for reduce motion. Reduce motion is about reducing motion (e.g., movement, like translation, scale, etc.). Simple animations like opacity and colors should still work. It shouldn't disable animations entirely.

It probably makes more sense to handle it at the component level, because how each component changes its animation in response to this depends on the component, e.g. menu should change to simple fade instead of scale etc.

src/theme/accessibility/useAccessibleTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+
+function applyReducedMotion(theme: Theme): Theme {
+  return {
+    ...theme,
+    animation: { ...theme.animation, scale: 0 },
+    motion: reducedMotion,
+  };
+}
+
+export function useAccessibleTheme(theme: Theme, enabled = true): Theme {
```
satya164: I don't think this should be wrapping `theme` in `useTheme`. This causes every component using this hook  to subscribe to reduce motion.

If this is handled, then it should be done in `PaperProvider` in a single place so there's a single subscription.

But also, if `prefersReducedMotion` is in the theme, it conflicts with such an API if it's also handled internally. When the user passes a `theme`, which will already have `prefersReducedMotion`, is it also handled internally and overridden?

It probably makes sense to expose a separate prop `reduceMotion="auto"` or something like that on `PaperProvider`, then internally it can subscribe and expose the actual value in context. Then, on the component level, components can adapt their animations to reduce motion when necessary.

There is also the problem of the value not being known on initial render, as it's async, so if some component has animation on render, it won't properly respect this.

satya164 (changes requested):

src/theme/tokens/sys/motion.ts
adrcotfas: Agree

src/theme/accessibility/useAccessibleTheme.ts
adrcotfas: I refactored that part. Previously we had a bug because the reduce motion handling was listener only. The state stayed false forever until the user toggled reduce motion in the OS during the running session.  
Now one frame is still wrong. Do we accept that for now or do you see a solution? Native modules might be an overkill for this edge case.

satya164: @adrcotfas i guess it's acceptable for now as it's rare to have motion animation fr initial render. there's also option to create native module but keep it optional, and fallback to react native's API if it's unavailable. i don't see another option.

src/core/PaperProvider.tsx
```diff
@@ -1,90 +1,44 @@
...
+      ...base,
       ...props.theme,
       animation: {
         ...props.theme?.animation,
-        scale: reduceMotionEnabled ? 0 : 1,
+        scale: resolvedReduceMotion ? 0 : userScale,
       },
-    };
-  }, [colorScheme, props.theme, reduceMotionEnabled]);
+    } as Theme;
```
satya164: Avoid `as` as it's not typesafe. A return type annotation on the function should handle it.

satya164 (approved):

src/core/PaperProvider.tsx
adrcotfas: My bad for that - the cast was wrong but there was a pre-existing issue here so I reverted it.  
A return annotation doesn't actually fix it though: `props.theme` is a deep-partial, so the shallow spread overwrites colors/fonts/motion/elevation with partials, and `(): Theme =>` rejects the body for that reason.  
Will fix properly (deep-merge per top-level key) in a follow-up.

PR #4925, opened 2026-05-07
refactor(theme): consolidate DynamicTheme role map and drop DefaultTheme export

adrcotfas:
### Motivation

Before this PR, using dynamic colors required every consumer to wire up the same boilerplate: import `DynamicLightTheme/DynamicDarkTheme`, detect Android API 31+ themselves, write a useMemo to switch between them, and keep it in sync with dark mode. The library provided the building blocks but left the assembly to the app.

This PR absorbs that complexity into `PaperProvider` via a single `dynamicColor` prop. Pass `dynamicColor` and the provider handles palette selection, API level detection, dark/light switching, and graceful fallback on iOS/web -- with no app-side logic needed.

Alongside that, the `DynamicTheme.android.tsx` implementation is refactored from ~500 lines of repetitive pick() calls into a typed role-map table with Palette.* ref constants as ground truth. This fixes two latent bugs in the original: error roles and MCL-only surface roles were calling PlatformColor with hex strings on API 31-33 (invalid, silently transparent), and `surfaceVariant` was missing its API 31 accent resources entirely.

Other changes in scope: `ThemeBase` inlined into `Theme` (kept as a deprecated Pick<> alias), and `isDynamicColorSupported` exported from the library root so consumers can conditionally show a dynamic theme toggle without re-implementing the platform check.

### Related issue

#### Merge order: 

- https://github.com/callstack/react-native-paper/pull/4945

### Test plan

### DIFF
--- a/src/core/PaperProvider.tsx
+++ b/src/core/PaperProvider.tsx
@@ -29,17 +29,15 @@ const PaperProvider = (props: Props) => {
   const resolvedReduceMotion = useResolvedReduceMotion(reduceMotion);
 
   const theme = React.useMemo(() => {
-    const scheme = colorScheme === 'dark' ? 'dark' : 'light';
-    const defaultThemeBase = defaultThemes[scheme];
-    const userScale = props.theme?.animation?.scale ?? 1;
+    const isDark = props.theme?.dark ?? colorScheme === 'dark';
+    const base = defaultThemes[isDark ? 'dark' : 'light'];
+    const scale = resolvedReduceMotion ? 0 : props.theme?.animation?.scale ?? 1;
 
     return {
-      ...defaultThemeBase,
+      ...base,
       ...props.theme,
-      animation: {
-        ...props.theme?.animation,
-        scale: resolvedReduceMotion ? 0 : userScale,
-      },
+      colors: { ...base.colors, ...props.theme?.colors },
+      animation: { ...props.theme?.animation, scale },
     };
   }, [colorScheme, props.theme, resolvedReduceMotion]);
 
--- a/src/core/__tests__/PaperProvider.test.tsx
+++ b/src/core/__tests__/PaperProvider.test.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {
   Appearance,
   AccessibilityInfo,
+  Platform,
   View,
   ColorSchemeName,
 } from 'react-native';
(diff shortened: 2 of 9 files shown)


src/core/PaperProvider.tsx
```diff
@@ -11,34 +11,45 @@ import { useSystemColorScheme } from './useSystemColorScheme';
...
+  darkDynamicColors,
+} from '../theme/schemes/DynamicTheme';
 import type { Theme, ThemeProp } from '../types';
 
 export type Props = {
   children: React.ReactNode;
   theme?: ThemeProp;
   settings?: Settings;
   reduceMotion?: ReduceMotionPreference;
+  dynamicColor?: boolean;
```
satya164: This shouldn't be a prop. We should export dynamic themes like any other theme, which makes it possible for the consumer to customize it, provide their own dynamic theme without using our helpers, and keep the API consistent, same as non-dynamic themes. I don't see a reason to make it work differently.

The merging can be done before exporting and theme can handle fallbacks for older version/other platforms etc. to reduce boilerplate. So the user just needs to import and pass it (and optionally customize it).

src/theme/schemes/DynamicTheme.android.tsx
```diff
@@ -2,501 +2,486 @@ import { Platform, PlatformColor } from 'react-native';
...
-import type { Theme } from '../types';
+import { Palette } from '../tokens';
+import type { Theme, ThemeColors } from '../types';
 
-const isApi34 = (Platform.Version as number) >= 34;
-const isApi31 = (Platform.Version as number) >= 31;
+const apiLevel = Platform.Version as number;
 
 const ac = (name: string) =>
   PlatformColor(`@android:color/${name}`) as unknown as string;
```
satya164, line 11: Not done in this PR but this is a problem. The theme should have colors as `ColorValue`, not `string`. There shouldn't be any type casting.

satya164 (changes requested):

src/theme/schemes/DynamicTheme.android.tsx
adrcotfas: I fixed that in a different PR: https://github.com/callstack/react-native-paper/pull/4945

src/core/PaperProvider.tsx
adrcotfas: Fixed in a separate commit for easy re-review. I'll adapt the commit message to reflect what was changed before merging.

src/core/__tests__/PaperProvider.test.tsx
```diff
@@ -213,6 +213,12 @@ describe('PaperProvider', () => {
     ).toStrictEqual(0);
   });
 
+  it('DynamicLightTheme falls back to LightTheme on unsupported platforms', () => {
+    // On non-Android (and Android API <31), DynamicLightTheme is structurally
+    // identical to LightTheme, so consumers can import it unconditionally.
+    expect(DynamicLightTheme.colors).toStrictEqual(LightTheme.colors);
```
satya164: Let's mock `Platform.OS` explicitly, since we're relying on implicit default of `ios` from react native's jest preset.

satya164 (approved):

PR #4942, opened 2026-05-11
chore: upgrade example app to Expo SDK 54 and align core dependencies

azizbecha:
### Motivation

The example app was pinned to Expo SDK 52. This bumps it to Expo SDK 54 and
aligns the root dev dependencies (`react`, `react-dom`, `react-native`,
`react-native-safe-area-context`, `@react-native/babel-preset`,
`react-test-renderer`) with the versions that SDK 54 ships, so the example app
and the library build/test against a consistent toolchain.

Also includes some cleanup that came up along the way:
- Removed deprecated `app.json` fields (`privacy`, `packagerOpts`).
- Switched `runtimeVersion` from the hardcoded `exposdk:52.0.0` to the
  `{ "policy": "sdkVersion" }` policy.
- Replaced `StyleSheet.absoluteFill` (a style array) with
  `StyleSheet.absoluteFillObject` where it was being spread into a
  `StyleSheet.create` object — spreading the array form is incorrect and the
  object form is what's intended.

### Related issue

N/A

### Test plan

- `yarn lint`, `yarn typescript`, `yarn test` pass.
- Build the example app with Expo SDK 54 (iOS / Android / web) and verify
  ActivityIndicator, BottomNavigation, FABGroup, Menu, Modal and
  TouchableRipple still render/layout correctly.


satya164 (changes requested):
I believe the task was to update to Expo SDK 56 for the v6 branch. Updating the main branch is also fine, but if we're updating, lets update to the current stable version, which is SDK 55 instead of SDK 54.

We also need to delete the `android` folder and use expo prebuilds, and install `expo-dev-client` (for both upgrades). This changes the runtime SDK version, but the actual runtime is still SDK 52, which is in the `android` folder.

The `absoluteFill` -> `absoluteFillObject` change is wrong. `absoluteFill` isn't a style array. It's same as `absoluteFillObject`, and `absoluteFillObject` is removed in the latest React Native versions, so it will break in the latest React Native version.

PR #4217, opened 2023-12-07
feat: add props to FAB, Button and Chip to support prefetching

robinlindhout:
# Summary
The other components like IconButton and ListItem already supports these props as they use Pressable props and spread them.

Fixes: https://github.com/callstack/react-native-paper/issues/3748
Continuation of https://github.com/callstack/react-native-paper/pull/3750

PR #4943, opened 2026-05-12
feat: modernize Button API with label and iconPosition props

### LINKED ISSUE #4928: refactor(button): improve MD3 compliance and modernize content API
Refactor Button to improve Material Design 3 compliance, remove problematic children patterns, improve performance characteristics, and modernize the public API.

**Problems**

**Focus Areas**

- Remove implicit text child handling
- Improve icon + label composition
- Standardize loading and disabled states
- Improve ripple and state layer behavior
- Reduce nested render structure

**Proposed API Direction**

Instead of:

`<Button icon="plus">Create</Button>`

Prefer:

`<Button label="Create" icon="plus" />`

**Performance Opportunities**

- Memoize content container
- Avoid cloning children
- Reduce dynamic style allocations
- Optimize ripple rendering

azizbecha:
### Motivation

`Button` set its label through `children`, which coupled the internal layout to arbitrary child structures, made `uppercase` unreliable (it couldn't transform React elements), and forced an extra `Text` wrapper. Icon placement on the trailing edge relied on the undocumented `contentStyle={{ flexDirection: 'row-reverse' }}` hack, the ripple/state-layer color didn't follow the label color as MD3 specifies, and the render derived a number of style objects on every render.

This PR (part of the v6 work) modernizes the Button API and rendering without removing anything:
- Adds a string `label` prop as the primary way to set the button text. `children` keeps working as a **deprecated** fallback (when both are set, `label` wins) and emits a dev-only warning, so existing code is unaffected.
- Adds an `iconPosition?: 'leading' | 'trailing'` prop. The previous `contentStyle` row-reverse approach still works but is deprecated (dev-only warning) — the icon-margin matrix is extracted into a `getButtonIconStyle` helper.
- Adds a `rippleColor?: ColorValue` prop; by default the ripple / state layer now uses the label color at the pressed-state opacity per MD3 (instead of `TouchableRipple's` `onSurface`-based default), with a graceful fallback when the label color is a `PlatformColor`.
- Wraps the expensive derived values (`getButtonColors`, border-radius extraction, ripple color, icon style, touchable-ripple style, flattened
  styles) in `useMemo`, memoizes the press handlers, and drops the `isMode useCallback`.
- Migrates internal consumers (`Banner`, `Snackbar`, `DataTablePagination`), the example app, and the guide/docs-site snippets to the new props.

Migration: `<Button>Text</Button>` → `<Button label="Text" />`; `contentStyle={{ flexDirection: 'row-reverse' }}` → `iconPosition="trailing"`.

### Related issue

closes #4928 

### Test plan

- `yarn typescript`, `yarn lint`, `yarn test` all pass (snapshots updated where the `iconPosition` restructure changed the rendered tree; reviewed —
  the only diffs are a short-circuit `false` slot in the content style array, the icon-container style going from a 3-element array to an equivalent single object, and `row-reverse` moving into a `styles.contentReverse`).
- New unit tests cover: `label` renders / takes precedence over `children`; deprecated children still renders and warns; `iconPosition="trailing"` and the legacy `contentStyle` fallback (+ warning); `getButtonRippleColor` (custom color, default = label color @ pressed opacity, `PlatformColor` → undefined).
- Manual: open the Button example screen on iOS / Android / web — all five modes, leading vs trailing icon (both `iconPosition` and the legacy `contentStyle` path), loading spinner placement, disabled appearance, compact, custom radius, elevated press-elevation animation, and the ripple color matching the label color.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -31,7 +31,10 @@ import TouchableRipple, {
 } from '../TouchableRipple/TouchableRipple';
 import Text from '../Typography/Text';
 
-export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
+export type Props = $Omit<
+  React.ComponentProps<typeof Surface>,
+  'mode' | 'children'
+> & {
   /**
    * Mode of the button. You can change the mode to adjust the styling to give it desired emphasis.
    * - `text` - flat button without background or outline, used for the lowest priority actions, especially when presenting multiple options.
@@ -74,9 +77,14 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
   /**
    * Label text of the button.
    */
-  children: React.ReactNode;
+  label?: string;
   /**
-   * Make the label text uppercased. Note that this won't work if you pass React elements as children.
+   * @deprecated Use `label` instead. When both `label` and `children` are set, `label` is used.
+   * Label text of the button.
+   */
+  children?: React.ReactNode;
+  /**
+   * Make the label text uppercased.
    */
   uppercase?: boolean;
   /**
@@ -157,9 +165,12 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
  * import { Button } from 'react-native-paper';
  *
  * const MyComponent = () => (
- *   <Button icon="camera" mode="contained" onPress={() => console.log('Pressed')}>
- *     Press me
- *   </Button>
+ *   <Button
+ *     icon="camera"
+ *     mode="contained"
+ *     onPress={() => console.log('Pressed')}
+ *     label="Press me"
+ *   />
  * );
  *
  * export default MyComponent;
@@ -175,6 +186,7 @@ const Button = (
     icon,
     buttonColor: customButtonColor,
     textColor: customTextColor,
+    label,
     children,
     accessibilityLabel,
     accessibilityHint,
@@ -209,6 +221,14 @@ const Button = (
   const { animation } = theme;
   const uppercase = uppercaseProp ?? false;
   const isWeb = Platform.OS === 'web';
+
+  if (process.env.NODE_ENV !== 'production' && children != null) {
+    console.warn(
+      'Button: the `children` prop is deprecated and will be removed in a future release. Use the `label` prop instead.'
+    );
+  }
+
+  const labelContent = label != null ? label : children;
 
   const hasPassedTouchHandler = hasTouchHandler({
     onPress,
@@ -416,7 +436,7 @@ const Button = (
             ]}
             maxFontSizeMultiplier={maxFontSizeMultiplier}
           >
-            {children}
+            {labelContent}
           </Text>
         </View>
       </TouchableRipple>
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -230,6 +230,9 @@ const initialElevation = 1;
 const activeElevation = 2;
 // MD3 leading/trailing icon size for the legacy (no-`size`) button.
 const iconSize = 20;
+// Minimum accessible touch target (dp). Extra-small/small buttons are shorter
+// than this and get expanded via hitSlop.
+const MIN_TOUCH_TARGET = 48;
 
 const Button = (
   {
@@ -448,6 +451,30 @@ const Button = (
     [size]
   );
 
+  // Extra-small/small buttons are shorter than the 48dp minimum accessible
+  // touch target, so expand the press area with hitSlop without changing the
+  // visual size. A user-supplied `hitSlop` wins on the axes it sets.
+  const hitSlopWithMinTarget = React.useMemo(() => {
+    const verticalSlop = sizeStyle
+      ? Math.max(0, (MIN_TOUCH_TARGET - sizeStyle.minHeight) / 2)
+      : 0;
+    if (verticalSlop === 0) {
+      return hitSlop;
+    }
+    if (hitSlop == null) {
+      return { top: verticalSlop, bottom: verticalSlop };
+    }
+    // A numeric hitSlop is an explicit uniform override — respect it as-is.
+    if (typeof hitSlop === 'number') {
+      return hitSlop;
+    }
+    return {
+      ...hitSlop,
+      top: hitSlop.top ?? verticalSlop,
+      bottom: hitSlop.bottom ?? verticalSlop,
+    };
+  }, [hitSlop, sizeStyle]);
+
   const labelTypeStyle = React.useMemo(
     () => ({
       color: labelColor,
@@ -511,7 +538,7 @@ const Button = (
         accessibilityRole={accessibilityRole}
         accessibilityState={{ disabled, selected }}
         accessible={accessible}
-        hitSlop={hitSlop}
+        hitSlop={hitSlopWithMinTarget}
         disabled={disabled}
         style={touchableRippleStyle}
         testID={testID}
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -15,9 +15,11 @@ import {
 
 import {
   ButtonMode,
+  ButtonSize,
   getButtonColors,
   getButtonIconStyle,
   getButtonRippleColor,
+  getButtonSizeStyle,
   getButtonTouchableRippleStyle,
 } from './utils';
 import { useInternalTheme } from '../../core/theming';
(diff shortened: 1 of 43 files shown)


satya164 (changes requested):
> Current API relies heavily on children composition in ways that are difficult to optimize

@ruben-rebelo can you clarify what the optimization problem is?

I'm against changing the API from `children` to `label: string`. It's practically the same except for the type, which limits usage of other components. I don't see the benefit of this restriction, and it's also a big breaking change as `Button` is a common component.

satya164:
To clarify removal of children patterns that we discussed, it's about the usage of `React.Children` API - which breaks composition, not about `children` props.

adrcotfas (changes requested):
Hey @azizbecha, thanks for the PR!  
I've updated https://github.com/callstack/react-native-paper/issues/4928 with a link to a component vs specs review we did internally. You'll find more details about `Button` there.

Some general comments:
- It would be a great moment to revamp the example page for Buttons with a way to toggle all possible states (sizes, configurations, shapes, toggle etc) in a more compact way; now it's huge and messy IMHO. This will also help a lot with testing the modernization.
- please make sure you're following the [specs](https://m3.material.io/components/buttons/specs) in terms of colors, spacing etc; For example,  the outlined variant should use `onSurfaceVariant` as  label color, not primary; 
- use the same naming as in the specs (filled instead of contained, tonal etc)
- default mode is text now but specs hierarchy implies filled as default
- we're using the legacy icon size of 18dp when no size is set: "extra-small" should have the iconSize of 20, not 16.
- we're using the legacy horizontal padding of 24dp instead of 16dp
- it would be nice to extract all "component specific tokens" in the same place for easier access/update if needed;
- use the theme shape tokens instead of magic numbers in `BUTTON_SHAPE_RADIUS`, for example here are the values for the `small` button variant; I can share the other values if you want to; observe that there's a separate field for `pressed`:
```kotlin
// extract from the native Jetpack Compose library
internal object ButtonSmallTokens {
    val ContainerHeight = 40.0.dp
    val ContainerShapeRound = ShapeKeyTokens.CornerFull
    val ContainerShapeSquare = ShapeKeyTokens.CornerMedium
    val IconLabelSpace = 8.0.dp
    val IconSize = 20.0.dp
    val LeadingSpace = 16.0.dp
    val OutlinedOutlineWidth = 1.0.dp
    val PressedContainerShape = ShapeKeyTokens.CornerSmall
    val SelectedContainerShapeRound = ShapeKeyTokens.CornerFull
    val SelectedContainerShapeSquare = ShapeKeyTokens.CornerMedium
    val TrailingSpace = 16.0.dp
}
``` 
- I'm glad you added shapes but we're missing the shape morph animation on press and on toggle; we have motion tokens for that BTW
- drop the `md3` prefix from styles
- make sure we respect "Extra small and small icon buttons must have a target size of 48x48dp or larger to be accessible." as mentioned in the specs
- we should make use of `useLocale()` for toggling the icon position
- consolidate the naming; we're currently using `text`(textColor, textOpacity etc) and `label` to refer to the same thing; Let's use label as in the MD specs
- @satya164 do you think we should look into `hovered` / `focused` states too? something relevant for screen readers, pointing device attached to device 
- It would be nice to consolidate all buttons under src/components/Buttons; See https://m3.material.io/components/all-buttons; Also, I believe we could reuse code between button types as they have a lot in common

satya164:
> @satya164 do you think we should look into hovered / focused states too? something relevant for screen readers, pointing device attached to device

yes. we also support web where this is much more important.

RichardLindhout:
Hopefully children will be kept as this a very breaking change IMHO. Also uglier than the children pattern. 

I have not looked in code yet but can we also expose these props on all touchables of react-native-paper?

https://github.com/callstack/react-native-paper/pull/4217

I need this to support prefetching. As should other navigation libraries when they finally support pre-fetching onPressIn etc.

adrcotfas:
The new playground looks great!

Did a quick visual review before looking into the code:
- the outline button has a different height than the other variants - maybe because of the stroke? this reveals a problem in the "Toggle (expressive)" example for the first button where both width and height differ between states (but look good for the text and tonal buttons)
- Android: in dark mode the outlined and text variants reveal a transparent square behind the text; Web and iOS look good
- Android and iOS: ripple does not follow the shape when morphing; on Web it looks fine
- we could use a way to disable morphing in the playground example - the default expressive buttons don't have the motion (except for the toggle ones).
- Something I didn't find explicit in the specs but I recommend you take a look at how the native morphing expressive buttons behave on Android so we get the native feel: the morphing is not instant on mobile but needs a bit of press (see attached video)
- Android: the Shape toggles in the Playground acts weird because selecting "unset" reuses the previous selected shape type; on Web and iOS it behaves as expected

[Screen_recording_20260521_120607.webm](https://github.com/user-attachments/assets/7210531a-c2a4-415f-906a-673bd7bb32eb)

src/components/Button/Button.tsx
```diff
@@ -67,23 +106,37 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
...
   /**
    * Label text of the button.
    */
-  children: React.ReactNode;
+  label?: string;
+  /**
+   * @deprecated Use `label` instead. When both `label` and `children` are set, `label` is used.
+   * Label text of the button.
+   */
+  children?: React.ReactNode;
```
satya164, line 125: Please revert this API change as per my initial comment. Remove `label` and keep `children` non-deprecated.

src/components/Button/Button.tsx
```diff
@@ -363,60 +623,81 @@ const Button = (
...
+                ? styles.sizedLabel
+                : isMode('text')
                 ? icon || loading
-                  ? styles.md3LabelTextAddons
-                  : styles.md3LabelText
-                : styles.md3Label,
-              compact && styles.compactLabel,
+                  ? styles.legacyLabelTextAddons
+                  : styles.legacyLabelText
+                : styles.legacyLabel,
```
satya164, line 692: what makes these "legacy"?

src/components/Button/Button.tsx
```diff
@@ -233,95 +331,254 @@ const Button = (
...
+    [
+      onPressOut,
+      animateShape,
+      springRadiusTo,
+      mode,
+      animation,
+      elevation,
+      isWeb,
+    ]
+  );
```
satya164, line 444: Use Reanimated for animations as we're moving to Reanimated

src/components/Button/Button.tsx
```diff
@@ -67,23 +106,37 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
...
+  /**
+   * @deprecated Use `label` instead. When both `label` and `children` are set, `label` is used.
+   * Label text of the button.
+   */
+  children?: React.ReactNode;
   /**
-   * Make the label text uppercased. Note that this won't work if you pass React elements as children.
+   * Make the label text uppercased.
    */
   uppercase?: boolean;
```
satya164, line 129: this prop is no longer needed. it was added when react native didn't have `textTransform`. since this is a breaking change, drop this as well

src/components/Button/Button.tsx
```diff
@@ -118,7 +171,10 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
   delayLongPress?: number;
   /**
    * Style of button's inner content.
-   * Use this prop to apply custom height and width, to set a custom padding or to set the icon on the right with `flexDirection: 'row-reverse'`.
+   * Use this prop to apply custom height and width or to set a custom padding.
+   *
+   * Note: setting `flexDirection: 'row-reverse'` here to move the icon to the
+   * trailing edge is deprecated — use the `iconPosition` prop instead.
```
satya164, line 177: No need to be mentioning `flexDirection` here as support is dropped. The first senetence and the note are also contradictory. Just remove mention of `flexDirection` altogether.

src/components/Button/Button.tsx
```diff
@@ -193,23 +268,48 @@ const Button = (
...
+    [contentStyle]
+  );
+  const usesReverseContentStyle =
+    flattenedContentStyle?.flexDirection === 'row-reverse';
+
+  if (process.env.NODE_ENV !== 'production' && usesReverseContentStyle) {
+    console.warn(
+      'Button: setting `flexDirection: \'row-reverse\'` in `contentStyle` to move the icon to the trailing edge is deprecated. Use the `iconPosition="trailing"` prop instead.'
+    );
+  }
```
satya164, line 304: There are many breaking changes in this PR. Just do the same for `row-revers`. No need for deprecation warning for this when there are other things without deprecation.

satya164 (changes requested):
your commit message will be part of the changelog when releasing, so make sure to reword it. keep it concise and keep the description of changes user facing.

focus on describing:
- breaking changes, what's removed, what behavior changed
- migration instructions if possible (in a diff format in codeblock)
- new features, what's added

for the PR title and description, follow conventional commit message format:
- remove `(v6)` from the title. otherwise changelog will create v6 section
- change `refactor` to `feat` - as it's not a plain refactor
- when describing breaking changes in PR description, prefix it with "BREAKING CHANGE:"

PR #4944, opened 2026-05-13
chore(v6): upgrade to Expo SDK 56

azizbecha:
### Motivation

- Example: Expo `~56.0.0-preview.10`, RN `0.85.3`, React `19.2.3`, expo-* aligned via `expo install --fix`.
- Core devDeps aligned to match; added `@react-native/jest-preset` (extracted from RN in 0.85).
- Deleted `example/android` / `example/ios`, gitignored. `expo run:*` regenerates them via prebuild.
- Dropped `@expo/webpack-config`; web served by Metro (config aliases for vector icons + crypto preserved).
- `app.json` cleanup (deprecated `privacy`/`packagerOpts`/`splash`/`newArchEnabled` removed, splash moved into the plugin). `expo-doctor` 18/18.
- Fixed deprecations surfaced by RN 0.85 / React 19.2:
  - `absoluteFillObject` → `absoluteFill`
  - Image `style.tintColor`/`resizeMode` → props
  - `shadow*` → `boxShadow` (token still sourced from `Palette.primary0`)
- Fixed React 19.2 hydration error in `CustomFABControls` (nested `<button>`).
- Fixed pre-existing typo in `DrawerItems` (Dynamic Theme row used a JS comma expression instead of a style array).

### Test plan

  - `yarn typescript`, `yarn lint`, `yarn test` all green (55 suites / 726 tests / 161 snapshots).
  - `cd example && npx expo-doctor` → 18/18 ✓.
  - `cd example && yarn ios`, `yarn android`, `yarn web` all boot.
  - All components render correctly on all platforms.


example/app.json
```diff
@@ -23,27 +21,35 @@
...
-      "backgroundColor": "#6200EE",
-      "resizeMode": "cover"
-    },
     "extra": {
       "eas": {
         "projectId": "ba8f5139-58fe-48f0-b2f2-2d675b6eb2d4"
       }
     },
-    "runtimeVersion": "exposdk:52.0.0",
+    "runtimeVersion": "56.0.0",
```
satya164: Shouldn't it use `sdkVersion`?

example/metro.config.js
```diff
@@ -1,9 +1,40 @@
...
+// Web-only aliases (replaces the dropped @expo/webpack-config setup):
+// - route @react-native-vector-icons/material-design-icons through @expo/vector-icons
+//   so its bundled font file is picked up on web
+// - shim Node's crypto with expo-crypto so libraries expecting it on web work
+const webAliases = {
+  '@react-native-vector-icons/material-design-icons': require.resolve(
+    '@expo/vector-icons/MaterialCommunityIcons'
+  ),
+  crypto: require.resolve('expo-crypto'),
+};
```
satya164: We should be able to use `@react-native-vector-icons/material-design-icons` directly in Expo SDK 56, drop `@expo/vector-icons` and remove such logic.

example/metro.config.js
```diff
@@ -1,9 +1,40 @@
-const { getDefaultConfig } = require('@expo/metro-config');
+const { getDefaultConfig } = require('expo/metro-config');
 const path = require('path');
-const { withMetroConfig } = require('react-native-monorepo-config');
```
satya164, line 3: What's the issue with `react-native-monorepo-config`? Why is it removed?

example/package.json
```diff
@@ -6,48 +6,43 @@
...
     "eas-build-pre-install": "cd .. && yarn",
-    "start": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --go",
-    "android": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --android",
+    "start": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --dev-client",
+    "android": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --android --dev-client",
     "android:native": "EXPO_NO_TYPESCRIPT_SETUP=1 expo run:android",
     "ios:native": "EXPO_NO_TYPESCRIPT_SETUP=1 expo run:ios",
-    "ios": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --ios",
+    "ios": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --ios --dev-client",
     "web": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --web"
```
satya164: I'd remove `android:native` and `ios:native` and replace current `ios` and `android` scripts with `run:` variant. No reason we need separate scripts with `--android` and `--ios`. Also don't need the `web` script. Devs can press w in terminal.

example/package.json
```diff
@@ -6,48 +6,43 @@
...
+    "expo-updates": "~56.0.10",
+    "react": "19.2.3",
+    "react-dom": "19.2.3",
+    "react-native": "0.85.3",
+    "react-native-gesture-handler": "~2.31.1",
+    "react-native-reanimated": "4.3.1",
+    "react-native-safe-area-context": "~5.7.0",
+    "react-native-screens": "4.25.0",
     "react-native-web": "^0.21.0",
-    "react-native-worklets": "0.5.1",
```
satya164: We should keep worklets here for explicitness.

src/components/Appbar/AppbarBackIcon.tsx
```diff
@@ -16,16 +13,15 @@ const AppbarBackIcon = ({ size, color }: { size: number; color: string }) => {
...
     >
       <Image
         source={require('../../assets/back-chevron.png')}
-        style={[
-          styles.icon,
-          { tintColor: color, width: iosIconSize, height: iosIconSize },
-        ]}
+        tintColor={color}
+        resizeMode="contain"
+        style={{ width: iosIconSize, height: iosIconSize }}
```
satya164: These changes seem unnecessary. Are there deprecation warnings?

src/components/Icon.tsx
satya164: Are these changes needed?

src/theme/shadow.tsx
```diff
@@ -1,44 +1,65 @@
...
+    boxShadow: [
+      {
+        offsetX: 0,
+        offsetY: shadowHeight[elevation],
+        blurRadius: shadowRadius[elevation],
+        color: SHADOW_COLOR,
+      },
+    ],
+  };
 }
```
satya164: Don't change anything here. They don't seem necessary for the upgrade and will conflict with other ongoing work.

satya164 (changes requested):
Thanks for the PR. I left some comments. Also, please revert any changes that aren't necessary in the library's code.

example/metro.config.js
```diff
@@ -1,9 +1,40 @@
...
+const upstreamResolveRequest = config.resolver.resolveRequest;
+config.resolver.resolveRequest = (context, moduleName, platform) => {
+  if (platform === 'web' && webAliases[moduleName]) {
+    return context.resolveRequest(context, webAliases[moduleName], platform);
+  }
+  if (upstreamResolveRequest) {
+    return upstreamResolveRequest(context, moduleName, platform);
+  }
+  return context.resolveRequest(context, moduleName, platform);
+};
```
satya164: I'd also add that we shouldn't have such alias logic in metro config. If we need it for our example, then consumers will need it for their apps, which is a problem. It should work without any custom metro code.

The metro config needs to stay minimal and only handle monorepo-specific logic.

example/package.json
azizbecha: It's still there, we just bumped the version to "0.8.3".

src/components/Appbar/AppbarBackIcon.tsx
azizbecha: Yes, React Native 0.85 throws warnings about both fields being on the style object:
- Image: `style.resizeMode` is deprecated. Please use `props.resizeMode`.
- Image: `style.tintColor` is deprecated. Please use `props.tintColor`.
Both fire when this Image renders. The fix here just moves the two values from the style array onto the props.

src/components/Icon.tsx
azizbecha: Same as AppbarIcon, React Native 0.85 throws warnings about both fields being on the style object:

Image: `style.resizeMode` is deprecated. Please use `props.resizeMode`.
Image: `style.tintColor` is deprecated. Please use `props.tintColor`.
So we moved the values from style array onto the props.

src/theme/shadow.tsx
azizbecha: Same, React Native 0.85 warnings about replacing `shadow*` with `boxShadow`.

example/metro.config.js
azizbecha: `react-native-monorepo-config@0.1.6`'s `withMetroConfig` overwrites `watchFolders` instead of appending, so `expo-doctor `flags it:

Screenshot 2026-05-14 at 2 00 05 PM

example/metro.config.js
azizbecha: Done!

example/metro.config.js
satya164: You can update it back in the config by modifying the return value. It shouldn't really be an actual problem tho given that the `watchFolders` is root of the repo.

satya164 (approved):
thank you

PR #4945, opened 2026-05-14
refactor: replace string with ColorValue for ThemeColors

adrcotfas:
### Motivation

Mentioned here: https://github.com/callstack/react-native-paper/pull/4925#discussion_r3235341180
### Related issue
https://github.com/callstack/react-native-paper/pull/4925

### Test plan


example/src/Examples/BottomNavigationBarExample.tsx
```diff
@@ -53,7 +53,7 @@ export default function BottomNavigationBarExample() {
           renderIcon={({ route, focused, color }) =>
             descriptors[route.key].options.tabBarIcon?.({
               focused,
-              color,
+              color: typeof color === 'string' ? color : '',
```
satya164: updated react navigation to latest alpha so `ColorValue` will work https://github.com/callstack/react-native-paper/pull/4948

ps: in future throw an error for such scenarios instead of using empty string, so any problems due to future changes are not suppressed

src/components/Chip/helpers.tsx
```diff
@@ -38,7 +38,11 @@ const getBorderColor = ({
   }
 
   if (isSelectedColor) {
-    return color(selectedColor).alpha(0.29).rgb().string();
+    if (typeof selectedColor === 'string') {
+      return color(selectedColor).alpha(0.29).rgb().string();
```
satya164, line 42: we'll need to tweak how we handle color operations so we also support things like `var(` (for web), which are also strings. could create a utility function that handles these properly.

satya164 (approved):

src/components/Chip/helpers.tsx
adrcotfas: This part is legacy. We'll anyway modernize each component and this old way of applying alpha will be replaced by using theme color tokens.  
I'm not sure about web but I assume that we'll pass the colors in the theme and not in components.  

Since you approved this PR I assume you're not requesting an utility function here and now.

example/src/Examples/BottomNavigationBarExample.tsx
adrcotfas: I see you modified this part in that PR.  
I assume I can merge this first since you approved it.

PR #4952, opened 2026-05-18
refactor: improve structure of reference and system tokens

adrcotfas:
### Motivation

Foundation prep for the Switch component modernization.

  1. Add focus-indicator tokens at `md.sys.state.focusIndicator` (thickness: 3, outerOffset: 2). Matches MD3 spec. Switch will be the first consumer.
  2. Move misplaced tokens from `ref` to `sys` per the MD3 spec:
    - `ref.stateOpacity` -> `sys.state.opacity`
    - `ref.scrimAlpha` -> `sys.scrim.alpha`
  3. Split tokens/index.ts (was 364 LOC, mixed ref values + sys decisions + barrel).
  4. ~~add `useFocusVisible` hook: required to detect when to show a focus ring~~
  Also in this PR:

  - New helper `getStateLayer(theme, role, state)` in theme/utils/state.ts. Returns { color, opacity }. Four components migrated: HelperText, TextInput/Adornment, CheckboxItem, RadioButtonItem.
  - New types `ColorRole = keyof ThemeColors and StateOpacityKey = keyof typeof tokens.md.sys.state.opacity`. Component token files can use them to validate string keys.
  - theme/types/utils.ts moved to src/types.tsx. The helpers there ($Omit, $RemoveChildren, EllipsizeProp) are not theme-specific.

No change to Theme. Renames are internal only.

### Related issue

  N/A. Prep for Switch modernization.

### Test plan
  
  - yarn typescript clean
  - yarn lint clean
  - yarn test green, snapshots stable


src/utils/useFocusVisible.ts
```diff
@@ -0,0 +1,55 @@
...
+  return keyboardActive;
+}
+
+/**
+ * Convenience hook for components that gate a focus indicator on
+ * `:focus-visible` semantics. Returns `{ focusVisible, onFocus, onBlur }`;
+ * wire `onFocus` / `onBlur` to a `Pressable` (or equivalent) and gate the
+ * indicator on `focusVisible`. `focusVisible` is `true` only while the
+ * element is focused AND the focus came from keyboard input.
+ */
```
satya164: Why not use `:focus-visible` in CSS directly?

adrcotfas: You're right - too much over-engineering.  
You can see it in action here: #4957 and it will be reused in other similar components requiring focus rings.

satya164: We should use plain CSS for anything that can be done in plain CSS, without JS code that also needs to keep state and trigger re-renders. I left a comment on the Switch PR on how you can use CSS directly.

satya164 (changes requested):

src/utils/useFocusVisible.ts
adrcotfas: I removed this file.

satya164 (approved):

PR #4957, opened 2026-05-20
feat: modernize Switch to MD3

### LINKED ISSUE #4892: fix(iOS): Enforce Material Design Consistency
At least the Checkbox and Switch components currently adopt native iOS styling on Apple devices. While platform-native UI is helpful for some, it creates visual inconsistency for projects aiming for a strict Material Design, which is part of the library's description:
> React Native Paper is the cross-platform UI kit library containing a collection of customizable and production-ready components, which by default are following and respecting the Google’s Material Design guidelines.

Respect the MD guidelines on all platforms by using the same style or at least provide a configurable style.

**Additional context**
Following the Material 3 guidelines, the design language is intended to be a unified system. For example, Google’s own web and mobile implementations maintain 1:1 parity in their selection controls, the same with Compose Multiplatform which looks the same on all platforms.

### LINKED ISSUE #4939: refactor(switch): align interaction model with MD3 guidelines
Refactor Switch to improve Material Design 3 compliance, remove problematic children patterns, improve performance characteristics, and modernize the public API.

**Problems**

**Focus Areas**

- Improve thumb and track animations
- Improve state layer behavior
- Improve disabled and pressed states
- Standardize sizing and spacing

**Performance Opportunities**

- Reduce animation allocations
- Improve controlled state rendering

adrcotfas:
### Motivation

Update the Switch to use the latest MD specs, including motion and RTL. 

### Related issue

Closes #4939
Partially closes #4892

### Test plan

- lint, existing tests and new tests pass
- manual testing on Android, iOS and web in all states, including hovered and focused

### Demo

[Android](https://github.com/user-attachments/assets/a2f44e95-8b92-4cd9-91ed-04a3477d5262)

[iOS](https://github.com/user-attachments/assets/9ef0471f-29d4-4b0b-a6c7-aefa20553d76)

### DIFF
--- a/src/components/Switch/Switch.tsx
+++ b/src/components/Switch/Switch.tsx
@@ -1,108 +1,540 @@
 import * as React from 'react';
 import {
-  NativeModules,
   Platform,
-  StyleProp,
-  Switch as NativeSwitch,
-  ViewStyle,
+  Pressable,
+  StyleSheet,
+  type StyleProp,
+  View,
+  type ViewStyle,
 } from 'react-native';
 
-import { getSwitchColor } from './utils';
+import Animated, {
+  Easing,
+  ReduceMotion,
+  useAnimatedReaction,
+  useAnimatedStyle,
+  useDerivedValue,
+  useSharedValue,
+  withDelay,
+  withSequence,
+  withSpring,
+  withTiming,
+} from 'react-native-reanimated';
+
+import { SwitchTokens } from './tokens';
+import { getDefaultSwitchColors, resolveSwitchPaint } from './utils';
+import { useLocale } from '../../core/locale';
 import { useInternalTheme } from '../../core/theming';
-import type { ThemeProp } from '../../types';
+import { useReduceMotion } from '../../theme/accessibility/ReduceMotionContext';
+import { tokens } from '../../theme/tokens';
+import { toRawSpring } from '../../theme/tokens/sys/motion';
+import { cornerFull } from '../../theme/tokens/sys/shape';
+import type { StateOpacityKey, ThemeProp } from '../../types';
+import { isKeyboardFocusEvent } from '../../utils/isKeyboardFocusEvent';
+import Icon, { type IconSource } from '../Icon';
 
-const version = NativeModules.PlatformConstants
-  ? NativeModules.PlatformConstants.reactNativeVersion
-  : undefined;
-
-export type Props = React.ComponentPropsWithRef<typeof NativeSwitch> & {
+export type Props = {
   /**
-   * Disable toggling the switch.
-   */
-  disabled?: boolean;
-  /**
-   * Value of the switch, true means 'on', false means 'off'.
+   * Whether the switch is on.
    */
   value?: boolean;
   /**
-   * Custom color for switch.
+   * Called with the new value when the user toggles the switch.
    */
-  color?: string;
+  onValueChange?: (value: boolean) => void;
   /**
-   * Callback called with the new value when it changes.
+   * Disables interaction and renders the disabled visual state.
    */
-  onValueChange?: Function;
+  disabled?: boolean;
+  /**
+   * Icon shown inside the handle when checked
+   */
+  checkedIcon?: IconSource;
+  /**
+   * Icon shown inside the handle when unchecked.
+   */
+  uncheckedIcon?: IconSource;
   style?: StyleProp<ViewStyle>;
-  /**
-   * @optional
-   */
+  testID?: string;
   theme?: ThemeProp;
+  accessibilityLabel?: string;
 };
 
+const {
+  trackWidth: TRACK_WIDTH,
+  trackHeight: TRACK_HEIGHT,
+  trackOutlineWidth: TRACK_OUTLINE_WIDTH,
+  stateLayerSize: STATE_LAYER_SIZE,
+  selectedHandleSize: SELECTED_HANDLE,
+  unselectedHandleSize: UNSELECTED_HANDLE,
+  iconHandleSize: ICON_HANDLE,
+  pressedHandleSize: PRESSED_HANDLE,
+  selectedIconSize: SELECTED_ICON,
+  unselectedIconSize: UNSELECTED_ICON,
+  disabledSelectedHandleOpacity: DISABLED_SELECTED_HANDLE_OPACITY,
+  disabledUnselectedHandleOpacity: DISABLED_UNSELECTED_HANDLE_OPACITY,
+  disabledSelectedIconOpacity: DISABLED_SELECTED_ICON_OPACITY,
+  disabledUnselectedIconOpacity: DISABLED_UNSELECTED_ICON_OPACITY,
+  disabledTrackOpacity: DISABLED_TRACK_OPACITY,
+} = SwitchTokens;
+
+const { state: stateTokens } = tokens.md.sys;
+const stateOpacity = stateTokens.opacity;
+const { thickness: FOCUS_THICKNESS, outerOffset: FOCUS_OUTER_OFFSET } =
+  stateTokens.focusIndicator;
+const FOCUS_RING_INSET = -(FOCUS_OUTER_OFFSET + FOCUS_THICKNESS);
+const OVERLAY_TOP = (STATE_LAYER_SIZE - TRACK_HEIGHT) / 2;
+
+// Hold-then-grow: a brief delay before snapping to PRESSED_HANDLE so a quick
+// tap doesn't flash the press-grow visual.
+const PRESS_GROW_DELAY = 100;
+
+function restingHandleSize(checked: boolean, hasIcon: boolean): number {
+  if (hasIcon) return ICON_HANDLE;
+  return checked ? SELECTED_HANDLE : UNSELECTED_HANDLE;
+}
+
+const HANDLE_PADDING = (TRACK_HEIGHT - SELECTED_HANDLE) / 2;
+const UNCHECKED_CENTER = TRACK_HEIGHT / 2;
+const CHECKED_CENTER = TRACK_WIDTH - HANDLE_PADDING - SELECTED_HANDLE / 2;
+
 /**
- * Switch is a visual toggle between two mutually exclusive states — on and off.
+ * Material 3 toggle between two mutually exclusive states (on / off).
  *
  * ## Usage
  * ```js
  * import * as React from 'react';
  * import { Switch } from 'react-native-paper';
  *
- * const MyComponent = () => {
- *   const [isSwitchOn, setIsSwitchOn] = React.useState(false);
- *
- *   const onToggleSwitch = () => setIsSwitchOn(!isSwitchOn);
- *
- *   return <Switch value={isSwitchOn} onValueChange={onToggleSwitch} />;
+ * const Example = () => {
+ *   const [on, setOn] = React.useState(false);
+ *   return <Switch value={on} onValueChange={setOn} />;
  * };
- *
- * export default MyComponent;
  * ```
+ *
+ * ## Theming
+ * Customize by overriding these `theme.colors` roles:
+ * - `primary` / `onPrimary`: selected track + icon / selected handle
+ * - `primaryContainer`: selected handle on hover, press
+ * - `surfaceContainerHighest`: unselected track + icon
+ * - `outline`: unselected resting handle, unselected track outline
+ * - `onSurfaceVariant`: unselected handle on hover, press
+ * - `onSurface`: disabled track, handle, and icon fills
+ * - `surface`: disabled selected handle
+ * - `secondary`: focus indicator
  */
 const Switch = ({
   value,
   disabled,
   onValueChange,
-  color,
+  checkedIcon,
+  uncheckedIcon,
+  style,
+  testID,
   theme: themeOverrides,
-  ...rest
+  accessibilityLabel,
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
-  const { checkedColor, onTintColor, thumbTintColor } = getSwitchColor({
-    theme,
-    disabled,
-    value,
-    color,
-  });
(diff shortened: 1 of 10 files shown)


src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,492 @@
...
+  /** Per-slot color overrides. Unspecified keys fall back to theme defaults. */
+  colors?: Partial<SwitchColors>;
   style?: StyleProp<ViewStyle>;
-  /**
-   * @optional
-   */
+  testID?: string;
   theme?: ThemeProp;
+  accessibilityLabel?: string;
 };
```
satya164, line 61: Please use the multiline comment pattern already used in the codebase. Don't collapse them to a single line.

src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,492 @@
...
+  /** Called with the new value when the user toggles the switch. */
+  onValueChange?: (value: boolean) => void;
+  /** Disables interaction and renders the disabled visual state. */
+  disabled?: boolean;
+  /** Icon shown inside the handle when checked */
+  checkedIcon?: IconSource;
+  /** Icon shown inside the handle when unchecked. */
+  uncheckedIcon?: IconSource;
+  /** Per-slot color overrides. Unspecified keys fall back to theme defaults. */
+  colors?: Partial<SwitchColors>;
```
satya164: I don't think we need this prop, given that all components already receive `theme` as a prop. It's 2 ways to achieve the same thing.

src/index.tsx
```diff
@@ -126,7 +126,10 @@ export type { Props as RadioButtonItemProps } from './components/RadioButton/Rad
 export type { Props as SearchbarProps } from './components/Searchbar';
 export type { Props as SnackbarProps } from './components/Snackbar';
 export type { Props as SurfaceProps } from './components/Surface';
-export type { Props as SwitchProps } from './components/Switch/Switch';
+export type {
+  Props as SwitchProps,
+  SwitchColors,
```
satya164: I don't think we need this exposed

src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,492 @@
...
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+        rippleColor="transparent"
+        accessibilityRole="switch"
+        accessibilityState={{ disabled: isDisabled, checked }}
+        accessibilityLiveRegion="polite"
+        accessibilityLabel={accessibilityLabel}
+        testID={testID}
+        style={styles.touchable}
+        theme={theme}
```
satya164: This can just be a `Pressable`. No ripple is visible, so the overhead of `TouchableRipple` is unnecessary.

src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,492 @@
...
+              borderWidth: FOCUS_THICKNESS,
+              top: OVERLAY_TOP + FOCUS_RING_INSET,
+              left: FOCUS_RING_INSET,
+              right: FOCUS_RING_INSET,
+              bottom: OVERLAY_TOP + FOCUS_RING_INSET,
+              borderRadius: cornerFull,
+            },
+          ]}
+        />
+      ) : null}
```
satya164: Why isn't this just done in CSS instead of JS? We shouldn't be reimplementing CSS features in JS.

Add a class name to the wrapper and embed a style tag:

```js
const CLASS_NAME = 'react-native-paper-switch'

const CSS_TEXT = /* css */ `
  .${CLASS_NAME}:focus-visible {
    /* focus styling */
  }
`

// ...

<style href={CLASS_NAME} precedence="react-native-paper">
  {CSS_TEXT}
</style>
```

With `href` and `precedence`, React will deduplicate styles across the tree making sure only single instance is present. Abstract `style` to a component to reuse it that renders a `style` on web and nothing on native.

src/components/Switch/utils.ts
```diff
@@ -1,106 +1,154 @@
...
+    disabledUncheckedIconColor: c[t.disabledUnselectedIconColor],
+  };
+}
+
+export function resolveSwitchColors(
+  theme: InternalTheme,
+  overrides: Partial<SwitchColors> | undefined
+): SwitchColors {
+  return { ...getDefaultSwitchColors(theme), ...overrides };
+}
```
satya164: Remove such unnecessary abstractions. It's used only once. That function can just call `{ ...getDefaultSwitchColors(theme), ...overrides }` instead of abstracting it out.

src/components/__tests__/Switch.test.tsx
```diff
@@ -1,224 +1,121 @@
...
-      onTintColor: 'rgb(178, 175, 177)',
-    });
+  it('resolves all color roles against the dark theme palette', () => {
+    const colors = getDefaultSwitchColors(DarkTheme);
+    expect(colors.checkedTrackColor).toBe(DarkTheme.colors.primary);
+    expect(colors.checkedHandleColor).toBe(DarkTheme.colors.onPrimary);
+    expect(colors.uncheckedTrackColor).toBe(
+      DarkTheme.colors.surfaceContainerHighest
+    );
   });
```
satya164: We don't need tests for such straightforward logic. They will be covered by the snapshot tests for the switch itself.

Tests should test the public API and the publicly visible output. Not internal implementation details, unless the implementation detail is complex enough to need a test. Otherwise, changing any implementation detail means also changing tests, which is wasteful and don't really help with confidence.

satya164 (changes requested):

src/components/Switch/Switch.tsx
adrcotfas: Previously we had a `color` prop and this is what replaces it for custom colors.  
I'm fine with removing it but having `colors` is more developer friendly since we're using explicit switch specific naming instead of having to guess what theme color token corresponds to which component token.  

As a developer using this lib, I would not use this unless I want custom colors for the switch and in general I would not pass non-default/global theme to components.

What do you think? 

Edit: I would prefer to remove the `theme` prop and use `useInternalTheme()` instead for the other tokens we need

src/index.tsx
adrcotfas: https://github.com/callstack/react-native-paper/pull/4957#discussion_r3296476887

src/components/Switch/Switch.tsx
satya164: @adrcotfas a `color` prop for single color feels different than a palette - which is also provided by theme. i'd remove it; it doesn't seem necessary.

> I would prefer to remove the theme prop

i don't have strong opinion to keep this from the API point of view, but it has been exposed for a long time. so to avoid unnecessary breaking changes, I would keep it, unless it's actually causing us problems or maintenance issues - which it shouldn't be per my understanding.

src/components/Switch/Switch.tsx
adrcotfas: Done

src/components/Switch/Switch.tsx
adrcotfas: I removed `colors` prop and added documentation on how to use theme to customize.

adrcotfas:
@satya164 Findings were fixed. I discovered an issue with the dynamic theme when testing custom switch colors and applied a fix as a separate commit.

src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,491 @@
...
+
+  const checked = !!value;
+  const isDisabled = !!disabled;
+  const isEnabled = !isDisabled;
+  const iconSource = checked ? checkedIcon : uncheckedIcon;
+  const hasIcon = iconSource !== undefined;
+
+  const [hovered, setHovered] = React.useState(false);
+  const [pressed, setPressed] = React.useState(false);
+  const [focused, setFocused] = React.useState(false);
```
satya164: make these shared values. ideally, we should have event -> animation rather than event -> state update -> effect -> animation. user interactions should be as responsive as possible, even if JS thread is busy.

src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,491 @@
...
+            borderColor: colors.focusIndicatorColor,
+            borderWidth: FOCUS_THICKNESS,
+            top: OVERLAY_TOP + FOCUS_RING_INSET,
+            left: FOCUS_RING_INSET,
+            right: FOCUS_RING_INSET,
+            bottom: OVERLAY_TOP + FOCUS_RING_INSET,
+            borderRadius: cornerFull,
+          },
+        ]}
+      />
```
satya164: i'd prefer to do this with CSS instead of react state. it'd be a relatively similar amount of code to write these styles inside `<style>`, and even less code with a reusable utility. is there any benefit to doing this with react state?

we also need to hide the outline from the default styles, as both of them are visible:

Image

src/components/Switch/SwitchStyle.web.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { Pressable, type PressableProps } from 'react-native';
+
+export function SwitchPressable({ onFocus, ...props }: PressableProps) {
+  const handleFocus: PressableProps['onFocus'] = (e) => {
+    if (document.activeElement?.matches(':focus-visible')) {
+      onFocus?.(e);
+    }
+  };
```
satya164: it seems a bit too much abstraction for this small component. the web code could be just a conditional with `Platform.OS === 'web'` check. in the same component instead of 2 platform specific files, and for such small logic, the whole thing could be inlined and it'd be less code.

src/components/Switch/SwitchStyle.web.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
```
satya164: don't use `.web` extension. metro supports `.native` by default. web bundlers don't support `.web` by default. even though most configs configure it, it also becomes problematic when more tools come into picture such as importing on node (e.g. for SSR).

use `.native` extension whenever a native/web split is needed.

src/components/Switch/SwitchFocusRing.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export function SwitchFocusRing({
+  focused,
+  style,
+}: {
+  focused: boolean;
+  style: StyleProp<ViewStyle>;
+}) {
+  if (!focused) return null;
+  return <View pointerEvents="none" style={style} />;
```
satya164: all of the styling logic is in `Switch`, so a separate component for just a `View` with `pointerEvents="none"` seems a bit unnecessary abstraction.

also `pointerEvents` prop warns on web. check the console warnings. it needs to be astyle.

satya164 (changes requested):
Quickly clicking on the Switch feels bad, probably because the Switch grows and changes position unanimated before the animation is visible

https://github.com/user-attachments/assets/f615f034-850c-46c6-8295-3d1acb76a843

RTL seems to be broken on Web

https://github.com/user-attachments/assets/d733cdfc-cedb-4edd-8ab4-1939544fe7ec

I've merged some updates to `main` which should let you toggle RTL on Web for testing

src/components/Switch/Switch.tsx
adrcotfas: Done

adrcotfas:
> Quickly clicking on the Switch feels bad, probably because the Switch grows and changes position unanimated before the animation is visible

It should be fine now.
 
> RTL seems to be broken on Web
> 

Fixed

src/components/Switch/SwitchFocusRing.tsx
adrcotfas: Done

src/components/Switch/SwitchStyle.web.tsx
adrcotfas: Done

src/components/Switch/SwitchStyle.web.tsx
adrcotfas: Done

satya164 (approved):

src/theme/provider.tsx
```diff
@@ -17,9 +18,43 @@ export function useTheme<T = Theme>(overrides?: $DeepPartial<T>) {
...
+  }
+  const out: Record<string, unknown> = { ...(base as Record<string, unknown>) };
+  for (const key of Object.keys(overrides as Record<string, unknown>)) {
+    out[key] = safeMerge(
+      (base as Record<string, unknown>)[key],
+      (overrides as Record<string, unknown>)[key]
+    );
+  }
+  return out as T;
+};
```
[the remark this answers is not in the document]
adrcotfas: Documented in the latest changes.

src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,546 @@
...
+  const xSign = direction === 'ltr' ? 1 : -1;
+
+  const checked = !!value;
+  const isDisabled = !!disabled;
+  const isEnabled = !isDisabled;
+  const iconSource = checked ? checkedIcon : uncheckedIcon;
+  const hasIcon = iconSource !== undefined;
+
+  const anchorStyle: ViewStyle =
+    Platform.OS === 'web' && direction === 'rtl' ? { right: 0 } : { left: 0 };
```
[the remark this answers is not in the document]
adrcotfas: Behaviour was verified and is fine.

src/components/Switch/Switch.tsx
```diff
@@ -1,108 +1,546 @@
...
+        onHoverOut={() => {
+          hoveredSV.value = 0;
+        }}
+        onFocus={() => {
+          // Only react to keyboard focus on web; ignore focus from a mouse click.
+          if (
+            Platform.OS === 'web' &&
+            !document.activeElement?.matches(':focus-visible')
+          ) {
+            return;
```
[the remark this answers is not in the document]
adrcotfas: I've extracted this to a reusable utility.

PR #4962, opened 2026-05-22
feat(checkbox): modernize for md3 spec compliance

fabriziocucci:
## Summary

Stacked on **#4952** (token reorg + `useFocusVisible` hook).

> **Reviewers:** because #4952 hasn't merged yet, the diff against `main` here also includes the 4 foundation commits from that PR (token reorg, shadow overrides, `useFocusVisible`, motion mass). **The actual Checkbox change is just the final commit: [`9ec676c`](https://github.com/callstack/react-native-paper/pull/4962/commits/9ec676c).** When #4952 lands, this PR rebases and the diff shrinks accordingly.

Bring the `Checkbox` component up to the [Material Design 3 spec](https://m3.material.io/components/checkbox/specs), end-to-end:

- 18dp container with 2dp outline (unselected) → 0dp outline + theme primary fill (selected), inside a 40dp state-layer tap target.
- State-layer overlay renders hover (8%), focus (10%) and pressed (10%) layers in the color the spec defines for each: selected → primary, unselected → onSurface, pressed-selected flips to onSurface, error always wins.
- Focus indicator: 3dp ring at `theme.colors.secondary` with the 2dp outer-offset from `md.sys.state.focusIndicator`. Gated on `:focus-visible` via the `useFocusVisible` hook from #4952.
- Animations approximate `androidx.compose.material3.Checkbox.kt`: 100ms fill transition + 150ms checkmark draw. The checkmark is sequenced short-leg then long-leg via `scaleY` to suggest the stroke fraction. Indeterminate uses a `scaleX`-animated dash.

## Why not SVG

Compose Material3 draws the checkmark as a `Path` with `strokeFraction` 0 → 1. Doing the same in RN would mean adding `react-native-svg` as a Paper peer-dep, an ecosystem-level change. This PR uses a reveal-mask instead: a static L-shape (borderLeftWidth + borderBottomWidth rotated -45°) inside a left-anchored View whose width animates 0 → 18dp with a matching opacity fade. The visual suggests the stroke drawing left-to-right without the SVG dependency. Same precedent will apply to RadioButton when it's modernized. If you'd prefer the SVG-based approach for v6, happy to switch. Picking once now sets the precedent.

## Files

- `src/components/Checkbox/Checkbox.tsx`: full rewrite (`Pressable` 40dp tap target, state-layer overlay, animated 18dp container, focus ring, view-based checkmark + dash).
- `src/components/Checkbox/utils.ts`: new `getSelectionVisualState` returns the full color + opacity + outline-width picture for any `(selected × hovered × focused × pressed × disabled × error × customColor)` combination. Legacy `getSelectionControlColor` kept as a back-compat export for `RadioButtonAndroid`.
- 9 snapshots auto-updated to reflect the new render tree.

## Out of scope

- RadioButton modernization (separate PR with the same pattern).
- Switch modernization (already in flight from #4952 onward).
- Migrating animations from `Animated` to `react-native-reanimated`.
- Extracting `<StateLayer />` as a shared primitive (can follow once Radio + Switch land and the pattern is proven).

## Test plan

- `yarn typescript` clean
- `yarn lint` clean
- `yarn jest`: 736/737 (1 skipped, pre-existing); 159/159 snapshots pass (9 updated for the new renderer).

## Visuals

| Android | iOS |
|---|---|
| checkbox-md3-android | checkbox-md3-ios |


src/utils/useFocusVisible.ts
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { Platform } from 'react-native';
+
+const isWeb = Platform.OS === 'web';
+
+type FocusVisibleEvent = {
+  currentTarget: object;
+};
```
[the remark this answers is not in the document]
fabriziocucci: This file was introduced by the foundation PR #4952 that this branch is stacked on. The type issue will be resolved when #4952 merges (or in a follow-up on that PR). Leaving this file untouched in the Checkbox PR to avoid modifying the shared foundation.

src/components/__tests__/Checkbox/utils.test.tsx
```diff
@@ -1,7 +1,8 @@
 import { getTheme } from '../../../core/theming';
 import { tokens } from '../../../theme/tokens';
 import { getSelectionControlColor } from '../../Checkbox/utils';
-const { stateOpacity } = tokens.md.ref;
+
+const stateOpacity = tokens.md.sys.state.opacity;
```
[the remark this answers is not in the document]
fabriziocucci: The helper is covered by the snapshot tests in `Checkbox.test.tsx` which exercise the full state matrix (unchecked/checked/indeterminate × disabled/error × hover/focus/pressed). The snapshots assert the exact RGBA values produced for each combination, so the helper's decision logic is verified. Dedicated unit tests for the helper would be valuable but aren't blocking for this PR — the snapshots already guard against regressions. Can add if you prefer.

satya164:
If you're stacking on PR against another, then please open your PR against the branch you're stacking instead of main, so it only includes your changes.

fabriziocucci:
> If you're stacking on PR against another, then please open your PR against the branch you're stacking instead of main, so it only includes your changes.

Thanks @satya164. Unfortunately targeting main here was intentional because @adrcotfas/tokens_structure (the head of #4952) is currently based on e860605 and doesn't include 6d92d77 (refactor(checkbox): unify platform-specific files), which is now on main and which this PR's modernization sits on top of. Rebasing onto #4952's branch today produces multiple conflicts.

Happy to wait for #4952 to land on main and then rebase this PR on top, at that point the diff will naturally shrink to just the Checkbox change.

adrcotfas:
> > If you're stacking on PR against another, then please open your PR against the branch you're stacking instead of main, so it only includes your changes.
> 
> Thanks @satya164. Unfortunately targeting main here was intentional because @adrcotfas/tokens_structure (the head of #4952) is currently based on [e860605](https://github.com/callstack/react-native-paper/commit/e8606051954a109be7ba30fb693fa3e2e550fc11) and doesn't include [6d92d77](https://github.com/callstack/react-native-paper/commit/6d92d77598b61e9d257952c0d3caa24513afebc8) (refactor(checkbox): unify platform-specific files), which is now on main and which this PR's modernization sits on top of. Rebasing onto #4952's branch today produces multiple conflicts.
> 
> Happy to wait for #4952 to land on main and then rebase this PR on top, at that point the diff will naturally shrink to just the Checkbox change.

I rebased @adrcotfas/tokens_structure on the latest main.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +93,261 @@ const Checkbox = ({
...
+  const [pressed, setPressed] = React.useState(false);
+
+  const selected = status === 'checked' || status === 'indeterminate';
 
   const {
     animation: { scale },
   } = theme;
 
+  // 0 = unselected (outline only), 1 = selected (filled + drawn icon).
+  const fillAnim = React.useRef(new Animated.Value(selected ? 1 : 0)).current;
```
adrcotfas: Use `useAnimatedValue` instead to avoid allocating a new `Animated.Value` on every render. See `src/utils/useAnimatedValue.tsx`.

src/components/Checkbox/utils.ts
```diff
@@ -3,71 +3,220 @@ import type { ColorValue } from 'react-native';
...
+      color: error
+        ? theme.colors.error
+        : selected
+        ? theme.colors.primary
+        : theme.colors.onSurface,
+      opacity: stateOpacity.hovered,
+    };
+  }
+  return { color: TRANSPARENT, opacity: 0 };
 };
```
adrcotfas, line 112: This shadows the `getStateLayer` from `src/theme/utils/state.ts`. We could reuse that here to reduce the code.

Something like this maybe:
```suggestion
const resolveStateLayer = ({
  theme,
  selected,
  hovered,
  focused,
  pressed,
  error,
}: Omit<SelectionState, 'customColor' | 'customUncheckedColor' | 'disabled'>): {
  color: ColorValue;
  opacity: number;
} => {
  const state = pressed
    ? 'pressed'
    : focused
    ? 'focused'
    : hovered
    ? 'hovered'
    : null;
  if (!state) return { color: TRANSPARENT, opacity: 0 };

  const role: ColorRole = error
    ? 'error'
    : (pressed ? !selected : selected)
    ? 'primary'
    : 'onSurface';
  return getStateLayer(theme, role, state);
};
```

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +93,261 @@ const Checkbox = ({
...
+ * animation without an SVG dependency.
+ */
+const Checkmark = ({
+  color,
+  progress,
+}: {
+  color: ColorValue;
+  progress: Animated.Value;
+}) => {
+  const maskWidth = progress.interpolate({
```
adrcotfas, line 236: We have duplicate code here to lines 216-237.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +93,261 @@ const Checkbox = ({
...
   );
 };
 
 const styles = StyleSheet.create({
-  container: {
-    borderRadius: 18,
-    width: 36,
-    height: 36,
-    padding: 6,
+  tapTarget: {
```
adrcotfas, line 250: This is identical to `tapTargetInner`.

src/components/Checkbox/utils.ts
```diff
@@ -3,71 +3,220 @@ import type { ColorValue } from 'react-native';
...
+  customUncheckedColor?: ColorValue;
+};
+
+type SelectionState = StaticState & {
+  hovered?: boolean;
+  focused?: boolean;
+  pressed?: boolean;
+};
+
+export type SelectionVisualState = {
```
adrcotfas: No need to export this.

src/components/Checkbox/utils.ts
```diff
@@ -3,71 +3,220 @@ import type { ColorValue } from 'react-native';
...
+ */
 
-const getCheckedColor = ({
-  theme,
-  customColor,
-  error,
-}: {
+const stateOpacity = tokens.md.sys.state.opacity;
+
+const TRANSPARENT: ColorValue = 'transparent';
```
adrcotfas: This could be inlined.

src/components/Checkbox/utils.ts
```diff
@@ -3,71 +3,220 @@ import type { ColorValue } from 'react-native';
 import { tokens } from '../../theme/tokens';
 import type { InternalTheme } from '../../types';
 
-const { stateOpacity } = tokens.md.ref;
+/*
```
adrcotfas: This is much too verbose. We don't need this level of detail and please don't add magic numbers in documentation which can drift.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -51,7 +52,15 @@ export type Props = $RemoveChildren<typeof TouchableRipple> & {
   testID?: string;
 };
 
-const ANIMATION_DURATION = 100;
+// Spec dimensions (https://m3.material.io/components/checkbox/specs).
+const CONTAINER_SIZE = 18;
```
adrcotfas: Please consider adding a tokens file that holds the component specific tokens for a better separation of concerns and easier updating.  

Maybe you'll find the structure from #4957 useful which adds a `Record<string, ColorRole>` as an intermediate step from theme tokens to component tokens.

I see that the native implementation uses this set of colors:
```
class CheckboxColors
constructor(
    val checkedCheckmarkColor: Color,
    val uncheckedCheckmarkColor: Color,
    val disabledCheckmarkColor: Color,
    val checkedBoxColor: Color,
    val uncheckedBoxColor: Color,
    val disabledCheckedBoxColor: Color,
    val disabledUncheckedBoxColor: Color,
    val disabledIndeterminateBoxColor: Color,
    val checkedBorderColor: Color,
    val uncheckedBorderColor: Color,
    val disabledBorderColor: Color,
    val disabledUncheckedBorderColor: Color,
    val disabledIndeterminateBorderColor: Color,
)
```

src/components/Checkbox/Checkbox.tsx
```diff
@@ -3,17 +3,18 @@ import {
   Animated,
```
adrcotfas, line 3: Image

Focused state does not look right on Android. See specs screenshot: 

Image

adrcotfas (changes requested):

fabriziocucci:
Superseded by https://github.com/callstack/react-native-paper/pull/4966.

Issue #4959, opened 2026-05-21
refactor(FAB): update to latest MD specs


adrcotfas:
See https://m3.material.io/components/floating-action-button/overview.
Include FAB Menu: https://m3.material.io/components/fab-menu/overview and Extended FAB: https://m3.material.io/components/extended-fab/overview

See https://github.com/callstack/react-native-paper/discussions/4949 for more details

PR #4963, opened 2026-05-22
feat(fab): modernize FAB to MD3

adrcotfas:
### Motivation

Rework the FAB family to follow Material Design 3, replacing the legacy
  FAB, FAB.Group, and AnimatedFAB with a token-driven implementation.

  - New `variant` prop (primary, secondary, tertiary, tonalPrimary,
    tonalSecondary, tonalTertiary) drives role-based coloring, defaulting
    to `tonalPrimary`.
  - New `size` prop (default, medium, large) replaces the old `small` /
    `customSize` sizing, with shape and metrics sourced from spec tokens.
  - `visible` toggling now animates the spec'd enter/exit (scale + alpha)
    on the FAB itself.
  - Add `FAB.Extended` for the extended (icon + label) FAB.
  - Add `FAB.Menu` (with Menu, MenuItem, MenuTrigger) to replace the old
    `FAB.Group` speed-dial pattern.
  - Add a keyboard focus ring (web) and shared visibility animation hook.
  - Split the implementation into Shell, Content, Extended, Menu, tokens,
    useFocusRing, and useVisibility for clarity.

  BREAKING CHANGE: The FAB API has been redesigned for MD3 with no
  deprecation aliases.

  - `AnimatedFAB` and `AnimatedFABProps` are removed. Use `FAB.Extended`
    for the icon + label FAB.
  - `FAB.Group` and `FABGroupProps` are removed. Use `FAB.Menu` instead.
  - The legacy FAB props (`mode`, `small`, `color`, `label`, `uppercase`,
    `customSize`, `animated`) are removed. Use `variant`, `size`,
    `containerColor`, and `contentColor` instead; for a labeled FAB switch
    to `FAB.Extended`.
  - New exported types: `FABExtendedProps`, `FABMenuProps`,
    `FABMenuItemProps`, `FABMenuTriggerProps`, `FABVariant`, `FABSize`.

### Related issue
https://github.com/callstack/react-native-paper/issues/4959

### Test plan

- Lint, typescript, existing and new tests pass
- manual visual inspection on all platforms


src/index.tsx
```diff
@@ -34,8 +32,9 @@ export { default as Chip } from './components/Chip/Chip';
 export { default as DataTable } from './components/DataTable/DataTable';
 export { default as Dialog } from './components/Dialog/Dialog';
 export { default as Divider } from './components/Divider';
-export { default as FAB } from './components/FAB';
-export { default as AnimatedFAB } from './components/FAB/AnimatedFAB';
+export { default as FloatingActionButton } from './components/FAB/FloatingActionButton';
+export { default as ExtendedFloatingActionButton } from './components/FAB/ExtendedFloatingActionButton';
+export { default as FloatingActionButtonMenu } from './components/FAB/FloatingActionButtonMenu';
```
satya164: I'm not convinced we need to make a breaking change here. `FAB` still has mostly the same API, so I'd keep the same component name. `FloatingActionButton` is also quite long to type, takes more tokens etc.

satya164: traditionally, the library has used format like `Header`, `Header.Title`, `FAB`, `FAB.Group` etc. any reason to not use `FAB.Menu`, `FAB.Extended` instead of a separate export?

example/src/Examples/AppbarExample.tsx
```diff
@@ -23,6 +23,7 @@ const MORE_ICON = Platform.OS === 'ios' ? 'dots-horizontal' : 'dots-vertical';
 const MEDIUM_FAB_HEIGHT = 56;
 
 const AppbarExample = () => {
+  // @ts-ignore
```
satya164: don't ignore types. if it's necessary to ignore, use `@ts-expect-error` and leave a comment saying why you're ignoring it. what's the problem here?

src/components/FAB/ExtendedFloatingActionButton.tsx
```diff
@@ -0,0 +1,273 @@
...
+      background,
+      style,
+      testID = 'extended-floating-action-button',
+      theme: themeOverrides,
+    },
+    ref
+  ) => {
+    const theme = useInternalTheme(themeOverrides);
+    const reduceMotion = useReduceMotion();
+    const isWeb = Platform.OS === 'web';
```
satya164: Don't assign `Platform.OS` checks to variables. Use `Platform.OS`or `Platform.select` checks directly - which allows a bundler to remove code for unused platforms.

src/components/FAB/utils.ts
```diff
@@ -364,3 +103,16 @@ export const getLabelSizeWeb = (ref: MutableRefObject<HTMLElement | null>) => {
...
+const getCanvasContext = () => {
+  if (cachedContext) {
+    return cachedContext;
+  }
+
+  const canvas = document.createElement('canvas');
+  cachedContext = canvas.getContext('2d');
+
+  return cachedContext;
+};
```
satya164: why doesn't it use the same approach as Android for text measurement? canvas seems a bit too much complexity compared to using the same code. I see it was using canvas before but since this is a rewrite, it's good opportunity to simplify and consolidate.

src/components/FAB/ExtendedFloatingActionButton.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+    const dimensions = React.useMemo(
+      () => getDimensions({ theme, size }),
+      [theme, size]
+    );
```
satya164: `getDimensions` seems like a very simple function creating a basic object. it's unnecessary to memo such code.

src/components/FAB/ExtendedFloatingActionButton.tsx
```diff
@@ -0,0 +1,273 @@
...
+            setLabelWidth(measured.width);
+          }
+        }
+      };
+      updateLabelSize();
+      window.addEventListener('resize', updateLabelSize);
+      return () => {
+        window.removeEventListener('resize', updateLabelSize);
+      };
+    }, [isWeb, label]);
```
satya164: typically for animations, it's best to measure the dimensions just before you start the animation. it keeps the code simple - you don't need to listen to various events, keep track of state, etc., and still get the most up-to-date values.

can you check if this is possible here to simplify the logic? reanimated even has a method to perform the measurement on UI thread directly https://docs.swmansion.com/react-native-reanimated/docs/advanced/measure/

src/components/FAB/FloatingActionButtonMenu.tsx
```diff
@@ -0,0 +1,767 @@
...
+  /**
+   * Trigger FAB. Pass a `<FloatingActionButton .../>`. The menu reads its
+   * `variant`, `size`, `icon`, and `onPress` and renders a single morphing
+   * FAB that animates between the trigger and the spec'd close button.
+   */
+  button: React.ReactElement;
+  /**
+   * Horizontal side the menu sits on. Default `'end'`.
+   */
+  horizontalAlignment?: 'start' | 'center' | 'end';
```
satya164: Just `alignment` sounds better since it's only single axis anyway

```suggestion
  alignment?: 'start' | 'center' | 'end';
```

src/components/FAB/FloatingActionButtonMenu.tsx
```diff
@@ -0,0 +1,767 @@
...
+  /**
+   * Called when the user taps the close button or taps an item.
+   */
+  onDismiss?: () => void;
+  /**
+   * Trigger FAB. Pass a `<FloatingActionButton .../>`. The menu reads its
+   * `variant`, `size`, `icon`, and `onPress` and renders a single morphing
+   * FAB that animates between the trigger and the spec'd close button.
+   */
+  button: React.ReactElement;
```
satya164: I assume the react element is being read. we shouldn't use React Children APIs or read react elements. it makes components both non-composable and non-type safe.

it was one of the wrong decisions in previous implementations, and we should change that in the refactors. we already have some tickets to change this iirc.

in this case, `button` can accept an object that's props of the `FAB` to render. the element is not being rendered either way, so it's misleading to accept a react element. and even if it was, accepting an object would make it type-safe (even if it's still not composable).

src/components/FAB/FloatingActionButtonMenu.tsx
```diff
@@ -0,0 +1,767 @@
...
+  /**
+   * Icon used by the close button when the menu is expanded. Default
+   * `'close'`.
+   */
+  closeIcon?: IconSource;
+  /**
+   * Menu items as `<FloatingActionButtonMenu.Item />`. Spec calls for 2 to 6
+   * items; a dev-mode warning fires outside that range.
+   */
+  children: React.ReactNode;
```
satya164: same here. accept a type-safe array of objects instead of react children and reading them. only accept react elements where you don't actually need to read or manipulate what they are.

the 2-6 items limit can also be enforced by the type system.

src/components/FAB/FloatingActionButtonMenu.tsx
```diff
@@ -0,0 +1,767 @@
...
+
+export type FloatingActionButtonMenuProps = {
+  /**
+   * Whether the menu is open.
+   */
+  expanded: boolean;
+  /**
+   * Called when the user taps the close button or taps an item.
+   */
+  onDismiss?: () => void;
```
satya164: since the component is controlled with `expanded`, `onDismiss` should also be required

src/components/FAB/FloatingActionButtonMenu.tsx
satya164: `trigger` maybe a better name

satya164 (changes requested):
the commit messages (from PR description) will be used for changelog, so lets make it a habit to include user facing description in the commit description. e.g. what removed, what changed, replacements, what user should change in their code, diff example when relevant etc.

for breaking changes, you can put "BREAKING CHANGE:" at the top of the description which makes the changelog generator pick it up.

satya164:
also noticed the button changing size back and forth on android 

https://github.com/user-attachments/assets/fcb97bfb-3113-4f21-bac1-2d257ea848ee

the shadows don't look correct on iOS and Web. it makes the component look blurry with light background

CleanShot 2026-06-02 at 13 25 12@2x

on android the shadows look better but don't match the ones in MD guidelines - looks darker and less spread out

CleanShot 2026-06-02 at 13 27 13@2x

example/src/Examples/AppbarExample.tsx
adrcotfas: I'll remove that. Something must have changed in the meantime because I don't see the error anymore but it was something about complexity.

adrcotfas:
> also noticed the button changing size back and forth on android

Fixed

> the shadows don't look correct on iOS and Web. it makes the component look blurry with light background
> 

I suspect that the image examples you shared don't follow the specs exactly.  It's not the first issue I see in the MD docs. iOS and Web were broken, indeed.  

Android delegates the shadow to the OS's top-positioned virtual light model while iOS and Web let us draw a fully parameterized, centered drop shadow, so the looks can't match exactly.

I'm using a native app with FAB to compare and here's the current state.  
Web screenshot is pixelated because I increased its size according to the zoomed in emulator/simulator screenshots.

fab_small

src/components/FAB/FloatingActionButtonMenu.tsx
adrcotfas: Good point. It's fixed.

satya164 (approved):

Issue #4946, opened 2026-05-14
Performance: theme color computations run on every render in List.Item, TextInput, Chip, Divider, Appbar.Action


ayobamiseun:
### Current behaviour

Several theme-derived colors in commonly-rendered components are recomputed via the color package (color(...).alpha(...).rgb().string()) on every render, even when their inputs don't change across renders. The result is wasted CPU on every keystroke in TextInput, on every parent re-render of any list of List.Item / Chip / Divider, and on every render of Appbar.Action. We noticed it while profiling typing latency on a search screen and scroll/interaction jank on a settings list, where calls into the color package showed up much more frequently than expected in flamegraphs.

The affected sites we found (all in the v2 theme path or other input-independent branches):

- [src/components/Divider.tsx:67-70](vscode-webview://104p3er6308m39fmlbmbq03n8jkb5u26aetvdm327jqvb2cklg06/src/components/Divider.tsx#L67-L70) — color(isDarkTheme ? white : black).alpha(0.12).rgb().string(). The inputs are constants plus a boolean; there are exactly two possible results, but we recompute on every render of every divider.

- [src/components/List/ListItem.tsx:209-211](vscode-webview://104p3er6308m39fmlbmbq03n8jkb5u26aetvdm327jqvb2cklg06/src/components/List/ListItem.tsx#L209-L211) and [233-235](vscode-webview://104p3er6308m39fmlbmbq03n8jkb5u26aetvdm327jqvb2cklg06/src/components/List/ListItem.tsx#L233-L235) — titleColor and descriptionColor run a full color() chain on every render. List.Item is rendered N times in lists, so the cost multiplies.

- [src/components/Appbar/AppbarAction.tsx:108](vscode-webview://104p3er6308m39fmlbmbq03n8jkb5u26aetvdm327jqvb2cklg06/src/components/Appbar/AppbarAction.tsx#L108) — color(black).alpha(0.54).rgb().string(). This branch has no dependencies at all; both arguments are constants, yet the chain runs on every render.

- [src/components/TextInput/helpers.tsx:332, 367, 401, 410, 421-422](vscode-webview://104p3er6308m39fmlbmbq03n8jkb5u26aetvdm327jqvb2cklg06/src/components/TextInput/helpers.tsx) — getTextColor, getActiveColor, getSelectionColor, getFlatBackgroundColor each contain color() chains and are called on every TextInput render. Since TextInput re-renders on every keystroke, this fires constantly during typing.

[...]

PR #4964, opened 2026-05-23
fix: memoize chip color resolution

popsiclelmlm:
### Motivation

`Chip` resolves multiple theme-derived colors during render. The current main branch has already removed most of the color chains mentioned in issue #4946, but `Chip` still calls `getChipColors` every render even when its color inputs are unchanged.

This wraps the color resolution in `React.useMemo` so the derived color object is recomputed only when the relevant inputs change. There is no API or visual behavior change.

### Related issue

Related to #4946.

### Test plan

- `yarn test src/components/__tests__/Chip.test.tsx --runInBand`
- `yarn lint-no-fix src/components/Chip/Chip.tsx`
- `yarn typescript`
- Full test suite also ran during the pre-commit hook: 55 suites / 736 tests passed.

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -259,13 +259,17 @@ const Chip = ({
     contentOpacity,
     selectedBackgroundColor,
     backgroundColor,
-  } = getChipColors({
-    isOutlined,
-    theme,
-    selectedColor,
-    customBackgroundColor,
-    disabled,
-  });
+  } = React.useMemo(
+    () =>
+      getChipColors({
+        isOutlined,
+        theme,
+        selectedColor,
+        customBackgroundColor,
+        disabled,
+      }),
+    [customBackgroundColor, disabled, isOutlined, selectedColor, theme]
+  );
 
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR. Are you seeing any slowdowns or is this hypothetical?

Understanding it is critical to know whether rest of the code should be updated.

popsiclelmlm:
Thanks for asking. This isn't based on a benchmark or a user-reported slowdown. I noticed that Chip recalculates derived colors during render, so this PR is a preventative cleanup to avoid repeating that color resolution when the relevant inputs haven't changed.

If you'd prefer changes like this to be backed by measured slowdown before merging, I'm happy to close this or keep it scoped to Chip rather than updating similar code elsewhere.

satya164:
@popsiclelmlm let's close it for now as it'll result in conflicts with ongoing work unnecessarily. happy to merge changes later if we can measure it.

popsiclelmlm:
Thanks for clarifying. I'll close this for now to avoid creating conflicts with the ongoing work. Happy to revisit later if there is measured impact to address.

PR #4966, opened 2026-05-26
feat(checkbox): modernize for md3 spec compliance

fabriziocucci:
## Summary

Bring the `Checkbox` component up to the [Material Design 3 spec](https://m3.material.io/components/checkbox/specs), end-to-end:

- 18dp container with 2dp outline (unselected) → 0dp outline + theme primary fill (selected), inside a 40dp state-layer tap target.
- The tap target uses `TouchableRipple` (borderless + centered), matching `IconButton`/`Chip`/`Button`. This gives the native Android ripple animation per @adrcotfas's review (the ripple *color* won't fully respect MD3 alpha until RN 86 ships `PlatformColor` + alpha for `android_ripple`, but the animation lands today). `rippleColor` is fed by `getSelectionVisualState({ ..., pressed: true })` so the ripple honours the MD3 selected/unselected colour flip on press.
- Focus indicator: 3dp ring at `theme.colors.secondary`, drawn at the 40dp state-layer boundary (the `borderless` `TouchableRipple` clips anything outside that, so we sit the ring on the boundary rather than applying `md.sys.state.focusIndicator.outerOffset`). On web, the ring is gated on `:focus-visible` via the shared `isKeyboardFocusEvent` helper (introduced in #4957) so it doesn't appear after a pointer press; on native it shows on any focus. The browser's default focus outline is suppressed via Adrian's `webNoOutline = { outline: 'none' } as unknown as ViewStyle` pattern from #4957.
- RTL: on web, the reveal-mask anchor flips via `useLocale().direction === 'rtl'` so the checkmark appears to draw from the leading edge. On native, `I18nManager` already mirrors the layout so the default `left: 0` anchor does the right thing automatically (matches the approach used in #4957). The glyph itself is not mirrored (Compose M3 keeps checkmark orientation across locales) — we explicitly pre-swap the L-shape's `borderLeftWidth` / `borderRightWidth` in RTL so RN's automatic swap brings them back to the LTR orientation and the rotated stroke renders as a `✓` instead of a `<` on Android (which honours `I18nManager.swapLeftAndRightInRTL`).
- Accessibility: the inner `<Checkbox>` now emits `accessibilityState.checked === 'mixed'` for `status="indeterminate"` (per W3C ARIA for tri-state controls). `Checkbox.Item`'s outer row no longer duplicates the `'mixed'` announcement — it defers to the inner Checkbox, so screen readers announce "mixed" once. Also `disabled` is now coerced via `!!disabled` to avoid leaking `undefined`.
- New `style?: StyleProp<ViewStyle>` prop forwarded to the underlying `Pressable`, per Copilot's review (the rest of the additional `Pressable` props — `accessibilityLabel`, `hitSlop`, `onLongPress` — are still out of scope for this PR).
- Animations: 100ms fill + 150ms checkmark reveal driven by `react-native-reanimated` (`useSharedValue` + `withTiming` + `useAnimatedStyle`), now using the canonical pattern established by #4957 — motion tokens (`theme.motion.duration.short2` / `short3`, `theme.motion.easing.standard`) and the new `useReduceMotion()` hook + `ReduceMotion.Always` config so the OS / `PaperProvider` reduce-motion setting is honored. Reads the new `isKeyboardFocusEvent` shared helper from `src/utils/` and Adrian's `webNoOutline = { outline: 'none' } as unknown as ViewStyle` pattern for the browser default-outline suppression.

## Why not SVG

Compose Material3 draws the checkmark as a `Path` with `strokeFraction` 0 → 1. Doing the same in RN would mean adding `react-native-svg` as a Paper peer-dep, an ecosystem-level change. This PR uses a reveal-mask instead: a static L-shape (borderLeftWidth + borderBottomWidth rotated -45°) inside a left-anchored View whose width animates 0 → 18dp with a matching opacity fade. The visual suggests the stroke drawing left-to-right without the SVG dependency. Same precedent will apply to RadioButton when it's modernized. If you'd prefer the SVG-based approach for v6, happy to switch. Picking once now sets the precedent.

## Files

- `src/components/Checkbox/Checkbox.tsx`: full rewrite. `TouchableRipple` 40dp tap target (borderless + centered, matching `IconButton`); animated 18dp container with cross-fade outline↔fill; focus ring as a circular outline matching the state-layer shape (— circle, not a rounded-square around the 18dp container — matches the rendering in Compose Material3 and the spec's focus-state reference); `:focus-visible` filter on web via the shared `isKeyboardFocusEvent` helper; locale-aware reveal-mask; view-based checkmark + dash; forwarded `style` plus all `Pressable`/`TouchableRipple` props via `...rest`; `'mixed'` a11y state for indeterminate. Animations driven by `react-native-reanimated` with motion tokens + `useReduceMotion`. No Platform-specific code paths.
- **NEW** `src/components/Checkbox/tokens.ts`: `CheckboxTokens` exporting the 4 spec dimensions (`containerSize`, `containerRadius`, `outlineWidth`, `stateLayerSize`) plus the colour-role table (`containerColor`, `outlineColor`, etc.). Merged via `{ ...sizes, ...colors }` to mirror the `SwitchTokens` pattern from #4957.
- `src/components/Checkbox/utils.ts`: `getSelectionVisualState` returns the full colour + opacity picture for any `(selected × hovered × pressed × disabled × error × customColor)` combination. Focus is rendered by the component (per the MD3 spec note: focus is an outline ring, not a state-layer fill), so it deliberately doesn't appear in the visual-state helper. The legacy `getSelectionControlColor` helper that used to live here (only consumed by `RadioButtonAndroid`) has been moved to `RadioButton/utils.ts`.
- `src/components/RadioButton/utils.ts`: now hosts `getSelectionControlColor` (moved from `Checkbox/utils.ts`). Cleans up the awkward back-compat re-export and keeps Checkbox/utils.ts focused on MD3 checkbox concerns.
- `src/components/RadioButton/RadioButtonAndroid.tsx`: import path updated to the local `./utils`.
- Tests for `getSelectionControlColor` migrated to `src/components/__tests__/RadioButton/utils.test.tsx`; the old `src/components/__tests__/Checkbox/utils.test.tsx` is removed.
- `example/src/Examples/CheckboxExample.tsx`: rewritten as a comprehensive showcase — "Tap to toggle" interactive section (unchecked↔checked, unchecked↔indeterminate), a 4-row × 3-status grid (Default / Disabled / Custom color tertiary / Error), and a Parent/children section demonstrating an indeterminate parent that aggregates state from three children.
- `example/package.json`: added `web` script (`expo start --web`) so reviewers can `yarn example web` to validate the focus ring, the `:focus-visible` filter, and RTL behaviour on web in addition to iOS / Android.
- 9 snapshots auto-updated to reflect the new render tree.

## Out of scope

- RadioButton modernization (separate PR with the same pattern).
- Extracting `<StateLayer />` as a shared primitive (can follow once Radio + Switch land and the pattern is proven).
- Forwarding additional `Pressable` props (`accessibilityLabel`, `hitSlop`, `onLongPress`, etc.). Already added `style`; the rest can land in a follow-up if reviewers want a wider API surface.

## Test plan

- Rebased onto `main` after #4957 landed; now consumes the shared `isKeyboardFocusEvent` helper, `ReduceMotionContext`, and motion tokens introduced there.
- `yarn typescript` clean (the `example/src/Examples/AppbarExample.tsx` TS2590 error is pre-existing on `main` and unaffected by this branch).
- `yarn lint` clean
- `yarn jest`: 738/739 (1 skipped, pre-existing); 161/161 snapshots pass. The legacy `getSelectionControlColor` test suite was moved from `__tests__/Checkbox/utils.test.tsx` to `__tests__/RadioButton/utils.test.tsx` alongside its sibling `getSelectionControlIOSColor` tests — same coverage, just relocated to match the helper's new home.

## Visuals

| iOS | Android | Web |
|---|---|---|
|  |  |  |

Each video cycles through the new states: unchecked / checked / indeterminate, the error variant, the disabled variants, and the Parent / children section showing an indeterminate parent updating from child changes.

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -1,16 +1,30 @@
 import * as React from 'react';
-import {
-  Animated,
+import { Platform, StyleSheet, View } from 'react-native';
+import type {
   ColorValue,
   GestureResponderEvent,
-  StyleSheet,
-  View,
+  NativeSyntheticEvent,
+  StyleProp,
+  TargetedEvent,
+  ViewStyle,
 } from 'react-native';
 
-import { getSelectionControlColor } from './utils';
+import Animated, {
+  Easing,
+  ReduceMotion,
+  useAnimatedStyle,
+  useSharedValue,
+  withTiming,
+} from 'react-native-reanimated';
+
+import { CheckboxTokens } from './tokens';
+import { getSelectionVisualState } from './utils';
+import { useLocale } from '../../core/locale';
 import { useInternalTheme } from '../../core/theming';
+import { useReduceMotion } from '../../theme/accessibility/ReduceMotionContext';
+import { tokens } from '../../theme/tokens';
 import type { $RemoveChildren, ThemeProp } from '../../types';
-import MaterialCommunityIcon from '../MaterialCommunityIcon';
+import { isKeyboardFocusEvent } from '../../utils/isKeyboardFocusEvent';
 import TouchableRipple from '../TouchableRipple/TouchableRipple';
 
 export type Props = $RemoveChildren<typeof TouchableRipple> & {
@@ -36,9 +50,9 @@ export type Props = $RemoveChildren<typeof TouchableRipple> & {
   color?: ColorValue;
   /**
    * Whether the checkbox is in an error state. When true, the outline
-   * (unchecked) and container (checked / indeterminate) use
-   * `theme.colors.error`. `disabled` and explicit `color`/`uncheckedColor`
-   * overrides take precedence.
+   * (unchecked) and container (selected) use `theme.colors.error`.
+   * `disabled` and explicit `color`/`uncheckedColor` overrides take
+   * precedence.
    */
   error?: boolean;
   /**
@@ -49,9 +63,28 @@ export type Props = $RemoveChildren<typeof TouchableRipple> & {
    * testID to be used on tests.
    */
   testID?: string;
+  /**
+   * Custom style to override the default tap target. Passed through to
+   * the underlying `TouchableRipple`.
+   */
+  style?: StyleProp<ViewStyle>;
 };
 
-const ANIMATION_DURATION = 100;
+// Spec dimensions (https://m3.material.io/components/checkbox/specs).
+const {
+  containerSize: CONTAINER_SIZE,
+  containerRadius: CONTAINER_RADIUS,
+  outlineWidth: OUTLINE_WIDTH,
+  stateLayerSize: STATE_LAYER_SIZE,
+} = CheckboxTokens;
+
+const FOCUS_THICKNESS = tokens.md.sys.state.focusIndicator.thickness;
+// Focus indicator is a circular ring at the 40dp state-layer boundary.
+// We don't apply `focusIndicator.outerOffset` here because the surrounding
+// `TouchableRipple borderless` clips overflow to the tap-target shape,
+// so a ring drawn outside the 40dp circle would be cropped.
+const FOCUS_RING_SIZE = STATE_LAYER_SIZE;
+const FOCUS_RING_RADIUS = STATE_LAYER_SIZE / 2;
 
 /**
  * Checkboxes allow the selection of multiple options from a set.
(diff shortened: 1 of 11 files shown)


adrcotfas:
Focus ring is now gone. Check https://github.com/callstack/react-native-paper/pull/4957 review comments for how it was handled there.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +95,235 @@ const Checkbox = ({
...
+  color,
+  uncheckedColor,
+  style,
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
-  const { current: scaleAnim } = React.useRef<Animated.Value>(
-    new Animated.Value(1)
-  );
-  const isFirstRendering = React.useRef<boolean>(true);
+  const [hovered, setHovered] = React.useState(false);
```
adrcotfas: @satya164 Might request the same approach with shared value as [here](https://github.com/callstack/react-native-paper/pull/4957#discussion_r3304377282) + Reanimated instead.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +95,235 @@ const Checkbox = ({
...
+          styles.stateLayer,
+          {
+            backgroundColor: visual.stateLayerColor,
+            opacity: visual.stateLayerOpacity,
+          },
+        ]}
+      />
+      <View style={[styles.container, { opacity: visual.containerOpacity }]}>
+        <Animated.View
+          pointerEvents="none"
```
adrcotfas: Warns on rn-web. Same fix as Switch: move into the style.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -3,17 +3,18 @@ import {
   Animated,
```
adrcotfas: Some general comments: 
- Missing focus handling
- No RTL handling. The checkmark mask animates width anchored at left: 0; on web that won't draw right-to-left in RTL. check `useLocale().direction`
- No outline: none on web Pressable: same default-outline problem we fixed in Switch
- See `useReduceMotion` hook for motion scaling

adrcotfas (changes requested):

adrcotfas:
- Focus ring does not have the right shape as in the [specs](https://m3.material.io/components/checkbox/specs) and on Android there's an extra transparent rectangle (see screenshot of modified example screen)
Screenshot_1780057122

- We should be using `TouchableRipple` to respect the specs (see screenshot); we won't be able to respect the ripple color until RN 86 which included PlatformColor and alpha for `android_ripple` but at least the animation is there (see other components)
- Please update the example app so we showcase all the new states; an indeterminate parent checkbox adapting according to children would be awesome along with showcasing the error variant.
image

adrcotfas (changes requested):
See previous comment.

src/components/RadioButton/utils.ts
```diff
@@ -93,3 +93,44 @@ export const getSelectionControlIOSColor = ({
     checkedColorOpacity,
   };
 };
+
+/**
```
adrcotfas: A general comment: please remove or simplify where valuable the overly verbose comments.    
Some mentioning previous state/history or unrelated logic(other components) like this one for which we don't loose any value if we remove it.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +121,341 @@ const Checkbox = ({
...
 };
 
+/**
+ * Reveal-mask checkmark: a static L-shape (borderLeftWidth +
+ * borderBottomWidth rotated -45deg) inside a directional-anchored View
+ * whose width animates 0 -> CONTAINER_SIZE. The checkmark "draws in"
+ * along the writing direction, approximating Compose Material3's
+ * stroke-fraction animation without an SVG dependency.
+ *
+ * The glyph itself is not mirrored in RTL (Compose M3 keeps the
```
adrcotfas: Now it's broken on:

- web: 
Image
- iOS:
Image

adrcotfas (changes requested):

example/package.json
```diff
@@ -8,7 +8,8 @@
     "eas-build-pre-install": "cd .. && yarn",
     "start": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --dev-client",
     "android": "EXPO_NO_TYPESCRIPT_SETUP=1 expo run:android",
-    "ios": "EXPO_NO_TYPESCRIPT_SETUP=1 expo run:ios"
+    "ios": "EXPO_NO_TYPESCRIPT_SETUP=1 expo run:ios",
+    "web": "EXPO_NO_TYPESCRIPT_SETUP=1 expo start --web"
```
satya164: just press w in the terminal (like i for ios and a for android), or run `yarn start --web`. don't need a separate script for this

src/components/__tests__/Checkbox/CheckboxItem.test.tsx
```diff
@@ -42,13 +42,16 @@ it('should have `accessibilityState={ checked: false }` when `status="unchecked"
...
     <Checkbox.Item status="indeterminate" label="Indeterminate Button" />
   );
 
-  const elements = getAllByA11yState({ checked: false });
-  expect(elements).toHaveLength(2);
+  // The inner Checkbox exposes `checked: "mixed"` (per W3C ARIA spec for
+  // tri-state controls), while the outer row Pressable exposes
+  // `checked: false`.
+  expect(getAllByA11yState({ checked: 'mixed' })).toHaveLength(1);
+  expect(getAllByA11yState({ checked: false })).toHaveLength(1);
```
satya164: I don't think this makes sense. The same element shouldn't expose 2 separate `checked` status, and the a11y state should only be on the interactable element.

src/components/__tests__/RadioButton/utils.test.tsx
```diff
@@ -85,3 +88,162 @@ describe('getSelectionControlIOSColor - checked color', () => {
     });
   });
 });
+
+describe('getSelectionControlColor - radio color (legacy, shared with pre-MD3 Checkbox)', () => {
```
satya164: we don't need such tests. these should be covered by actual checkbox tests already

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +118,319 @@ const Checkbox = ({
...
+    width: checkProgress.value * CONTAINER_SIZE,
+    opacity: checkProgress.value,
+  }));
+
+  // Remember the last drawn glyph so the reveal-mask can finish collapsing
+  // when `selected` flips back to false.
+  const lastGlyph = React.useRef<'check' | 'indeterminate'>('check');
+  if (status === 'checked') lastGlyph.current = 'check';
+  else if (status === 'indeterminate') lastGlyph.current = 'indeterminate';
+  const showIndeterminate = lastGlyph.current === 'indeterminate';
```
satya164: Refs shouldn't be mutated during render.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +118,319 @@ const Checkbox = ({
...
       accessibilityLiveRegion="polite"
-      style={styles.container}
       testID={testID}
-      theme={theme}
+      style={[
+        styles.tapTarget,
+        Platform.OS === 'web' ? webNoOutline : undefined,
+        style,
+      ]}
+      {...rest}
```
satya164: The spread should be the first one, not last. Otherwise, all of the previously passed can be overwritten by user and break the component.

satya164 (changes requested):

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +119,343 @@ const Checkbox = ({
...
+  );
+  const nextGlyph: 'check' | 'indeterminate' =
+    status === 'checked'
+      ? 'check'
+      : status === 'indeterminate'
+      ? 'indeterminate'
+      : lastGlyph;
+  if (nextGlyph !== lastGlyph) {
+    setLastGlyph(nextGlyph);
+  }
```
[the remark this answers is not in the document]
fabriziocucci: Keeping this as-is. It's React's officially-endorsed ["derive state during render" pattern](https://react.dev/reference/react/useState#storing-information-from-previous-renders) for storing information from previous renders. The docs explicitly call out this exact shape (bail-out + render-time `setState` call) for the case where you need to remember a value from a prior render that can't be computed from current props.

Why an effect doesn't work here: effects run after paint, so on the `indeterminate → unchecked` transition the first frame would render with the new `lastGlyph` value, the mask reveal would snap, and only the second frame (after the effect commits) would have the right glyph for the collapse animation. The render-time `setState` avoids that intermediate frame, React bails out the in-progress render and re-renders in the same commit, so it's a single paint either way.

Walked through the docs' three preferred alternatives before reaching for this:

1. Compute from current props: doesn't work; `lastGlyph` needs the previous `status`, which isn't derivable from `status="unchecked"`.
2. Pass a different `key` to reset: would reset the in-flight reanimated values too, breaking the collapse.
3. Update in event handlers: `status` is a prop, not internal state.

So we genuinely hit the "none of these apply" case the docs describe.

src/components/__tests__/Checkbox/utils.test.tsx
```diff
@@ -1,163 +1,199 @@
...
+        selected: true,
+        hovered: true,
+        pressed: true,
+        disabled: true,
+      });
+      // The renderer itself zeroes out hover/pressed when disabled before
+      // calling this helper; verify the helper still returns a sensible value
+      // when explicitly called with `disabled: true` (no special-casing).
+      expect(typeof v.stateLayerOpacity).toBe('number');
     });
```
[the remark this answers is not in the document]
fabriziocucci: Addressed in `b58cf07` by removing the state-layer fields from `getSelectionVisualState` entirely (per the sibling comment about unused returns) — the misleading test and its weak assertion are gone with it. `TouchableRipple`'s default ripple is the sole owner of hover/pressed visuals on the tap target.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +117,343 @@ const Checkbox = ({
...
+  // `pressed` aren't tracked here — `TouchableRipple` owns the press ripple
+  // and hover overlay.
+  const visual = getSelectionVisualState({
+    theme,
+    selected,
+    disabled,
+    error,
+    customColor: color,
+    customUncheckedColor: uncheckedColor,
   });
```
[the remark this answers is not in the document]
fabriziocucci: Went with option (b) in `b58cf07`: removed `stateLayerColor`/`stateLayerOpacity` from the helper's return type, dropped the unused `hovered`/`pressed` params and the private `resolveStateLayer` function. `TouchableRipple`'s built-in ripple owns hover/pressed visuals, so the helper now only exposes the four static values the renderer actually consumes.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +117,343 @@ const Checkbox = ({
...
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
-  const { current: scaleAnim } = React.useRef<Animated.Value>(
-    new Animated.Value(1)
+  const reduceMotion = useReduceMotion();
+  const { direction } = useLocale();
+  // Web (react-native-web) doesn't auto-mirror layout, so flip the mask
+  // anchor manually for RTL. Native handles it via `I18nManager`.
+  const flipMaskForWebRTL = Platform.OS === 'web' && direction === 'rtl';
+  const [focused, setFocused] = React.useState(false);
```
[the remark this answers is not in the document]
fabriziocucci: Considered this. The unit-test path is non-trivial here:

1. **Focus ring** — requires mocking `e.currentTarget.matches(':focus-visible')` (and `Platform.OS = 'web'`) on a Pressable. I prototyped this and the tests work, but only with full `PaperProvider` + `LocaleProvider` wrapping for the rendered Checkbox, plus reaching into the TouchableRipple's internal Pressable via `UNSAFE_getAllByType`. It's brittle and couples tests to test-renderer internals.

2. **RTL mask** — requires `jest.spyOn(Locale, 'useLocale').mockReturnValue({ direction: 'rtl' })` plus `Platform.OS = 'web'` reassignment, plus the same provider plumbing.

Both behaviors **are** covered today via:
- The snapshot tests (`Checkbox.test.tsx.snap`) capture the rendered tree, which includes the focus ring + the `checkmarkMask` / `checkmarkMaskWebRTL` style choice.
- Manual cross-platform verification — iOS RTL via the system Arabic locale + `-AppleTextDirection YES` launch arg, Android RTL via Developer Options → Force RTL, web RTL via the in-sandbox toggle. All verified post-fix.

Happy to add the targeted unit tests if the team prefers — open to feedback on whether the cost of the test infra plumbing is worth it for branches the snapshots already capture.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -84,121 +117,343 @@ const Checkbox = ({
...
+  const showIndeterminate = nextGlyph === 'indeterminate';
+
+  const handleFocus = React.useCallback(
+    (e: NativeSyntheticEvent<TargetedEvent>) => {
+      if (disabled) return;
+      if (!isKeyboardFocusEvent(e)) return;
+      setFocused(true);
+    },
+    [disabled]
+  );
```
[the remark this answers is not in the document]
fabriziocucci: Considered this. The unit-test path is non-trivial here:

1. **Focus ring** — requires mocking `e.currentTarget.matches(':focus-visible')` (and `Platform.OS = 'web'`) on a Pressable. I prototyped this and the tests work, but only with full `PaperProvider` + `LocaleProvider` wrapping for the rendered Checkbox, plus reaching into the TouchableRipple's internal Pressable via `UNSAFE_getAllByType`. It's brittle and couples tests to test-renderer internals.

2. **RTL mask** — requires `jest.spyOn(Locale, 'useLocale').mockReturnValue({ direction: 'rtl' })` plus `Platform.OS = 'web'` reassignment, plus the same provider plumbing.

Both behaviors **are** covered today via:
- The snapshot tests (`Checkbox.test.tsx.snap`) capture the rendered tree, which includes the focus ring + the `checkmarkMask` / `checkmarkMaskWebRTL` style choice.
- Manual cross-platform verification — iOS RTL via the system Arabic locale + `-AppleTextDirection YES` launch arg, Android RTL via Developer Options → Force RTL, web RTL via the in-sandbox toggle. All verified post-fix.

Happy to add the targeted unit tests if the team prefers — open to feedback on whether the cost of the test infra plumbing is worth it for branches the snapshots already capture.

src/components/__tests__/Checkbox/CheckboxItem.test.tsx
```diff
@@ -29,26 +29,27 @@ it('should have `accessibilityState={ checked: true }` when `status="checked"`',
     <Checkbox.Item status="checked" label="Checked Button" />
   );
 
-  const elements = getAllByA11yState({ checked: true });
-  expect(elements).toHaveLength(2);
+  // Only the outer row (the interactable element) exposes a11y state.
+  // The inner Checkbox is rendered with `accessible={false}` so the same
+  // logical control doesn't expose two `checked` states.
```
satya164: this comment seems like implementation/detail specific to recent changes detail and not necessary in the test

src/components/__tests__/Checkbox/CheckboxItem.test.tsx
```diff
@@ -29,26 +29,27 @@ it('should have `accessibilityState={ checked: true }` when `status="checked"`',
...
 
-it('should have `accessibilityState={ checked: false }` when `status="indeterminate"', () => {
+it('should have `accessibilityState={ checked: "mixed" }` when `status="indeterminate"`', () => {
   const { getAllByA11yState } = render(
     <Checkbox.Item status="indeterminate" label="Indeterminate Button" />
   );
 
-  const elements = getAllByA11yState({ checked: false });
-  expect(elements).toHaveLength(2);
+  // Per W3C ARIA, tri-state controls report `mixed` for indeterminate.
```
satya164: also seems like unnecessary comment

satya164 (approved):

adrcotfas (approved):

PR #4968, opened 2026-05-28
fix(docs): resolve SSR crash and webpack warnings in docs build

adrcotfas:
- Inject polyfills/dev.js via getClientModules to define globalThis.__DEV__ for both client and SSR bundles; needed because the new Switch pulls in react-native-reanimated which assumes this global exists
- Alias @expo/vector-icons/MaterialCommunityIcons to material-design-icons (already present) and @react-native-vector-icons/get-image to false
- Flatten CSS nesting in custom.css to fix CSS minimizer warnings about unsupported & selector syntax
- Add isDynamicColorSupported = false to the non-Android DynamicTheme fallback, which was missing the re-exported symbol
- Fix TypescaleKey re-export to use the type keyword

### DIFF
--- a/docs/plugins/docusaurus-react-native-plugin.js
+++ b/docs/plugins/docusaurus-react-native-plugin.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const webpack = require('webpack');
 
 module.exports = function () {
   return {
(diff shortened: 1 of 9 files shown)


docs/plugins/docusaurus-react-native-plugin.js
```diff
@@ -14,10 +14,17 @@ module.exports = function () {
             'react-native-vector-icons/MaterialCommunityIcons': path.resolve(
               'node_modules/@react-native-vector-icons/material-design-icons'
             ),
+            '@expo/vector-icons/MaterialCommunityIcons': path.resolve(
+              'node_modules/@react-native-vector-icons/material-design-icons'
+            ),
```
satya164: afaik we should now be able to drop `@expo/vector-icons` and `react-native-vector-icons` entirely and only import from `@react-native-vector-icons/material-design-icons` to support all platforms.

it should simplify both bundler logic and try/catch and conditions in library code.

docs/plugins/polyfills/dev.js
```diff
@@ -0,0 +1 @@
+globalThis.__DEV__ = process.env.NODE_ENV !== 'production';
```
satya164: this shouldn't be a runtime variable, but defined using [`definePlugin`](https://webpack.js.org/plugins/define-plugin/) based on dev/prod builds so dev specific code is treeshaken properly.

satya164 (changes requested):

satya164 (approved):

PR #4994, opened 2026-06-10
feat(tooltip): modernize Tooltip to Material Design 3

### LINKED ISSUE #4980: feat: modernize Tooltip to the latest Material Design specs
The goal is to modernize the Tooltip according to the latest specs - see below.

# Current State

## Tooltip

**MD3 spec:** Tooltips
**Specs link:** https://m3.material.io/components/tooltips/specs

### Color tokens

- [ ]  Container background uses `theme.colors.onSurface`; spec requires `inverseSurface`
- [ ]  Text color uses `theme.colors.surface`; spec requires `inverseOnSurface`

### Shape

- `borderRadius: theme.roundness` = 4dp matches `corner.extraSmall` ✅

### Spacing / sizing

- Container height 32dp ✅
- [ ]  Text variant is `labelLarge` (14sp); spec requires `bodySmall` (12sp)

### Behavior

- [ ]  Rich tooltip variant not implemented. MD3 defines two variants: plain (text only) and rich (optional title, body text, link, and action buttons). The current `title: string` prop cannot support rich content
- No pointer/caret indicator (not required by spec) ✅

### Animation

- [ ]  No enter/exit animation. MD3 specifies a fade transition on show/hide

burczu:
### Motivation

The `Tooltip` component had drifted from the current [Material Design 3 tooltip spec](https://m3.material.io/components/tooltips/specs), and only supported the plain (text-only) variant. As part of the v6 MD3 modernization effort (alongside TextInput, Switch, Checkbox, FAB), this brings the tooltip up to spec and adds the missing rich variant:

- **Plain tooltip** used the wrong color roles (`onSurface`/`surface`) and type style (`labelLarge`), and had no show/hide transition.
- **Rich tooltips** (title + supporting text + actions) weren't possible at all — the `title: string` prop can't express them.

### Related issue

Closes #4980.

### What changed

- **Plain tooltip — spec fixes:** container `onSurface` → **`inverseSurface`**, text `surface` → **`inverseOnSurface`**, type style `labelLarge` → **`bodySmall`** (12sp). Public API unchanged.
- **Fade animation:** plain (and rich) tooltips now fade in/out via Reanimated, driven by the MD3 motion tokens (`short3`/`standardDecelerate` in, `short2`/`standardAccelerate` out), and honor reduce-motion.
- **New `Tooltip.Rich` (compound component):** an optional `title` subhead, `content` (string → `bodyMedium`, or a custom element for inline links), and `actions` (button row). Rendered on a `surfaceContainer` `Surface` at elevation level 2 with a 12dp corner and 312dp max-width. Uncontrolled tap-to-toggle; dismisses on outside tap, action press, or re-tap; on web it opens on hover and bridges the trigger→tooltip gap.
- **Tokens** extracted to `Tooltip/tokens.ts`; shared fade lifecycle extracted to a `useTooltipFade` hook used by both variants.
- **Example + docs:** new "Rich tooltips" section in the example app; `RichTooltip` registered in the docs component map; JSDoc usage for both variants.

### Test plan

- `yarn test` — 743 passing / 170 snapshots (extended `Tooltip.test.tsx` with MD3 color/typography, fade mount-through-exit, and a `Tooltip.Rich` suite: toggle, custom content, color roles, backdrop/action dismiss, web hover open + gap-bridge).
- `yarn typescript`, `yarn lint` — clean.
- Verified on the iOS simulator (iPhone 17 Pro):
  - Plain: long-press an icon button → dark `inverseSurface` container, smaller `bodySmall` text, fades in/out.
  - Rich: tap `+` on the new "Rich tooltips" row → `surfaceContainer` card with title/body/actions; dismisses on outside-tap, action press, or re-tap; info icon shows a body-only variant.

### Videos

https://github.com/user-attachments/assets/e1e39979-d86a-4198-bf70-4b42fb923479

https://github.com/user-attachments/assets/c05d3678-00a0-4c64-81b9-5fc660421453

https://github.com/user-attachments/assets/282b5e6d-34ee-4b1b-913c-c1c960733572


src/components/Tooltip/tokens.ts
```diff
@@ -0,0 +1,72 @@
...
+  action: ColorRole;
+  shape: ShapeKey;
+  elevation: Elevation;
+  maxWidth: number;
+  paddingHorizontal: number;
+  paddingVertical: number;
+  titleTypescale: TypescaleKey;
+  contentTypescale: TypescaleKey;
+  gap: number;
+};
```
satya164: Doesn't seem like these `satisfies` are necessary. As these types aren't used anywhere else, it's essentially duplicating the structure of the object for not much. If the goal is to typecheck `surfaceContainer`, `onSurface` etc., I assume they are already checked at usage site.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,321 @@
...
+      )}
+      <Pressable
+        ref={childrenWrapperRef}
+        style={styles.pressContainer}
+        {...(isWeb ? webPressProps : mobilePressProps)}
+      >
+        {React.cloneElement(children, {
+          ...rest,
+          ...(isWeb ? webPressProps : mobilePressProps),
+        })}
```
satya164: The API/implementation needs to change to avoid `React.cloneElement` usage (or reading children directly, as they are not type safe and hurt composition.

It is used in existing components, but that's something we're revisiting.

There can be various approaches:
- split the component into parts that the user should render themselves and compose together
- accept render prop instead of children prop
- accept objects instead of react elements

Consider them based on how the final API may look like and what's most ergonomic as well as flexible.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,321 @@
...
+                  </Text>
+                ) : (
+                  content
+                )}
+                {actions ? (
+                  // `onTouchEnd` bubbles from the pressed action up to this
+                  // wrapper, so selecting any action dismisses the tooltip.
+                  <View
+                    style={styles.actions}
+                    onTouchEnd={hide}
```
satya164: does `onTouchEnd` trigger for mouse actions on web? we also need to consider keyboard on web

src/components/Tooltip/tokens.ts
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Rich tooltip — an optional subhead, supporting text and action buttons on a
+ * surface-container container at elevation level 2.
+ * https://m3.material.io/components/tooltips/specs#8e6cf915
+ */
+const rich = {
+  container: 'surfaceContainer',
+  title: 'onSurface',
+  content: 'onSurfaceVariant',
+  action: 'primary',
```
satya164: is this used?

satya164 (changes requested):

src/components/Tooltip/tokens.ts
burczu: Fixed.

src/components/Tooltip/tokens.ts
burczu: Fixed.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,321 @@
...
+    <>
+      {rendered && (
+        <Portal>
+          <Pressable
+            accessibilityRole="button"
+            onPress={hide}
+            pointerEvents={visible ? 'auto' : 'none'}
+            style={StyleSheet.absoluteFill}
+            testID="tooltip-rich-backdrop"
+          />
```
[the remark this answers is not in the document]
burczu: Fixed.

src/components/Tooltip/RichTooltip.tsx
burczu: Went with the render-prop approach — most ergonomic and flexible of the options, and the only one that actually removes the `cloneElement`/`children.props` reading rather than relocating it (a `Tooltip.Trigger` part would still need to clone its child internally). The trigger handlers and the `actions`' `dismiss` are handed to the consumer to compose, so it's type-safe and ref-free.

I scoped this to `Tooltip.Rich` and left the plain `Tooltip` on `cloneElement` since you mentioned that's being revisited separately — do you want me to revisit it in this PR too, or keep it separate?

src/components/Tooltip/RichTooltip.tsx
burczu: Reworked: actions now dismiss via an explicit `dismiss()` callback instead of `onTouchEnd`, so it works for press/click/keyboard everywhere. Web also wires `onFocus`/`onBlur` now, so the tooltip opens on keyboard focus too.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,321 @@
...
+  const handlePress = React.useCallback(() => {
+    if (visible) {
+      hide();
+    } else {
+      show();
+    }
+    if (isValidChild) {
+      (children.props as TooltipChildProps).onPress?.();
+    }
+  }, [visible, hide, show, isValidChild, children.props]);
```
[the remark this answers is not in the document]
burczu: No longer read or forward the child's `onPress` — the consumer owns the trigger element and spreads the provided handlers, so a disabled trigger doesn't fire. No more divergence from the plain tooltip.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,321 @@
...
+                {actions ? (
+                  // `onTouchEnd` bubbles from the pressed action up to this
+                  // wrapper, so selecting any action dismisses the tooltip.
+                  <View
+                    style={styles.actions}
+                    onTouchEnd={hide}
+                    testID="tooltip-rich-actions"
+                  >
+                    {actions}
+                  </View>
```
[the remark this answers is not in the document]
burczu: Replaced with an explicit `dismiss()` passed to the `actions` render function, so dismissal fires for click and keyboard, not just touch.

src/components/Tooltip/RichTooltip.tsx
satya164: > I scoped this to Tooltip.Rich and left the plain Tooltip on cloneElement since you mentioned that's being revisited separately

if you're refactoring the component already, then update it here. by "existing components" i mean other components that you're not working on.

src/components/Tooltip/hooks.ts
```diff
@@ -0,0 +1,112 @@
...
+    childrenWrapperRef.current?.measure(
+      (_x, _y, width, height, pageX, pageY) => {
+        setMeasurement({
+          children: { pageX, pageY, height, width },
+          tooltip: { ...layout },
+          measured: true,
+        });
+      }
+    );
+  };
```
satya164, line 193: use `useLayoutEffect` to synchronously measure when the tooltip gets rendered instead of waiting for `onLayout`

src/components/Tooltip/hooks.ts
```diff
@@ -0,0 +1,112 @@
...
+    : theme.motion.duration[Tokens.motion.exit.duration];
+
+  const animatedStyle = useAnimatedStyle(() => ({ opacity: opacity.value }));
+
+  // Mount as soon as the tooltip is requested.
+  React.useEffect(() => {
+    if (visible) {
+      setRendered(true);
+    }
+  }, [visible]);
```
satya164: don't update state in effects

```suggestion
  if (visible && !rendered) {
    setRendered(true);
  }
```

src/components/Tooltip/hooks.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (!rendered) {
+      return;
+    }
+
+    if (visible) {
+      opacity.value = measurement.measured ? withTiming(1, enterConfig) : 0;
+      return;
+    }
+
+    opacity.value = withTiming(0, exitConfig);
```
satya164: it seems like this can use reanimated css transitions instead of shared values

https://docs.swmansion.com/react-native-reanimated/docs/category/css-transitions/

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,326 @@
...
+  title,
+  content,
+  actions,
+  enterTouchDelay = 100,
+  leaveTouchDelay = 500,
+  titleMaxFontSizeMultiplier,
+  contentMaxFontSizeMultiplier,
+  theme: themeOverrides,
+}: Props) => {
+  const isWeb = Platform.OS === 'web';
```
satya164: don't use `isWeb`. inline `Platform.OS` checks so the bundler can tree-shake unused platform code.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,326 @@
...
+    hideTooltipTimer.current.forEach((t) => clearTimeout(t));
+    hideTooltipTimer.current = [];
+  }, []);
+
+  React.useEffect(() => {
+    return () => {
+      clearShowTimers();
+      clearHideTimers();
+    };
+  }, [clearShowTimers, clearHideTimers]);
```
satya164: this seems unnecessarily complex. a simpler design would clear the hide timer when the tooltip is shown again instead of tracking multiple timers.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,326 @@
...
+    clearShowTimers();
+    setVisible(false);
+  }, [clearShowTimers]);
+
+  const scheduleHide = React.useCallback(() => {
+    clearShowTimers();
+    const id = setTimeout(
+      () => setVisible(false),
+      leaveTouchDelay
+    ) as unknown as NodeJS.Timeout;
```
satya164: don't use `as`. use proper types for the ref

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,326 @@
...
+    clearHideTimers();
+  }, [clearShowTimers, clearHideTimers]);
+
+  // Web: open on hover (with a short enter delay) and on keyboard focus.
+  const handleHoverIn = React.useCallback(() => {
+    clearHideTimers();
+    const id = setTimeout(
+      () => setVisible(true),
+      enterTouchDelay
+    ) as unknown as NodeJS.Timeout;
```
satya164: don't use `as`. use proper types for the ref

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,326 @@
...
+            onPress={hide}
+            pointerEvents={visible ? 'auto' : 'none'}
+            style={StyleSheet.absoluteFill}
+            testID="tooltip-rich-backdrop"
+          />
+          <Animated.View
+            onLayout={onLayout}
+            style={[
+              styles.container,
+              getTooltipPosition(measurement as Measurement),
```
satya164: don't use `as`

satya164 (commented):
the tooltip is broken on web
- it keeps switching between hovered/unhovered
- it doesn't react to clicking on the button, so it's inaccessible if you're using a touchscreen.
- if the whole page is scrolled (the example seems to have a bug that makes it scrollable beyond the cpntent height - though it'd likely happen for normal whole page scroll as well), the tooltip is positioned wrongly

https://github.com/user-attachments/assets/3634be1f-3ad6-4e8b-94d4-920b1655a56e

the animation is broken on android with elevation:

https://github.com/user-attachments/assets/4addd5d6-3ebc-42f0-8c94-f3f75866d27c

the tooltip should also not be attached to the edge. there needs to be some spacing.

there can be multiple tooltips shown at the same time:

https://github.com/user-attachments/assets/1fad7503-9067-4582-a454-75e4e0a21005

the FAB doesn't show tooltips

also please apply safe area insets to the the scrollable view.

src/components/Tooltip/RichTooltip.tsx
burczu: Done for both.

src/components/Tooltip/hooks.ts
burczu: Done for the trigger measurement — moved to `useLayoutEffect` so it fires synchronously. The tooltip's own dimensions still come from `onLayout` because that callback receives the layout values directly from the event, with no extra `.measure()` call needed.

src/components/Tooltip/hooks.ts
burczu: Done.

src/components/Tooltip/hooks.ts
burczu: Done.

src/components/Tooltip/RichTooltip.tsx
burczu: Done.

src/components/Tooltip/RichTooltip.tsx
burczu: Simplified — `show()` clears the hide timer, `hide()`/`scheduleHide()` clear the show timer. Two timers are the minimum for hover debouncing (enter delay + leave delay).

src/components/Tooltip/RichTooltip.tsx
burczu: Ref casts removed. Only `as ViewStyle` in `StyleSheet.create` remains — needed for the platform-conditional spread.

src/components/Tooltip/RichTooltip.tsx
burczu: Ref casts removed. Only `as ViewStyle` in `StyleSheet.create` remains — needed for the platform-conditional spread.

src/components/Tooltip/RichTooltip.tsx
burczu: Ref casts removed. Only `as ViewStyle` in `StyleSheet.create` remains — needed for the platform-conditional spread.

docs/docusaurus.config.js
```diff
@@ -175,6 +175,7 @@ const config = {
           },
           Tooltip: {
             Tooltip: 'Tooltip/Tooltip',
+            TooltipRich: 'Tooltip/RichTooltip',
           },
```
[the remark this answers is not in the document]
burczu: File was removed by the rspress migration — not applicable.

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,326 @@
...
+                ) : null}
+              </Surface>
+            </Pressable>
+          </Animated.View>
+        </Portal>
+      )}
+      <Pressable
+        ref={childrenWrapperRef}
+        style={styles.pressContainer}
+        testID="tooltip-rich-trigger"
```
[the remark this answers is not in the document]
burczu: Done.

tsconfig.json
```diff
@@ -28,8 +28,7 @@
     "skipLibCheck": true,
     "strict": true,
     "target": "esnext",
-    "types": ["node"],
-    "verbatimModuleSyntax": true
```
JKobrynski, line 32: Why remove this one? It was deliberately added in #4999

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,26 +1,42 @@
...
+   * tooltip's show/hide behavior and must be spread onto the returned element:
+   *
+   * ```js
+   * <Tooltip title="Selected Camera">
+   *   {(props) => <IconButton {...props} icon="camera" selected size={24} onPress={() => {}} />}
+   * </Tooltip>
+   * ```
    */
-  children: React.ReactElement;
+  children: (props: TooltipTriggerProps) => React.ReactElement;
```
JKobrynski, line 39: `children` changes from `React.ReactElement` to a render function. That's a hard break for every existing `Tooltip` consumer — your own `TooltipExample.tsx` diff rewrites eleven call sites to prove it — and there's no entry for it in `docs/6.x/docs/guides/migration.md`, which currently only documents the TextInput changes.

I think the change itself is right and I'd keep it: it's what lets you delete the `React.cloneElement` call, and cloning is on the banned list. But the house rule is that a removed API gets deleted outright *with* a migration-notes entry, so please add a `### Tooltip` section to `docs/6.x/docs/guides/migration.md` with the before/after:

```jsx
// before
<Tooltip title="Print"><Appbar.Action icon="printer" onPress={...} /></Tooltip>
// after
<Tooltip title="Print">{(props) => <Appbar.Action {...props} icon="printer" onPress={...} />}</Tooltip>
```

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,435 @@
...
+   * Render function returning the trigger element. The provided props wire the
+   * tooltip's show/hide behavior and must be spread onto the returned element:
+   *
+   * ```js
+   * <Tooltip.Rich content="...">
+   *   {(props) => <IconButton {...props} icon="plus" />}
+   * </Tooltip.Rich>
+   * ```
+   */
+  children: (props: TooltipRichTriggerProps) => React.ReactElement;
```
JKobrynski, line 55: Same here as **Tooltip.tsx:39**

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,435 @@
...
+  );
+
+  return (
+    <>
+      {rendered && (
+        <Portal>
+          <Pressable
+            accessibilityRole="button"
+            accessibilityLabel="Close"
+            accessibilityHint="Dismisses the tooltip"
```
JKobrynski: The legacy `accessibility*` props are the house rule violation — #5005 ("refactor: migrate accessibility props to aria props") just moved the library onto `aria-*`. This should be `role` / `aria-label`, matching what `FAB/Shell.tsx` does.

src/components/FAB/Extended.tsx
```diff
@@ -110,7 +111,10 @@ export type Props = {
    * @optional
    */
   theme?: ThemeProp;
-  ref?: React.Ref<View>;
+  /**
+   * @optional
+   */
+  ref?: React.RefObject<View>;
```
JKobrynski: I think we should stick to `React.Ref` as `React.RefObject` rejects callback refs

src/components/Tooltip/Tooltip.tsx
```diff
@@ -115,128 +114,115 @@ const Tooltip = ({
...
+          </Animated.View>
         </Portal>
       )}
-      <Pressable
+      <View
         ref={childrenWrapperRef}
+        collapsable={false}
         style={styles.pressContainer}
-        {...(isWeb ? webPressProps : mobilePressProps)}
+        testID="tooltip-trigger"
```
JKobrynski: I don't think we should be hardcoding testIDs

src/components/Tooltip/RichTooltip.tsx
```diff
@@ -0,0 +1,435 @@
...
+            onPress={handleBackdropPress}
+            pointerEvents={visible && Platform.OS !== 'web' ? 'auto' : 'none'}
+            style={StyleSheet.absoluteFill}
+            testID="tooltip-rich-backdrop"
```
JKobrynski: There is a bunch of hardcoded testIDs in this file, similar to **Tooltip.tsx** I don't think we should be hardcoding them

JKobrynski (changes requested):

src/components/Tooltip/RichTooltip.tsx
MikitasK: cleaned it up 🧹

src/components/Tooltip/Tooltip.tsx
MikitasK: cleaned it up 👌

src/components/FAB/Extended.tsx
MikitasK: yep, done ✅

tsconfig.json
MikitasK: good catch 👍 
it was removed unintentionally during conflict resolution 
I’ve restored `verbatimModuleSyntax` & updated affected FAB imports to use type-only imports where required

src/components/Tooltip/Tooltip.tsx
MikitasK: added a Tooltip section to the 6.x migration guide with before-and-after examples showing how to migrate from element child to a render function & spread the supplied trigger props ✅

src/components/Tooltip/RichTooltip.tsx
MikitasK: yep, covered this in same migration section

src/components/Tooltip/RichTooltip.tsx
MikitasK: updated backdrop to use `role="button"` & `aria-label="Close"` consistent with current ARIA convention used elsewhere in the library 👌

MikitasK:
- restored `verbatimModuleSyntax` and updated the affected FAB imports to use type-only imports
- documented new tooltip render-prop API in the 6.x migration guide, including examples for `Tooltip` & `Tooltip.Rich`
- updated rich-tooltip dismiss backdrop to use `role` & `aria-label` instead of legacy accessibility properties
- removed hardcoded test IDs from both tooltip implementations & updated affected tests to use semantic queries & observable behavior
- updated extended FAB actions to support both object & callback refs through `React.Ref`
- stabilized native tooltip positioning by ignoring redundant layout updates & snapping coordinates to the physical pixel grid
- fixed Android rich-tooltip shadows so the elevated surface & its shadow fade together without jumping, appearing late, briefly flashing black
- added focused regression coverage for native positioning & Android shadow rendering

https://github.com/user-attachments/assets/d8e0cd31-75b2-4eb3-b1d8-52cea639fac4

cc @JKobrynski @satya164

src/components/FAB/FAB.tsx
```diff
@@ -82,7 +102,7 @@ export type Props = {
    * @optional
    */
   theme?: ThemeProp;
-  ref?: React.Ref<View>;
+  ref?: React.RefObject<View>;
```
JKobrynski: Can we keep `React.Ref`?

MikitasK: done 👌

src/components/Tooltip/tokens.ts
```diff
@@ -0,0 +1,42 @@
...
+} as const;
+
+/**
+ * Rich tooltip — an optional subhead, supporting text and action buttons on a
+ * surface-container container at elevation level 2.
+ * https://m3.material.io/components/tooltips/specs#8e6cf915
+ */
+const rich = {
+  container: 'surfaceContainer',
+  title: 'onSurface',
```
JKobrynski:
```diff
-  title: 'onSurface',
+  title: 'onSurfaceVariant',
```
`md.comp.rich-tooltip.subhead.color` resolves to `on-surface-variant`, the same role as the supporting text. `Tooltip.test.tsx:683` asserts the old value and needs updating with it.

MikitasK: nice catch 👍 updated 👌

src/components/Tooltip/tokens.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+const rich = {
+  container: 'surfaceContainer',
+  title: 'onSurfaceVariant',
+  content: 'onSurfaceVariant',
+  shape: 'medium',
+  elevation: 2,
+  maxWidth: 312,
+  paddingHorizontal: 16,
+  paddingVertical: 12,
```
JKobrynski: Try 12dp top and 8dp bottom - the [specs page](https://m3.material.io/components/tooltips/specs)  gives the rich tooltip asymmetric padding, and this is a symmetric 12.

src/components/Tooltip/hooks.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    childrenWrapperRef.current?.measure(
+      (_x, _y, width, height, pageX, pageY) => {
+        // On web, measure() returns viewport-relative coords but the Portal
+        // container is positioned at the document origin — add scroll offset.
+        const scrollX =
+          Platform.OS === 'web' ? ((window as Window).scrollX ?? 0) : 0;
+        const scrollY =
+          Platform.OS === 'web' ? ((window as Window).scrollY ?? 0) : 0;
```
JKobrynski: Is it possible to avoid `as` here?

src/components/Tooltip/tokens.ts
MikitasK: done

src/components/Tooltip/hooks.ts
MikitasK: yep, done

src/components/__tests__/FABMenu.test.tsx
```diff
@@ -176,11 +181,12 @@ it('calls trigger onPress when menu is closed', async () => {
     />
   );
   // Shell's TouchableRipple uses the default testID 'fab-shell'
-  await userEvent.press(screen.getByTestId('fab-shell'));
+  await user.press(screen.getByTestId('fab-shell'));
```
JKobrynski, line 184: These two tests use test IDs that I don't think exist on main at this point, https://github.com/callstack/react-native-paper/pull/5088 removed default IDs. Can you verify that?

src/index.tsx
```diff
@@ -146,5 +146,9 @@ export type { Props as TextProps } from './components/Typography/Text';
 export type { Props as SegmentedButtonsProps } from './components/SegmentedButtons/SegmentedButtons';
 export type { Props as ListImageProps } from './components/List/ListImage';
 export type { Props as TooltipProps } from './components/Tooltip/Tooltip';
+export type {
+  Props as TooltipRichProps,
+  TooltipRichTriggerProps,
+} from './components/Tooltip/RichTooltip';
```
JKobrynski, line 152: Do we have an entry for `Tooltip.Rich` in the docs site?

JKobrynski:
@MikitasK we've got conflicts

src/index.tsx
MikitasK: good catch, added

src/components/Tooltip/RichTooltip.tsx
JKobrynski: This landed for `Tooltip.Rich` only. The plain `Tooltip` still opens on hover alone, with no `onFocus`/`onBlur`. Could it get the same?

src/components/Tooltip/RichTooltip.tsx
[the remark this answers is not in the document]
JKobrynski: Have you tried the backdrop with a screen reader? I'm afraid it now announces a full-screen "Close" button before the tooltip text. Would `aria-hidden` be better? CC: @MikitasK

src/components/Tooltip/RichTooltip.tsx
JKobrynski: CC: @MikitasK

PR #4995, opened 2026-06-11
feat: modernize progress indicators to latest Material Design specs

adrcotfas:
https://github.com/user-attachments/assets/292472bb-66b2-426b-adeb-089bfbd377a2

### Motivation

Replace the legacy ProgressBar and ActivityIndicator with a token-driven
Material Design 3 progress indicator family. Four components cover the
linear and circular shapes, each in a standard and an expressive "wavy"
variant, with determinate and indeterminate modes.

- Add `LinearProgressIndicator` and `CircularProgressIndicator` (standard)
  and `LinearWavyProgressIndicator` and `CircularWavyProgressIndicator`
  (expressive wave), built on react-native-svg with shared spec tokens.
- `progress` accepts a number or a Reanimated `SharedValue<number>`, so a
   caller-animated value glides on the UI thread with no re-renders.
- Indeterminate timing follows the MD3 motion spec (linear spin + breathing
  sweep for circular; two-segment travel for linear).
- Wavy variants animate a traveling wave that flattens near 0% and 100%;
  `amplitude`, `wavelength`, and `waveSpeed` are configurable. Wave geometry
  is computed in worklets (`wavePath.ts`).
- Migrate the internal loading spinner in Button, IconButton, and Searchbar
  to the standard indeterminate `CircularProgressIndicator`.
- Add `react-native-svg` as a dependency (peer >= 15.0.0).
- Rework the example screen
    
BREAKING CHANGE: The progress components were redesigned for MD3 with no deprecation aliases.
    
- `ProgressBar` and `ProgressBarProps` are removed. Use
  `LinearProgressIndicator` (or `LinearWavyProgressIndicator`). The
  `progress` prop is the same 0..1 range; `indeterminate` replaces the old
  indeterminate behavior.
- `ActivityIndicator` and `ActivityIndicatorProps` are removed. Use
  `CircularProgressIndicator indeterminate`.

### Test plan
- Lint, typescript, existing and new tests pass
- manual visual inspection on all platforms


satya164:
React Native SVG is quite heavy and can cause performance issues. I'd avoid depending on it. Is it possible to use something like https://github.com/software-mansion-labs/react-native-nano-icons to convert these shapes to fonts and use them? Maybe some masking could help with the edges, tho seems very tricky.

satya164:
Maybe we can use variable fonts and manipulate the font metrics to perform the animation. Thanks to @stachbial for the idea. I'll do some experimentation with it.

adrcotfas:
Hey @satya164, do you want me to explore your suggestion or will you take it over since you showed me an alternative implementation?

satya164:
@adrcotfas the font approach won't work with react native because it doesn't expose the necessary style properties to customize variable fonts. and i'm not keen on adding `react-native-svg` dependency for only progress indicator. i need to sit on this one for a while to figure what direction we go.

JKobrynski:
@adrcotfas can you resolve the conflicts?

src/components/Searchbar.tsx
```diff
@@ -301,8 +301,9 @@ const Searchbar = forwardRef<TextInputHandles, Props>(
           {...rest}
         />
         {loading ? (
-          <ActivityIndicator
-            testID="activity-indicator"
+          <CircularProgressIndicator
```
JKobrynski, line 304: The searchbar spinner goes from 24 to 40dp here - the old `ActivityIndicator` defaulted to 24 and `CircularProgressIndicator` defaults to `size: 40`. Your own snapshot shows it. Pass `size={24}` and a matching thickness.

src/components/Button/Button.tsx
```diff
@@ -387,8 +387,10 @@ const Button = (
             </View>
           ) : null}
           {loading ? (
-            <ActivityIndicator
+            <CircularProgressIndicator
+              indeterminate
               size={customLabelSize ?? iconSize}
+              thickness={(customLabelSize ?? iconSize) / 10}
```
JKobrynski, line 393: `size / 10` is the proportional stroke this component exists to remove - `ActivityIndicator.tsx:207` on `main` is the same formula. At the default `iconSize` of 18 that's a 1.8dp stroke against a 4dp token. Same at `IconButton.tsx:218`. If you want a thinner inline spinner, pick one value.

src/components/ProgressIndicator/LinearProgressIndicator/LinearProgressIndicator.tsx
```diff
@@ -0,0 +1,355 @@
...
+        />
+      </>
+    );
+  }
+
+  return (
+    <View
+      onLayout={onLayout}
+      accessible
+      accessibilityRole="progressbar"
```
JKobrynski, line 323: These are the legacy props - [#5005](https://github.com/callstack/react-native-paper/pull/5005) moved the library to `role` / `aria-busy` / `aria-valuemin|max|now`, and the public prop is `'aria-label'`, not `accessibilityLabel`. `ProgressBar.tsx:198-202` on `main` is the shape to copy. Same four lines in the other three components.

src/components/ProgressIndicator/LinearWavyProgressIndicator/LinearWavyProgressIndicator.tsx
```diff
@@ -0,0 +1,390 @@
...
+          LINES.secondHead.duration
+        ) * width;
+      const secondTail =
+        lineFraction(
+          elapsed,
+          LINES.secondTail.delay,
+          LINES.secondTail.duration
+        ) * width;
+      const right = buildStraightPath(
+        clip(firstHead > 0 ? firstHead + gap : min),
```
JKobrynski, line 296: The indeterminate track offsets don't compensate for the round caps the way the determinate branch does at `:314` (`head + gap + 2 * capWidth`), so at the default 4dp thickness the 4dp track/active gap renders as zero, and at thickness 8 the two overlap.

src/components/ProgressIndicator/hooks.ts
```diff
@@ -0,0 +1,78 @@
...
+import { scheduleOnRN } from 'react-native-worklets';
+
+const clamp = (value: number) => Math.min(1, Math.max(0, value));
+
+/**
+ * One clock that drives every indeterminate animation. Returns a `phase` SharedValue that
+ * loops `0 -> 1` over `duration` ms (linear) while `indeterminate`, and holds `0` otherwise.
+ * Deriving every edge/angle from a single clock keeps the moving pieces in sync.
+ */
+export const useIndeterminatePhase = (
```
JKobrynski, line 21: This clock never consults reduce motion and the two flat components don't import `useReduceMotion` at all, so the wavy pair flattens the wave while its sweep keeps travelling. `Switch.tsx` and `Checkbox.tsx` on `main` show the pattern. If a spinner should keep moving, say so and drop the partial handling.

JKobrynski (changes requested):

PR #4996, opened 2026-06-12
chore(deps): upgrade to React Native 0.86

adrcotfas:
Bump react-native and the @react-native/* presets to 0.86.0. React stays at 19.2.3 and there are no native template changes between 0.85.3 and 0.86.0.

Bump react-native-reanimated to 4.4.1 and react-native-worklets to 0.9.2 for RN 0.86 support. In the example, bump react-native-gesture-handler to 2.32.0 (RN 0.86 removed the legacy renderer shim that 2.31.x imported) and align the Expo SDK 56 packages to their latest patch versions.

### Motivation

RN 0.86 comes with `PlatformColor` and `alpha` support for `Pressable`. We need this to fully comply with the MD3 specs.

### Test plan

lint, typescript and tests pass


satya164:
Expo usually only supports specific React Native versions. They can be deeply coupled due to how autolinking works, metro works etc. So we can't bump React Native independently unless Expo mentions that they support the version. Otherwise, we'd be testing with an unsupported setup.

If you run `npx expo-doctor@latest`, it'll show the incompatibilities.

adrcotfas:
> npx expo-doctor@latest

Output is:
```
⚠️ Minor version mismatches
package                       expected  found   
react-native                  0.85.3    0.86.0  
react-native-gesture-handler  ~2.31.1   2.32.0  
react-native-reanimated       4.3.1     4.4.1   
react-native-worklets         0.8.3     0.9.2
```

The bump is so we can proceed with the PlatformColor + alpha for ripples.

Expo 57.0.0 is currently in canary and I can see that the example app works fine.  
How do you recommend to proceed? Wait until full compatibility? Only bump in the root package.json?

satya164:
@adrcotfas if canary works fine then lets update to it

adrcotfas:
Expo 57 canary seems to target RN 0.85.3 so we'll have to wait on this.  The iOS app is broken with the current changes - Android works fine.

JKobrynski:
I will take this over

JKobrynski:
@satya164 I updated this PR, can you take a look?

adam-sajko:
LGTM! 💪 

One thing this doesn't cover. `peerDependencies` still say
`react-native-reanimated: ">=4.3.0"` and `react-native-worklets: ">=0.8.1"`, and per the
[compatibility table](https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility/) 4.3.x stops at RN 0.85

A flat floor can't really express it, `>=4.4.0` would drop RN 0.81 and 0.82 users who are fine on 4.3.x. So I would leave the range alone and put it in the getting started guide instead, the Minimum requirements list there still says >= 4.3.0.

Something like `react-native-reanimated >= 4.3.0 (>= 4.4.0 on React Native 0.86)` and the same for worklets

PR #5000, opened 2026-06-15
feat: add slider component

adrcotfas:
### Motivation

Introduces a new component, the slider.
See https://m3.material.io/components/sliders/overview


Doberjohn:
Leaving two early findings after playing around with the slider in iOS simulator.

1. If we drag both handles to the far right or far left side they get both stuck and cannot move anymore.
2. Disabled toggle does not actually disable any of the slider, only the color changes.

Doberjohn:
Open issue: Dragging slider to the right, enables iOS swipe to go back gesture, both events fire at the same time. Paper has no gesture-handler dependency currently, and the slider is the only PanResponder in the library.

Doberjohn:
> Leaving two early findings after playing around with the slider in iOS simulator.
> 
> 1. If we drag both handles to the far right or far left side they get both stuck and cannot move anymore.
> 2. Disabled toggle does not actually disable any of the slider, only the color changes.

Fixed in 8a2a69233e7ed2b124a8f5c72eca4ef8071a286f

Doberjohn:
> Open issue: Dragging slider to the right, enables iOS swipe to go back gesture, both events fire at the same time. Paper has no gesture-handler dependency currently, and the slider is the only PanResponder in the library.

Fixed in 8badc4f3c

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+    spec,
+    timingConfig,
+    handleInsetStart,
+    handleInsetEnd,
+  };
+
+  // Callbacks are read through their own ref so that a consumer passing an
+  // inline arrow gets the current one rather than the first render's.
+  const propsRef = React.useRef(props);
+  propsRef.current = props;
```
satya164, line 569: refs shouldn't be assigned during render

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+
+  const grantTouchRef = React.useRef(0);
+  // 'pending' means the handles overlapped at touch-down, so which one moves is
+  // decided by the first deliberate drag direction instead.
+  const activeHandleRef = React.useRef<'start' | 'end' | 'pending'>('end');
+
+  const panResponder = React.useRef(
+    PanResponder.create({
```
satya164, line 577: We need to use gesture handler, not `PanResponder` which is JS based.

satya164 (changes requested):
Copy the code from https://github.com/satya164/react-native-range-slider which has a reanimated + gesture handler setup to use it as a base and then tweak it to match material design.

We need to use gesture handler so the slider gestures doesn't run on JS thread. This component needs to be rewritten to match that.

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+  return (
+    <View style={[outerStyle, style]} testID={testID}>
+      {/* Gesture and accessibility wrapper */}
+      <View
+        {...panResponder.panHandlers}
+        style={StyleSheet.absoluteFill}
+        // A bare `View` is not an accessibility element on its own, so the role,
+        // value and actions below would never reach a screen reader without this.
+        accessible
+        accessibilityRole="adjustable"
```
JKobrynski, line 915: `accessibilityValue` isn't in react-native-web's forwarded props, so on web this renders `role="slider"` with no `aria-valuenow`, `aria-valuemin` or `aria-valuemax`. Use the aria props the way `ProgressBar.tsx:200-202` does - [#5005](https://github.com/callstack/react-native-paper/pull/5005) took the legacy ones out everywhere else, and `accessibilityLabel` here is public API. Range also needs one accessible element per handle: `onAccessibilityAction` never calls `onValueChange` for it today.

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+    ? ReduceMotion.Always
+    : ReduceMotion.Never;
+
+  const timingConfig = React.useMemo(
+    () => ({ duration: 100, reduceMotion: reanimatedReduceMotion }),
+    [reanimatedReduceMotion]
+  );
+
+  const initialEndValue = isRange
+    ? (props as RangeValue).value[1]
```
JKobrynski, line 261: Don't use `as` casts. There are many in this PR.

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+
+  const inactiveColor = disabled
+    ? colors.disabledContent
+    : colors.inactiveTrack;
+  const activeColor = disabled ? colors.disabledContent : colors.activeTrack;
+
+  return (
+    <View style={[outerStyle, style]} testID={testID}>
+      {/* Gesture and accessibility wrapper */}
+      <View
```
JKobrynski, line 909: The slider can't be operated by keyboard on web: `slider` isn't one of the roles react-native-web makes a tab stop, and there are no key handlers. MD3 specifies Tab to the handle, arrows to step, Home/End for the ends - https://m3.material.io/components/sliders/accessibility

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+                      activeLead * usableTrackLength +
+                      iconLeadingOffset,
+                  },
+            ]}
+            pointerEvents="none"
+          >
+            <Icon
+              source={icon}
+              size={spec.iconSize}
+              color={colors.stopOnActive}
```
JKobrynski, line 1034: At low values the inset icon sits over the inactive track but stays `on-primary` - white on `secondary-container` measures 1.29:1 in the default light theme. The spec moves the icon onto the inactive track instead, and says not to use it on centered or range sliders at all: https://m3.material.io/components/sliders/guidelines

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+                left: pixelCenter - STOP_SIZE / 2,
+              };
+          return (
+            <View
+              key={f}
+              style={[
+                styles.stopDot,
+                dotStyle,
+                {
+                  backgroundColor: disabled
```
JKobrynski, line 996: Disabled stops on the active track should be `inverse-on-surface` (`md.comp.slider.disabled.active.stop-indicator.container.color`). As written they're `on-surface` at 38% on an `on-surface` 38% track, so they disappear.

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,1119 @@
...
+      disabledContent: c[t.disabledContent],
+    };
+  }, [theme, color]);
+
+  const reanimatedReduceMotion = reduceMotion
+    ? ReduceMotion.Always
+    : ReduceMotion.Never;
+
+  const timingConfig = React.useMemo(
+    () => ({ duration: 100, reduceMotion: reanimatedReduceMotion }),
```
JKobrynski, line 256: `theme.motion.duration.short2` is 100 - use the token rather than a literal. There's no `easing` either, so this falls back to Reanimated's default curve instead of an MD3 one.

src/components/Slider/utils.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * The slider renders left-to-right in every direction: nothing in its handle,
+ * track, corner radii or icon placement is mirrored for RTL. Honouring `isRTL`
+ * here while that is true inverts the drag against the handle, so that dragging
+ * right moves the handle right and lowers the value.
+ *
+ * Flip this to `true` in the same change that mirrors the rendering.
+ */
+const RTL_MIRRORING_IMPLEMENTED = false;
```
JKobrynski, line 43: `RTL_MIRRORING_IMPLEMENTED = false` means the slider renders left-to-right in every locale. "For right-to-left (RTL) languages, this is reversed" - https://m3.material.io/components/sliders/guidelines. Is there a follow-up issue for this, or should it land here?

JKobrynski (changes requested):

Doberjohn:
@satya164 @JKobrynski I saw your comments thanks for taking the time to review! I will address them as soon as possible!

PR #5005, opened 2026-06-17
refactor: migrate accessibility props to aria props

MrMuzyk:

### Motivation

<!-- Explain the **motivation** for making this change. What existing problem does the pull request solve? -->

React Native added aria-* props as web-compatible aliases for the legacy accessibility* props. This PR migrates the library's internal usage across all components to the modern API, so components work correctly with React Native Web and align with the direction RN is taking the accessibility API.

### Changes

Mapping applied uniformly:                                                                                                                                              
                                                                                                                                                 
- `accessibilityLabel` → `aria-label`                                                                                                                                   
- `accessibilityRole` → `role`                                                                                                                                          
- `accessibilityState={{ disabled }}` → `aria-disabled`                                                                                                                 
- `accessibilityState={{ selected }}` → `aria-selected`                                                                                                                 
- `accessibilityState={{ checked }}` → `aria-checked`                                                                                                                   
- `accessibilityState={{ expanded }}` → `aria-expanded`                                                                                                                 
                                                                                                                                                 
`TextInput` additionally replaces `accessibilityState: { disabled, invalid }` with `aria-disabled` and `aria-invalid`. Because `aria-invalid` is not yet in RN's `AccessibilityProps` type (RN 0.85.3), `GetAccessibilityDataReturn.input` uses an intersection type `AccessibilityProps & { 'aria-invalid'?: boolean }`.                                                                                          
                                                                                                                                                 
BREAKING CHANGE: `aria-*` props take precedence over `accessibility*` props in React Native when both are set on the same element. Consumers who pass `accessibilityLabel`, `accessibilityRole`, or `accessibilityState` to Paper components to override defaults will be silently ignored after this bump.  

### Related issue

<!-- If this pull request addresses an existing issue, link to the issue. If an issue is not present, describe the issue here. -->

N/A

### Test plan

<!-- Describe the **steps to test this change**, so that a reviewer can verify it. Provide screenshots or videos if the change affects UI. -->

<!-- Keep in mind that PR changes must pass lint, typescript and tests. -->

`yarn typescript` clean                                                                                                                                               
`yarn lint` clean                                                                                                                                                     
`yarn jest` — 9 test files updated to query/assert on `aria-*` props instead of `accessibilityState`/`accessibilityLabel`; 28 snapshot files regenerated to reflect the prop rename in the render tree. Full suite passes.


satya164 (approved):
Thank you!

PR #5006, opened 2026-06-18
feat(navigation-bar): modernize to Material Design 3

### LINKED ISSUE #4975: feat: modernize BottomNavigationBar
Modernize the existing BottomNavigationBar to reuse theme tokens (color, shape, motion, typography etc) and make sure the MD3 specs are respected. See the expressive update too: https://m3.material.io/components/navigation-bar/overview

# Current State

## Bottom Navigation

**MD3 spec:** Navigation bar
**Specs link:** https://m3.material.io/components/navigation-bar/specs

### Notes

- `BottomNavigation` is a router-aware wrapper around `BottomNavigation.Bar` that handles scene management and tab transitions. MD3 compliance is governed by `BottomNavigation.Bar` below.

---

## Bottom Navigation Bar

**MD3 spec:** Navigation bar
**Specs link:** https://m3.material.io/components/navigation-bar/specs

### Color tokens

- Container `surfaceContainer` ✅
- Active icon `onSecondaryContainer` ✅
- Active indicator `secondaryContainer` ✅
- Inactive icon `onSurfaceVariant` ✅
- Inactive label `onSurfaceVariant` ✅
- [ ]  Active label uses `onSurface`. Spec token `Nav bar item active label text color` maps to `secondary`

### Spacing / sizing

- Active indicator 64x32dp with 16dp corner radius ✅
- Icon-to-label gap 4dp (`marginBottom: 4` on icon container) ✅
- [ ]  `v3NoLabelContainer` height is hardcoded to 80dp. Spec requires 64dp
- [ ]  `BAR_HEIGHT = 56` constant is used for label height styling. Spec bar height is 64dp

### Behavior

- [ ]  `shifting` mode (inactive tabs hide labels) is an MD2 pattern not present in MD3
- `labeled={false}` hides all labels (not in MD3 spec, but a deliberate extension used by Google’s own products e.g. Gmail) ✅
- [ ]  `rippleColor: 'transparent'` suppresses state layer feedback. MD3 requires visible state layers (8% hovered, 10% focused/pressed)
- [ ]  Flexible navigation bar (horizontal items for medium windows) not implemented. M3 Expressive requires horizontal layout support
- [ ]  Baseline navigation bar is not recommended in M3 Expressive; flexible variant should replace it

### Notes

- Rename to match MD3: **`NavigationBar`**

burczu:
### Motivation

\`BottomNavigation.Bar\` had drifted from the current [Material Design 3 navigation bar spec](https://m3.material.io/components/navigation-bar/specs), including the M3 Expressive update. As part of the v6 MD3 modernization effort (alongside TextInput, Switch, Checkbox, FAB), this brings it up to spec, renames it, and adds the flexible (horizontal) variant:

- Active label used the wrong color role (\`onSurface\` instead of **\`secondary\`**), and the bar height (56dp) and no-label container height (80dp) didn't match the **64dp** spec.
- \`rippleColor: 'transparent'\` suppressed all interaction feedback — MD3 requires visible **state layers**.
- \`shifting\` is a Material Design 2 pattern with no MD3 equivalent.
- The **flexible navigation bar** (horizontal items for medium windows) wasn't supported, and the component name (\`BottomNavigation.Bar\`) no longer matches the spec's "navigation bar".

### Related issue

Closes #4975.

### What changed

- **New top-level \`NavigationBar\` export** — the MD3 *flexible* navigation bar, intended for direct use with React Navigation or any custom navigator. \`BottomNavigation\` (the scene-transition wrapper) is removed; most apps use a navigation library for scene management anyway.
- **Spec fixes (token-driven):** active label \`onSurface\` → **\`secondary\`**, bar height 56 → **64dp**, no-label container 80 → **64dp**. Values extracted to \`NavigationBar/tokens.ts\`.
- **Visible MD3 state layers:** pill-shaped hover (8%) / focus / press (10%) layers via the shared \`getStateLayer\` util, replacing the suppressed ripple. Keyboard focus is represented by the focus state layer.
- **Motion:** durations/easings now come from the motion tokens, and the active indicator animates with the M3-Expressive spring on selection.
- **Flexible \`variant\` prop** (\`'stacked' | 'horizontal'\`): horizontal lays the icon beside the label inside a content-hugging indicator pill, for medium-width windows. No-op without labels.
- **Internal cleanup:** collapsed a vestigial active/inactive cross-fade, deduped item rendering across both layouts, and consolidated color tokens — all behavior-preserving and guarded by characterization tests.
- **Example + docs:** new \`NavigationBarExample\` with a responsive stacked↔horizontal toggle; JSDoc for \`NavigationBar\` including the flexible-vs-original migration note.

### Test plan

- \`yarn test\` — 715 passing / 159 snapshots. \`BottomNavigation.test.tsx\` removed; \`NavigationBar.test.tsx\` covers variant layouts, state-layer opacities, per-state label/indicator colors, and badge rendering.
- \`yarn typescript\`, \`yarn lint\` — clean.
- Verified on the iOS simulator (iPhone 17 Pro):
  - Stacked: MD3 colors/sizing, indicator springs on tab change, state layers on interaction.
  - Horizontal: icon beside label with a hugging pill; indicator follows selection.

### Videos

https://github.com/user-attachments/assets/da091e3f-61b2-40ab-976a-24079c1eaad9

https://github.com/user-attachments/assets/2f31233b-9402-4d31-9a98-3ac4bddccea0

https://github.com/user-attachments/assets/29fd73e5-8f8c-41cc-b79d-c3f24c9c2241


docs/docusaurus.config.js
```diff
@@ -148,6 +147,9 @@ const config = {
             MenuItem: 'Menu/MenuItem',
           },
           Modal: 'Modal',
+          NavigationBar: {
+            NavigationBar: 'NavigationBar/NavigationBar',
+          },
```
satya164: Is the `BottomNavigation` component still needed? it'd be inconsistent to have a mix of old and new APIs

src/components/BottomNavigation/BottomNavigation.tsx
```diff
@@ -612,6 +601,11 @@ BottomNavigation.SceneMap = <Route extends BaseRoute>(scenes: {
   );
 };
 
+/**
+ * @deprecated Use the top-level `NavigationBar` export instead.
+ * `BottomNavigation.Bar` is the M3 "original" navigation bar, superseded by the
+ * flexible `NavigationBar`. Kept as an alias for backwards compatibility.
+ */
```
satya164: There shouldn't be any deprecations. Remove any old code entirely

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,942 @@
+import * as React from 'react';
+import {
+  Animated,
+  Easing,
```
satya164: Lets move to reanimated

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,942 @@
...
+   */
+  activeColor?: string;
+  /**
+   * Custom color for icon and label in the inactive tab.
+   */
+  inactiveColor?: string;
+  /**
+   * The scene animation Easing.
+   */
+  animationEasing?: EasingFunction | undefined;
```
satya164: not sure this needs to be customizable. lets remove it

satya164 (changes requested):
Looks like this now removes the separate bar component and combines both the view and bar to a single component. It's important that the bar is standalone so it can be easily used in react navigation.

The view that animates screen transitions may not even be needed as react navigation also has transition animations for tab switches.

Also since you're working on UI components, please share videos on android, ios and web.

src/components/BottomNavigation/BottomNavigation.tsx
burczu: Done in 13d34e9f5 — removed the `BottomNavigation.Bar` alias, the `BottomNavigationBar` module, and the dead no-op `shifting` prop. No deprecated surface left.

src/components/NavigationBar/NavigationBar.tsx
burczu: Done in c6bc35f36 — migrated the bar's motion to react-native-reanimated (shared values + worklets). This also let me delete the per-tab `Animated.Value` array, `animateToIndex`, and the old native-driver reset workaround; each item now springs its own selection progress from its `focused` prop.

src/components/NavigationBar/NavigationBar.tsx
burczu: Done in c6bc35f36 — removed `animationEasing` (and its `BottomNavigation` passthrough). Selection always uses the M3-expressive spring now; reduce-motion is honored via reanimated's `ReduceMotion`.

docs/docusaurus.config.js
burczu: Kept intentionally. `BottomNavigation` is the router-integrated scene component (owns tab screens + transitions); `NavigationBar` is the standalone bar for use with React Navigation's own navigator — same layering split as React Navigation itself, so it isn't an old-vs-new API mix. Every deprecation/alias is gone (13d34e9f5), so there's no MD2 surface remaining.

src/components/BottomNavigation/BottomNavigationBar.tsx
```diff
@@ -1,862 +1,8 @@
...
- *         getLabelText={({ route }) => route.title}
- *       />
- *     </Provider>
- *   );
- * }
- * ```
+ * @deprecated Use `NavigationBar` instead. `BottomNavigation.Bar` is the M3
+ * "original" navigation bar and has been superseded by the flexible
+ * `NavigationBar`. This module re-exports `NavigationBar` for backwards
+ * compatibility and will be removed in a future major version.
```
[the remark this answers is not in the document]
burczu: Resolved in 13d34e9f5 — the `BottomNavigationBar` module is deleted entirely (no deprecated alias).

src/components/BottomNavigation/BottomNavigation.tsx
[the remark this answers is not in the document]
burczu: Resolved in 13d34e9f5 — the deprecation block is removed; `BottomNavigation` renders `NavigationBar` directly.

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,942 @@
...
+              })
+            : Animated.spring(tabsAnims[i], {
+                toValue,
+                ...toRawSpring(motion.spring.fast.spatial),
+                useNativeDriver: true,
+              });
+        })
+      ).start(() => {
+        // Workaround a bug in native animations where this is reset after first animation
+        tabsAnims.map((tab, i) => tab.setValue(i === index ? 1 : 0));
```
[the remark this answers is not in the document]
burczu: Fixed in c6bc35f36 — that whole `animateToIndex` path (and the `.map` side-effect) is gone with the reanimated migration.

src/components/__tests__/BottomNavigation.test.tsx
```diff
@@ -449,36 +388,174 @@ it('does not apply maxTabBarWidth styling if compact prop is falsy', () => {
...
+        routes: [
+          { key: 'a', title: 'Alpha', focusedIcon: 'magnify' },
+          { key: 'b', title: 'Beta', focusedIcon: 'camera' },
+        ],
+      }}
+      onTabPress={jest.fn()}
+    />
+  );
+
+  // Each tab renders an active and inactive label layer, so both match.
```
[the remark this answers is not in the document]
burczu: Fixed in 13d34e9f5 — comment now reads "Each tab renders a single label (no cross-fade layers)."

burczu:
Thanks @satya164 — addressed the round:

**Standalone bar.** The bar *is* standalone: `NavigationBar` is its own module and the top-level export; `BottomNavigation` just consumes it. The earlier re-export shim made it look merged — that's gone now, and I migrated the React Navigation example to use `NavigationBar` directly to show the standalone path.

**Scene-transition view.** Left `BottomNavigation` in place — it's still useful when not relying on React Navigation's own tab transitions. Happy to drop it in a follow-up if you'd rather not ship both.

**No deprecations / old code.** Removed the `.Bar` alias, the `BottomNavigationBar` module, the no-op `shifting` prop, and `animationEasing` (13d34e9f5, c6bc35f36).

**Reanimated.** Motion is fully on react-native-reanimated now (c6bc35f36).

**Videos** (android/ios/web) — added to the description.

docs/docusaurus.config.js
satya164: > Kept intentionally. BottomNavigation is the router-integrated scene component (owns tab screens + transitions); NavigationBar is the standalone bar for use with 

then the naming is inconsistent. both should use the same naming scheme.

satya164:
> Left BottomNavigation in place — it's still useful when not relying on React Navigation's own tab transitions

i'd rather remove it. most people use navigation libraries which they would want to integrate the navigation bar with. so it's a very niche component and mostly unnecessary.

burczu:
Done — `BottomNavigation` is removed entirely in the latest push.

docs/docusaurus.config.js
burczu: Agreed — removed `BottomNavigation` entirely, so there's no naming mismatch anymore. `NavigationBar` is the only public surface.

CatLover01:
Minor issue with the MD3 state layer: in the press-recording(`navbar_and_small.mp4`, the ios video seems fine)  you can see the pressed state renders as a sharp-cornered rectangle instead of the rounded pill. I believe it should match the active indicator's shape (64×32, radius 16).

src/components/NavigationBar/tokens.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+// Container heights (dp). The flexible navigation bar is shorter than the
+// original navigation bar (which was 80dp).
+export const BAR_HEIGHT = 64;
+export const NO_LABEL_BAR_HEIGHT = 64;
+
+// Active indicator / state-layer pill (dp). Matches the M3 `large` corner (16).
+export const INDICATOR_WIDTH = 64;
+export const INDICATOR_HEIGHT = 32;
```
MikitasK, line 15: what about using separate M3 Expressive indicator dimensions?

current sizes are 64×32 for stacked & 32dp high for horizontal, but spec defines 56×32 stacked & HUG×56 horizontal ([source](https://github.com/material-components/material-components-android/blob/master/docs/components/BottomNavigation.md#active-indicator-attributes))

so could we update it to:
```suggestion
export const INDICATOR_WIDTH = 56;
export const INDICATOR_HEIGHT = 32;
export const HORIZONTAL_INDICATOR_HEIGHT = 56;
```

and then use `HORIZONTAL_INDICATOR_HEIGHT` in `horizontalItem`

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,887 @@
...
+  horizontalContainer: {
+    height: BAR_HEIGHT,
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  horizontalItem: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'center',
+    height: INDICATOR_HEIGHT,
```
MikitasK: check out [this](https://github.com/callstack/react-native-paper/pull/5006/changes#r3865675351) comment
```suggestion
    height: HORIZONTAL_INDICATOR_HEIGHT,
```

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,887 @@
...
+    borderless: true,
+    centered: true,
+    rippleColor: 'transparent',
+    onPress,
+    onLongPress,
+    onPressIn: () => setPressed(true),
+    onPressOut: () => setPressed(false),
+    onHoverIn: () => setHovered(true),
+    onHoverOut: () => setHovered(false),
+    onFocus: () => setKeyboardFocused(true),
```
MikitasK: wdyt about gating focus state with `isKeyboardFocusEvent`?

on web `onFocus` also fires after pointer clicks, leaving the focus layer visible. 
existing [`isKeyboardFocusEvent`](https://github.com/callstack/react-native-paper/blob/main/src/utils/isKeyboardFocusEvent.ts) helper already handles `:focus-visible` and is used by `Switch`/`Checkbox`

```suggestion
    onFocus: (event) => {
      if (isKeyboardFocusEvent(event)) { 
          setKeyboardFocused(true);
        }
      },
```


`onFocus` field should also use `TouchableRippleProps['onFocus']` for its event type

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,887 @@
...
+    ...(Platform.OS === 'web'
+      ? {
+          width: '100%',
+        }
+      : null),
+  },
+  item: {
+    flex: 1,
+    paddingVertical: 0,
+  },
```
MikitasK, line 825: could we make horizontal items content-sized instead of inheriting `flex: 1`?

```
style: horizontal ? styles.horizontalTouchable : styles.item;
```


M3 Expressive specifies content-based item widths with maximum ([source](https://github.com/material-components/material-components-android/blob/master/docs/components/BottomNavigation.md#m3-expressive-styles))
currently they expand across available width (particularly on web)


```suggestion
  item: {
    paddingVertical: 0,
  },
  stackedTouchable: {
    flex: 1,
  },
  horizontalTouchable: {
    flexGrow: 0,
    flexShrink: 1,
    maxWidth: MAX_TAB_WIDTH,
  },
```

`horizontalItems` could be applied to the items wrapper for the horizontal variant

MikitasK: done

src/components/NavigationBar/tokens.ts
MikitasK: done

MikitasK (approved):
- updated stacked & horizontal active-indicator dimensions to match M3 Expressive tokens
- made horizontal navigation items content-sized, centered and capped at `168px`
- fixed horizontal items collapsing to zero width on web
- kept horizontal labels on 1 line with tail truncation on narrow screens
- limited focus state layers to keyboard-originated focus events on web
- moved consumer layout styles to the outer animated wrapper while keeping visual styles on the `Surface`
- clipped pressed, hovered & focused state layers to rounded indicator shape (good catch @CatLover01 👍)

https://github.com/user-attachments/assets/362fb318-c5aa-4fe1-8397-9fe1bc99a7a0

cc @JKobrynski @satya164

src/components/NavigationBar/tokens.ts
JKobrynski: Use 40dp - `md.comp.nav-bar.item.horizontal.active-indicator.height` on [m3.material.io](https://m3.material.io/components/navigation-bar/specs). The 56dp is androidx's `expandedHeight`, and where the two disagree m3 governs; the stacked move to 56 is right either way. `NavigationBar.tsx:915` and `NavigationBar.test.tsx:82` pin 56 too. And since `md.comp.nav-bar.item.active-indicator.shape` is `corner.full`, `cornerFull` from `theme/tokens/sys/shape` would replace both radius literals rather than tracking half the height by hand.

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,935 @@
...
+      badge != null && typeof badge !== 'boolean'
+        ? String(badge).length * -2
+        : 0,
+  };
+
+  const font = theme.fonts.labelMedium;
+
+  // MD3 state layer: visible on hover (8%) and focus/press (10%). Active items
+  // use the on-secondary-container role, inactive the on-surface-variant role.
+  const stateLayerRole = focused
```
JKobrynski: `md.comp.nav-bar.item.inactive.{hovered,focused,pressed}.state-layer.color` is
`on-secondary-container` — the same role as active, so this is a constant, not a `focused` branch. Baseline says `on-surface`; neither namespace says `on-surface-variant`. The comment above needs the same fix.

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,935 @@
...
+    onHoverIn: () => setHovered(true),
+    onHoverOut: () => setHovered(false),
+    onFocus: (event) => {
+      if (isKeyboardFocusEvent(event)) {
+        setKeyboardFocused(true);
+      }
+    },
+    onBlur: () => setKeyboardFocused(false),
+    testID: itemTestID,
+    accessibilityLabel: getAccessibilityLabel({ route }),
```
JKobrynski: [#5005](https://github.com/callstack/react-native-paper/pull/5005) moved the library to aria props - `BottomNavigation/BottomNavigationBar.tsx:592-594` on `main` already reads `'aria-label'` / `role` / `'aria-selected'`, and `:514` is `role={'tablist'}`. `BaseRoute` should keep `'aria-label'` too.

src/components/NavigationBar/tokens.ts
MikitasK: oh yep, good catch
updated it 👌

src/components/NavigationBar/NavigationBar.tsx
MikitasK: updated - active & inactive state layers now use `onSecondaryContainer`

src/components/NavigationBar/NavigationBar.tsx
MikitasK: yep, sure, updated it to match ARIA conventions introduced in #5005

src/index.tsx
```diff
@@ -37,6 +36,7 @@ export { default as Icon } from './components/Icon';
 export { default as IconButton } from './components/IconButton/IconButton';
 export { default as Menu } from './components/Menu/Menu';
 export { default as Modal } from './components/Modal';
+export { default as NavigationBar } from './components/NavigationBar';
```
JKobrynski, line 39: `BottomNavigation`, `BottomNavigation.Bar`, `SceneMap`, `renderScene`, `onIndexChange`, `getLazy`, `shifting`, `barStyle` and the `sceneAnimation*` props all go away here, I think we should mention that in `docs/6.x/docs/guides/migration.md`.

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,930 @@
...
+    top?: number;
+    right?: number;
+    bottom?: number;
+    left?: number;
+  };
+  /**
+   * Specifies the largest possible scale a label font can reach.
+   */
+  labelMaxFontSizeMultiplier?: number;
+  style?: RNAnimated.WithAnimatedValue<StyleProp<ViewStyle>>;
```
JKobrynski: Is there not a way to get rid of RN's Animated here? Like e.g. `StyleProp<ViewStyle>`? I think it would also resolve the type suppression on line 696

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,930 @@
...
+  style,
+  activeIndicatorStyle,
+  labeled = true,
+  variant = 'stacked',
+  onTabPress,
+  onTabLongPress,
+  safeAreaInsets,
+  labelMaxFontSizeMultiplier = 1,
+  compact: compactProp,
+  testID = 'bottom-navigation-bar',
```
JKobrynski: Drop the default `testID`

src/index.tsx
MikitasK: done

src/components/NavigationBar/NavigationBar.tsx
MikitasK: done

src/components/NavigationBar/NavigationBar.tsx
MikitasK: done

docs/6.x/docs/guides/migration.md
```diff
@@ -86,6 +86,40 @@ The `style` props for `Appbar` and `Appbar.Header` no longer accept `Animated.Va
...
++   {renderScene({ route: routes[index] })}
++   <NavigationBar
++     navigationState={{ index, routes }}
++     onTabPress={({ route }) => setIndex(routes.findIndex((r) => r.key === route.key))}
++     style={{ backgroundColor: theme.colors.surfaceContainer }}
++   />
++ </>
+```
+
+`NavigationBar` also does not set a default `testID`. If your tests looked up `bottom-navigation-bar` (or its `-container` / `-content` / `-content-wrapper` children), pass a `testID` explicitly.
```
JKobrynski, line 121: I think we need one more docs change - `docs/6.x/docs/guides/bottom-navigation.md` still imports `BottomNavigation` and renders it, which is what this PR removes right? I think we should either rewrite it to include the new `NavigationBar` or drop it completely and point at at the migration entry.

docs/6.x/docs/guides/migration.md
```diff
@@ -86,6 +86,40 @@ The `style` props for `Appbar` and `Appbar.Header` no longer accept `Animated.Va
...
+
+| v5                                                                    | v6                                                                    |
+| --------------------------------------------------------------------- | --------------------------------------------------------------------- |
+| `BottomNavigation`, `BottomNavigation.Bar`                            | `NavigationBar`                                                       |
+| `BottomNavigationProps`, `BottomNavigationRoute`                      | `NavigationBarProps`, `NavigationBarRoute`                            |
+| `BottomNavigation.SceneMap`, `renderScene`, `onIndexChange`           | render the active scene yourself, alongside the bar                   |
+| `getLazy`, `route.lazy`                                               | removed — the bar no longer owns the scenes                           |
+| `sceneAnimationEnabled`, `sceneAnimationType`, `sceneAnimationEasing`  | removed with the scene container                                      |
+| `shifting`                                                            | removed (it was already a no-op)                                      |
```
JKobrynski, line 100: I don't think it was a no-op, on main it still shifts the icon, fades the inactive label and hides labels on unselected tabs right?

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,921 @@
...
+    onHide: handleKeyboardHide,
+  });
+
+  const eventForIndex = (index: number) => {
+    const event = {
+      route: navigationState.routes[index],
+      defaultPrevented: false,
+      preventDefault: () => {
+        event.defaultPrevented = true;
+      },
```
JKobrynski, line 680: Is `preventDefault()` still necessary?

JKobrynski (commented):
A few comments + we've got conflicts

JKobrynski:
@MikitasK we've got conflicts

docs/6.x/docs/guides/migration.md
MikitasK: done

docs/6.x/docs/guides/migration.md
MikitasK: corrected

src/components/NavigationBar/NavigationBar.tsx
MikitasK: good point, removed

PR #5007, opened 2026-06-19
feat: modernize RadioButton for MD3

### LINKED ISSUE #4938: refactor(radio-button): modernize selection model and improve MD3 compliance
Refactor RadioButton to improve Material Design 3 compliance, remove problematic children patterns, improve performance characteristics, and modernize the public API.

**Problems**

**Focus Areas**

- Improve RadioButton.Group architecture
- Improve spacing and density consistency
- Simplify internal selection propagation

**Proposed API Direction**

`<RadioButton.Group
  value={value}
  onValueChange={setValue}
/>`

**Performance Opportunities**

- Reduce context-driven rerenders
- Improve group memoization
- Simplify animation paths

burczu:
### Motivation

`RadioButton` still shipped as two platform-specific implementations (`RadioButtonAndroid`, `RadioButtonIOS`) with the iOS variant rendering a checkmark glyph rather than a Material radio. This is inconsistent with the rest of the v6 MD3 work (Checkbox, Switch, TextInput, FAB) and diverges from the [M3 radio button spec](https://m3.material.io/components/radio-button/specs).

This PR unifies RadioButton into a single MD3 component that mirrors the modernized **Checkbox** structure (single component, `tokens.ts`, `error` prop, `accessible={false}` inner control), without pulling in Checkbox's reanimated/focus-ring machinery (out of scope — radio ships with none today).

### Related issue

Closes #4938

### What changed

- **Unified component** — merged `RadioButtonAndroid` + `RadioButtonIOS` into a single MD3 `RadioButton` (Android impl as base). Dropped `RadioButton.Android`, `RadioButton.IOS`, `RadioButtonIOS`, and the `RadioButton.Item` `mode` prop.
- **Tokens** — new `RadioButton/tokens.ts` (MD3 dims `20`/`10`/`2` + color roles); colors resolved via tokens in `utils`.
- **`error` prop** — added to `RadioButton` and `RadioButton.Item` (ring + dot use `theme.colors.error`; `disabled`/custom colors take precedence).
- **Single animation path (reanimated)** — collapsed two RN `Animated.Value`s into one shared value and migrated to `react-native-reanimated` (matching Checkbox): dot scales in with overshoot on selection, keyed on `checked` so it also animates inside a `RadioButton.Group`. Uses `theme.motion` MD3 duration/easing tokens and respects reduce-motion.
- **Disabled opacity** — the control now dims to 38% (`selectionControlOpacity` was previously computed but never applied).
- **Group perf** — memoized the provider value + stabilized `onValueChange` with `use-latest-callback`; consumers switched from `Context.Consumer` to `useContext`.
- **Item a11y** — inner control is `accessible={false}` + container `importantForAccessibility="no-hide-descendants"`, so a screen reader sees **one** radio per row; dropped per-radio `accessibilityLiveRegion`; `labelMaxFontSizeMultiplier=1.5` + `${testID}-text` for CheckboxItem parity.
- **Docs/examples** — updated examples (Group/Item, leading/trailing, error checked+unchecked, disabled); removed the iOS-mode demo and the `RadioButtonAndroid`/`RadioButtonIOS` doc pages.

### Breaking changes

- Removed `RadioButton.Android`, `RadioButton.IOS`, `RadioButtonIOS` (and `RadioButtonAndroidProps`/`RadioButtonIOSProps` exports).
- Removed the `RadioButton.Item` `mode` prop.
- `error` prop added (no visual change for existing usage).

### Test plan

- `yarn typescript`, `yarn lint`, `yarn test` (snapshots updated) — green.
- `yarn --cwd docs build` — green.
- Manually verified on **iOS sim, Android emulator, and web**: single selection, dot scale-in (standalone **and** in a Group), error (checked/unchecked), disabled dimming, leading/trailing, ripple, single a11y node per Item row.

### Videos

https://github.com/user-attachments/assets/2852a49b-b384-4ead-8c75-953c9cf63c95

https://github.com/user-attachments/assets/442cae35-fb7d-440b-948a-11b58ef16123

https://github.com/user-attachments/assets/d762bde8-d74f-4817-b14e-afae12acb112


satya164:
can you rebase the PR against main?

src/components/RadioButton/RadioButtonItem.tsx
```diff
@@ -184,53 +179,53 @@ const RadioButtonItem = ({
...
+        pointerEvents="none"
+        importantForAccessibility="no-hide-descendants"
+      >
+        {isLeading && radioButton}
+        <Text
+          variant={labelVariant}
+          testID={`${testID}-text`}
+          style={[styles.label, computedStyle, labelStyle]}
+          maxFontSizeMultiplier={labelMaxFontSizeMultiplier}
+        >
```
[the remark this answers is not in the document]
burczu: fixed

src/components/RadioButton/RadioButton.tsx
```diff
@@ -71,16 +90,159 @@ export type Props = {
...
+          value,
+          event,
+        });
+      }}
+      disabled={disabled}
+      {...accessibilityProps}
+      style={styles.container}
+      testID={testID}
+      theme={theme}
+    >
```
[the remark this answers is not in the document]
burczu: fixed

src/components/RadioButton/RadioButton.tsx
```diff
@@ -22,7 +33,7 @@ export type Props = {
   /**
    * Function to execute on press.
    */
-  onPress?: (e: GestureResponderEvent) => void;
+  onPress?: (param?: any) => void;
   /**
```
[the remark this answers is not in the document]
burczu: done

src/components/RadioButton/RadioButtonItem.tsx
```diff
@@ -147,35 +145,32 @@ const RadioButtonItem = ({
...
-    radioButton = <RadioButtonAndroid {...radioButtonProps} />;
-  } else if (mode === 'ios') {
-    radioButton = <RadioButtonIOS {...radioButtonProps} />;
-  } else {
-    radioButton = <RadioButton {...radioButtonProps} />;
-  }
+  // The outer TouchableRipple is the interactable element + a11y radio; the
+  // inner control is purely visual, so exclude it from the a11y tree to
+  // avoid duplicate `checked` states.
+  const radioButton = <RadioButton {...radioButtonProps} accessible={false} />;
```
MikitasK: what about preventing visual-only radio from receiving keyboard focus?

`accessible={false}` removes its accessibility semantics, but updated snapshot still shows `focusable={true}` ([source](https://github.com/callstack/react-native-paper/blob/ea2a6e99422f86025ee152423f302f89a411aefd/src/components/__tests__/RadioButton/__snapshots__/RadioButtonItem.test.tsx.snap#L77-L81))


additionally, `react-native-web` defaults enabled `Pressable` to `tabIndex={0}` ([RNW sourcecode](https://github.com/necolas/react-native-web/blob/0.21.2/packages/react-native-web/src/exports/Pressable/index.js#L193-L214)), leaving unlabeled second tab stop inside `RadioButton.Item`

so what about this change?


```suggestion
  const radioButton = <RadioButton {...radioButtonProps} accessible={false} focusable={false} tabIndex={-1} />;
```

src/components/RadioButton/tokens.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * MD3 Radio button spec dimensions and color-role tokens.
+ * @see https://m3.material.io/components/radio-button/specs
+ */
+const sizes = {
+  ringSize: 20,
+  dotSize: 10,
+  outlineWidth: 2,
+} as const;
```
MikitasK, line 12: what about adding an explicit state-layer size token & sizing touchable accordingly?

current 20dp ring with 8dp margins produces 36dp state layer, while MD3 defines radio state layer as 40dp.
modernized `Checkbox` component already uses the same 40dp state-layer token ([source](https://github.com/callstack/react-native-paper/blob/df3cdfd7a1feb48260b946d6c0fd63cced020ebc/src/components/Checkbox/tokens.ts#L11))

so could we change it to:
```suggestion
const sizes = {
  ringSize: 20,
  dotSize: 10,
  outlineWidth: 2,
  stateLayerSize: 40,
} as const;
```
and use like that in `RadioButton.tsx`:
```tsx
const {
  ringSize,
  dotSize,
  outlineWidth: OUTLINE_WIDTH,
  stateLayerSize,
} = RadioButtonTokens;

  container: {
    width: stateLayerSize,
    height: stateLayerSize,
    borderRadius: stateLayerSize / 2,
    alignItems: 'center',
    justifyContent: 'center',
  },
  radio: {
    height: ringSize,
    width: ringSize,
    borderRadius: ringSize / 2,
    borderWidth: OUTLINE_WIDTH,
  },
});
```

MikitasK (approved):
- updated radio-button state layer to MD3 specified 40dp size while keeping visible ring centered at 20dp
- added reusable `stateLayerSize` token & removed previous margin-based state-layer sizing
- made nested visual control inside `RadioButton.Item` non-focusable with `focusable={false}` and `tabIndex={-1}` keeping outer item as the only interactive radio
- added regression coverage for `40dp` state layer & updated affected accessibility & component snapshots

src/components/RadioButton/RadioButton.tsx
```diff
@@ -71,16 +97,171 @@ export type Props = {
...
+  // which owns the a11y tree for the wrapped row), suppress our own role and
+  // state so the same logical control doesn't expose two `checked` states to
+  // assistive tech.
+  const accessibilityProps =
+    rest.accessible === false
+      ? {}
+      : {
+          accessibilityRole: 'radio' as const,
+          accessibilityState: { disabled: !!disabled, checked },
+        };
```
JKobrynski, line 196:
```diff
-          accessibilityRole: 'radio' as const,
-          accessibilityState: { disabled: !!disabled, checked },
+          role: 'radio' as const,
+          'aria-disabled': !!disabled,
+          'aria-checked': checked,
```

`466d9c608` (#5005) moved the library to `aria-*` and `src/` on `main` has none of these props left. Same revert at `RadioButtonItem.tsx:207-212` (`accessibilityLabel` / `Role` / `State`) and `RadioButtonGroup.tsx:69` (`accessibilityRole="radiogroup"`; `main` has `role`). `Checkbox.tsx:249-256` is the shape to copy under the identical `rest.accessible === false` guard, and `CheckboxItem.tsx:157-160` for the row. Snapshots need regenerating after.

src/components/RadioButton/RadioButton.tsx
```diff
@@ -71,16 +97,171 @@ export type Props = {
...
+          <View style={[StyleSheet.absoluteFill, styles.radioContainer]}>
+            <Animated.View
+              style={[
+                styles.dot,
+                { backgroundColor: selectionControlColor },
+                dotAnimatedStyle,
+              ]}
+            />
+          </View>
+        ) : null}
```
JKobrynski, line 243: The `StyleSheet.absoluteFill` + `radioContainer` wrapper only existed because the old code animated `borderWidth`. The outline is a constant now, so it can go: put `alignItems: 'center'` / `justifyContent: 'center'` on `styles.radio`, render the `Animated.View` dot directly as its child, and drop `styles.radioContainer`.

JKobrynski (changes requested):

src/components/RadioButton/RadioButton.tsx
MikitasK: done ✅

src/components/RadioButton/RadioButton.tsx
MikitasK: done

JKobrynski:
@MikitasK is this ready for another review?

JKobrynski (commented):
Add a `### RadioButton` entry to `docs/6.x/docs/guides/migration.md`. The iOS variant drew a checkmark, so anyone on it gets a visibly different control. [#5059](https://github.com/callstack/react-native-paper/pull/5059) has a `### Checkbox` section covering the identical removals that would work as the template, but it skips RadioButton and still lists `RadioButton.Android` / `RadioButton.IOS` as live. Worth syncing with @mikolajadamowicz on which PR carries it.

MikitasK:
> Add a `### RadioButton` entry to `docs/6.x/docs/guides/migration.md`. The iOS variant drew a checkmark, so anyone on it gets a visibly different control. [#5059](https://github.com/callstack/react-native-paper/pull/5059) has a `### Checkbox` section covering the identical removals that would work as the template, but it skips RadioButton and still lists `RadioButton.Android` / `RadioButton.IOS` as live. Worth syncing with @mikolajadamowicz on which PR carries it.

done ✅ added `### RadioButton` section following the structure used for Checkbox in #5059 
@mikolajadamowicz this PR now carries `RadioButton`-specific migration guidance so we can avoid duplicating it in #5059

src/components/RadioButton/RadioButton.tsx
```diff
@@ -71,16 +97,168 @@ export type Props = {
...
+      </View>
+    </TouchableRipple>
+  );
 };
 
+RadioButton.displayName = 'RadioButton';
+
+const styles = StyleSheet.create({
+  container: {
+    width: stateLayerSize,
```
JKobrynski, line 253: The [spec's](https://m3.material.io/components/radio-button/specs) Measurements table lists a 48dp target next to the 40dp state layer, and the same page warns against dropping targets below 48x48. This sizes the whole touchable to 40. Would `hitSlop` cover the difference? Checkbox has the same 40dp target, so it may be one change for both.

JKobrynski (commented):
We've got a bunch of conflicts

src/components/RadioButton/RadioButton.tsx
MikitasK: good catch! 👍 done

src/utils/touchTarget.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Symmetric `hitSlop` that grows a `size` x `size` control up to
+ * `MIN_TOUCH_TARGET_SIZE` without changing its visual bounds. Returns
+ * `undefined` when the control already meets the minimum, so no redundant
+ * prop reaches the underlying pressable.
+ *
+ * Note that on Android a child's `hitSlop` cannot extend past its parent's
+ * bounds, so a control placed in a tightly clipped container may still
+ * receive a smaller effective target.
```
JKobrynski, line 21: I'm pretty sure this doesn't change much on the web. AFAIK react-native-web's `Pressable` never reads `hitSlop`. So the touchable should probably have padding on the web instead, or at least a note next to this Android caveat. Wdyt?

MikitasK: nice catch!
documented this limitation next to Android caveat & noted that web consumers should use wrapper padding when `48dp` target is required

PR #5011, opened 2026-06-22
fix: avoid disabled accessibility state for active chips

### LINKED ISSUE #4844: Chip announces to screen readers as "dimmed" even when active
### Current behaviour
Chips announce as "dimmed" to screen readers even when they have an `onPress`

### Expected behaviour
They probably shouldn't ever announce as "dimmed", since chips tend to be either active buttons, or informational; rarely disabled buttons. I'm assuming this is an unintentional side-effect of the construction of this component.

### How to reproduce?
Interact with a `<Chip>` component with VoiceOver.

Setting `disabled={false}` does not address the issue. (It's `false` by default anyway.)

PRs aren't being approved right now, so I haven't bothered working one up; this is to make sure the issue is recorded and known. If the library starts being maintained again and PRs are approved and merged, I can look into the internals and try to address the issue in a PR.

### Your Environment

| ios                                | 26
| android                        | 16
| react-native                | 0.7.9

vivekjm:
### Motivation

Fixes an accessibility regression where Chips without an explicit disabled prop can still be exposed as disabled/dimmed to screen readers. `TouchableRipple` marks itself disabled when no touch handler is present, so informational or selected Chips could inherit a disabled accessibility state even though the Chip itself is not disabled.

This keeps explicitly disabled Chips disabled, while allowing non-disabled Chips to remain active from an accessibility perspective.

### Related issue

Fixes #4844

### Test plan

- `yarn test src/components/__tests__/Chip.test.tsx --runInBand -u`
- `yarn eslint src/components/Chip/Chip.tsx src/components/__tests__/Chip.test.tsx`
- `yarn typescript`

Also attempted the default pre-commit hook, which runs full lint/types/tests. Lint and most tests ran, but the full Jest run failed in `src/babel/__tests__/index.js` because this checkout does not have `lib/mappings.json` generated.


matkoson (commented):
src/components/Chip/Chip.tsx:217 @vivekjm wouldn't this make non-interactive chips ripple when pressed?


generated by an automated pr-review pipe, not written by a human. reply and tag the repo owner if it got something wrong.

vivekjm:
Thanks for checking this. The intent is not to make non-interactive chips behave as actionable controls, only to prevent them from being exposed as disabled to accessibility APIs when the Chip itself is not disabled.\n\nThe fallback no-op handler is only used when the chip is not disabled and no touch handler was passed. It avoids TouchableRipple deriving a disabled accessibility state from the missing handler; explicitly disabled chips still pass through as disabled, and chips with real handlers still use their provided onPress/onPressIn/onPressOut behavior.\n\nIf you prefer avoiding a no-op onPress entirely, I can rework this to keep the accessibility state independent from TouchableRipple's disabled inference instead.

src/components/__tests__/Chip.test.tsx
```diff
@@ -398,7 +425,5 @@ it('animated value changes correctly', async () => {
   await act(() => {
     jest.advanceTimersByTime(200);
   });
-  expect(screen.getByTestId('chip-container-outer-layer')).toHaveStyle({
-    transform: [{ scale: 1.5 }],
-  });
+  expect(getScaleValue('chip-container-outer-layer')).toBe(1.5);
 });
```
satya164: Please revert these changes. Tests shouldn't be reading internal implementation details and private properties.

The only test worth adding is verifying that the touchable doesn't have `disabled`.

src/components/Chip/Chip.tsx
```diff
@@ -300,7 +302,7 @@ const Chip = ({
         borderless
         background={background}
         style={[{ borderRadius }, styles.touchable]}
-        onPress={onPress}
+        onPress={touchableOnPress}
```
satya164: this turns all chips pressables even when they aren't intended to be interactive. the proper fix would be not to render a `TouchableRipple` for informational/close-only chips

satya164 (changes requested):
thanks for the pr. please provide a video showing before/after

PR #5015, opened 2026-06-28
refactor: migrate components to Reanimated

satya164:


JKobrynski:
Is this still active? Happy to help push it over the line if it's parked — a fair bit of the MD3 work queues behind it.

Also confirming my reading: there's no withSpring anywhere here, so adopting the MD3 spring tokens is a deliberate follow-up rather than part of this?

satya164:
@JKobrynski I'll update and merge it next week. This PR only focuses on moving the existing animations to Reanimated. Using material tokens would be separate.

PR #5017, opened 2026-06-30
docs: screenshots and mobile menu

adamTrz:
## Summary

Fixes the issues reported in the Slack thread against the Rspress/Callstack-themed React Native Paper docs.

- Restores the current-root legacy TextInput screenshots so the existing 5.x `TextInput`, `TextInput.Affix`, and `TextInput.Icon` screenshot references render instead of broken image icons.
- Fixes the mobile docs menu duplicate hamburger by hiding the redundant pseudo-icon when the theme already renders an SVG icon.
- Makes the version selector dropdown use an opaque background and positions it below the sticky docs menu on mobile docs pages.
- Updates the home banner `Let's talk` CTA from the stale `/contact-us` URL to the live `/contact` URL.


satya164 (approved):
thanks

Issue #4218, opened 2023-12-08
Modals on iOS do not respond to the presence of the on-screen keyboard 


JPStrydom:
### Current behaviour
On web and Android, the Modal component respects the state of the keyboard, adjusting it's height accordingly. On iOS, however, the Modal does not move up when the keyboard is opened.

This issue is likely related to [2172](https://github.com/callstack/react-native-paper/issues/2172)

### Expected behaviour
The Modal component should respect the positions of the keyboard and respond accordingly on all platforms.

### How to reproduce?
1. Add a Modal component inside a Portal component.
2. Add content to the Modal component and open the Modal.
3. Open the device on-screen keyboard on an iOS device.
4. Notice how the Modal doesn't respond (move up) to accommodate the keyboard, even though there is empty space available above the modal.

### Preview
iOS:
![image](https://github.com/callstack/react-native-paper/assets/25905330/ba086873-7684-4754-924c-d3992e5f7690)

Android:
![image](https://github.com/callstack/react-native-paper/assets/25905330/91c79e81-6c68-4bb4-b7e8-dd6682963bde)

### What have you tried so far?
We're currently following the workaround listed [here](https://github.com/callstack/react-native-paper/issues/2172#issuecomment-725769818).

### Your Environment

| software                      | version
| --------------------- | -------
| ios                                | x
| android                        | x
| react-native                | 0.71.14
| react-native-paper    | 5.11.2
| node                            | 20.9.0
| npm or yarn                | 10.1.0
| expo sdk                     | 48.0.20

Bi0max:
Having the same issue

lukewalczak:
@Bi0max please add your issue reproduction

Sanal:
Hello! I confirm the issue. It seems there is a problem with Portal and its keyboard avoiding behavior on iOS. It also make SnackBar (placed in Portal) appear under the keyboard on iOS. Reproduced the same way as the issue's author described.

https://github.com/callstack/react-native-paper/assets/32746079/49953d6b-45e7-4d0a-9168-d4b5ea6c6b34

## My environment
software | version
-- | --
ios | 17.2
android | 13.0
react-native | 0.72.10
react-native-paper | 5.12.2
node | 21.3.0
yarn | 1.22.21

foxysolutions:
Same issue after upgrading to latest and placing modal within Portal component.

Abdalrzakalsouki:
[...]

PR #5023, opened 2026-07-03
fix(modal): avoid keyboard in edge-to-edge Android

### LINKED ISSUE #5021: In edge-to-edge Android, Dialogs do not move when keyboard is opened
### Current behavior
On Android, when edge-to-edge mode is enabled (either by setting `edgeToEdgeEnabled=true` in `android/gradle.properties` or by running the app on a recent Android version where edge-to-edge is enforced), Dialog objects do not move when the keyboard is opened. See the screenshots below.

Note: A similar issue (https://github.com/callstack/react-native-paper/issues/4218) has already been reported for react-native-paper on iOS.

#### Android 14 (API 34) with `edgeToEdgeEnabled=false` (expected behavior):

Image

#### Android 14 (API 34) with `edgeToEdgeEnabled=true`:

Image

#### Android 16 (API 36.1):

(Edge-to-edge mode is of course always enabled by Android 16, regardless of the value of `edgeToEdgeEnabled`.)

Image

### Expected behavior
It would be great if the dialog always moved out of the way of the keyboard, regardless of whether edge-to-edge was enabled.

### How to reproduce?
1. Create a new React Native project: `npx @react-native-community/cli@latest init DialogBugReproducer`
2. Enter the new project directory: `cd DialogBugReproducer`
3. Install react-native-paper: `npm install react-native-paper`
4. Delete all content in App.tsx and replace it with the following code. (I copied this code from the react-native-paper docs earlier this year, but I have made a few changes. The original code appears to have recently been moved to https://oss.callstack.com/react-native-paper/docs/components/Dialog/Dialog#usage.)
```js
// Code is from https://oss.callstack.com/react-native-paper/docs/components/Dialog/#usage
// with minor changes.

import { Button, Dialog, Portal, PaperProvider, Text, TextInput } from 'react-native-paper';

const App = () => {
  const [visible, setVisible] = React.useState(false);

  const showDialog = () => setVisible(true);

  const hideDialog = () => setVisible(false);

  return (
    <PaperProvider>
      <View>
        <Button style={{marginTop: 100}} onPress={showDialog}>Show Dialog</Button>
        <Portal>
          <Dialog visible={visible} onDismiss={hideDialog}>
            <Dialog.Title>Alert</Dialog.Title>
            <Dialog.Content>
              <Text variant="bodyMedium">This is simple dialog</Text>
              <TextInput />
            </Dialog.Content>
            <Dialog.Actions>
              <Button onPress={hideDialog}>Done</Button>
            </Dialog.Actions>
          </Dialog>
        </Portal>
      </View>
    </PaperProvider>
  );
};

export default App;
```
5. Run the app in an Android 16 emulator (or set `edgeToEdgeEnabled=true` in `android/gradle.properties` and run the app in an Android 14 emulator)
6. Click the "Show Dialog" button in the app
7. Click the `TextInput` to show the keyboard, and notice that the dialog does not move

- I tried running this example in an Android 7 (API 24) emulator with `edgeToEdgeEnabled=true`. The problem still occurred: the dialog did not move when the keyboard was opened, and the keyboard covered up the "Done" button in the dialog. As such, it seems this problem occurs even on very old versions of Android when `edgeToEdgeEnabled` is `true`.
- I tried importing `TextInput` from `'react-native'` instead of `'react-native-paper'` (i.e., I modified the above example to use the standard `TextInput` provided by React Native). This did not fix the problem.
- In the above example, I tried replacing the `<View>` inside `<PaperProvider>` with `<KeyboardAvoidingView>`. I tried all three values of [`behavior`](https://reactnative.dev/docs/keyboardavoidingview#behavior): `"height"`, `"position"`, and `"padding"`. None of these options fixed the problem.
- I later tried putting the `<Button>` outside the `<KeyboardAvoidingView>` and trying all three values for `behavior` again. This did not work either.
- I then tried moving the `<KeyboardAvoidingView>` inside the `<Portal>` but outside the `<Dialog>` and trying all three values for `behavior` again. This completely messed up the dialog's appearance (the entire dialog displayed as a thin horizontal line with no visible content).
- I then tried moving the `<KeyboardAvoidingView>` inside the `<Dialog>` and trying all three values for `behavior` again. The dialog contents were visible again, but the problem was still not fixed.

### Notes

If this is a React Native issue rather than a react-native-paper issue, please let me know and I will create a React Native bug report instead. Also, feel free to close this issue if there are no plans to implement this out of the box in react-native-paper.

Some webpages that may possibly be relevant:

- https://github.com/react/react-native/pull/55855
- [Discussion on StackOverflow about `adjustResize` no longer being sufficient](https://stackoverflow.com/a/79364446)

### Your Environment

| ios                                | N/A. For iOS, see https://github.com/callstack/react-native-paper/issues/4218 instead.
| android                        | 14 (API 34) and 16 (API 36.1), both emulated
| react-native                | 0.86.0
| react-native-paper    | 5.15.3
| node                            | 26.3.1
| npm or yarn                | npm 11.18.0
| expo sdk                     | Not using Expo

PRATHAMESH75:
### Motivation

On Android with **edge-to-edge** enabled (the default on Android 15+, or `edgeToEdgeEnabled=true` in `gradle.properties`), `Dialog`s do not move out of the way when the soft keyboard opens, so the keyboard covers dialog content such as a `TextInput`.

`Dialog` renders through `Modal`, which centers its content in a full-screen `absoluteFill` view with no keyboard avoidance. In legacy Android the OS resizes the window (`adjustResize`), so the `absoluteFill` shrinks and the centered content shifts up on its own. In edge-to-edge mode the window is no longer resized — the app draws behind the keyboard — so the content stays centered behind it. The same gap exists on iOS, which never had this avoidance.

The fix wraps the modal content in a `KeyboardAvoidingView` (`behavior="padding"`). It measures its own frame against the keyboard, so it reserves space in edge-to-edge mode without double-shifting when `adjustResize` is active, and fixes the behavior on iOS too.

### Related issue

Fixes #5021

### Test plan

- `yarn typecheck` — passes
- `yarn lint` — passes
- `yarn test` (Modal + Dialog suites) — 26/26 pass, no regressions

Manual (Android 15+ emulator or `edgeToEdgeEnabled=true`):
1. Open a `Dialog` containing a `TextInput`.
2. Focus the `TextInput` to bring up the soft keyboard.
3. The dialog shifts up and stays clear of the keyboard (previously it stayed centered behind it). Confirm behavior is unchanged in non-edge-to-edge mode and on iOS.

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { Animated, Easing, StyleSheet, Pressable, View } from 'react-native';
+import {
+  Animated,
+  Easing,
+  KeyboardAvoidingView,
+  StyleSheet,
+  Pressable,
+} from 'react-native';
 import type { StyleProp, ViewStyle } from 'react-native';
 
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
@@ -206,7 +212,16 @@ function Modal({
         ]}
         testID={`${testID}-backdrop`}
       />
-      <View
+      {/*
+       * Keep the modal content clear of the soft keyboard. In legacy Android
+       * (non edge-to-edge) the window is resized by `adjustResize`, so the
+       * centered content shifts up on its own. In edge-to-edge mode (Android
+       * 15+ default) the window is no longer resized, so we rely on
+       * `KeyboardAvoidingView` to reserve space for the keyboard instead. It
+       * measures its own frame, so it doesn't double up with `adjustResize`.
+       */}
+      <KeyboardAvoidingView
+        behavior="padding"
         style={[
           styles.wrapper,
           { marginTop: top, marginBottom: bottom },
@@ -223,7 +238,7 @@ function Modal({
         >
           {children}
         </Surface>
-      </View>
+      </KeyboardAvoidingView>
     </Animated.View>
   );
 }


sandeepsengupta:
I did a quick test on Android 16 and on iOS 26, and I can confirm that replacing the `<View>` with a `<KeyboardAvoidingView behavior="padding">` as shown in your commit fixes the issue in both cases. Thanks for your work with this!

Although I have not tested it with a generic Modal, I believe this PR fixes https://github.com/callstack/react-native-paper/issues/4218 as well.

src/components/Modal.tsx
```diff
@@ -206,7 +212,16 @@ function Modal({
...
       />
-      <View
+      {/*
+       * Keep the modal content clear of the soft keyboard. In legacy Android
+       * (non edge-to-edge) the window is resized by `adjustResize`, so the
+       * centered content shifts up on its own. In edge-to-edge mode (Android
+       * 15+ default) the window is no longer resized, so we rely on
+       * `KeyboardAvoidingView` to reserve space for the keyboard instead. It
+       * measures its own frame, so it doesn't double up with `adjustResize`.
+       */}
```
satya164, line 222: Please remove unnecessary comments

src/components/Modal.tsx
```diff
@@ -206,7 +212,16 @@ function Modal({
...
+      <KeyboardAvoidingView
+        behavior="padding"
```
satya164, line 224: Wouldn't this also override padding specified in `style`?

satya164 (changes requested):
thanks for the PR. can you provide before/after videos showing the fix, and also a video on non-edge-edge android, and on iOS showing it doesn't break?

PR #5031, opened 2026-07-20
feat: add rich tooltip mode to Tooltip

### LINKED ISSUE #4074: feat: rich `Tooltip` type

According to the Material Design documentation there is new type for `Tooltip` component, called `rich`.

* Rich tooltips provide more details, like describing the value of a feature
* Rich tooltips can include an optional title, link, and buttons

https://m3.material.io/components/tooltips/overview

* Follow component anatomy from: https://m3.material.io/components/tooltips/guidelines

* Create all possible variants:

image

**Additional context**

Should be available in both MD generations.

hazrid93:
feat: add rich tooltip mode to Tooltip

Closes #4074.

Extends Tooltip with a mode that selects between the current plain behaviour and a new rich behaviour. Rich mode renders an optional heading above multi line supporting text and up to two action buttons.

Each action button is individually pressable, exposes a button accessibility role and label, invokes its own callback, and dismisses the tooltip. A rich tooltip is persistent, staying open after the trigger touch ends, and dismisses on an action press or an outside tap, firing a dismissal callback on close. Plain mode keeps the current fixed 32px container height and single line label, and rich only props have no effect in plain mode. The new mode, action, and dismiss types are exported from the public surface.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -16,6 +16,20 @@ import { addEventListener } from '../../utils/addEventListener';
 import Portal from '../Portal/Portal';
 import Text from '../Typography/Text';
 
+export type TooltipMode = 'plain' | 'rich';
+
+export type TooltipAction = {
+  /**
+   * Label shown on the action button.
+   */
+  label: string;
+  /**
+   * Called when the action button is pressed. The tooltip is dismissed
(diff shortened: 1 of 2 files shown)


JKobrynski:
Have you seen #4994? It adds rich tooltips as `Tooltip.Rich` rather than a prop. Worth agreeing the API before either lands.

Issue #4256, opened 2024-01-05
DropDowns and DatePickers


ananduremanan:
**Is your feature request related to a problem? Please describe.**
I need a way to select a date and time easily along with material design dropdown in my application, but the current options are limited and couldn't find a component for both the need.

**Describe the solution you'd like**
I would like to see a dropdown component that allows users to select from a list of options, and a datepicker component that enables users to pick a date from a calendar..

**Describe alternatives you've considered**
Tried couple of packages dependent to React Native Paper but didn't Satisfied My needs.

**Additional context**
-Nil-

apicgg:
These are much needed components. Thanks for this. @lukewalczak Do you guys have any plan on this?

robertoosantos:
As [react-native-dropdown-picker](https://www.npmjs.com/package/react-native-dropdown-picker) is not maintained anymore (last publish 3 years ago) I think it would be a great incorporation on React Native Paper.

apicgg:
> As [react-native-dropdown-picker](https://www.npmjs.com/package/react-native-dropdown-picker) is not maintained anymore (last publish 3 years ago) I think it would be a great incorporation on React Native Paper.

Thinking to start the same.

PR #5036, opened 2026-07-20
feat: add TimePicker component

hazrid93:
feat: add TimePicker component

Addresses #4256.

Adds a Material 3 time picker that lets the user choose an hour and minute. Controlled by a value Date, reading only the hour and minute, and calls onChange with a new Date carrying the chosen time.

Renders hour and minute fields, each with up and down steppers that increment or decrement and wrap at the field boundary. In 12 hour mode shows an AM/PM toggle that shifts the period by 12 hours, with midnight shown as 12; in 24 hour mode the toggle is hidden. In range adjustments never change the date part of the value. An optional minute step controls how many minutes a single step moves. The component is exported from the public surface. Companion to the date picker proposal, together covering the linked issue.

### DIFF
--- a/src/components/TimePicker/TimePicker.tsx
+++ b/src/components/TimePicker/TimePicker.tsx
@@ -1,0 +1,209 @@
+import * as React from 'react';
+import { Pressable, StyleSheet, Text, View, ViewStyle } from 'react-native';
+
+import type { TimePickerProps } from './types';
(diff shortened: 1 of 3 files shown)


bartekkrok (commented):
Thanks for Your commitment, however it's not expected result. 
Please follow these rules: https://m3.material.io/components/time-pickers/overview
Examples are also appreciated.

Issue #4933, opened 2026-05-11
refactor(dialog): improve composition model and accessibility


ruben-rebelo:
Refactor Dialog to improve Material Design 3 compliance, remove problematic children patterns, improve performance characteristics, and modernize the public API.

**Problems**

- Current API relies heavily on children composition in ways that are difficult to optimize
- Internal layout logic is tightly coupled to arbitrary child structures
- MD3 behaviors and visuals are not fully aligned with current Material guidelines
- Rendering performance can degrade in large lists or frequently updated screens
- Type safety and discoverability of the API can be improved
- The component has inconsistent behavior between platforms and edge cases

**Focus Areas**

- Simplify Dialog.Title / Dialog.Content / Dialog.Actions hierarchy
- Improve keyboard and accessibility handling
- Improve MD3 spacing and tonal elevation
- Reduce excessive wrappers

**Proposed API Direction**

`<Dialog
  title="Delete item"
  content="Are you sure?"
  actions={[...]}
/>`

**Performance Opportunities**

- Reduce nested layout complexity
- Improve portal interaction handling
- Reduce unnecessary rerenders during visibility updates

adrcotfas:
> [!IMPORTANT]  
See https://github.com/callstack/react-native-paper/discussions/4949 for more details

PR #5038, opened 2026-07-27
feat(dialog): add declarative title/content/actions API

avenali5:
Adds an optional declarative API to Dialog via `icon`, `title`, `content` and `actions` props, alongside the existing composition-based `Dialog.Title` / `Dialog.Content` / `Dialog.Actions` children. When `children` are provided they take precedence and the declarative props are ignored (with a dev-only warning), so the change is fully backward compatible.

String `content` is wrapped in a themed `onSurfaceVariant` Text, and the title/content are centered automatically when an `icon` is present, per MD3.

Refs #4933

### Motivation

Creating a dialog today requires composing Dialog.Title / Dialog.Content / Dialog.Actions children for even the most common case (a title, a line of supporting text, and one or two buttons). That is verbose, easy to get subtly wrong (spacing, button props, centering when an icon is present), and harder to type and discover.

This PR adds an optional declarative API to Dialog so those common dialogs can be expressed as plain props, while leaving the existing composition API fully intact. It implements the direction proposed in #4933 additively, no breaking changes.

### Related issue

Refs #4933.

New optional props on Dialog:

icon — renders a centered Dialog.Icon above the title (MD3).
title — string or node, rendered inside Dialog.Title.
content — string or node, rendered inside Dialog.Content. A string is wrapped in a themed onSurfaceVariant Text.
actions — a DialogAction[] ({ label, onPress, mode?, icon?, loading?, disabled?, labelStyle?, testID? }) rendered as Buttons inside Dialog.Actions.
Behavior:

When children are provided they take precedence and the declarative props are ignored, with a __DEV__-only warning. This keeps every existing usage working unchanged.
When an icon is present, the title and string content are centered automatically, per MD3.
```
<Portal>
  <Dialog
    visible={visible}
    onDismiss={hideDialog}
    icon="alert"
    title="Delete item"
    content="Are you sure? This action cannot be undone."
    actions={[
      { label: 'Cancel', onPress: hideDialog },
      { label: 'Delete', onPress: onConfirmDelete, mode: 'contained' },
    ]}
  />
</Portal>

```

### Test plan

yarn test — added 4 Dialog tests (declarative title/content render, action button press, React-node content, children-take-precedence + dev warning). Full suite passes (736 tests).
yarn typecheck — passes.
yarn lint — passes.
Manual (example app): Dialog → "Declarative API" button opens a dialog built entirely via the new props (example/src/Examples/Dialogs/DialogWithDeclarativeApi.tsx).

### DIFF
--- a/example/src/Examples/DialogExample.tsx
+++ b/example/src/Examples/DialogExample.tsx
@@ -5,6 +5,7 @@ import { Button } from 'react-native-paper';
 
 import {
   DialogWithCustomColors,
+  DialogWithDeclarativeApi,
   DialogWithDismissableBackButton,
   DialogWithIcon,
   DialogWithLoadingIndicator,
@@ -79,6 +80,13 @@ const DialogExample = () => {
           Dismissable back button
         </Button>
       )}
+      <Button
+        mode="outlined"
+        onPress={_toggleDialog('dialog8')}
+        style={styles.button}
+      >
+        Declarative API
+      </Button>
       <DialogWithLongText
         visible={_getVisible('dialog1')}
         close={_toggleDialog('dialog1')}
@@ -106,6 +114,10 @@ const DialogExample = () => {
       <DialogWithDismissableBackButton
         visible={_getVisible('dialog7')}
         close={_toggleDialog('dialog7')}
+      />
+      <DialogWithDeclarativeApi
+        visible={_getVisible('dialog8')}
+        close={_toggleDialog('dialog8')}
       />
     </ScreenWrapper>
   );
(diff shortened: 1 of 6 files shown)


alpharius-ck:
This dual API (icon / title / content / actions or Dialog.Title / Content / Actions) is the same change maintainers later asked to drop on [#5072](https://github.com/callstack/react-native-paper/pull/5072): keep the subcomponents, matching Card, List, and DataTable. [#5076](https://github.com/callstack/react-native-paper/issues/5076) is the preferred composition cleanup (Children/cloneElement → context/layout), including Dialog.

This one could be closed

PR #5040, opened 2026-08-07
fix(text): inherit parent styles in nested Text

### LINKED ISSUE #4351: Nested Text components do not inherit style

Doberjohn:
### Motivation

Two related bugs in nested `Text`, both reproduced on `main` first.

**A variant parent lost its typography.** When a `Text` with a `variant` wrapped a
nested Paper `Text`, it replaced its own font with the child's style instead of
layering on top of it. With an unstyled child there was no child style to take
over, so the parent rendered with no typography at all. This reached shipped
components: `Chip` renders `<Text variant="labelLarge">` around its children, so
`<Chip><Text>label</Text></Chip>` dropped the entire label font block
(`fontSize: 14`, `fontWeight: 500`, `lineHeight: 20`, `letterSpacing: 0.1`).

**A nested `Text` did not inherit.** A nested `Text` or `AnimatedText` without a
`variant` applied its fallback font unconditionally, overriding the family,
weight, letter spacing and color it should have inherited from the enclosing text.

On the approach: React Native's `Text` already inherits from an enclosing `Text`
natively, so nothing needs to merge styles. The bug was Paper *overriding* what
would otherwise be inherited. A nested component now declares only its own
`style`. The shared context therefore carries a boolean rather than a style, and
lives in its own module so the `Text` and `AnimatedText` imports stay
one-directional.

`AnimatedText` was broken in both directions and is fixed too. It falls back to
`bodyMedium` where `Text` falls back to `fonts.default`, so an animated child
inside a `displayLarge` parent was snapping to `fontSize: 14`. That fallback
inconsistency between the two components still exists, but is no longer
reachable through nesting.

### Related issue

Fixes #4351.

#4356 is an earlier attempt at the same issue, taking the opposite approach of
merging the parent's style into the child. It also only publishes the parent
style from the variant-less branch, so a child under a variant parent would
still not inherit, and it is written against a version of `Text.tsx` that no
longer exists on `main`.

### Test plan

Example app, Typography screen, new "Nested text" section with four cases:
unstyled child, styled child, child with its own variant, and a bold parent
whose nested child inherits the weight.

- 6 new unit tests. Each behaviour-changing one was confirmed to fail without its fix.
- The 3 pre-existing nesting tests still pass, so the intentional
  child-overrides-parent behaviour is preserved.
- `AnimatedText` had no test coverage before this; it now has 3 tests.
- Full suite: 55 suites, 740 tests, 169 snapshots, with no snapshot churn.
- `yarn lint` and `yarn typecheck` clean.

### DIFF
--- a/src/components/Typography/NestedTextContext.tsx
+++ b/src/components/Typography/NestedTextContext.tsx
@@ -12,3 +12,31 @@ import * as React from 'react';
  * checks, so importing the context from either of them would form a cycle.
  */
 export const NestedTextContext = React.createContext(false);
+
+/**
+ * Whether `children` can hold a nested `Text`, so that the provider above is
+ * only rendered when something is there to consume it.
+ *
+ * Mirrors the check React Native runs before wrapping its own text ancestor
+ * context, down to the array cutoff. The point of the check is to save the
+ * provider's overhead, so it has to stay cheaper than what it saves: three
+ * children covers the common cases without walking a long list on every render.
+ *
+ * Takes `unknown` rather than `ReactNode` because `AnimatedText` widens its
+ * `children` with animated values, and this check only looks at the shape.
+ */
+export const canContainNestedText = (children: unknown) => {
+  if (children == null) {
+    return false;
+  }
+
+  if (Array.isArray(children) && children.length <= 3) {
+    return children.some(
+      (child: unknown) => child != null && typeof child === 'object'
+    );
+  }
+
+  // Anything longer is assumed to contain an element, since an array is an
+  // object itself.
+  return typeof children === 'object';
+};
--- a/src/components/Typography/NestedTextContext.tsx
+++ b/src/components/Typography/NestedTextContext.tsx
@@ -1,0 +1,14 @@
+import * as React from 'react';
+
+/**
+ * Tells a `Text` or `AnimatedText` that it is rendered inside another one.
+ *
+ * React Native's `Text` inherits the resolved style of an enclosing `Text`, so a
+ * nested one only has to declare what it wants to change. Applying the default
+ * font here regardless would overwrite everything it should have inherited, so a
+ * nested component without a `variant` leaves those properties unset instead.
+ *
+ * Lives in its own module because `Text` renders `AnimatedText` in its nesting
+ * checks, so importing the context from either of them would form a cycle.
+ */
+export const NestedTextContext = React.createContext(false);
--- a/src/components/Typography/Text.tsx
+++ b/src/components/Typography/Text.tsx
@@ -4,6 +4,7 @@ import { StyleSheet, Text as NativeText } from 'react-native';
 import type { StyleProp, TextStyle } from 'react-native';
 
 import AnimatedText from './AnimatedText';
+import { NestedTextContext } from './NestedTextContext';
 import type { VariantProp } from './types';
 import { useLocale } from '../../core/locale';
 import { useInternalTheme } from '../../core/theming';
@@ -87,10 +88,13 @@ const Text = ({
   // FIXME: destructure it in TS 4.6+
   const theme = useInternalTheme(initialTheme);
   const { direction: writingDirection } = useLocale();
+  const isNested = React.useContext(NestedTextContext);
 
   React.useImperativeHandle(ref, () => ({
     setNativeProps: (args: Object) => root.current?.setNativeProps(args),
   }));
+
+  let element: React.ReactElement;
 
   if (variant) {
     let font = theme.fonts[variant];
@@ -143,7 +147,7 @@ const Text = ({
       );
     }
 
-    return (
+    element = (
       <NativeText
         ref={root}
         style={[
@@ -154,13 +158,17 @@ const Text = ({
         {...rest}
       />
     );
+  } else if (isNested) {
+    // Declare only what this `Text` was asked for. Everything else, including
+    // the font and the color, inherits from the enclosing `Text`.
+    element = <NativeText {...rest} ref={root} style={style} />;
   } else {
     const font = theme.fonts.default;
     const textStyle = {
       ...font,
       color: theme.colors?.onSurface,
     };
-    return (
+    element = (
       <NativeText
         {...rest}
         ref={root}
@@ -168,6 +176,12 @@ const Text = ({
       />
     );
   }
+
+  return (
+    <NestedTextContext.Provider value={true}>
+      {element}
+    </NestedTextContext.Provider>
+  );
 };
 
 const styles = StyleSheet.create({
--- a/src/components/Typography/Text.tsx
+++ b/src/components/Typography/Text.tsx
@@ -4,7 +4,7 @@ import { StyleSheet, Text as NativeText } from 'react-native';
 import type { StyleProp, TextStyle } from 'react-native';
 
 import AnimatedText from './AnimatedText';
-import { NestedTextContext } from './NestedTextContext';
+import { canContainNestedText, NestedTextContext } from './NestedTextContext';
 import type { VariantProp } from './types';
 import { useLocale } from '../../core/locale';
 import { useInternalTheme } from '../../core/theming';
--- a/src/components/Typography/Text.tsx
+++ b/src/components/Typography/Text.tsx
@@ -177,6 +177,10 @@ const Text = ({
     );
   }
 
+  if (!canContainNestedText(rest.children)) {
+    return element;
+  }
+
   return (
     <NestedTextContext.Provider value={true}>
       {element}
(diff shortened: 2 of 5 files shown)


Doberjohn:
A quick note on the Typography work: because a nested `Text` **without a variant** no longer applies `fonts.default` at all, this fix is immune to Body variants switching typeface. Before it, that switch would have caused a visible `sans-serif` inside `sans-serif-medium` mismatch on Android. A nested `Text` that declares its own `variant` still applies it, which is unchanged here.

src/components/__tests__/Typography/Text.test.tsx
```diff
@@ -116,6 +117,110 @@ it("nested text without variant, but with styles, should override parent's style
...
+  );
+
+  // React Native's `Text` inherits from the enclosing `Text`, so the child must
+  // not restate the default font, which would override what it inherits.
+  const { fontFamily, fontWeight, letterSpacing } = LightTheme.fonts.default;
+
+  expect(screen.getByTestId('child-text')).not.toHaveStyle({ fontFamily });
+  expect(screen.getByTestId('child-text')).not.toHaveStyle({ fontWeight });
+  expect(screen.getByTestId('child-text')).not.toHaveStyle({ letterSpacing });
+});
```
MikitasK, line 172: could you cover inherited color in this test as well? current assertions only protect font properties, but  reported regression also includes parent color & both previous fallback branches inject `theme.colors.onSurface`

```suggestion
it('nested text alongside other content inherits instead of resetting', async () => {
  const parentColor = 'tomato';

  await render(
    <Text
      testID="parent-text"
      variant="displayLarge"
      style={{ color: parentColor }}
    >
      Parent <Text testID="child-text">child</Text>
    </Text>
  );

  const { fontFamily, fontWeight, letterSpacing } = LightTheme.fonts.default;

  expect(screen.getByTestId('parent-text')).toHaveStyle({
    color: parentColor,
  });
  expect(screen.getByTestId('child-text')).not.toHaveStyle({ fontFamily });
  expect(screen.getByTestId('child-text')).not.toHaveStyle({ fontWeight });
  expect(screen.getByTestId('child-text')).not.toHaveStyle({ letterSpacing });
  expect(screen.getByTestId('child-text')).not.toHaveStyle({
    color: LightTheme.colors.onSurface,
  });
});
```

src/components/Typography/Text.tsx
```diff
@@ -150,20 +158,30 @@ const Text = ({
...
         style={[styles.text, textStyle, { writingDirection }, style]}
       />
     );
   }
+
+  return (
+    <NestedTextContext.Provider value={true}>
+      {element}
+    </NestedTextContext.Provider>
+  );
```
MikitasK, line 188: could you avoid adding this provider when children cannot contain nested text? every Paper `Text` currently gets context overhead even for string-only content. 
RN skips its equivalent provider in these cases ([source](https://github.com/react/react-native/blob/v0.85.3/packages/react-native/Libraries/Text/Text.js#L312-L338)). so what about extracting  shared helper that performs the same children check & using it here and in `AnimatedText`?
```suggestion
  if (!canContainNestedText(rest.children)) {
    return element;
  }

  return (
    <NestedTextContext.Provider value={true}>
      {element}
    </NestedTextContext.Provider>
  );
```

`canContainNestedText` might live in `NestedTextContext.tsx` & mirror RN’s lightweight check:
```
export const canContainNestedText = (children: React.ReactNode) => {
  if (children == null) {
    return false;
  }

  if (Array.isArray(children) && children.length <= 3) {
    return children.some(
      (child) => child != null && typeof child === 'object'
    );
  }

  return typeof children === 'object';
};
```

MikitasK (commented):
overall, these changes looks good to me 👌  the approach looks solid & aligns with RN's text inheritance behavior 👍 
just a few point to address before merge:

Doberjohn:
@MikitasK thanks for the review, your proposed changes sounded good!

MikitasK (approved):
LGTM ✅

JKobrynski (approved):
LGTM, good job!

JKobrynski:
@satya164 @artus9033 all yours

src/components/Typography/NestedTextContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  if (Array.isArray(children) && children.length <= 3) {
+    return children.some(
+      (child: unknown) => child != null && typeof child === 'object'
+    );
+  }
+
+  // Anything longer is assumed to contain an element, since an array is an
+  // object itself.
+  return typeof children === 'object';
```
satya164, line 41: this doesn't make much sense. what makes children longer than 3 different?

src/components/__tests__/Typography/Text.test.tsx
```diff
@@ -116,6 +117,120 @@ it("nested text without variant, but with styles, should override parent's style
...
+
+  expect(screen.getByTestId('parent-text')).toHaveStyle({
+    color: parentColor,
+  });
+  expect(screen.getByTestId('child-text')).not.toHaveStyle({ fontFamily });
+  expect(screen.getByTestId('child-text')).not.toHaveStyle({ fontWeight });
+  expect(screen.getByTestId('child-text')).not.toHaveStyle({ letterSpacing });
+  expect(screen.getByTestId('child-text')).not.toHaveStyle({
+    color: LightTheme.colors.onSurface,
+  });
```
satya164, line 171: shouldn't this inherit parent colors?

src/components/__tests__/Typography/Text.test.tsx
```diff
@@ -116,6 +117,120 @@ it("nested text without variant, but with styles, should override parent's style
...
+
+it('text outside of another text still gets the default font', async () => {
+  await render(<Text testID="lone-text">{content}</Text>);
+  const { fontFamily, fontWeight } = LightTheme.fonts.default;
+
+  expect(screen.getByTestId('lone-text')).toHaveStyle({
+    fontFamily,
+    fontWeight,
+  });
+});
```
satya164, line 197: doesn't this duplicate "renders v3 Text component without variant..."?

src/components/__tests__/Typography/Text.test.tsx
```diff
@@ -116,6 +117,120 @@ it("nested text without variant, but with styles, should override parent's style
...
+      <Text testID="child-text" style={{ fontStyle: 'italic' }}>
+        child
+      </Text>
+    </Text>
+  );
+
+  expect(screen.getByTestId('child-text')).toHaveStyle({
+    fontStyle: 'italic',
+  });
+});
```
satya164, line 187: this doesn't assert what everything the title says. it should just be consolidated into the above test

src/components/__tests__/Typography/Text.test.tsx
```diff
@@ -116,6 +117,120 @@ it("nested text without variant, but with styles, should override parent's style
...
+
+it('animated text outside of any text still gets its fallback font', async () => {
+  await render(<AnimatedText testID="lone-animated">{content}</AnimatedText>);
+  const { fontFamily, fontSize } = LightTheme.fonts.bodyMedium;
+
+  expect(screen.getByTestId('lone-animated')).toHaveStyle({
+    fontFamily,
+    fontSize,
+  });
+});
```
satya164, line 232: not sure we need separate tests for animated text.

satya164 (changes requested):
please remove all LLM generated comments, they are either unnecessary or could be replaced with 1-2 lines describing the purpose.

only add comments where you think it's needed for understanding something and write it by hand.

for tests: add tests where
- `Text` isn't a direct child and can be nested inside another component that's used inside `Text`
- child text also sets its variant and theme

this implementation also isn't accurate to what react native does. a `View` nested inside `Text` will reset the inheritance, though it might be a reasonable limitation.

Issue #4976, opened 2026-06-09
feat: modernize List related components to the latest MD3 specs


adrcotfas:
The list specific components are spread out in the places mentioned below.  
We need a refactoring to reuse the theme tokens, update to the latest specs(https://m3.material.io/components/lists) and simplify the implementation by reusing/composing subcomponents.
The specs mention draggable, swipeable and toggleable list items.  
Given the complexity, this task can be split into multiple PRs.

# Current State

## List Accordion

**MD3 spec:** Lists (expand interaction)
**Specs link:** https://m3.material.io/components/lists/specs

### Color tokens

- Description / supporting text: `onSurfaceVariant` ✅
- [ ]  Title color when expanded: hardcoded to `primary`; spec label text uses `onSurface` regardless of expand state
- [ ]  Left icon color when expanded: hardcoded to `primary`; spec uses `onSurfaceVariant` for leading icons
- [ ]  Container background: `theme.colors.background`; should be `surface`
- No selected state colors (`primaryContainer` bg / `onPrimaryContainer` text/icon) ❌

### Spacing / sizing

- `paddingVertical: 8` (container) + `marginVertical: 6` (row) = 14dp each side; spec 1-line item is 56dp total height, requiring 16dp top/bottom padding ❌
- `paddingRight: 24` trailing ✅
- Content `paddingLeft: 16` ✅
- [ ]  Title and description use hardcoded `fontSize: 16` / `fontSize: 14`; should use `theme.fonts.bodyLarge` / `theme.fonts.bodyMedium`

### Behavior

- Expand/collapse toggling via press ✅
- Controlled and uncontrolled modes ✅
- [ ]  No expand/collapse animation; children appear and disappear instantly
- [ ]  Chevron direction reversal is the only visual feedback for expanded state; no container highlight or shape change
- [ ]  No multi-select or single-select mode

### Notes

- Should be renamed to align with MD3 terminology (e.g., expandable list item) after the audit
- Animations and selected state are the most impactful missing features

---

## List Accordion Group

**MD3 spec:** Lists (expand interaction – single-expand group)
**Specs link:** https://m3.material.io/components/lists/specs

### Color tokens

No color logic of its own – delegates to `List.Accordion`. ✅

### Spacing / sizing

No sizing logic of its own. ✅

### Behavior

- At most one accordion expanded at a time (single-expand) ✅
- Controlled and uncontrolled modes ✅
[...]

PR #5041, opened 2026-08-07
refactor: align List tokens, spacing and typography with MD3

adam-sajko:
### Motivation

`List.Item` and `List.Accordion` now follow the [MD3 list spec](https://m3.material.io/components/lists/specs) for padding, type and colour, instead of hardcoded sizes and the old expanded primary tint.

### Related issue

Related to #4976

### Screenshots / Videos

image
image
image

### Test plan

Open the List example and compare one, two and three line rows against the spec.

### DIFF
--- a/src/components/List/ListAccordion.tsx
+++ b/src/components/List/ListAccordion.tsx
@@ -212,16 +212,6 @@ const ListAccordion = ({
     setIsDescriptionMultiline(nativeEvent.lines.length >= 2);
   };
 
-  const getVerticalPaddingStyle = () => {
-    if (!description) {
-      return styles.containerOneLine;
-    }
-
-    return isDescriptionMultiline
-      ? styles.containerThreeLine
-      : styles.containerTwoLine;
-  };
-
   const handlePressAction = (e: GestureResponderEvent) => {
     onPress?.(e);
 
@@ -257,7 +247,11 @@ const ListAccordion = ({
         style={{ backgroundColor: theme.colors[ListTokens.containerColor] }}
       >
         <TouchableRipple
-          style={[styles.container, getVerticalPaddingStyle(), style]}
+          style={[
+            styles.container,
+            description ? styles.containerTwoLine : styles.containerOneLine,
+            style,
+          ]}
           onPress={handlePress}
           onLongPress={onLongPress}
           delayLongPress={delayLongPress}
@@ -323,7 +317,7 @@ const ListAccordion = ({
               ) : (
                 <MaterialCommunityIcon
                   name={isExpanded ? 'chevron-up' : 'chevron-down'}
-                  color={theme.colors[ListTokens.trailingIconColor]}
+                  color={theme.colors[ListTokens.expandTrailingIconColor]}
                   size={24}
                   direction={direction}
                 />
@@ -358,16 +352,15 @@ ListAccordion.displayName = 'List.Accordion';
 
 const styles = StyleSheet.create({
   container: {
+    paddingVertical: ListTokens.verticalPadding,
     paddingRight: ListTokens.trailingSpace,
+    justifyContent: 'center',
   },
   containerOneLine: {
-    paddingVertical: ListTokens.oneLineVerticalPadding,
+    minHeight: ListTokens.oneLineContainerHeight,
   },
   containerTwoLine: {
-    paddingVertical: ListTokens.twoLineVerticalPadding,
-  },
-  containerThreeLine: {
-    paddingVertical: ListTokens.threeLineVerticalPadding,
+    minHeight: ListTokens.twoLineContainerHeight,
   },
   row: {
     flexDirection: 'row',
--- a/src/components/List/ListAccordion.tsx
+++ b/src/components/List/ListAccordion.tsx
@@ -202,13 +202,24 @@ const ListAccordion = ({
   const [expanded, setExpanded] = React.useState<boolean>(
     expandedProp || false
   );
-  const [alignToTop, setAlignToTop] = React.useState(false);
+  const [isDescriptionMultiline, setIsDescriptionMultiline] =
+    React.useState(false);
 
   const onDescriptionTextLayout = (
     event: NativeSyntheticEvent<TextLayoutEventData>
   ) => {
     const { nativeEvent } = event;
-    setAlignToTop(nativeEvent.lines.length >= 2);
+    setIsDescriptionMultiline(nativeEvent.lines.length >= 2);
+  };
+
+  const getVerticalPaddingStyle = () => {
+    if (!description) {
+      return styles.containerOneLine;
+    }
+
+    return isDescriptionMultiline
+      ? styles.containerThreeLine
+      : styles.containerTwoLine;
   };
 
   const handlePressAction = (e: GestureResponderEvent) => {
@@ -246,11 +257,7 @@ const ListAccordion = ({
         style={{ backgroundColor: theme.colors[ListTokens.containerColor] }}
       >
         <TouchableRipple
-          style={[
-            styles.container,
-            description ? styles.containerTwoLine : styles.containerOneLine,
-            style,
-          ]}
+          style={[styles.container, getVerticalPaddingStyle(), style]}
           onPress={handlePress}
           onLongPress={onLongPress}
           delayLongPress={delayLongPress}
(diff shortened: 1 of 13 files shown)


src/components/List/ListItem.tsx
```diff
@@ -179,18 +180,15 @@ const ListItem = ({
...
         ellipsizeMode={descriptionEllipsizeMode}
-        style={[
-          styles.description,
-          { color: descriptionColor },
-          descriptionStyle,
-        ]}
+        style={[{ color: descriptionColor }, descriptionStyle]}
         onTextLayout={onDescriptionTextLayout}
         maxFontSizeMultiplier={descriptionMaxFontSizeMultiplier}
       >
```
MikitasK, line 201: what about passing resolved `theme` to these `Text` components?

```
<Text theme={theme} variant="bodyMedium" ... >
```

in current implementation component-level font overrides are used for render-prop titles & descriptions, but plain text still reads typography from the surrounding provider

src/components/List/ListItem.tsx
```diff
@@ -200,35 +198,40 @@ const ListItem = ({
...
+        variant="bodyLarge"
         selectable={false}
         ellipsizeMode={titleEllipsizeMode}
         numberOfLines={titleNumberOfLines}
-        style={[styles.title, { color: titleColor }, titleStyle]}
+        style={[{ color: titleColor }, titleStyle]}
         maxFontSizeMultiplier={titleMaxFontSizeMultiplier}
       >
         {title}
       </Text>
```
MikitasK, line 228: [same suggestion](https://github.com/callstack/react-native-paper/pull/5041/changes#r3805116266) here

src/components/List/ListAccordion.tsx
```diff
@@ -301,20 +306,15 @@ const ListAccordion = ({
                 </Text>
```
MikitasK: [same suggestion](https://github.com/callstack/react-native-paper/pull/5041/changes#r3805116266) here

src/components/List/ListItem.tsx
```diff
@@ -200,35 +198,40 @@ const ListItem = ({
...
+  const descriptionColor = theme.colors[ListTokens.supportingTextColor];
 
   return (
     <TouchableRipple
       {...rest}
       ref={ref}
-      style={[styles.container, style]}
+      style={[
+        styles.container,
+        description ? styles.containerTwoLine : styles.containerOneLine,
```
MikitasK: what about descriptions with 2+ lines? they currently keep `14dp` 2-line padding, resulting in `92dp` item instead of MD3 `88dp` minimum ([source](https://m3.material.io/components/lists/specs))
so can we use `12dp` vertical padding (`multilineVerticalPadding: 12`) whenever `lines.length >= 2`?

```suggestion
        description  
           ? isDescriptionMultiline
              ? styles.containerMultiline
              : styles.containerTwoLine
           : styles.containerOneLine
```

src/components/List/ListAccordion.tsx
```diff
@@ -232,20 +233,24 @@ const ListAccordion = ({
...
     <View>
-      <View style={{ backgroundColor: theme?.colors?.background }}>
+      <View
+        style={{ backgroundColor: theme.colors[ListTokens.containerColor] }}
+      >
         <TouchableRipple
-          style={[styles.container, style]}
+          style={[
+            styles.container,
+            description ? styles.containerTwoLine : styles.containerOneLine,
```
MikitasK: [same suggestion](https://github.com/callstack/react-native-paper/pull/5041/changes#r3805404500) here

MikitasK (commented):
good job 👏  overall, this refactoring looks solid 👍 
just a few thing to consider before merge:

src/components/List/ListItem.tsx
adam-sajko: done

src/components/List/ListItem.tsx
adam-sajko: done

src/components/List/ListAccordion.tsx
adam-sajko: done

src/components/List/ListAccordion.tsx
adam-sajko: done

src/components/List/ListItem.tsx
adam-sajko: Done, also corrected end inset 24 -> 16 to match current MD3/Compose

Just one caveat: onTextLayout isn’t implemented in react-native-web, so this tier only applies on native. That’s pre-existing (alignToTop already does nothing on web before)

MikitasK (approved):
LGTM ✅

bartekkrok (approved):

src/components/List/ListAccordion.tsx
```diff
@@ -301,20 +315,15 @@ const ListAccordion = ({
...
+            <View style={styles.trailingItem}>
               {right ? (
                 right({
                   isExpanded: isExpanded,
                 })
               ) : (
                 <MaterialCommunityIcon
                   name={isExpanded ? 'chevron-up' : 'chevron-down'}
-                  color={descriptionColor}
+                  color={theme.colors[ListTokens.trailingIconColor]}
```
JKobrynski: Should the chevron use the expand set rather than the generic list-item one? On the
 [Lists specs](https://m3.material.io/components/lists/specs) `md.comp.list.expand.collapsed.list-item.trailing-icon.icon.color` and its expanded twin both resolve to `md.sys.color.on-surface`, where `md.comp.list.list-item.trailing-icon.color` is `on-surface-variant`.

src/components/List/ListItem.tsx
```diff
@@ -161,13 +162,24 @@ const ListItem = ({
...
 
   const onDescriptionTextLayout = (
     event: NativeSyntheticEvent<TextLayoutEventData>
   ) => {
     const { nativeEvent } = event;
-    setAlignToTop(nativeEvent.lines.length >= 2);
+    setIsDescriptionMultiline(nativeEvent.lines.length >= 2);
+  };
+
+  const getVerticalPaddingStyle = () => {
```
JKobrynski: `react-native-web` doesn't implement `onTextLayout`, and a function `description` never gets it either, so in both cases a wrapped description stays on the 14dp tier and renders 92dp against `md.comp.list.list-item.three-line.container.height` = 88dp. Would `minHeight: description ? 72 : 56` with a flat 12dp padding work instead? It lands on 56 / 72 / 88 without measuring anything.

src/components/List/ListAccordion.tsx
adam-sajko: Good catch, done

src/components/List/ListItem.tsx
adam-sajko: I did what you suggested with one exception. I kept onTextLayout. The height no longer reads it, but the same flag still top-aligns the left and right elements on three line items, which was all it did before this PR. Please take another look :)

src/components/__tests__/Tooltip.test.tsx
```diff
@@ -163,7 +163,7 @@ describe('Tooltip', () => {
       it('hides the tooltip when the user stop pressing the component', async () => {
         const {
           wrapper: { queryByText, getByText, findByText },
-        } = await setup({ enterTouchDelay: 50, leaveTouchDelay: 0 });
+        } = await setup({ enterTouchDelay: 50, leaveTouchDelay: 100 });
```
adam-sajko, line 166: Stolen from #4996 (CI fix)

src/components/List/tokens.ts
```diff
@@ -0,0 +1,20 @@
+import type { ColorRole } from '../../theme/types';
+
+const sizes = {
+  verticalPadding: 12,
```
MikitasK: could we avoid applying 12dp root padding to every row? 

with the PR’s own 40dp avatar example, one-line item becomes 64dp (40 + 12 + 12) instead of 56dp, `List.Image` (56dp) becomes 80dp instead of 72dp. compose uses 8dp vertical padding for one/two-line items & 12dp only for 3-line items

maybe we can preserve measurement-free 56/72/88dp text rows by using 8dp container padding plus 4dp content padding with the same 4dp offset for top-aligned accessories? and could we also cover a 40dp avatar or `List.Image` case?

JKobrynski (approved):
Good job, LGTM! @satya164 want to take a look?

src/components/List/tokens.ts
adam-sajko: Done. However, I haven't added the 4dp offset for accessories because it would be tricky, and the MD3 spec images in Figma directly state 8dp or 12dp vertical padding for the entire container

adam-sajko:
Note: I didn't modify the checkbox element itself, but it renders differently than specified in the MD3 specs. Our checkbox is normally rendered within a 40x40dp container. However, for three-line text that is top-aligned, the 40x40dp container is dismissed, and the checkbox scales down to 24dp. I'd like to do it separately.

Reference:
image

Issue #4988, opened 2026-06-09
feat: implement Toolbar


adrcotfas:
- Implement the new component according to specs: https://m3.material.io/components/toolbars/overview
- Reuse newly introduced tokens (shape, motion, typography, color etc)
- Extract component specific tokens for easier maintenance
- Get inspired from TextInput/TextField, Switch, Checkbox, FAB components which have gone through modernization
- Showcase the component in the example app
- Update the documentation

Issue #4989, opened 2026-06-09
refactor: usage of React.Children


adrcotfas:
The goal is to review and update the components using `React.Children` that make composition harder.  
For an example, see how the FAB Menu component (Previously FAB Group) handled this migration. 
See https://github.com/callstack/react-native-paper/pull/4963.

PR #5043, opened 2026-08-11
feat: implement Toolbar component

k0ndee:
## Motivation

Introduces `Toolbar`, implementing the Material Design 3 toolbars spec. It reuses existing theme tokens (shape, color roles, elevation) and follows the same per-component-tokens pattern as `FAB`/`Checkbox`.

## Spec re-check (M3 toolbars)

Re-checked the [M3 toolbars spec](https://m3.material.io/components/toolbars/specs):

- Two `variant`s: **`floating`** (self-positioned pill, `corner.full`, elevation level 3) and **`docked`** (full-width bar pinned to the bottom edge, `corner.none`, no elevation, extends into safe-area insets).
- `floating` supports `horizontal`/`vertical` `orientation`; `docked` is always horizontal per spec.
- **`standard`** (default) colors: container/unselected-button `surfaceContainer`, icon/label `onSurfaceVariant`, selected button `secondaryContainer` with `onSecondaryContainer` icon.
- **`vibrant`** colorScheme: container/unselected-button `primaryContainer`, icon/label `onPrimaryContainer`, selected button falls back to `surfaceContainer` with `onSurface` icon.
- Mode-less `IconButton`/`Button` descendants pick up the toolbar's `colorScheme` automatically via context - not just direct children - unless they already set their own `mode` or an explicit color prop.

### `Toolbar` / tokens / utils
- New `src/components/Toolbar/{Toolbar.tsx,tokens.ts,utils.ts,ToolbarColorContext.ts}`: `variant` (`floating`/`docked`), `orientation` (`horizontal`/`vertical`, floating-only), `colorScheme` (`standard`/`vibrant`), `containerColor` override, `style`/`contentContainerStyle`, `testID`, `aria-label`, `theme`, `ref`
- Container shape/elevation/spacing resolved from `ToolbarTokens`. `docked`'s outer wrapper has no inset margin, so its background naturally spans edge-to-edge into the safe area, while the content row gets `insets`-based margin to stay clear of it - the 64dp icon band itself stays untouched
- `ToolbarColorContext` carries `theme`/`colorScheme` down to descendants; `IconButton`/`Button` each read it directly and resolve their own color, backing off if they already have a `mode` or explicit color prop
- Exported from `src/index.tsx`

### Example / docs / tests
- Example screen (`ToolbarExample.tsx`) covering both variants, both orientations, both color schemes, over a scrollable list
- Prop docs and a theme-colors table entry (auto-generated docs page); screenshots for floating (horizontal/vertical) and docked, each in standard/vibrant
- Unit tests (`Toolbar.test.tsx`) covering shape/elevation per variant, color resolution across light/dark themes, and child auto-recoloring (including nested/non-direct descendants, opt-out via `mode` or explicit color)

## Scope note

Per Satyajit Sahoo's recommendation, this lands as a series of PRs instead of one large change. This is **PR 1** of the series:

- **PR 1** (this PR) - Floating + docked variant, static (no animation, no FAB pairing), `colorScheme` standard/vibrant. Component, tokens, tests, docs, example.
- **PR 2** - Show/hide animation (offscreen slide + spring), generic scroll shared-value primitive.
- **PR 3** - FAB pairing, FAB/key-action collapse on scroll.
- **PR 4** - Trailing-edge overflow menu.

## Related issue

Related to #4988

## Test plan

- [x] `yarn typecheck` / `yarn lint` / Toolbar unit tests
- [x] Full unit test suite green
- [x] Manually verified in the example app: floating (horizontal/vertical) and docked, standard/vibrant, over scrolling content, light and dark themes

## Visual verification

| | standard | vibrant |
| :--- | :---: | :---: |
| **floating, horizontal** | floating-h-standard | floating-h-vibrant |
| **floating, vertical** | floating-v-standard | floating-v-vibrant |
| **docked** | docked-standard | docked-vibrant |


src/components/Toolbar/tokens.ts
```diff
@@ -0,0 +1,86 @@
...
+  container: { light: 'primaryContainer', dark: 'primaryContainer' },
+  buttonContainer: { light: 'primaryContainer', dark: 'primaryContainer' },
+  selectedButtonContainer: {
+    light: 'surfaceContainer',
+    dark: 'surfaceContainer',
+  },
+  icon: { light: 'onPrimaryFixedVariant', dark: 'onPrimaryContainer' },
+  selectedIcon: { light: 'onSurface', dark: 'onSurface' },
+  label: { light: 'onPrimaryFixedVariant', dark: 'onPrimaryContainer' },
+} as const satisfies Record<string, ToneRole>;
```
MikitasK: could we use color roles from M3 spec directly? 
standard buttons should use `primary`, selected buttons should use `onSecondaryContainer`, vibrant buttons should use `onPrimaryContainer`. 
using `FixedVariant` roles may also produce incorrect colors with custom themes

```suggestion
const standardColors = {
  container: { light: 'surfaceContainer', dark: 'surfaceContainer' },
  buttonContainer: { light: 'surfaceContainer', dark: 'surfaceContainer' },
  selectedButtonContainer: {
    light: 'secondaryContainer',
    dark: 'secondaryContainer',
  },
  icon: { light: 'primary', dark: 'primary' },
  selectedIcon: {
    light: 'onSecondaryContainer',
    dark: 'onSecondaryContainer',
  },
  label: { light: 'primary', dark: 'primary' },
} as const satisfies Record<string, ToneRole>;

const vibrantColors = {
  container: { light: 'primaryContainer', dark: 'primaryContainer' },
  buttonContainer: { light: 'primaryContainer', dark: 'primaryContainer' },
  selectedButtonContainer: {
    light: 'surfaceContainer',
    dark: 'surfaceContainer',
  },
  icon: { light: 'onPrimaryContainer', dark: 'onPrimaryContainer' },
  selectedIcon: { light: 'onSurface', dark: 'onSurface' },
  label: { light: 'onPrimaryContainer', dark: 'onPrimaryContainer' },
} as const satisfies Record<string, ToneRole>;
```

src/components/Toolbar/utils.ts
```diff
@@ -0,0 +1,181 @@
...
+      ) {
+        return child;
+      }
+
+      const { iconColor, containerColor } = resolveIconColors({
+        theme,
+        colorScheme,
+        selected: child.props.selected ?? false,
+      });
+      return React.cloneElement(child, { iconColor, containerColor });
```
MikitasK: what about selected & disabled icon buttons? explicit `containerColor` prevents `IconButton` from applying its normal disabled container treatment. could we use tonal mode for selected buttons?
```suggestion
      return React.cloneElement(child, { 
         iconColor, 
         containerColor, 
         ...(child.props.selected ? { mode: 'contained-tonal' } : null),
      });
```

src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,267 @@
...
+      }
+    : null;
+
+  const pill = (
+    <Surface
+      // Keying on the axis forces a fresh mount when `floating`'s pill
+      // jumps position (e.g. bottom-center to a vertical trailing edge),
+      // avoiding a stale shadow "ghost" that `Surface`'s iOS shadow can
+      // leave at the old frame when an existing view is resized in place.
+      key={isVertical ? 'vertical' : 'horizontal'}
```
MikitasK: could we avoid changing `key`? any other options? 
this approach implies that toolbar & its children remounts when orientation changes. it might cause loosing of child state & focus

MikitasK (commented):
nice work! 👍 overall, this looks like a solid foundation for the new `Toolbar` components
just a few points to address before merge:

src/components/Toolbar/tokens.ts
k0ndee: The spec for Toolbar tells that the standard colors should be:
`icon: { light: 'onSurfaceVariant', dark: 'onSurfaceVariant' },`
`label: { light: 'onSurfaceVariant', dark: 'onSurfaceVariant' },`

image


The vibrant colors according to spec are:
`icon: { light: 'onPrimaryContainer', dark: 'onPrimaryContainer' },`
`label: { light: 'onPrimaryContainer', dark: 'onPrimaryContainer' },`

image

I was confused about this, because these colors, like You commented are not the primary colors for Buttons and IconButtons according to the current spec for Buttons, but the Toolbar has it's own spec for included actions, that is why I used these tokens. 

Now that the fixedVariants are removed, I can see that the light and dark are using the same roles for colors. I am thinking that I should remove the light and dark fields.

src/components/Toolbar/Toolbar.tsx
k0ndee: Without remount, when switching from floating vertical to horizontal on iOS there is this bug that the inner Animated.View of SurfaceIOS inflated. See screenshots below:

Simulator Screenshot - iPhone 17 - 2026-08-12 at 13 19 11
Simulator Screenshot - iPhone 17 - 2026-08-12 at 13 19 32

I can't figure out why that is happening. On the first mount of floating horizontal Toolbar it looks fine, but when I switch between orientation that inflated pill is appearing. I will do some more debugging and try to find a more sutiable solution for that.

MikitasK: maybe we can extract pill background into a separate component & key only that component by orientation while keeping the toolbar content outside it? 
I suppose, this should reset `SurfaceIOS` without remounting the buttons & losing their state or focus

```
<View style={pillStyle}>
  <PillBackground
    key={isVertical ? 'vertical' : 'horizontal'}
    elevation={elevation}
  />
  <View role="toolbar" style={contentStyle}>
    {children}
  </View>
</View>
```

src/components/Toolbar/tokens.ts
MikitasK: you’re right, thanks for clarifying 🙏  toolbar-specific spec uses `onSurfaceVariant` for standard actions & `onPrimaryContainer` for vibrant actions. removing `light` and `dark` fields totally makes sense - these semantic roles already adapt to the theme 👌

src/components/Toolbar/Toolbar.tsx
k0ndee: I found a better solution. I moved the height and width control to the inner View inside Surface. Surface requires children prop so I would have to pass a `{null}`, which is I think is not an elegant approach here. With the latest commit I uploaded we keep the same code structure, where Surface wraps the inner View.

```
<Surface
      ref={isDocked ? undefined : ref}
      elevation={elevation}
      style={[
        {
          backgroundColor,
          borderRadius,
        },
        isDocked && styles.dockedFill,
        styles.content,
        !isDocked && style,
      ]}
      testID={testID}
    >
      <View
        role="toolbar"
        aria-label={ariaLabel}
        testID={`${testID}-content`}
        style={[
          styles.content,
          isVertical ? styles.column : styles.row,
          { ...contentPadding, gap },
          dockedInsetMargin,
          // Cross-axis thickness is the spec default (see `thickness`
          // above). Deliberately set here rather than on `Surface` (which
          // wraps this `View` with no size of its own, so it just hugs
          // it) — giving `Surface` an explicit width/height that flips
          // between renders is what previously left a stale shadow
          // "ghost" on iOS when `floating`'s `orientation` changed axis;
          // that no longer happens with the fixed dimension living here
          // instead.
          isDocked && { height: thickness },
          !isDocked &&
            (isVertical ? { width: thickness } : { height: thickness }),
          contentContainerStyle,
        ]}
      >
        {withToolbarChildColors({ children, theme, colorScheme })}
      </View>
    </Surface>
```

src/components/Toolbar/utils.ts
k0ndee: That's a good catch. Fixed it in [6a695d7](https://github.com/callstack/react-native-paper/pull/5043/commits/6a695d7ae91f2c9f0d64316a0bed68e41e4fbd6b)

MikitasK (approved):
LGTM 👍

src/components/Toolbar/utils.ts
```diff
@@ -0,0 +1,179 @@
...
+      if (
+        (child.props.mode != null && child.props.mode !== 'text') ||
+        child.props.textColor != null ||
+        child.props.buttonColor != null
+      ) {
+        return child;
+      }
+
+      const textColor = resolveLabelColor({ theme, colorScheme });
+      return React.cloneElement(child, { textColor });
```
JKobrynski: `React.Children` and `cloneElement` are both off-limits in v6 — see #4989, which is open specifically to unwind them where they hurt composition, with the FAB Menu migration in #4963 as the reference. The occurrences still in `src/` (`Card`, `CardActions`, `Appbar/utils.ts`, `Dialog`, `ToggleButtonRow`) are what that issue exists to remove, so they aren't precedent for new code.

Beyond the rule, the traversal has the failure mode context exists to avoid: it only matches `Fragment`, `IconButton` and `Button`, and returns everything else untouched. Write `<Toolbar><MyBoldButton /></Toolbar>`, or wrap two actions in a `View` to group them, and the recolouring silently stops — the children keep `IconButton`'s default `onSurfaceVariant` on the toolbar's container.

src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,272 @@
...
+ */
+const Toolbar = ({
+  children,
+  variant = 'floating',
+  orientation = 'horizontal',
+  colorScheme = 'standard',
+  containerColor,
+  style,
+  contentContainerStyle,
+  testID = 'toolbar',
```
JKobrynski: No default `testID`s in new components — `Switch.tsx:58` is the v6 reference.

```diff
-  testID = 'toolbar',
+  testID,
```

`:198` and `:237` derive `${testID}-content` / `-container`, so those need to become conditional too.

docs/src/data/themeColors.ts
```diff
@@ -320,6 +320,33 @@ export const themeColors = {
...
+          'theme.colors.onSecondaryFixedVariant (light) / theme.colors.onSecondaryContainer (dark)',
+      },
+    },
+    vibrant: {
+      unselected: {
+        backgroundColor: 'theme.colors.primaryContainer',
+        iconColor:
+          'theme.colors.onPrimaryFixedVariant (light) / theme.colors.onPrimaryContainer (dark)',
+        textColor:
+          'theme.colors.onPrimaryFixedVariant (light) / theme.colors.onPrimaryContainer (dark)',
```
JKobrynski: `Toolbar.standard.selected`, lines 332-333:

```diff
-          iconColor:
-            'theme.colors.onSecondaryFixedVariant (light) / theme.colors.onSecondaryContainer (dark)',
+          iconColor: 'theme.colors.onSecondaryContainer',
```

`Toolbar.vibrant.unselected`, lines 339-342:

```diff
-        iconColor:
-          'theme.colors.onPrimaryFixedVariant (light) / theme.colors.onPrimaryContainer (dark)',
-        textColor:
-          'theme.colors.onPrimaryFixedVariant (light) / theme.colors.onPrimaryContainer (dark)',
+        iconColor: 'theme.colors.onPrimaryContainer',
+        textColor: 'theme.colors.onPrimaryContainer',
```

`tokens.ts:49,57,59` already resolve these unconditionally, and there's no `*-fixed-variant` token anywhere in `md.comp.toolbar.*`. Then `yarn docs generate` — the same strings are inlined at `Toolbar.mdx:160` and need committing alongside. The other two rows in the block are already correct.

src/components/Toolbar/tokens.ts
```diff
@@ -0,0 +1,68 @@
...
+  containerHeight: 64,
+  containerShape: 'none' as ShapeToken,
+  containerLeadingSpace: 16,
+  containerTrailingSpace: 16,
+  defaultSpacing: 32,
+} as const;
+
+const floating = {
+  containerHeight: 64,
+  containerShape: 'full' as ShapeToken,
```
JKobrynski: `as` casts aren't used anywhere in `src/` — and these two do nothing anyway. `ShapeToken` is `keyof ThemeShapeCorners | 'none' | 'full'`, so the literal is already assignable; the cast only widens it and loses the literal under `as const`.

```diff
-  containerShape: 'none' as ShapeToken,
+  containerShape: 'none',
```
```diff
-  containerShape: 'full' as ShapeToken,
+  containerShape: 'full',
```

Deleting them is the whole fix — no `satisfies` needed, the values are checked at the usage site in `resolveCornerRadius`. Same at `Toolbar.test.tsx:392`: type the `it.each` table as a `const` instead of casting it.

JKobrynski (changes requested):

src/components/Toolbar/utils.ts
k0ndee: I added a ToolbarColorContext and modified Button and IconButton to pick up the Toolbar colors. Now they will work regardless of how deep they are nested inside any Views or custom components.

src/components/Toolbar/Toolbar.tsx
k0ndee: I removed the default testID

docs/src/data/themeColors.ts
k0ndee: Good catch, I forgot to udpate the docs when I changed tokens. Now they are correct.

src/components/Toolbar/tokens.ts
k0ndee: I removed casts.

src/components/__tests__/Toolbar.test.tsx
```diff
@@ -0,0 +1,413 @@
...
+    await render(
+      <Toolbar colorScheme="vibrant">
+        <IconButton icon="format-bold" mode="outlined" onPress={() => {}} />
+      </Toolbar>
+    )
+  ).toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it('recolors IconButtons through a Fragment held in a variable, not just direct children', async () => {
```
JKobrynski, line 192: The point of moving to context is that arbitrarily nested children get recoloured, not just direct ones. Nothing covers that. The suite still only has the case on this line, "recolors IconButtons through a Fragment held in a variable" — which the old `cloneElement` traversal already handled via its `Fragment` branch.

Please add the case that was actually broken:

```tsx
<Toolbar colorScheme="vibrant">
  <View>
    <IconButton icon="magnify" testID="wrapped" />
  </View>
</Toolbar>
```

and assert the icon resolves to `onPrimaryContainer`. Without it the regression can come back silently.

JKobrynski (commented):
One minor comment

src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,271 @@
...
+  /**
+   * Accessibility label for the toolbar group. `children` still need
+   * their own `aria-label`s.
+   */
+  'aria-label'?: string;
+  /**
+   * @optional
+   */
+  theme?: ThemeProp;
+  ref?: React.RefObject<View>;
```
JKobrynski, line 78:
```diff
-  ref?: React.RefObject<View>;
+  ref?: React.Ref<View>;
```

`RefObject` rejects callback refs. `Surface.tsx:56`, `IconButton.tsx:73` and `FAB.tsx:85` all use the wider type. Also worth a JSDoc line that the ref lands on `Surface` for `floating` (`:182`) but the positioning wrapper for `docked` (`:232`).

JKobrynski (commented):
I think it's the last one, and it's minor

src/components/Toolbar/Toolbar.tsx
k0ndee: Widening the ref type to `React.Ref<View>` broke `yarn typecheck` (`tsc -b`) on the `docked` wrapper's raw `<View ref={ref}>` (cross-project `@types/react` binding issue with the callback-ref cleanup type). Fixing it means swapping that wrapper off a plain `View`, and I'm not sure which way you'd prefer:

- **RN core `Animated.View`** — matches the existing pattern in `Surface.tsx`/`FAB.tsx`/`IconButton.tsx`, and gets `collapsable={false}` for free (needed since this node is now a real ref target).
- **`react-native-reanimated`'s `Reanimated.View`** — matches where v6 seems to be heading overall, but doesn't auto-add `collapsable={false}` (needs it added by hand), and this file currently has zero Reanimated usage otherwise (that's slated for a later PR).

I will change the wrapper to `Reanimated.View` in the following PR that will bring animation to this component anyway, but I'd rather clarify how we should handle that situation for now. Let me know which fits the intended direction better and I'll settle on it.

JKobrynski: Good catch, and thanks for not just forcing it through. I'd rather not restructure the render tree to satisfy a type error, though — I think the cause is elsewhere. Root has @types/react 19.2.17 and example/ has 19.2.7, with no resolutions pinning them, and typecheck is tsc -b across both projects. RefCallback carries an augmentable interface, so two copies of the types give you structurally incompatible callback refs — that's most likely what you hit on the plain <View>, and it'll hit the next component that widens a ref too.

On your two options: Animated.View from react-native is the legacy Animated we're moving off, so I'd avoid it in new code even though Surface still uses it. And pulling in Reanimated here ahead of the animation PR seems premature.

Simplest path: leave ref?: React.RefObject<View> as it is for now and I'll open a separate issue for aligning @types/react plus widening the refs across the board. Not worth blocking this PR on. If you'd rather fix it here, adding a resolutions entry for @types/react and keeping the plain <View> should do it — worth checking yarn typecheck passes with that alone.

k0ndee: Sounds good, leaving `ref?: React.RefObject<View>` as-is for this PR and I updated the JSDoc to include information where the ref lands depending on variant.

Tried the resolutions pin as a sanity check. I added `resolutions: { "@types/react": "19.2.7" }` and reinstalled. Root and `example` do resolve to the identical `@types/react` package on disk afterward, but `yarn typecheck` still fails with the same "two different types with this name exist, but they are unrelated" error on the callback-ref's cleanup type. So it's not version skew, `tsc -b` compiles root and `example` as separate project-reference graphs, and `RefCallback`'s cleanup type (built off a branded/unique-symbol type) gets a fresh, structurally-incompatible instantiation per graph regardless of whether the underlying package version matches.

JKobrynski:
LGTM! @satya164 do you want to take a look?

src/components/__tests__/__snapshots__/Toolbar.test.tsx.snap
```diff
@@ -0,0 +1,3388 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
JKobrynski, line 1: I think you need to rerun these with `-u` after rebasing, because default testIDs were removed

src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,274 @@
...
+  colorScheme?: ColorScheme;
+  /**
+   * Override the container (background) color.
+   */
+  containerColor?: ColorValue;
+  /**
+   * Style for positioning `floating`'s pill, or overriding `docked`'s
+   * default anchoring.
+   */
+  style?: StyleProp<ViewStyle>;
```
JKobrynski: What happens when you pass `backgroundColor|borderRadius` through this prop? I'm afraid they'd silently do nothing, other components type `style` as `styleProp<SurfaceStyle>`

src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,274 @@
...
+import { ToolbarTokens } from './tokens';
+import type { ColorScheme, Orientation, Variant } from './tokens';
+import { ToolbarColorContext } from './ToolbarColorContext';
+import { getSpacing, resolveContainerColor, resolveElevation } from './utils';
+import { useInternalTheme } from '../../core/theming';
+import type { ThemeProp } from '../../theme/types';
+import { resolveCornerRadius } from '../../theme/utils/shape';
+import Surface from '../Surface';
+
+export type Props = {
```
JKobrynski: Other components use `Omit<ViewProps, 'style'> to allow props like `onLayout`, does that work here?

example/src/Examples/ToolbarExample.tsx
```diff
@@ -0,0 +1,239 @@
...
+      />
+      {isFloating ? (
+        // `floating` doesn't anchor itself (same as `FAB`)—position it with
+        // a wrapping `View`, same as the component's own doc example.
+        <View
+          pointerEvents="box-none"
+          style={
+            isVertical
+              ? [styles.verticalAnchor, { right: insets.right + 16 }]
+              : [styles.horizontalAnchor, { bottom: insets.bottom + 24 }]
```
JKobrynski: Have you checked the edge offsets against the [toolbar specs](https://m3.material.io/components/toolbars/specs)? I'm afraid the two are swapped

src/components/__tests__/__snapshots__/Toolbar.test.tsx.snap
k0ndee: You were right, the PR was still behind main branch. I rebased it and now it is up to date.

example/src/Examples/ToolbarExample.tsx
k0ndee: The horizontal edge offset is incorrect, it should be 16. As for the vertical one the only spec in MD3 is that the minimum offset from screen edge should be equal to 16 so I left it as it is.

src/components/Toolbar/Toolbar.tsx
k0ndee: You are correct, fixed it by changing `style` type to `StyleProp<SurfaceStyle> `

src/components/Toolbar/Toolbar.tsx
k0ndee: Fixed it, changed `Props` to `Omit<ViewProps, 'style'>`

src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,274 @@
...
+    return pill;
+  }
+
+  return (
+    <View
+      {...rest}
+      ref={ref}
+      pointerEvents="box-none"
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      style={[styles.dockedContainer, style as StyleProp<ViewStyle>]}
```
JKobrynski, line 240: Do we really need to use `as` here?

k0ndee: Yes, it's needed. `style` is typed as `StyleProp<SurfaceStyle>`, and `SurfaceStyle` wraps `ViewStyle` in Reanimated's `AnimatedStyle` (so it typechecks when forwarded to `Surface`, which is an `Animated.View`). Here it's applied to a plain `View` instead, which only accepts plain `ViewStyle` - hence the mismatch.

The cast is safe in practice: this wrapper isn't animated, so no shared-value-typed style ever reaches it. I also checked whether swapping this to Reanimated's `Animated.View` would avoid the cast - it does typecheck, but it always renders with `collapsable={false}`, which is a permanent Android view-flattening cost for every Toolbar (docked or not) just to accommodate a case that doesn't apply here.

JKobrynski: Have you tried a Reanimated style on a `docked` toolbar? I'm afraid it's silently dropped here, though `floating` applies it.

example/src/Examples/ToolbarExample.tsx
JKobrynski: There's a separate vertical token of 24dp, against 16dp for horizontal ([specs](https://m3.material.io/components/toolbars/specs)). Worth using 24 here?

src/components/Toolbar/Toolbar.tsx
k0ndee: You're right. I changed it so the `docked` toolbar wrapper is `Animated.View`.

example/src/Examples/ToolbarExample.tsx
k0ndee: I looked at the wrong tokens. I can see it now and I updated it so that now it aligns with the spec.

JKobrynski (approved):
LGTM! @satya164

PR #5044, opened 2026-08-12
fix(tooltip): propagate press and hover event objects to children

### LINKED ISSUE #5003: Tooltip: onPress event object is not propagated to parent component on native
### Current behaviour
onPress (and probably other pressable-related) event object is not propagated to parent component on native (Android/iOS).

This line is suspicious:
https://github.com/callstack/react-native-paper/blob/main/src/components/Tooltip/Tooltip.tsx#L156C18-L156C25

### Expected behaviour

Event object should be passed to parent callbacks.

### How to reproduce?

https://snack.expo.dev/@szado/rnp-tooltip-onpress-bug

### Your Environment

| react-native-paper    | 5.15.3
| expo sdk                     | 55

aliyilmaztech:
### Motivation

`Tooltip` invokes the wrapped children's handlers itself, but it called them without any arguments:

```ts
return props.onPress?.();
```

So any consumer callback wrapped in a `Tooltip` received `undefined` instead of the event, e.g. `onPress={(e) => e.nativeEvent...}` crashed or silently misbehaved on Android/iOS. The same applied to `onHoverIn`/`onHoverOut` on web.

This PR forwards the original `GestureResponderEvent`/`MouseEvent` to the children handlers and types `TooltipChildProps` accordingly, so the event object reaches parent callbacks.

### Related issue

Closes #5003

### Test plan

- `yarn test`, `yarn lint` and `yarn typecheck` pass.
- Added unit tests in `src/components/__tests__/Tooltip.test.tsx`:
  - mobile: pressing the trigger calls the children's `onPress` with an event object (this test fails on `main`),
  - web: `onHoverIn`/`onHoverOut` receive the event object.
- Repro from the issue (https://snack.expo.dev/@szado/rnp-tooltip-onpress-bug): the child's `onPress` now receives the event instead of `undefined`.

No API change other than the widened (previously argument-less) handler types on `TooltipChildProps`.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -6,7 +6,12 @@ import {
   Platform,
   Pressable,
 } from 'react-native';
-import type { LayoutChangeEvent, ViewStyle } from 'react-native';
+import type {
+  GestureResponderEvent,
+  LayoutChangeEvent,
+  MouseEvent,
+  ViewStyle,
+} from 'react-native';
 
 import { getTooltipPosition } from './utils';
 import type { Measurement, TooltipChildProps } from './utils';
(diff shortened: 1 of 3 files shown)


JKobrynski:
@aliyilmaztech we've got conflicts, could you resolve them? 🙏

PR #5047, opened 2026-08-13
fix(provider): deep merge theme prop so a partial theme keeps defaults

### LINKED ISSUE #4589: Warning: Error: Variant titleLarge was not provided properly. Valid variants are regular, medium, light, thin.
### Current behaviour
After yarn react-native run-android is run, the following error occurs:
Image

### Preview
Program entry file 》index.js:
`import { AppRegistry } from 'react-native';
import { PaperProvider } from 'react-native-paper';
import App from './App';
import { name as appName } from './app.json';
import appTheme from './utils/appTheme';

export default function Main() {
    return (
        <PaperProvider theme={appTheme}>
            <App />
        </PaperProvider>
    );
}
AppRegistry.registerComponent(appName, () => Main);`
------------------------------------------------------------------

App.js：
`import React, { useState, useEffect } from 'react';
import { View, StyleSheet, TouchableOpacity } from 'react-native';
import { Appbar, Text, useTheme } from 'react-native-paper';
import api from './utils/netWorkTool';
import constants from './utils/constants';

const createStyles = (theme) => StyleSheet.create({
    container: {
        flex: 1,
        justifyContent: 'flex-start',
        backgroundColor: theme.colors.background,
    },
});

const App = () => {
    const theme = useTheme();

    const styles = createStyles(theme);

    const [klineData, setKlineData] = useState([]);
    const [tradingPair, setTradingPair] = useState(constants.TRADINGPAIR.BTCUSDT);
    const [timecycle, setCycle] = useState(constants.TIMECYCLE.ONE_H);
    const [error, setError] = useState(null);

    useEffect(() => {
        const fetchKlineData = async () => {
            try {
                const response = await api.get('/lines', {
                });
                setKlineData(response);
            } catch (error) {
                setError(error);
            }
        };

        fetchKlineData();
    }, []);

    const handleTitlePress = () => { };

    return (
            <Appbar.Header >
                <TouchableOpacity onPress={handleTitlePress}>
                    <Appbar.Content title={tradingPair} />
                </TouchableOpacity>
            </Appbar.Header>
            <Text style={{ color: theme.colors.primary }}>show data</Text>
        </View>
    );
};
export default App;`

I tried changing the <Appbar.Content title={tradingPair} /> in app.js to the following: 
<Appbar.Content title={tradingPair} titleVariant="'titleMedium'" />
But the error still exists

Environment

| OS                                | windows 11
| android                        | 9
| react-native                | 0.76.5
| react-native-paper    | ^5.12.5
| node                            | v18.20.5
| yarn                             | 1.22.22
| expo sdk                     | x.

giaBaoJS:
## Summary

`PaperProvider` replaces the whole theme object instead of merging it, so any partial `theme` prop silently drops the defaults it doesn't mention. The most visible casualty is `fonts`: supplying one makes every `<Text variant="…">` in the app throw.

Fixes #4589.

## Reproduction

@lukewalczak asked for a repro on the issue and never got one, so here it is as a runnable snippet — this is the whole thing, no app scaffolding needed:

```jsx
import { PaperProvider, Text } from 'react-native-paper';

// Anything partial will do. This is the v2 `configureFonts` shape that most
// people upgrading from v4/v5 still have in their codebase.
const theme = {
  fonts: {
    regular: { fontFamily: 'System', fontWeight: '400' },
    medium: { fontFamily: 'System', fontWeight: '500' },
    light: { fontFamily: 'System', fontWeight: '300' },
    thin: { fontFamily: 'System', fontWeight: '100' },
  },
};

export default () => (
  <PaperProvider theme={theme}>
    <Text variant="titleLarge">Hello</Text>
  </PaperProvider>
);
```

Throws immediately:

```
Variant titleLarge was not provided properly.
Valid variants are regular, medium, light, thin.
```

That trailing list is literally `Object.keys(theme.fonts)` in [`Text.tsx#L136-L138`](https://github.com/callstack/react-native-paper/blob/main/src/components/Typography/Text.tsx#L136-L138) printing back the user's four keys — proof that the 15 MD3 typescale variants are gone by the time `Text` reads the theme, not that the user's `fonts` was malformed. The in-thread workaround (pinning to 5.12.3) matches: this predates the typescale becoming mandatory.

It is committed as a regression test in this PR, so the repro lives in CI rather than in a Snack that can rot.

## Root cause

[`src/core/PaperProvider.tsx#L39-L44`](https://github.com/callstack/react-native-paper/blob/main/src/core/PaperProvider.tsx#L39-L44):

```ts
return {
  ...base,
  ...props.theme,
  colors: { ...base.colors, ...props.theme?.colors },   // merged
  animation: { ...props.theme?.animation, scale },
};
```

`colors` gets re-merged against `base` on the line below the spread. **Nothing else does.** `...props.theme` therefore replaces `fonts`, `shapes`, `motion` and `elevation` wholesale with whatever fragment the user passed.

Worth stressing that this is an internal inconsistency, not just a missing case. `src/theme/provider.tsx` already exports `safeMerge`, and `useInternalTheme` uses it to deep-merge the **per-component** `theme` prop. So today:

```jsx
<Button theme={{ fonts: { regular } }} />      {/* deep merged — fine   */}
<PaperProvider theme={{ fonts: { regular } }} /> {/* shallow — wipes it */}
```

The same prop, on the same shape, with two different merge semantics depending on where you put it.

`fonts` is what #4589 reports because it is the only sub-object with a hard `throw` behind it. `shapes`, `motion` and `elevation` fail identically but silently — and they're nested two or three levels deep (`shapes.corner.small`, `motion.spring.fast.spatial.stiffness`), so a one-level merge wouldn't save them either.

## The fix

```diff
-    return {
-      ...base,
-      ...props.theme,
-      colors: { ...base.colors, ...props.theme?.colors },
-      animation: { ...props.theme?.animation, scale },
-    };
+    const merged = safeMerge(base, props.theme);
+
+    return {
+      ...merged,
+      animation: { ...merged.animation, scale },
+    };
```

Reusing `safeMerge` rather than adding a `fonts:` line next to the `colors:` one, because:

- it fixes `shapes` / `motion` / `elevation` too, which have the identical defect and are too deeply nested for a shallow merge;
- it makes the provider-level and component-level `theme` props agree, which is the actual inconsistency here;
- it is the helper you already reach for in this exact situation, including the `PlatformColor` sentinel handling added in `8720eac` — no new merge logic enters the codebase;
- it is a smaller diff than enumerating five sub-objects.

`animation.scale` is still applied last so `reduceMotion` keeps overriding it.

I'm happy to switch to the minimal, `fonts`-only version if you'd rather keep the blast radius at exactly the reported bug:

```diff
       colors: { ...base.colors, ...props.theme?.colors },
+      fonts: { ...base.fonts, ...props.theme?.fonts },
```

Just say the word and I'll swap it — but it leaves the other three broken.

## Behaviour change to be aware of

Anyone deliberately **replacing** rather than extending a sub-object now gets the defaults merged underneath it. In practice this is hard to observe: a complete `configureFonts()` output overrides every key anyway (covered by a test below), and there's no way to express "and delete the rest" through a `$DeepPartial` prop today. But it is a real semantic change and worth a line in the release notes.

Deep merge also means individual *variants* inherit missing properties — `fonts: { titleLarge: { fontSize: 20 } }` now keeps the default `lineHeight`/`letterSpacing` instead of dropping them. That's the behaviour `useInternalTheme` already has, and it's the friendlier of the two.

## Tests

Five tests added to `src/core/__tests__/PaperProvider.test.tsx`, using the existing `FakeChild` + `useTheme` harness. Three fail on `main`, two pass on `main` as controls.

Fail before the fix:

- `keeps the base typescale when only part of theme.fonts is provided` — `theme.fonts.titleLarge` is `undefined` on `main`.
- `renders <Text variant> instead of throwing when theme.fonts is partial` — asserts the **user-visible** symptom. On `main` it fails with the issue's error string verbatim, which is the point: the shape assertion above could pass while the app still crashed.
- `keeps sibling tokens when a nested shape token is overridden` — demonstrates the non-`fonts` half of the bug.

Pass before *and* after, as non-vacuity controls:

- `still merges theme.colors with the base palette` — `colors` was already correct; this proves the harness observes real merging rather than always going green.
- `lets a complete fonts object override every default variant` — guards the behaviour change above: a full typescale still wins outright.

I deliberately did **not** add a test for the `animation` line. `...props.theme?.animation` drops `base.animation` for the same reason, but `Theme['animation']` currently has exactly one key (`scale`) which is recomputed on the next line, so the omission is unobservable and any test for it would pass with or without the change. The fix restores `...merged.animation` anyway so it stays correct if a second key is ever added.

## Verification

```
Before:  55 suites, 732 passed, 1 skipped, 169 snapshots — exit 0
After:   55 suites, 737 passed, 1 skipped, 169 snapshots — exit 0
```

Zero snapshot churn — all 169 pass unwritten, since they render under the default theme where deep and shallow merge coincide. `yarn lint` and `yarn typecheck` are clean.


src/core/PaperProvider.tsx
```diff
@@ -36,11 +36,14 @@ const PaperProvider = (props: Props) => {
       ? 0
       : (props.theme?.animation?.scale ?? 1);
 
+    // Deep merge so a partial theme extends the defaults instead of replacing
+    // them. `safeMerge` is the same helper `useInternalTheme` uses for the
+    // per-component `theme` prop, which keeps both levels consistent.
+    const merged = safeMerge(base, props.theme);
```
MikitasK, line 43: calling `safeMerge` at the theme root can drop all defaults when a custom theme contains `dynamic`, `semantic` or `resource_paths`

current [sentinel check](https://github.com/callstack/react-native-paper/blob/1b2b0433ea918b2949b192dfc52bef30f568430c/src/theme/provider.tsx#L27-L41) treats any object containing one of these keys as a native color, while [custom theme properties are supported](https://github.com/callstack/react-native-paper/blob/1b2b0433ea918b2949b192dfc52bef30f568430c/docs/6.x/docs/guides/theming.mdx#L146-L165)

```
safeMerge(base, { dynamic: true });
// => { dynamic: true }, dropping fonts, colors, shapes etc.
```

RN’s actual native-color values have narrower shapes, such as `{ semantic: string[] }` or `{ dynamic: { light, dark } }` ([source](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/StyleSheet/PlatformColorValueTypes.ios.js))

so could we validate those shapes before using `safeMerge` here & add regression test for custom `dynamic` property?

giaBaoJS: Good catch — you're right, and this is a real regression introduced by moving `safeMerge` to the theme root. Fixed in 25363ae.

Reproduced first: `safeMerge(base, { dynamic: true })` returned `{ dynamic: true }`, and `<PaperProvider theme={{ dynamic: true }}>` made `<Text variant="titleLarge">` throw again — the exact symptom this PR set out to fix.

`isPlatformColorSentinel` now validates the value, not just the key name. Checking the shapes React Native actually emits (`PlatformColorValueTypes.ios.js` L25-28 and L37-50, `PlatformColorValueTypes.android.js` L19-23), a native color value carries exactly one key and nothing else, so the check now accepts only:

- `{ semantic: string[] }` — iOS `PlatformColor`
- `{ resource_paths: string[] }` — Android `PlatformColor`
- `{ dynamic: { light, dark, highContrastLight?, highContrastDark? } }` — `DynamicColorIOS`, requiring both `light` and `dark` and rejecting any key outside that tuple

Everything else — including `{ dynamic: true }`, `{ semantic: 'label' }`, and any object that carries a sentinel key alongside siblings such as `colors`/`fonts` — is now an ordinary object and gets deep-merged, so a documented custom theme property called `dynamic` keeps every default.

Regression tests added, as requested:

- `PaperProvider.test.tsx`: a theme `{ dynamic: true, colors: { primary: 'tomato' } }` keeps the typescale, palette and shape tokens, plus a user-visible one asserting `<Text variant="titleLarge">` renders under `theme={{ dynamic: true }}`.
- `provider.test.ts`: rejection cases for the three key names with wrong-shaped values and with sibling keys.

Two of the new tests are deliberately non-vacuous controls that pass both before and after, so the tightening can't quietly degenerate into "nothing is ever a sentinel": one asserts real `PlatformColor('label')` and `DynamicColorIOS(...)` values are still detected, and one asserts a real `DynamicColorIOS` override still passes through `safeMerge` by identity, with no `highContrast*` keys inherited from the base sentinel underneath it. With the sentinel change reverted the six new tests fail and those two still pass.

Suite: 55 suites / 737 → 745 passed / 1 skipped, 169 snapshots unchanged; lint and typecheck clean.

MikitasK: thanks for covering this up 🙌 
just one narrow non-blocking edge seems to remain - a custom property that matches the sentinel shape exactly is still treated as the whole theme being a native color:

```ts
safeMerge(base, {
  dynamic: { light: '#fff', dark: '#000' },
});
// => only the override; fonts/colors/shapes are dropped
```

this is because `safeMerge` still checks `isPlatformColorSentinel(overrides)` ([source](https://github.com/callstack/react-native-paper/blob/25363ae13ed50752f910e77954513dc164c199aa/src/theme/provider.tsx#L80-L92)). 
since arbitrary custom theme properties are supported, could we avoid treating complete theme override as a color sentinel? or otherwise make the check root-aware?

the real native-color cases should remain covered: primitive/absent bases already take the leaf branch, while sentinel-to-sentinel replacement is handled by `isPlatformColorSentinel(base)`

MikitasK (approved):
LGTM 👍 
just one non-blocking comment from my side:

src/core/PaperProvider.tsx
giaBaoJS: Sorry for the slow turnaround here. You were right, and it reproduces exactly as you wrote it. Running your snippet against the branch as it stood:

```
safeMerge(
  { colors: {...}, fonts: {...}, shapes: {...} },
  { dynamic: { light: '#fff', dark: '#000' } }
)

// keys = ["dynamic"]
```

Every default gone, not just theoretical.

Fixed by making the check root-aware, as you suggested. The sentinel test on `overrides` now runs only on nested values, in a small helper the recursion goes through:

```ts
const mergeThemeValue = (base: unknown, overrides: unknown): unknown =>
  isPlatformColorSentinel(overrides) ? overrides : safeMerge(base, overrides);
```

`safeMerge` itself keeps `isPlatformColorSentinel(base)` and the rest of the leaf conditions. Same snippet after the change gives `["colors", "fonts", "shapes", "dynamic"]`.

Your reasoning about the native-color cases holds; I checked each one against the code rather than assuming:

- Root `overrides` is never a color. There are only two call sites, `PaperProvider`'s `theme` prop and `useInternalTheme`'s `themeOverrides`, and both pass a theme.
- Primitive or absent base already takes the leaf branch on `!base` / `typeof base !== 'object'`, so a sentinel landing on `'#000'` or on a key the base does not have is unaffected.
- Sentinel-to-sentinel replacement is handled by `isPlatformColorSentinel(base)`.

Three tests added in `src/theme/__tests__/provider.test.ts`: your case at the root for both the `dynamic` and `semantic` shapes, plus two that pin the nested check so this cannot be relaxed further. The one that matters is a sentinel replacing a plain object, where the base is mergeable and nothing except the check on `overrides` stops the recursion.

Counterfactual on both directions. Reverting only `provider.tsx` turns the new root test red for the right reason (`result.fonts` is `undefined`, base dropped) while the two nested tests stay green. Deleting the nested check instead turns the plain-object test red and the root test green. Full suite is green: 748 passed, 1 skipped, 55 suites.

giaBaoJS:
Rebased onto current main to clear the conflict with #5098 and #5087. One content change fell out of it: main now enables `@typescript-eslint/no-unsafe-type-assertion`, which flagged the `(v as Record<string, unknown>)[key]` line in `isPlatformColorSentinel`. Rewrote it as a single `Object.entries` destructure rather than widening the eslint-disable block, so no new suppression is added.

`yarn typecheck --force` and `yarn lint` clean, provider and PaperProvider suites 41/41. Counterfactual still holds on the rebased head: reverting only `src/theme/provider.tsx` fails `keeps the base when the whole override matches the sentinel shape`, 1 failed / 20 passed.

giaBaoJS:
Rebased on current `main`. One upstream change needed adapting to: `defaultThemes` was removed from `src/theme/provider.tsx`, so `PaperProvider` now takes its base theme from `DarkTheme`/`LightTheme` directly and imports only `safeMerge` from `./theming`. The deep merge itself is unchanged, and `useInternalTheme`'s new reference-preserving behaviour is kept as is.

I also added a commit for the Lint job, which was already failing on the previous head. The deliberately off-spec test themes (a v2 `fonts` object, and a theme with a custom `dynamic` property) now carry `@ts-expect-error` instead of `as ThemeProp`, which `@typescript-eslint/no-unsafe-type-assertion` rejects. `@ts-expect-error` also fails if those inputs ever become valid, so the intent stays checked.

PR #5050, opened 2026-08-15
fix: remove duplicate tabstop in RadioButton.Item

### LINKED ISSUE #4775: RadioButton.Item has duplicate tabstop
### Current behaviour
RadioButton.Item contains two elements with a tabIndex of 0, so it takes 2 tabs to go past it.

### Expected behaviour
There should be only one tabbable element, since they do the same thing

### How to reproduce?
Issue is visible in the example snack (<https://snack.expo.dev/@react-native-paper/react-native-paper-example_v5>)  under the "Radio Button Item" page, at least on the Web platform. Click an item, and then press Tab a few times to observe the behaviour

nothing

### Your Environment

| ios                                | n/a
| android                        | n/a
| react-native                | 0.80.1
| node                            | 22.13.1
| npm or yarn                | npm 10.9.2
| expo sdk                     | n/a

ErfanBagheri404:
Fixes #4775

RadioButton.Item renders an outer TouchableRipple plus the inner RadioButton, both focusable on web. Two tabstops per item.

Disable the inner radio button focusability so only the outer touchable is a tabstop.

### DIFF
--- a/src/components/RadioButton/RadioButtonItem.tsx
+++ b/src/components/RadioButton/RadioButtonItem.tsx
@@ -165,7 +165,12 @@ const RadioButtonItem = ({
     color,
     theme,
     uncheckedColor,
-  };
+    // The outer TouchableRipple in RadioButtonItem already provides the
+    // interactive surface. Disable the inner RadioButton's focusability
+    // so there is only one tabstop per radio button item on web.
+    focusable: false,
+    tabIndex: -1,
+  } as const;
   const isLeading = position === 'leading';
   let radioButton: any;
 
--- a/src/components/RadioButton/RadioButtonItem.tsx
+++ b/src/components/RadioButton/RadioButtonItem.tsx
@@ -166,10 +166,14 @@ const RadioButtonItem = ({
     theme,
     uncheckedColor,
     // The outer TouchableRipple in RadioButtonItem already provides the
-    // interactive surface. Disable the inner RadioButton's focusability
-    // so there is only one tabstop per radio button item on web.
+    // interactive surface. Hide the inner RadioButton from the
+    // accessibility tree so screen readers only encounter one control,
+    // and keep tabIndex: -1 so web keyboard users have a single
+    // tabstop per row (matches the approach used in Checkbox.Item).
     focusable: false,
     tabIndex: -1,
+    accessible: false,
+    'aria-hidden': true,
   } as const;
   const isLeading = position === 'leading';
   let radioButton: any;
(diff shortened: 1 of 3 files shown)


src/components/RadioButton/RadioButtonItem.tsx
```diff
@@ -165,7 +165,12 @@ const RadioButtonItem = ({
     color,
     theme,
     uncheckedColor,
-  };
+    // The outer TouchableRipple in RadioButtonItem already provides the
+    // interactive surface. Disable the inner RadioButton's focusability
+    // so there is only one tabstop per radio button item on web.
+    focusable: false,
+    tabIndex: -1,
```
MikitasK, line 174: could we also hide inner `RadioButton` from the accessibility tree?

`tabIndex: -1` only removes it from tab sequence - it remains focusable programmatically according to [W3C guidance](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex)
inner [Android](https://github.com/callstack/react-native-paper/blob/1b2b0433ea918b2949b192dfc52bef30f568430c/src/components/RadioButton/RadioButtonAndroid.tsx#L122-L141) & [iOS](https://github.com/callstack/react-native-paper/blob/1b2b0433ea918b2949b192dfc52bef30f568430c/src/components/RadioButton/RadioButtonIOS.tsx#L80-L99) controls still expose `role="radio"`  & `aria-checked`, so screen readers may encounter 2 radio controls

so could we explicitly hide inner visual control as documented for [aria-hidden](https://reactnative.dev/docs/accessibility#aria-hidden)?

```suggestion
    tabIndex: -1,
    accessible: false,
    'aria-hidden': true,
```

this follows the same approach already used by [Checkbox.Item](https://github.com/callstack/react-native-paper/blob/1b2b0433ea918b2949b192dfc52bef30f568430c/src/components/Checkbox/CheckboxItem.tsx#L143-L146)

src/components/RadioButton/RadioButtonItem.tsx
MikitasK, line 1: what about updating the snapshots & adding regression test here? 
existing `RadioButton.Item` tests use `toMatchSnapshot()` & those snapshots currently contain `focusable={true}` for inner radio. 

since this PR changes that rendered output but only updates the component file, the snapshot tests will fail when CI runs, I suppose

it would also be useful to guard against the duplicate control explicitly:
```
it('exposes only one radio control', async () => {
  await render(
    <RadioButton.Item
      label="Radio button"
      value="radio"
      status="unchecked"
    />
  );

  expect(screen.getAllByRole('radio')).toHaveLength(1);
});
```

ErfanBagheri404:
Both review points are already addressed in the latest push (b3b6612f): accessible=false + aria-hidden=true on the inner RadioButton, snapshot updated, and the regression test with getAllByRole radio assertion was added. The review was posted after the push so GitHub still showed the unresolved thread.

MikitasK (approved):
LGTM ✅

PR #5051, opened 2026-08-15
fix: allow Menu to reopen when hide animation callback never fires

### LINKED ISSUE #4763: Menu component closes and fails to open again on re-render

ErfanBagheri404:
Fixes #4763

On React Native >= 0.80 the hide animation completion callback is not delivered, so prevRendered.current stayed true and the menu could never be shown again after being dismissed.

Reset prevRendered.current before starting the hide animation instead of in its completion callback.

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -19,6 +19,7 @@ import type {
 } from 'react-native';
 
 import Animated, {
+  cancelAnimation,
   Easing,
   ReduceMotion,
   useAnimatedStyle,
@@ -225,6 +226,10 @@ const Menu = ({
   const anchorRef = React.useRef<View | null>(null);
   const menuRef = React.useRef<View | null>(null);
   const isShownRef = React.useRef(false);
+  // Mirrors the `rendered` state so animation callbacks (which may fire late
+  // or never on React Native >= 0.80, see #4763) can tell whether the reset
+  // already happened in `hide()`.
+  const renderedRef = React.useRef(rendered);
 
   const keyboardDidShow = React.useCallback((e: RNKeyboardEvent) => {
     const keyboardHeight = e.endCoordinates.height;
@@ -322,15 +327,14 @@ const Menu = ({
     focusFirstDOMNode(menuRef.current);
   });
 
-  const handleHideAnimationFinished = useLatestCallback((finished: boolean) => {
-    if (!finished || prevVisible.current) {
-      return;
+  const handleHideAnimationFinished = useLatestCallback(() => {
+    // The completion callback never fires on React Native >= 0.80 (see
+    // #4763), so `hide()` already reset the portal and focus synchronously.
+    // Only run the layout cleanup if the menu was not reopened before the
+    // animation finished (reopening resets `renderedRef` via the state path).
+    if (!renderedRef.current) {
+      setMenuLayout({ width: 0, height: 0 });
     }
-
-    setMenuLayout({ width: 0, height: 0 });
-    setRendered(false);
-    isShownRef.current = false;
-    focusFirstDOMNode(anchorRef.current);
   });
 
   const show = React.useCallback(async () => {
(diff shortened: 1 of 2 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -379,7 +385,6 @@ const Menu = ({
     }).start(() => {
       setMenuLayout({ width: 0, height: 0 });
       setRendered(false);
-      prevRendered.current = false;
       focusFirstDOMNode(anchorRef.current);
     });
```
MikitasK: I suppose, moving `prevRendered.current = false` before hide animation can introduce a race. if the menu closes while its opening animation is running, the interrupted opening callback can set this back to `true`, preventing the menu from reopening

could we handle interrupted opening & hiding animations using `finished` and add a test for quickly closing & reopening the menu?

```
}).start(({ finished }) => {
  if (!finished) {
    return;
  }
  setMenuLayout({ width: 0, height: 0 });
  setRendered(false);
  prevRendered.current = false;
  focusFirstDOMNode(anchorRef.current);
});
```

MikitasK (approved):
Thanks @ErfanBagheri404 🙏 
LGTM 👌 
just one non-blocking suggestion: regression test covering close & reopen behavior would be helpful

src/components/Menu/Menu.tsx
```diff
@@ -379,7 +387,6 @@ const Menu = ({
...
```
satya164: resetting `prevRendered.current` but leaving everything else like `rendered` state and focus restoration doesn't make sense and leaves the menu in a broken state if the callback is never delivered.

though the callback not firing sounds like a big regression in react native so a lot of other things should be broken

satya164 (changes requested):
according to this comment https://github.com/callstack/react-native-paper/issues/4763#issuecomment-3012224234 - the issue happens in react native 0.79

but the pr says React Native >= 0.80. can you verify that you're fixing the same issue?

also please provide before/after videos.

ErfanBagheri404:
On the version question: the reports in the thread are consistent about the boundary. bjeld hit it after upgrading to RN 0.80 with paper 5.14.5 and downgrading to 0.79.4 fixed it, atheck and adamgajzlerowicz both on 0.80.1, raiyanihiren on 0.81. Nobody reported it below 0.80, so "React Native >= 0.80" matches what users are seeing.

The PR is rebased onto the current main now that the Menu component was rewritten to Reanimated. In that rewrite the reopen deadlock is gone, but the broken state you described is still there and is what this fixes now: when the hide callback is not delivered, `setRendered(false)` never ran, so the portal stayed mounted after close and `focusFirstDOMNode(anchorRef.current)` never returned focus to the anchor. `hide()` now resets the portal and focus synchronously, cancels any in-flight show animation so a late `finished` callback cannot mark a hidden menu as shown, and the completion callback only zeroes `menuLayout` when the menu was not reopened meanwhile.

I can't record before/after videos myself, no RN setup here to film one. The close-reopen regression test in the PR covers the visible behavior; if a video is still a hard requirement for merging, someone with a 0.80 device would need to record it.

PR #4998, opened 2026-06-15
feat: add SplitButton component

draggie:
### Motivation

  Implements the new Material Design 3 SplitButton component as part of the v6 component modernization work.

  The component provides a primary leading action and a separate trailing action for contextual options. It reuses the modernized Button API direction
  from #4928/#4943: explicit `label`/`icon` props, MD3 mode names (`filled`, `tonal`, `elevated`, `outlined`), derived ripple/state-layer colors,
  extracted component tokens, size metrics, shape tokens, disabled/loading states, and separate accessibility/press handlers for each button segment.

  This also adds the component to the public exports, documentation generation config, theme color docs data, and the example app. The example includes a
  Menu-based playground, mode showcase, loading/disabled controls, and custom color/label styling.

  ### Related issue

  Closes #4986

  Related to #4928 and #4943

  ### Test plan

  - `yarn lint-no-fix`
    - Passes with one existing unrelated warning in `src/components/__tests__/TextInput.test.tsx`.
  - `yarn typescript`
  - `yarn test --watchman=false`
  - `yarn docs build`

  Manual verification in the example app:

  - Open the `SplitButton` example screen.
  - In the playground, tap the leading `Send` action and trailing menu action separately.
  - Verify the trailing menu opens below the SplitButton anchor and does not cover the button.
  - Toggle `Disabled` and verify both press targets are disabled and styled correctly.
  - Toggle `Loading` and verify the leading icon is replaced with a spinner.
  - Review all modes: `filled`, `tonal`, `elevated`, and `outlined`.
  - Review custom color and custom label style examples.
  
  
Screenshot_1781525547
Simulator Screenshot - iPhone 17 - 2026-06-15 at 14 15 22

PR #5052, opened 2026-08-16
feat: modernize drawer

### LINKED ISSUE #4981: feat: modernize Drawer to latest Material Design Specs
- The goal is to modernize the Drawer according to the latest specs
- See https://m3.material.io/components/navigation-drawer/overview
- Reuse newly introduced tokens (shape, motion, typography, color etc)
- Extract component specific tokens for easier maintenance.
- Get inspired from TextInput/TextField, Switch, Checkbox, FAB components which have gone through modernization.

The Drawer component is not recommended anymore after MD3 Expressive. Another task will handle a modern replacement.

ziarno:
### Motivation

**Spec alignment**

- **Trailing padding.** `Drawer.Item` reserved 36dp trailing / 28dp leading. Both are now 28dp per spec, so items sit 8dp wider.
- **Active label weight.** Active `Drawer.Item` labels now use `labelLargeEmphasized` (700). Inactive labels stay at 500.
- **Section divider colour.** `Drawer.Section` drew its divider in a fixed `Palette.neutralVariant50` grey that ignored the theme and was wrong in dark mode. It now uses `theme.colors.outlineVariant`, like every other `Divider`.
- **Active indicator shape.** `Drawer.CollapsedItem` computed its radius as `itemSize / 2`; it now resolves the `full` shape token through the theme.
- **Motion.** The collapsed item's indicator animation was a hardcoded `150ms` linear ramp. It now uses the `short3` duration and `standard` easing tokens.
- **Disabled state.** Disabled destinations were visually identical to enabled ones. They are now dimmed via the shared `state.opacity.disabled` token.
- **Keyboard focus.** `Drawer.Item` gains a focus ring in `theme.colors.secondary`.

**Token extraction**

New `src/components/Drawer/tokens.ts` exports `DrawerItemTokens`, `DrawerSectionTokens` and `DrawerCollapsedItemTokens`.

**Supporting theme changes**

- `TypescaleKey` gains the 15 `*Emphasized` keys. The values already existed in the token set — only the types were missing.
- `state.focusIndicator` gains `innerOffset`, needed to inset the focus ring.

**A note on drawer width.** The example app deliberately sets no drawer width. `react-native-drawer-layout`'s default — `min(360dp, containerWidth − 56dp)` — is MD3's 360dp container plus MD2's 56dp edge gap; a flat 360dp leaves almost no scrim to tap and dismiss on narrow phones (also noticed by [flutter/flutter#123380](https://github.com/flutter/flutter/issues/123380)).

### Related issue

Closes #4981

### Test plan

- `yarn test` (Drawer unit tests + snapshots updated in this branch)
- Visual check of `Drawer.Item`, `Drawer.CollapsedItem` and `Drawer.Section` in the example app (expanded, collapsed/rail, active/inactive/disabled states) in both light and dark themes


src/components/Drawer/DrawerItem.tsx
```diff
@@ -101,15 +106,32 @@ const DrawerItem = ({
...
+    ? theme.colors[DrawerItemTokens.activeIndicatorColor]
+    : undefined;
   const contentColor = active
-    ? theme.colors.onSecondaryContainer
-    : theme.colors.onSurfaceVariant;
+    ? theme.colors[DrawerItemTokens.activeIconColor]
+    : theme.colors[DrawerItemTokens.inactiveIconColor];
+  const labelColor = active
+    ? theme.colors[DrawerItemTokens.activeLabelTextColor]
+    : theme.colors[DrawerItemTokens.inactiveLabelTextColor];
```
MikitasK: [nit] could we also apply focused icon & label colors here? M3 spec [uses](https://m3.material.io/components/navigation-drawer/specs) `onSurface` for inactive focused items instead of `onSurfaceVariant` . it also specifies focus state layer (currently only focus ring rendered)

MikitasK (approved):
LGTM 👍 
just one small suggestion:

src/components/Drawer/DrawerItem.tsx
ziarno: sure, done!

src/components/Drawer/DrawerItem.tsx
```diff
@@ -130,30 +162,51 @@ const DrawerItem = ({
...
-          <View style={styles.content}>
-            {icon ? (
-              <Icon source={icon} size={24} color={contentColor} />
-            ) : null}
-            <Text
-              variant="labelLarge"
-              selectable={false}
-              numberOfLines={1}
+        <View style={styles.inner}>
+          <View style={styles.wrapper} testID="drawer-item-content">
```
JKobrynski, line 166: Don't add default test ids - such a default will leave multiple test ids and may conflict with the user's own test ids" (#4998). All three exist only so the new tests can query them.

src/components/Drawer/DrawerItem.tsx
```diff
@@ -130,30 +162,51 @@ const DrawerItem = ({
...
+                {label}
+              </Text>
+            </View>
+
+            {right?.({ color: contentColor })}
+          </View>
+
+          {isFocused ? (
+            <View
+              testID="drawer-item-focus-ring"
```
JKobrynski, line 199: same here

src/components/Drawer/DrawerSection.tsx
```diff
@@ -72,39 +72,32 @@ const DrawerSection = ({
...
       )}
       {children}
       {showDivider && (
         <Divider
           horizontalInset
           bold
-          style={[styles.divider, styles.v3Divider]}
+          style={styles.divider}
           theme={theme}
+          testID="drawer-section-divider"
```
JKobrynski, line 100: Same here with the default test id

src/components/Drawer/DrawerCollapsedItem.tsx
```diff
@@ -125,22 +137,36 @@ const DrawerCollapsedItem = ({
...
+      duration:
+        theme.motion.duration[
+          DrawerCollapsedItemTokens.activeIndicatorDuration
+        ] * scale,
+      easing: Easing.bezier(
+        ...theme.motion.easing[DrawerCollapsedItemTokens.activeIndicatorEasing]
+      ),
       useNativeDriver: true,
     }).start();
   };
```
JKobrynski, line 149: Use Reanimated here. #5015 already rewrites this exact block as `transitionProperty` / `transitionDuration` driven by plain state — adding `Easing` to `Animated.timing` will only make that harder to land.

src/components/Drawer/DrawerItem.tsx
JKobrynski, line 160: `aria-selected` isn't allowed on `role="button"`, so the active destination isn't exposed as current to a screen reader. `BottomNavigationBar` uses `role={Platform.OS === 'ios' ? 'button' : 'tab'}` for exactly this - worth matching, here and in `DrawerCollapsedItem`.

src/components/Drawer/DrawerItem.tsx
```diff
@@ -101,15 +106,40 @@ const DrawerItem = ({
...
-  const borderRadius = theme.shapes.corner.extraLarge;
-  const font = theme.fonts.labelLarge;
+  const borderRadius = resolveCornerRadius(
+    theme,
+    DrawerItemTokens.indicatorShape
+  );
+  const { inset } = DrawerItemTokens.focusIndicator;
+  const opacity = disabled
+    ? DrawerItemTokens.stateOpacity.disabled
+    : DrawerItemTokens.stateOpacity.enabled;
```
JKobrynski, line 142: `opacity` on the `TouchableRipple` dims the active indicator to 38% along with the content. `Menu.Item` puts it on the content row instead - should this move to `styles.inner`? Same question for `DrawerCollapsedItem`'s `Pressable`.

JKobrynski (changes requested):

Issue #4990, opened 2026-06-09
feat: update non standard components to the latest Material Design specs


adrcotfas:
The goal is to go through non standard components like the following:
- Avatar
- Banner
- Data Table

and reuse recently added theme tokens (shape, motion, typography, colors etc). As these components are non standard, explore similar spec components for consistency.  

Latest Material Design specs: https://m3.material.io/

PR #5056, opened 2026-08-18
fix(banner): make hidden content inert and fix live region semantics

### LINKED ISSUE #5055: refactor(banner): make hidden content inert and fix live region semantics
Part of #4990

Clean up Banner accessibility so hidden content is actually hidden/inert, live regions behave consistently across platforms, and animation callbacks only fire when a real show/hide transition happens.

Banner doesn't exist in MD3 anymore, but it's still a supported Paper extension, so this is mostly about making its behaviour internally sane and accessible rather than matching the Material spec exactly.

**Problems**

- Hidden Banner content is still mounted and reachable. It's only translated off-screen and clipped — screen readers can still see it and action buttons stay tabbable. Verified in the browser a11y tree: with `visible={false}`, both actions still have `tabIndex: 0` and can receive focus.
- `role="alert"` + `aria-live="polite"` doesn't really make sense. `alert` normally implies assertive + atomic, while Chrome currently ends up with `alert atomic live="polite"`.
- The live region currently does nothing on Android. RN maps `aria-live` to `accessibilityLiveRegion` on `View`, but not on `Text`, so the prop is effectively ignored there. It only works on web today.
- iOS has no live-region equivalent, so the message isn't announced there either.
- `actions` is unlimited and actions always sit below the message. Material allows max two.
- If a focused action disappears after the actions array changes, focus gets dumped at the top of the page.
- `onShowAnimationFinished` / `onHideAnimationFinished` fire on mount and again when `theme.animation.scale` changes, even though no visibility transition happened. Existing tests already call this out as probably a bug.

**Focus Areas**

- Make hidden content properly inert: no pointer interaction, no a11y exposure, no focus.
- Put the live region on a `View` so Android actually supports it, and keep it scoped to the message so action labels don't cause the whole banner to re-announce.
- Make `role` and `aria-live` agree: polite by default, assertive when explicitly requested.
- Announce manually on iOS without double-announcing on Android/web.
- Limit actions to two and let them reflow depending on available width.
- Restore focus if the currently focused action disappears.
- Run show/hide animations and callbacks only for actual `visible` transitions.

**Proposed API**

One new prop, everything else is behavioural:

```jsx
<Banner visible urgent actions={[...]}>
  Your payment failed.
</Banner>
```

- `urgent?: boolean`
  - `false` (default): `role="status"` + `aria-live="polite"`
  - `true`: `role="alert"` + `aria-live="assertive"`; on iOS it interrupts instead of queueing

**Breaking changes**

- Banner children are unmounted once fully hidden, so local child state is lost after hiding.
- More than two actions are ignored, with a dev warning.

**Notes**

- #5016 retokenizes Banner (motion tokens + easing), but doesn't touch any of this accessibility behaviour.
- This should be checked separately with VoiceOver, TalkBack and the browser accessibility tree. They're different enough that testing only one isn't very meaningful.

lukemorawski:
### Motivation

`Banner` was only visually hidden but content stayed mounted, screen readers could still reach it, and action buttons remained tabbable. The live region also had conflicting semantics and simply didn't work on Android.

- **Hidden content was still reachable.** The banner was just translated off-screen and clipped - no `aria-hidden`, `pointerEvents` guard or `inert`. With `visible={false}`, both actions still had `tabIndex: 0` and accepted focus in the browser a11y tree.
- **`role="alert"` + `aria-live="polite"` didn't agree.** `alert` implies assertive + atomic, while Chrome ended up with `alert atomic live="polite"`.
- **The live region was dead on Android.** RN maps `aria-live` → `accessibilityLiveRegion` on `View`, not `Text`, so the prop was ignored by the native text node. In practice it only worked on web.
- **iOS has no live regions**, so nothing announced the message there either.
- **`actions` was unbounded**, and animation callbacks fired on mount and on `theme.animation.scale` changes even when visibility didn't change. Two existing tests already called this out as probably a bug.

### Related issue

Fixes #5055  
Part of #4990

### Changes

**Inertness.** Content becomes `aria-hidden` + `pointerEvents="none"` + `inert` on web as soon as hiding starts, then unmounts after the exit animation - basically the same lifecycle `Snackbar` already uses. There's still one inert measuring pass when mounting hidden so the spacer gets the right height. Layout stays the same.

**Live region.** Moved from `Text` to a `View`, which makes it actually work on Android. It's also scoped to the message only, so action labels don't re-announce the whole banner. `role` and `aria-live` now match.

**New `urgent` prop.**

```jsx
<Banner visible urgent actions={[...]}>
  Your payment failed.
</Banner>
```

- `false` by default: `role="status"` + `aria-live="polite"`
- `true`: `role="alert"` + `aria-live="assertive"`

On iOS the message is announced explicitly with `announceForAccessibilityWithOptions({ queue: !urgent })`: queued for normal banners, interrupting for urgent ones. This is iOS-only to avoid double announcements elsewhere. Interpolated children like `<Banner>Hello {name}</Banner>` are flattened before announcing.

**Actions.** Limited to 2, with a development warning for extras. They can now sit inline with the message when there's enough room instead of always dropping below it.

**Focus.** If a focused action disappears, focus moves to the nearest remaining action, or the message region if there are none. During hiding the content is inert, so focus is simply released. Restoring it to whatever opened the banner needs consumer-owned API/state and is out of scope here.

**Callbacks.** `onShowAnimationFinished` / `onHideAnimationFinished` now only run after actual `visible` transitions - not on mount, animation-scale changes, or interrupted animations.

### Breaking changes

- Banner children unmount once hidden, so local child state is lost.
- More than 2 actions are ignored, with a dev warning.
- Actions may render inline instead of always below the message.

### Test plan

- [x] `yarn typecheck`
- [x] `yarn lint`
- [x] `yarn test` - 55 suites, 760 tests. Banner coverage went from 13 to 41 tests; the two "probably a bug" tests were inverted rather than removed.
- [x] Chrome browser accessibility tree (`expo start --web`)
- [x] Mobile Safari (iOS 18.3 simulator) + Chrome (Android 16 emulator)
- [x] Android native + **TalkBack**
- [x] iOS native + **VoiceOver**

**Hidden state, browser a11y tree.** Before, both actions were still focusable:

```text
{"totalTabbablesOnPage":60,
 "bannerActionsStillTabbable":[
  {"label":"Set custom theme","tabIndex":0,"receivedFocus":true,"insideInert":false,"insideAriaHidden":false},
  {"label":"Fix it","tabIndex":0,"receivedFocus":true,"insideInert":false,"insideAriaHidden":false}]}
```

Afterwards the message and actions disappear from the tree completely. The page tab count drops by exactly those two buttons (from 31 to 29), then returns when the banner is shown again.

**Visible state.** `alert atomic live="polite"` becomes `status atomic live="polite"`. The live region contains only the message; buttons are siblings.

**Android / TalkBack.** The native a11y tree contains `banner-content`, message and actions while visible, and none of them while hidden.

**iOS / VoiceOver.** Full show > hide > show cycle gives message + actions > `[]` > message + actions again. The `urgent` announcement was also checked.

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -7,7 +7,7 @@ import {
   StyleSheet,
   View,
 } from 'react-native';
-import type { StyleProp, ViewStyle } from 'react-native';
+import type { StyleProp, ViewProps, ViewStyle } from 'react-native';
 import type { LayoutChangeEvent } from 'react-native';
 
 import useLatestCallback from 'use-latest-callback';
@@ -23,6 +23,27 @@ import type { $Omit, $RemoveChildren, Theme, ThemeProp } from '../types';
 const DEFAULT_MAX_WIDTH = 960;
 // banners carry at most two actions per the material spec
 const MAX_ACTIONS = 2;
+// a tradeoff, not a safe number: too short and the set and clear can batch into
+// one a11y update, too long and a talkback swipe hears the message twice
+const ANNOUNCE_CLEAR_DELAY = 3000;
+
+// only android and web have a real live region. everywhere else announces by hand
+const hasLiveRegion = () => Platform.OS === 'android' || Platform.OS === 'web';
+
+// a nested <Text> would otherwise be dropped from the announcement
+const extractText = (node: React.ReactNode): string =>
+  React.Children.toArray(node)
+    .map((child) => {
+      if (typeof child === 'string' || typeof child === 'number') {
+        return String(child);
+      }
+      if (React.isValidElement(child)) {
+        const { children } = child.props as { children?: React.ReactNode };
+        return extractText(children);
+      }
+      return '';
+    })
+    .join('');
 
 export type Props = $Omit<$RemoveChildren<typeof Surface>, 'mode'> & {
   /**
@@ -37,6 +58,12 @@ export type Props = $Omit<$RemoveChildren<typeof Surface>, 'mode'> & {
    * Icon to display for the `Banner`. Can be an image.
    */
   icon?: IconSource;
+  /**
+   * Accessibility label for the icon. Leave it out when the icon is decorative
+   * and the message already says everything - the icon is then hidden from
+   * screen readers instead of being read out as an unlabelled image.
+   */
+  iconAccessibilityLabel?: string;
   /**
    * Action items to shown in the banner.
    * An action item should contain the following properties:
@@ -139,6 +166,7 @@ export type Props = $Omit<$RemoveChildren<typeof Surface>, 'mode'> & {
 const Banner = ({
   visible,
   icon,
+  iconAccessibilityLabel,
   children,
   actions = [],
   contentStyle,
@@ -233,22 +261,39 @@ const Banner = ({
     }
   }, [actions.length]);
 
-  const liveRegion = urgent ? 'assertive' : 'polite';
-  const message = React.Children.toArray(children)
-    .filter((child) => typeof child === 'string' || typeof child === 'number')
-    .join('');
+  const region = urgent
+    ? ({ role: 'alert', live: 'assertive' } as const)
+    : ({ role: 'status', live: 'polite' } as const);
+  const message = extractText(children);
 
-  // aria-live only reaches a real live region on android and web. ios has no
-  // equivalent, so announce there by hand whenever the message becomes
-  // available. doing it everywhere would double up with the live region
   React.useEffect(() => {
-    if (Platform.OS !== 'ios' || !visible || !message) {
+    if (hasLiveRegion() || !visible || !message) {
       return;
     }
 
     AccessibilityInfo.announceForAccessibilityWithOptions(message, {
       queue: !urgent,
     });
+  }, [visible, message, urgent]);
+
+  // a region only fires when its text changes while it is already in the tree.
+  // the banner unmounts when hidden, so the message can never be that change
+  const [announcement, setAnnouncement] = React.useState('');
+  React.useEffect(() => {
+    if (!hasLiveRegion() || !visible || !message) {
+      setAnnouncement('');
+      return;
+    }
+
+    // re-setting the same string is not a render, so empty it first or an
+    // unchanged message announces nothing
+    setAnnouncement('');
+    const fill = setTimeout(() => setAnnouncement(message), 0);
+    const clear = setTimeout(() => setAnnouncement(''), ANNOUNCE_CLEAR_DELAY);
+    return () => {
+      clearTimeout(fill);
+      clearTimeout(clear);
+    };
   }, [visible, message, urgent]);
 
   // one stable ref per action slot; the cap is what bounds the array
@@ -265,13 +310,16 @@ const Banner = ({
       return;
     }
 
+    // rnw's findNodeHandle throws, and the ref is already the dom node there
+    if (Platform.OS === 'web') {
+      (node as unknown as { focus?: () => void }).focus?.();
+      return;
+    }
+
     const handle = findNodeHandle(node);
     if (handle !== null) {
       AccessibilityInfo.setAccessibilityFocus(handle);
     }
-
-    // rnw resolves the ref to the dom node, which takes focus directly
-    (node as unknown as { focus?: () => void }).focus?.();
   };
 
   // a removed action would otherwise strand focus at the top of the document
@@ -322,6 +370,10 @@ const Banner = ({
   // tree and the tab order. native ignores the unknown prop
   const inertProps = visible ? null : ({ inert: true } as object);
 
+  // annotated, not cast: the literal -1 widens to number on its own
+  const messageFocusProps: Pick<ViewProps, 'tabIndex' | 'accessible'> =
+    Platform.OS === 'web' ? { tabIndex: -1 } : { accessible: true };
+
   return (
     <Surface
       {...rest}
@@ -358,20 +410,23 @@ const Banner = ({
               {/* icon and message travel together as one flex item, so only
                   the actions can be wrapped onto the next line */}
               <View style={styles.body}>
+                {/* rnw gives the icon role="img" with no name */}
                 {icon ? (
-                  <View style={styles.icon}>
+                  <View
+                    testID={`${testID ?? 'banner'}-icon`}
+                    style={styles.icon}
+                    aria-hidden={!iconAccessibilityLabel}
+                    accessible={iconAccessibilityLabel ? true : undefined}
+                    aria-label={iconAccessibilityLabel}
+                  >
                     <Icon source={icon} size={40} />
                   </View>
                 ) : null}
-                {/* the region is scoped to the message: status/alert imply
-                    aria-atomic, so keeping the actions out of it stops their
-                    labels from re-announcing the whole banner */}
                 <View
                   ref={messageRef}
                   testID={`${testID ?? 'banner'}-message`}
                   style={styles.message}
-                  role={urgent ? 'alert' : 'status'}
-                  aria-live={visible ? liveRegion : 'off'}
+                  {...messageFocusProps}
                 >
                   <Text
                     variant="bodyMedium"
@@ -415,6 +470,17 @@ const Banner = ({
             </View>
           </Animated.View>
         )}
+        {/* last, so a screen reader reaches the real message first */}
+        {hasLiveRegion() ? (
+          <View
+            testID={`${testID ?? 'banner'}-announcer`}
+            style={styles.announcer}
+            role={region.role}
+            aria-live={region.live}
+          >
+            <Text>{announcement}</Text>
+          </View>
+        ) : null}
       </View>
     </Surface>
   );
@@ -468,6 +534,14 @@ const styles = StyleSheet.create({
   },
   transparent: {
     opacity: 0,
+  },
+  announcer: {
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    width: 1,
+    height: 1,
+    overflow: 'hidden',
   },
 });
 
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -19,6 +19,7 @@ import Surface from './Surface';
 import Text from './Typography/Text';
 import { useInternalTheme } from '../core/theming';
 import type { $Omit, $RemoveChildren, Theme, ThemeProp } from '../types';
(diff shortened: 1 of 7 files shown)


src/components/Banner.tsx
```diff
@@ -169,14 +209,96 @@ const Banner = ({
         duration: 200 * scale,
         toValue: 0,
         useNativeDriver: false,
-      }).start(hideCallback);
+      }).start((result) => {
+        if (result.finished) {
+          setExited(true);
+        }
+        hideCallback(result);
+      });
```
MikitasK, line 251: maybe we should invoke `hideCallback` only when `result.finished === true` ?
```suggestion
      }).start((result) => {
        if (result.finished) {
          setExited(true);
          hideCallback(result);
        }
      });
```

src/components/Banner.tsx
```diff
@@ -192,69 +314,103 @@ const Banner = ({
...
+                {/* the region is scoped to the message: status/alert imply
+                    aria-atomic, so keeping the actions out of it stops their
+                    labels from re-announcing the whole banner */}
+                <View
+                  ref={messageRef}
+                  testID={`${testID ?? 'banner'}-message`}
+                  style={styles.message}
+                  role={urgent ? 'alert' : 'status'}
+                  aria-live={visible ? liveRegion : 'off'}
+                >
```
MikitasK, line 456: could we make message focusable on web? 
when last action is removed, we call `.focus()` on this `View`, but `role="status"` doesn't make rendered `<div>` focusable

maybe we can add `tabIndex={-1}` on web:

```suggestion
                <View
                  ref={messageRef}
                  testID={`${testID ?? 'banner'}-message`}
                  style={styles.message}
                  role={urgent ? 'alert' : 'status'}
                  aria-live={visible ? liveRegion : 'off'}
                  {...(Platform.OS === 'web' ? { tabIndex: -1 } : { accessible: true })}
                >
```

also it'd be great to verify this behavior in `Banner.text.tsx`

src/components/Banner.tsx
```diff
@@ -169,14 +209,96 @@ const Banner = ({
...
+      console.warn(
+        `Banner supports a maximum of ${MAX_ACTIONS} actions, received ${actions.length}. The extra actions are ignored.`
+      );
+    }
+  }, [actions.length]);
+
+  const liveRegion = urgent ? 'assertive' : 'polite';
+  const message = React.Children.toArray(children)
+    .filter((child) => typeof child === 'string' || typeof child === 'number')
+    .join('');
```
MikitasK: what about nested text elements? current code only collects direct string & number children, so messages like that would announce only "Payment " on iOS:

```
<Banner visible>
  Payment <Text>failed</Text>
</Banner>
```

maybe we can recursively collect text from nested children? or provide explicit announcement label?

src/components/Banner.tsx
```diff
@@ -192,69 +314,103 @@ const Banner = ({
...
+                  {visibleActions.map(({ label, ...others }, i) => (
+                    <Button
+                      key={i}
+                      compact
+                      mode="text"
+                      style={styles.button}
+                      textColor={colors.primary}
+                      theme={theme}
+                      {...others}
+                      touchableRef={actionRefs.current[i]}
```
MikitasK: could we preserve `touchableRef` passed through an action? internal `touchableRef` currently overrides the consumer’s ref. maybe we can merge both refs instead:

```suggestion
                      touchableRef={mergeRefs(actionRefs.current[i], touchableRef)}
```

MikitasK (commented):
looks pretty solid 👍 
just a few points to address before merge:

src/components/Banner.tsx
```diff
@@ -169,14 +209,96 @@ const Banner = ({
...
+    AccessibilityInfo.announceForAccessibilityWithOptions(message, {
+      queue: !urgent,
+    });
+  }, [visible, message, urgent]);
+
+  // one stable ref per action slot; the cap is what bounds the array
+  const actionRefs = React.useRef<Array<React.RefObject<View>>>([]);
+  for (let i = 0; i < MAX_ACTIONS; i++) {
+    // Button types touchableRef as non-nullable, but a ref always starts null
+    actionRefs.current[i] ??= React.createRef<View>() as React.RefObject<View>;
```
JKobrynski: Let's try not to use `as` casts if possible

src/components/Banner.tsx
```diff
@@ -169,14 +209,96 @@ const Banner = ({
...
+      return;
+    }
+
+    const handle = findNodeHandle(node);
+    if (handle !== null) {
+      AccessibilityInfo.setAccessibilityFocus(handle);
+    }
+
+    // rnw resolves the ref to the dom node, which takes focus directly
+    (node as unknown as { focus?: () => void }).focus?.();
```
JKobrynski: Same here

src/components/Banner.tsx
```diff
@@ -192,69 +314,103 @@ const Banner = ({
     Animated.add(position, -1),
     layout.height
   );
+  // rnw forwards `inert` to the dom, which drops the subtree from the a11y
+  // tree and the tab order. native ignores the unknown prop
+  const inertProps = visible ? null : ({ inert: true } as object);
```
JKobrynski: Same here

src/components/Banner.tsx
```diff
@@ -192,69 +370,125 @@ const Banner = ({
...
+              <View style={[styles.body, stacked ? null : styles.bodyInline]}>
+                {/* rnw gives the icon role="img" with no name */}
+                {icon ? (
+                  <View
+                    testID={`${testID ?? 'banner'}-icon`}
+                    style={styles.icon}
+                    aria-hidden={!iconAccessibilityLabel}
+                    accessible={iconAccessibilityLabel ? true : undefined}
+                    aria-label={iconAccessibilityLabel}
+                  >
```
MikitasK, line 447: could we add `role="img"` when `iconAccessibilityLabel` is provided?

this `View` renders as a generic `<div>` on web  
react native web doesn't implement `accessible` prop ([source](https://necolas.github.io/react-native-web/docs/accessibility/#keyboard-focus)) and a generic element cannot have an accessible name. that's why wrapper’s `aria-label` isn't guaranteed to label nested image (([source](https://www.w3.org/TR/wai-aria-1.2/#namefromprohibited))

```suggestion
                  <View
                    testID={`${testID ?? 'banner'}-icon`}
                    style={styles.icon}
                    aria-hidden={!iconAccessibilityLabel}
                    accessible={iconAccessibilityLabel ? true : undefined}
                    aria-label={iconAccessibilityLabel}
                    role={iconAccessibilityLabel ? 'img' : undefined}
                  >
```

src/components/Banner.tsx
```diff
@@ -192,69 +370,125 @@ const Banner = ({
...
-                {
-                  color: colors.onSurface,
-                },
-              ]}
-              aria-live={visible ? 'polite' : 'off'}
-              role="alert"
-              maxFontSizeMultiplier={maxFontSizeMultiplier}
+        {exited ? null : (
+          <Animated.View
+            testID={`${testID ?? 'banner'}-content`}
```
JKobrynski: Only derive these when `testID` is set - the snapshots show `banner-content`, `banner-row` and `banner-message` on banners that were given no `testID`. Default test ids conflict with the user's own (#4998).

src/components/Banner.tsx
```diff
@@ -11,9 +18,36 @@ import type { IconSource } from './Icon';
...
+const COMPACT_BREAKPOINT = 600;
+// a tradeoff, not a safe number: too short and the set and clear can batch into
+// one a11y update, too long and a talkback swipe hears the message twice
+const ANNOUNCE_CLEAR_DELAY = 3000;
+
+// only android and web have a real live region. everywhere else announces by hand
+const hasLiveRegion = () => Platform.OS === 'android' || Platform.OS === 'web';
+
+// a nested <Text> would otherwise be dropped from the announcement
+const extractText = (node: React.ReactNode): string =>
```
JKobrynski, line 38: `child.props.children` breaks on any child that doesn't render its own children, so `<Banner><Message /></Banner>` announces nothing on iOS. `React.Children` is the composition problem v6 is meant to remove (#4954) - what's the alternative here?

src/components/Banner.tsx
```diff
@@ -155,28 +210,151 @@ const Banner = ({
...
+    if (handle !== null) {
+      AccessibilityInfo.setAccessibilityFocus(handle);
+    }
+  };
+
+  // a removed action would otherwise strand focus at the top of the document
+  React.useEffect(() => {
+    const focused = focusedAction.current;
+
+    if (focused === null || focused < actionCount) {
```
JKobrynski: `focusedAction.current` isn't cleared on hide - `focused < actionCount` returns first. Hide, then show with `actions={[]}`, and focus jumps to the message. The test that covers this keeps one action, so it short-circuits too.

src/components/Banner.tsx
```diff
@@ -192,69 +370,125 @@ const Banner = ({
...
+                  {visibleActions.map(({ label, ...others }, i) => (
+                    <Button
+                      key={i}
+                      compact
+                      mode="text"
+                      style={styles.button}
+                      textColor={colors.primary}
+                      theme={theme}
+                      {...others}
+                      touchableRef={mergeRefs(
```
JKobrynski: `mergeRefs` returns a new callback every render, so React detaches and reattaches each action ref on every commit — a consumer's `touchableRef` callback gets `null` then the node each time. Maybe try to memoize it.

JKobrynski (changes requested):

src/components/Banner.tsx
lukemorawski: yep, and it looks like it's not just iOS. Both the android and web live regions will go empty too. Checked it: zero announceForAccessibility calls on iOS, empty announcer on Android, visible text renders fine. dead silent on every platform. Thanks for bringing that up!

i see it liek that:
- add `messageAccessibilityLabel`, matching `iconAccessibilityLabel` in this PR
- keep extraction as the default so plain text keeps working
- warn in dev when extraction is empty but children are not

React.Children stays as best effort with an opt out, not as main mechanism. but i'm happy to drop extraction entirely or park this for v6 instead. any preference?

src/components/Banner.tsx
MikitasK: thanks for addressing 🙏

MikitasK (approved):
LGTM 👌

PR #5016, opened 2026-06-29
refactor: modernize non standard components to the latest MD3 spec

MrMuzyk:
<!-- Please provide enough information so that others can review your pull request. -->
<!-- Keep pull requests small and focused on a single change. -->

### Motivation

<!-- Explain the **motivation** for making this change. What existing problem does the pull request solve? -->

Modernizes the remaining non-standard components — Avatar, Banner, and DataTable — to the latest MD3 spec by adopting the recently added theme tokens (shape, motion, typography, colors).

#### Changes

Avatar (AvatarIcon, AvatarText, AvatarImage, new utils.ts)                                                                                                               

- Colors: default primary + getContrastingColor(white) → MD3 container pair primaryContainer / onPrimaryContainer. Custom backgrounds keep the getContrastingColor       
- luminance fallback (arbitrary per-user colors have no on- role).                                                                                                         
- Shape: borderRadius: size / 2 → cornerFull (9999) from src/theme/tokens/sys/shape.                                                                                     
- Typography: AvatarText spreads theme.fonts.titleMedium (family/weight/letterSpacing) before its proportional fontSize/lineHeight.                                      
- Structure: new src/components/Avatar/utils.ts (DEFAULT_SIZE = 64, ICON_SIZE_RATIO = 0.6, resolveAvatarColors) shared by the three files; removes the duplicated        
- defaultSize constant.  

Banner (Banner.tsx)                                                                                                                                                      
- Motion: durations 250 * scale / 200 * scale → theme.motion.duration.medium1 / short4 (values unchanged), plus easing: Easing.bezier(...theme.motion.easing.standard).  Dropped the legacy * scale multiplier to match how Switch/Checkbox adopted motion tokens.                                                                                
- Icon size: size={40} → const ICON_SIZE = 40.                                                                                                                           
                                                                                                                                                                           
DataTable (DataTableTitle, DataTableCell, DataTablePagination)     

- Typography: header → variant="labelMedium" (keeps 12px; lineHeight: 24 retained to stay vertically aligned with the 24dp sort icon); cell → variant="bodyMedium"; pagination labels → variant="bodySmall".
- Motion: sort-arrow spin duration: 150 → theme.motion.duration.short3 (unchanged value) + easing.standard.                                                              
                                                                                                                                                                           

#### Visual / behavioral changes (no public API changed)                                                                                                                      
                                                                                                                                                                           
  - Avatar's default tone flips from solid primary + white content to the primaryContainer / onPrimaryContainer pair.                                                      
[...]

PR #5058, opened 2026-08-21
fix: define avatar identity and fallback semantics

oleksandrzavarzin-callstack:
### Motivation

Defines Avatar identity, contrast, and fallback semantics as a **Paper extension** (not a standalone MD3 family). Reuses shared theme tokens where they match, and fixes accessibility, initials, and image-error handling rather than chasing a spec that does not exist for Avatar.

Builds on the MD3 token work in [#5016](https://github.com/callstack/react-native-paper/pull/5016) (container pair, `cornerFull`, `titleMedium`) and finishes the contrast / a11y / fallback gaps that PR left open.

#### Changes

**Avatar (AvatarIcon, AvatarText, AvatarImage)**

- **Colors:** Custom **string** backgrounds still use the luminance heuristic. Opaque / dynamic values (`PlatformColor`, `DynamicColorIOS`) go through `contentColorFor` — a theme-role token pairs with its `on-` color; anything else falls back to `onSurface`. Explicit `color` still wins. `getContrastingColor` is string-only, so it no longer guesses white for unresolved dynamic colors.
- **Accessibility (`Avatar.Image`):** a11y props land on the rendered `Image` (or function source), not the wrapper `View`. An unlabeled / non-actionable avatar stays unfocusable (`accessible={false}`). When `fallback` is showing, a11y props move to the host so the replacement content is announced.
- **Accessibility (`Avatar.Text`):** initials are hidden from assistive tech (`accessibilityElementsHidden` / `importantForAccessibility="no-hide-descendants"`) so a parent label is not double-read.
- **Initials:** `takeGraphemes(label, 2)` — combining marks, emoji (ZWJ / skin tones), and flag sequences stay intact; long names are bounded to two graphemes.
- **Fallback API:** `fallback={({ size }) => …}` when the image fails. Function `source` now receives `{ size, style, onError, …a11y }` so custom images fill the circle and can trigger fallback. Source identity is keyed by URI (or `'function'`) so inline renderers and equal-URI objects do not reset error state; a real URI change retries.
- **Docs:** `fallback` + host `size`/`style` guidance

#### Visual / behavioral changes

- Custom `PlatformColor` / dynamic backgrounds no longer get white content by default.
- `Avatar.Text` shows at most two graphemes (emoji-safe); previously sliced by UTF-16 code unit.
- Image a11y is announced on the image, not a duplicate wrapper focus target.
- Failed images can swap to `fallback` instead of staying empty/broken.

#### Public API

- **New:** `Avatar.Image` `fallback?: (props: { size: number }) => React.ReactNode`
- **Extended:** function `source` props now include `style` and `onError` (plus a11y). Existing `{ size }` usage still works.

### Related issue

*(new — to be created)* · overlaps [#5016](https://github.com/callstack/react-native-paper/pull/5016) / [#4990](https://github.com/callstack/react-native-paper/issues/4990)

### Test plan

- [ ] `yarn typescript`
- [ ] `yarn lint`
- [ ] `yarn test`
- [ ] Visual verification (`yarn example android`):
  - **Avatar.Text** — default tonal pair; custom hex still contrasts; `PlatformColor` stays legible (not forced white); long names / emoji / combining marks show as two graphemes
  - **Avatar.Icon** — default + custom background/color still readable
  - **Avatar.Image** — a11y label is on the image; invalid URI shows `fallback`; function `source` fills the circle via host `style`; new URI retries, same URI keeps fallback

---

**Screenshots**:

1. **Custom / dynamic background** — platformColor, hex, palette color; content color is readable
Screenshot 2026-08-21 at 12 00 00

2. **Grapheme initials** — long name (`Jane Doe`), combining mark (`éva`), family emoji; content truncated to 2 graphemes
Screenshot 2026-08-21 at 11 57 34

3. **Image fallback** — no image no fallback, no image with fallback (Avatar.text), with original image; broken URI → custom component
Screenshot 2026-08-21 at 12 05 30


src/components/Avatar/AvatarText.tsx
```diff
@@ -97,8 +104,10 @@ const AvatarText = ({
         ]}
         numberOfLines={1}
         maxFontSizeMultiplier={maxFontSizeMultiplier}
+        accessibilityElementsHidden
+        importantForAccessibility="no-hide-descendants"
```
JKobrynski:
```
        accessibilityElementsHidden
        importantForAccessibility="no-hide-descendants"
```

This hides the initials from assistive tech unconditionally, so `<Avatar.Text label="JD" />` with no other props goes from announcing "JD" to announcing nothing. On iOS RN `Text` is an accessibility element unless `accessible={false}` (`Libraries/Text/Text.js:145-148`), and `accessibilityElementsHidden` removes it; on Android `no-hide-descendants` does the same. Nothing takes over, because `View.accessible` defaults to `false` in Fabric (`ReactCommon/.../AccessibilityProps.h:31`), so a host `aria-label` alone isn't announced either.

Put the name on the host and hide the glyphs, and use the aria form — `aria-hidden` maps to exactly these two props (`Text.js:138-141`):

```diff
       ]}
+      accessible
+      aria-label={label}
       {...rest}
     >
```
```diff
         numberOfLines={1}
         maxFontSizeMultiplier={maxFontSizeMultiplier}
-        accessibilityElementsHidden
-        importantForAccessibility="no-hide-descendants"
+        aria-hidden
```

`{...rest}` still lets a consumer override with `aria-label` (RN prefers `aria-label` over `accessibilityLabel` when both are set).

src/components/Avatar/AvatarImage.tsx
```diff
@@ -87,41 +127,88 @@ const AvatarImage = ({
...
+  const { accessibilityProps, rest: viewProps } = splitAccessibilityProps(rest);
+  const imageStyle = {
+    width: size,
+    height: size,
+    borderRadius: cornerFull,
+  };
+  const imageA11y =
+    Object.keys(accessibilityProps).length > 0
+      ? accessibilityProps
+      : { accessible: false as const };
```
JKobrynski:
```
  const imageA11y =
    Object.keys(accessibilityProps).length > 0
      ? accessibilityProps
      : { accessible: false as const };
```

Moving the label to the `Image` doesn't make it announceable on iOS. RN's `Image` sets `accessible` only from `alt` or an explicit `accessible` — `Image.ios.js:170-171` is `props.alt !== undefined ? true : props.accessible`, `Image.android.js:291-295` the same — so `<Avatar.Image aria-label="Jane" />` yields an `Image` with a label and `isAccessibilityElement = NO`, while the wrapper is now explicitly `accessible={false}`. Android still works (contentDescription makes the view focusable), iOS is silent. Same gap at `:176`: the fallback branch spreads `accessibilityProps` onto the wrapper `View` without `accessible`.

```diff
-      {...(showImage
-        ? { accessible: false, importantForAccessibility: 'no' as const }
-        : accessibilityProps)}
+      {...(showImage
+        ? { accessible: false, importantForAccessibility: 'no' as const }
+        : { accessible: true, ...accessibilityProps })}
```

and give the image `accessible` whenever a label came in. `alt` is the cleaner route — it sets both halves on both platforms — and would let `Avatar.Image` declare one documented prop instead of forwarding 40.

src/utils/splitAccessibilityProps.ts
```diff
@@ -0,0 +1,82 @@
...
+  accessibilityLargeContentTitle: true,
+  accessibilityRespondsToUserInteraction: true,
+} satisfies Record<keyof AccessibilityProps, true>;
+
+/**
+ * Keys of the `AccessibilityProps` interface.
+ */
+const ACCESSIBILITY_PROP_KEYS = Object.keys(
+  ACCESSIBILITY_PROP_PRESENCE
+) as (keyof AccessibilityProps)[];
```
JKobrynski:
```
    return (source as { uri: unknown }).uri;
```

Four type assertions. `src/` on `main` has none outside `as const`, and @satya164  has asked for them removed verbatim in #4998 and #4994. This one is unnecessary — TS narrows `'uri' in source` on an `object` since 4.9:

```diff
-    return (source as { uri: unknown }).uri;
+    return source.uri;
```

src/utils/splitAccessibilityProps.ts
```diff
@@ -0,0 +1,82 @@
...
+  const rest = { ...props };
+
+  for (const key of ACCESSIBILITY_PROP_KEYS) {
+    if (!Object.hasOwn(rest, key)) {
+      continue;
+    }
+
+    const value = rest[key];
+    if (value !== undefined) {
+      (accessibilityProps as Record<keyof AccessibilityProps, unknown>)[key] =
```
JKobrynski: Same here

src/utils/splitAccessibilityProps.ts
```diff
@@ -0,0 +1,82 @@
...
+        value;
+    }
+
+    delete rest[key];
+  }
+
+  return {
+    accessibilityProps,
+    rest: rest as Omit<T, keyof AccessibilityProps>,
```
JKobrynski: Same here

JKobrynski (changes requested):

JKobrynski:
@oleksandrzavarzin-callstack let me know when this is ready for another review!

oleksandrzavarzin-callstack:
> @oleksandrzavarzin-callstack let me know when this is ready for another review!

@JKobrynski It's ready for the another review. Let me know if you have any other suggestions

src/components/Avatar/AvatarImage.tsx
```diff
@@ -87,41 +145,89 @@ const AvatarImage = ({
...
+  const theme = useInternalTheme(themeOverrides);
+  const { backgroundColor } = StyleSheet.flatten(style) || {};
+  const { background } = resolveAvatarColors({ theme, backgroundColor });
+  const imageStyle = {
+    width: size,
+    height: size,
+    borderRadius: cornerFull,
+  };
+  const imageA11y =
+    alt === undefined ? { accessible: false as const } : { alt };
```
JKobrynski: Pass `aria-label={alt}` alongside `alt`. react-native-web's `Image` fills the inner `` from `aria-label` only; `alt` falls through to the wrapper `View`, whose forwarded-prop allowlist drops it. Web avatars render ``, which marks them decorative. [react-native-web `exports/Image/index.js`](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Image/index.js)

src/components/Avatar/AvatarImage.tsx
```diff
@@ -87,41 +145,89 @@ const AvatarImage = ({
...
+  }
+
+  const handleError: ImageProps['onError'] = (event) => {
+    setHasError(true);
+    onError?.(event);
+  };
+
+  const showImage = !(hasError && fallback !== undefined);
+
+  const hostA11y = showImage
```
JKobrynski, line 192: Fold `aria-label` / `accessibilityLabel` into `alt` here. They currently land on a host that is `accessible={false}` `importantForAccessibility="no"`, so nothing announces on native - while `Avatar.Text` and `Avatar.Icon` in this PR honour both. Your own icon test says why: "`View` is not an accessibility element by default".

oleksandrzavarzin-callstack:
@JKobrynski Thanks for the review! I've pushed the fix to your comments. 
Please review again.

src/components/Avatar/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+ *   onPrimaryContainer).
+ * - Explicit `color` wins.
+ * - String custom backgrounds keep the luminance heuristic (arbitrary
+ *   per-user colors have no on- role).
+ * - Opaque custom values (`PlatformColor` / `DynamicColorIOS`) go through
+ *   `contentColorFor`: a theme-role token pairs with its on-color; anything
+ *   else falls back to `onSurface`. Pass `color` when that fallback is not
+ *   appropriate.
+ */
+export const resolveAvatarColors = ({
```
JKobrynski, line 29: This file duplicates #5016's Avatar work - `DEFAULT_SIZE`, `ICON_SIZE_RATIO` and this function, whose `contentColorFor` branch is the only addition. The two conflict in ten places across four files. Which one lands first?

oleksandrzavarzin-callstack: @JKobrynski , as I mentioned in the PR description, this PR continues that #5016's work. That's why I’ve cherry-picked the connected changes. My changes are purely additive, so they are pretty interchangeable:

* If the original PR merges, we can just apply all my changes to the conflicts.
* If my PR merges, we can just drop the avatar changes from the original one.

If you have any suggestions - please let me know

JKobrynski:
@oleksandrzavarzin-callstack lint is failing

oleksandrzavarzin-callstack:
> @oleksandrzavarzin-callstack lint is failing

@JKobrynski The lint should be fixed now

JKobrynski (approved):
LGTM! @satya164

PR #5061, opened 2026-08-21
fix(example): render banner below the app bar on native

### LINKED ISSUE #5060: `Banner` example renders behind the app bar on iOS and Android
### Current behaviour

On the Banner example screen the banner is painted under the status bar and the app bar, so all you see is an empty gap between the header and the photo grid. The banner is not missing, it is just covered. Web is fine.

Minor, but it means you cannot eyeball the `Banner` component on either simulator, which is how a layout regression in #5056 went unnoticed for a whole review.

### Expected behaviour

The banner renders below the app bar, while still overlaying the scrollable content like the example intends.

### How to reproduce?

Needs a native run, it does not happen on web.

```sh
cd example
npx expo prebuild -p android # or -p ios
npx expo run:android # or run:ios
```

Then open the **Banner** example from the list.

### Preview

Image

Android, banner example screen. The gap between the app bar and the grid is where the banner should be.

Pre-existing on `main`, not caused by any in-flight Banner work.

`example/src/Examples/BannerExample.tsx` styles the banner as `position: 'absolute', top: 0`. On native that is the top of the screen rather than the top of the content, so it ends up under the custom header that `RootNavigator.tsx` supplies. Bumping `top` by hand puts the banner and both buttons back where they should be, so the component itself is fine.

Probably just needs offsetting by `useHeaderHeight()` from `@react-navigation/elements`, which the example already depends on. `BannerExample` is the only example screen using absolute + `top: 0`, so it looks like a one file fix.

### Your Environment

| software           | version                        |
| ------------------ | ------------------------------ |
| ios                | 18.3 (iPhone 16 Pro simulator) |
| android            | 16 (Pixel 10 Pro XL emulator)  |
| react-native       | 0.85.3                         |
| react-native-paper | 6.0.0-alpha.0 (`main`)         |
| node               | 24.13.0                        |
| npm or yarn        | yarn 4.9.1                     |
| expo sdk           | 56.0.0                         |

lukemorawski:
### Motivation

The Banner example is painted behind the app bar on iOS and Android, so the screen
just shows an empty gap where the banner should be. Web is fine.

`BannerExample` returns a fragment and styles the banner `position: 'absolute',
top: 0`. On web the navigator wraps the screen in a `flex: 1` container below the
header, so that means the top of the content. On native it does not, so it means
the top of the screen, and the header's `zIndex: 1` paints over it.

Wrapping the screen in a `flex: 1` `View` gives native the same containing block, so
`top: 0` lands below the app bar on both.

### Related issue

Fixes https://github.com/callstack/react-native-paper/issues/5060

### Test plan

- [x] `yarn lint`, `yarn typecheck`, `yarn test` (733 passed) all clean
- [x] Android emulator and iOS simulator: banner renders below the app bar, hide and
      show still works, landscape fine
- [x] Web unchanged, before and after screenshots match

<details>
<summary>Before</summary>
ios-before
</details>

<details>
<summary>After</summary>
ios-after
</details>

### DIFF
--- a/example/src/Examples/BannerExample.tsx
+++ b/example/src/Examples/BannerExample.tsx
@@ -36,7 +36,7 @@ const BannerExample = () => {
   };
 
   return (
-    <>
+    <View style={styles.container}>
       <ScreenWrapper>
         <View style={[styles.grid, { paddingTop: height }]}>
           {PHOTOS.map((uri) => (
@@ -78,13 +78,16 @@ const BannerExample = () => {
         Two line text string with two actions. One to two lines is preferable on
         mobile.
       </Banner>
-    </>
+    </View>
   );
 };
 
 BannerExample.title = 'Banner';
 
 const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
   ...Platform.select({
     web: {
       grid: {


MikitasK (approved):
LGTM 👌

JKobrynski (approved):
LGTM!

satya164 (approved):
Thanks

PR #5069, opened 2026-08-24
docs: replace FAB and Switch screenshots with live interactive examples

### LINKED ISSUE #4991: docs: revisit the documentation and update outdated sections
At the time of writing this, there are sections in the documentation that are outdated.  
For example, FAB and Switch sections use old screenshots, naming and token references.  

To improve the docs, we could drop these screenshots and inline the real interactive components instead.

hristototov:
### Motivation

Parts of the documentation have drifted out of sync with the v6 modernisation effort. The FAB and Switch pages were still illustrated with screenshots captured before those components were reworked for the latest Material Design 3 spec, so they showed outdated naming and token references:

- **FAB** had an "all modes" screenshot tab, but the component no longer has a `mode` prop — it's `variant` (six role-color presets) and `size` (three spec sizes) now.
- **Switch** was documented with a four-way `Android (enabled/disabled)` / `iOS (enabled/disabled)` screenshot split, which no longer reflects reality: the modernised component renders the same MD3 visuals on both platforms.

Static images also can't show what these components actually do - the Switch handle animation, the Extended FAB expand/collapse transition - and they silently rot every time a component changes.

This PR drops those screenshots and inlines the real components instead, so the docs render live from

### Related issue

Closes #4991

### Implementation notes

A general mechanism rather than a one-off, so the remaining outdated pages can follow:

- **`docs/src/data/liveExamples.ts`** - new registry mapping a component title to the example module and named exports to render. Adding an entry here automatically removes that component's screenshot tabs, so a component is documented either with screenshots or with live examples, never both. The invariant is enforced in one place in `generatePageMDX.ts`.
- **`docs/src/components/InteractiveExample.tsx`** - shared shell for embedded demos. Paper components reach browser APIs through `react-native-web`, so the tree mounts client-only behind `BrowserOnly` with a same-sized placeholder during SSR to avoid layout shift, and the Paper theme is kept in sync with the active docs color mode. Follows the existing pattern in `BannerExample.tsx`.
- **`FABExample.tsx` / `SwitchExample.tsx`** - the actual demos. The FAB variant and size lists mirror the `Variant` and `Size` unions in `src/components/FAB/tokens.ts` in order, so a spec change surfaces as a type error rather than a stale image.

Also includes one small runtime fix surfaced while building the demos: `Switch`'s `absoluteFill` style was missing `alignItems` / `justifyContent`, so the unselected icon which is smaller than the `SELECTED_ICON`-sized `iconWrap` it sits in - was pinned to the top-left instead of centered. Snapshots updated accordingly.

### Images

switch-after

switch-before

fab-after

fab-before

fab-extended-after

fab-extended-before


JKobrynski (approved):
LGTM!

docs/src/components/FABExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+            aria-label={`${variant} floating action button`}
+          />
+        </Labelled>
+      ))}
+    </ExampleRow>
+  </InteractiveExample>
+);
+
+export const FABSizesExample = () => (
+  <InteractiveExample title="All sizes — press for the ripple and state layer">
```
satya164:
```suggestion
  <InteractiveExample title="Sizes">
```

docs/src/components/FABExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * order as the `Variant` union in `src/components/FAB/tokens.ts`.
+ */
+const VARIANTS = [
+  'primary',
+  'secondary',
+  'tertiary',
+  'tonalPrimary',
+  'tonalSecondary',
+  'tonalTertiary',
+] as const;
```
satya164: Remove the comment. Annotate the `VARIANTS` with a type based on the `Variant` so it actually stays in sync instead of the comment saying it does

docs/src/components/FABExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Every spec size accepted by the `size` prop.
+ */
+const SIZES = ['default', 'medium', 'large'] as const;
```
satya164: Same as above

docs/src/components/FABExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const FABVariantsExample = () => (
+  <InteractiveExample title="All variants — press for the ripple and state layer">
```
satya164: Remove emdashes please

Also, just "Variants" is enough. And "press for the ripple and state layer" is unnecessary. It's an interactive demo, and we shouldn't be mentioning every detail in the title. 

```suggestion
  <InteractiveExample title="Variants">
```

docs/src/components/FABExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+      ))}
+    </ExampleRow>
+  </InteractiveExample>
+);
+
+export const FABExtendedExample = () => {
+  const [expanded, setExpanded] = React.useState(true);
+
+  return (
+    <InteractiveExample title="Extended — press to collapse and expand">
```
satya164:
```suggestion
    <InteractiveExample title="Extended">
```

docs/src/components/InteractiveExample.tsx
```diff
@@ -0,0 +1,92 @@
...
+  },
+  item: {
+    alignItems: 'center',
+    gap: 8,
+  },
+  itemLabel: {
+    fontSize: 12,
+    opacity: 0.7,
+  },
+});
```
satya164: styles should be at the end of the file

docs/src/components/InteractiveExample.tsx
```diff
@@ -0,0 +1,92 @@
...
+  title?: string;
+}>;
+
+/**
+ * Shared shell for live component demos embedded in the docs.
+ *
+ * Paper components depend on browser APIs through `react-native-web`, so the
+ * tree is only mounted on the client and a same-sized placeholder is rendered
+ * during SSR to avoid layout shift. The Paper theme is kept in sync with the
+ * active docs color mode so demos match the surrounding page.
```
satya164: The comment is wrong for the component's content. There's nothing related to SSR here.

But also why are we only rendering these on client and skipping on SSR? We need to render components both on SSR and client.

docs/src/css/custom.css
```diff
@@ -1286,3 +1286,26 @@ html:not(.dark) .paper-version-selector-menu {
...
+.paper-interactive-example__title {
+  margin-bottom: 16px;
+  color: var(--rp-c-text-2);
+  font-size: 13px;
+  font-weight: 600;
+}
+
+/* Nominal reserve so the page does not jump to zero height during SSR. Demos
+   vary in height, so this deliberately approximates a single row rather than
+   claiming to match any particular example. */
```
satya164: Use simple english

docs/src/data/liveExamples.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Components whose docs page renders live, interactive examples instead of
+ * static screenshots.
+ *
+ * Each entry maps a component title (as it appears in the generated docs) to
+ * the named exports of a module under `@docs/components` that should be
+ * rendered right below the page summary. Adding an entry here automatically
+ * removes the component's screenshot tabs, so a component is documented either
+ * with screenshots or with live examples — never both.
+ */
```
satya164: remove unnecessary comment

satya164 (changes requested):
Please avoid emdashes and claudism in text. Using LLM is fine but we shouldn't make our documentation look AI generated without human attention.

docs/src/components/FABExample.tsx
hristototov: updated

docs/src/components/FABExample.tsx
hristototov: updated

docs/src/components/FABExample.tsx
hristototov: updated

docs/src/components/FABExample.tsx
hristototov: updated

docs/src/components/InteractiveExample.tsx
hristototov: updated

docs/src/data/liveExamples.ts
hristototov: updated

docs/src/css/custom.css
hristototov: updated

docs/src/components/FABExample.tsx
hristototov: updated

docs/src/components/InteractiveExample.tsx
hristototov: updated

- Dropped `BrowserOnly` and the `Placeholder`, then collapsed `Frame `+ the pass-through `InteractiveExample` into a single component with no gate; the split had no purpose
- Removed the now-dead `.paper-interactive-example--loading` class
- Rewrote the comment

hristototov: updated

- Dropped `BrowserOnly` and the `Placeholder`, then collapsed `Frame `+ the pass-through `InteractiveExample` into a single component with no gate; the split had no purpose
- Removed the now-dead `.paper-interactive-example--loading` class
- Rewrote the comment

docs/src/components/SwitchExample.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Switch } from 'react-native-paper';
+
+import InteractiveExample, { ExampleRow, Labelled } from './InteractiveExample';
+
+export const SwitchStatesExample = () => {
+  const [on, setOn] = React.useState(true);
+  const [off, setOff] = React.useState(false);
+
+  return (
+    <InteractiveExample title="Enabled — toggle to see the handle animate">
```
JKobrynski: Same fix here as FABExample.tsx:26: remove the em dashes and shorten the titles. Lines 12, 26 and 43.

hristototov: updated

docs/src/components/InteractiveExample.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * Shared shell for live component demos embedded in the docs.
+ *
+ * The Paper theme follows the active docs color mode so demos match the
+ * surrounding page.
+ */
+const InteractiveExample = ({ title, children }: InteractiveExampleProps) => {
+  const isDarkTheme = useColorMode().colorMode === 'dark';
+
+  return (
+    <Provider theme={isDarkTheme ? DarkTheme : LightTheme}>
```
JKobrynski, line 46: Have you checked these in dark mode? I'm afraid they render with light-theme colors, so the caption under each switch ends up dark on dark and unreadable.

docs/src/components/FABExample.tsx
```diff
@@ -0,0 +1,65 @@
...
+    </ExampleRow>
+  </InteractiveExample>
+);
+
+export const FABExtendedExample = () => {
+  const [expanded, setExpanded] = React.useState(true);
+
+  return (
+    <InteractiveExample title="Extended">
+      <FAB.Extended
```
JKobrynski, line 57: Have you checked the extended FAB on the built page?

docs/src/components/InteractiveExample.tsx
hristototov: I have checked dark mode (even the screenshots are in dark mode) however, you're right about the text color - i didn't notice this. Fixed now 👍 

cr-1

docs/src/components/FABExample.tsx
hristototov: Yes, checked it on the built output (yarn build + yarn preview), it renders correctly, and the interaction is as expected 
cr-2

docs/src/components/InteractiveExample.tsx
JKobrynski: Captions read fine now, thanks. The switches still come up light-themed on a first dark-mode load, though. They only turn dark after navigating away and back. Try a hard reload with a dark OS theme and nothing stored.

PR #5071, opened 2026-08-24
fix: don't expose handler-less TouchableRipple as a disabled control

### LINKED ISSUE #5070: TouchableRipple exposes non-interactive surfaces as disabled controls
### Current behaviour

`TouchableRipple` treats "no touch handler was passed" as identical to "this control is disabled":

```ts
// src/components/TouchableRipple/TouchableRipple.native.tsx
const disabled = disabledProp || !hasPassedTouchHandler;
```

That single flag is then handed to `Pressable`, which derives `accessibilityState.disabled` from it. The result is that every non-interactive surface built on `TouchableRipple` is announced to assistive tech as a **disabled control**, even though nothing about it is disabled and it was never meant to be a control.

Nothing looks wrong on screen, which is why this has gone unnoticed. Paper never dimmed these views, so a screenshot is identical either way. The defect is only visible in the accessibility tree.

Measured on the example app, `TextInput` screen, with "Leading icon" enabled. The leading magnifier has no `onPress`; the trailing clear button does:

| platform   | probe                   | decorative magnifier                                   | clear button            |
| ---------- | ----------------------- | ------------------------------------------------------ | ----------------------- |
| iOS 18.3   | `AXButton.enabled`      | `false`                                                | `true`                  |
| Android 15 | `uiautomator` `enabled` | `"false"`                                              | `"true"`                |
| web        | DOM                     | `<button disabled aria-disabled="true" tabindex="-1">` | `<button tabindex="0">` |

The web case is the easiest to confirm and the most severe. A decorative icon becomes a genuinely `disabled` HTML `<button>` with `pointer-events: none`.

VoiceOver and TalkBack both read the disabled bit, so a screen reader user hears "search, dimmed, button" on an element that is purely decorative.

This is not limited to one component. It affects any of the 22 components that render `TouchableRipple` when they are used without a touch handler, including `Chip`, `Button`, `List.Item`, `DataTable.Cell`, `Drawer.Item`, `Menu.Item`, `Card` and `IconButton`. Across Paper's own test suite, 80 rendered elements carry the false `disabled` state.

Some consumers work around it by nulling their own handler to fake a disabled state, which stops working the moment the primitive is corrected. `TextInput.Icon` does exactly this:

```ts
// src/components/TextInput/TextInputIcon.tsx
const onPressHandler = disabled ? undefined : onPress;
```

### Expected behaviour

A component with no touch handler should not be exposed as a control at all: no `accessibilityState.disabled`, no keyboard or D-pad focus, no `<button disabled>` on web.

A component whose caller explicitly passed `disabled` should still be announced as a disabled control, exactly as it is today.

The two cases are different and should not share one flag.

### How to reproduce?

1. Run the example app: `cd example && yarn start`
2. Open the **TextInput** screen and turn on the "Leading icon" toggle. That renders a `TextInput.Icon` with no `onPress`, next to a clear button that has one.
3. Inspect the accessibility tree:
   - **iOS**: Xcode > Open Developer Tool > Accessibility Inspector, target the simulator, click the magnifier. Traits read `Button, Not Enabled`.
   - **Android**: `adb shell uiautomator dump` and read the node with `resource-id="icon-button"`. It reports `enabled="false"`.
   - **Web**: `yarn start --web`, then inspect the magnifier in devtools. It is `<button disabled aria-disabled="true" tabindex="-1">`.
4. With VoiceOver or TalkBack on, move to the magnifier. It is announced as disabled.

The same thing is visible anywhere a `TouchableRipple`-based component is rendered without a handler, for example `<Chip>Plain</Chip>` or a `List.Item` with no `onPress`.

### Preview

Not applicable. There is no visual difference to screenshot, which is the point: Paper applies no dimming to these views on native, and `styles.disabled` on web is only `cursor: auto`. The defect lives entirely in the accessibility tree, so the tables above are the evidence.

- Confirmed the mechanism in React Native's source. `Pressable.js:235-236` merges the `disabled` prop into `accessibilityState` and it always wins over `aria-disabled`/`accessibilityState`, so passing `undefined` is the only way to leave the key absent.
- Confirmed the blast radius by walking the accessibility tree across component configurations on a real simulator and emulator, plus the DOM on web.
- Looked at the existing narrow fixes. #5011 ("fix: avoid disabled accessibility state for active chips") addresses this same symptom, but only inside `Chip.tsx`, by injecting an empty `onPress` to defeat the primitive. That swaps a false "disabled" for a false "enabled" and leaves the other 21 components untouched. #5002 also modifies `Chip.tsx` in overlapping ways.
- Note for whoever picks this up: two of Paper's own tests currently encode the bug as intended behaviour (`renders disabled button if there is no touch handler passed` and its `Chip` equivalent, both asserting `toBeDisabled()`), so they need splitting into a "no handler" case and a "disabled prop" case rather than updating in place.

### Your Environment

lukemorawski:
### Motivation

`TouchableRipple` had `const disabled = disabledProp || !hasPassedTouchHandler`, so anything without a press handler got announced as a disabled control. Split into two flags:

```ts
const isControl = hasPassedTouchHandler || Boolean(disabledProp);
const isInteractive = hasPassedTouchHandler && !disabledProp;
```

Non-controls now render a plain `Animated.View` instead of a `Pressable`.

Worth knowing why it drops the `Pressable` instead of just passing the caller's `disabled` through: `disabled` also gates touch-responder claiming (`Pressability.js:448`, `onStartShouldSetResponder: () => !disabled`). Fixing only the announcement turns claiming *on*, so handler-less ripples start swallowing taps. `TextInput` wraps its field and icons in `<Pressable onPress={focusInput}>` (`TextInput.tsx:352`), so tapping a decorative icon stopped focusing the field. Tests stayed green through that.

Two things kept on purpose, both break stuff if dropped:
- `accessible={rest.accessible !== false}`, or a checked `CheckboxItem` with no handler loses its role and state.
- `focusable={rest.focusable ?? false}`, or on web the icon lands in the tab order, since RNW turns `role="button"` into a real `<button>`.

### Related issue

Closes #5070

Supersedes #5011, which fixes the same symptom in `Chip.tsx` only by injecting an empty `onPress`. That swaps a false "disabled" for a false "enabled" and leaves the other 21 consumers broken.

### Test plan

Lint, typecheck and tests pass: 55 suites, 742 tests, 169 snapshots.

The suite can't catch this though. It renders an element tree, never a DOM or a native view, and jest never loads `TouchableRipple.tsx` at all. So it's checked on device, example app, **TextInput** screen with "Leading icon" on. Leading magnifier has no handler, trailing clear button does.

| | announced disabled | tap reaches field | clear button still works |
| --- | --- | --- | --- |
| iOS 18.3 | `enabled: false` -> `true` | yes | yes |
| Android 15 | `enabled="false"` -> `"true"` | yes, 3/3 runs | yes |
| web | `<button disabled>` -> gone | yes | yes |

Web also keeps `tabindex="-1"` on the decorative icon.

Snapshots are noisy but mechanical: 80 elements drop the `Pressable` handlers and the always-`undefined` a11y scaffolding. Checked every style key/value pair, nothing net-added. Some `aria-*` show up raw now because `View.js` does that folding at runtime and the jest preset mocks `View`.

### Two extras, worth a look

`TextInputIcon` was faking disabled by nulling its own handler, which only worked because of this bug. Now passes `disabled` through.

`IconButton.getIconColor` returned on `disabled` before reading `customIconColor`, so once `TextInputIcon` forwarded `disabled` a disabled field's icon lost its custom and error colour. Reordered so an explicit colour wins, opacity still shows disabled. No snapshot churn, so nothing else passes both, but it is a public component behaviour change and the bit I'd most want checked.

### DIFF
--- a/src/components/IconButton/utils.ts
+++ b/src/components/IconButton/utils.ts
@@ -61,14 +61,12 @@ const getIconColor = ({
   selected,
   customIconColor,
 }: BaseProps & { customIconColor?: ColorValue }) => {
-  // An explicitly passed color is an instruction, so it outranks the disabled
-  // default. Disabled is still conveyed by the reduced icon opacity.
-  if (typeof customIconColor !== 'undefined') {
-    return customIconColor;
-  }
-
   if (disabled) {
     return theme.colors.onSurface;
+  }
+
+  if (typeof customIconColor !== 'undefined') {
+    return customIconColor;
   }
 
   if (isMode('contained')) {
--- a/src/components/IconButton/utils.ts
+++ b/src/components/IconButton/utils.ts
@@ -61,12 +61,14 @@ const getIconColor = ({
   selected,
   customIconColor,
 }: BaseProps & { customIconColor?: ColorValue }) => {
-  if (disabled) {
-    return theme.colors.onSurface;
-  }
-
+  // An explicitly passed color is an instruction, so it outranks the disabled
+  // default. Disabled is still conveyed by the reduced icon opacity.
   if (typeof customIconColor !== 'undefined') {
     return customIconColor;
+  }
+
+  if (disabled) {
+    return theme.colors.onSurface;
   }
 
   if (isMode('contained')) {
--- a/src/components/TextInput/TextInputIcon.tsx
+++ b/src/components/TextInput/TextInputIcon.tsx
@@ -78,8 +78,6 @@ const TextInputIcon = ({
     isDisabled: disabled,
   });
 
-  const onPressHandler = disabled ? undefined : onPress;
-
   return (
     <View style={styles.iconWrapper}>
       <IconButton
@@ -88,7 +86,8 @@ const TextInputIcon = ({
         iconColor={color}
         size={iconSize}
         style={[styles.icon, style]}
-        onPress={onPressHandler}
+        disabled={disabled}
+        onPress={onPress}
       />
     </View>
   );
--- a/src/components/TextInput/TextInputIcon.tsx
+++ b/src/components/TextInput/TextInputIcon.tsx
@@ -97,7 +97,15 @@ const TextInputIcon = ({
         icon={icon}
         iconColor={color}
         size={iconSize}
-        style={[styles.icon, style]}
+        style={[
+          styles.icon,
+          style,
+          // TextInput already dims the whole accessory when the field is
+          // disabled. Forwarding `disabled` makes IconButton dim the icon too,
+          // and the two multiply, so let IconButton own it and cancel the outer
+          // one. A decorative icon keeps it, nothing else dims it.
+          isInteractive && disabled ? styles.notDimmed : null,
+        ]}
         disabled={isInteractive ? disabled : undefined}
         onPress={onPress}
       />
--- a/src/components/TextInput/TextInputIcon.tsx
+++ b/src/components/TextInput/TextInputIcon.tsx
@@ -7,6 +7,7 @@ import { styles } from './styles';
 import { getIconColor } from './utils';
 import { useInternalTheme } from '../../core/theming';
 import type { $Omit } from '../../types';
+import hasTouchHandler from '../../utils/hasTouchHandler';
 import IconButton from '../IconButton/IconButton';
 
 export type TextInputAccessoryProps = {
@@ -78,6 +79,17 @@ const TextInputIcon = ({
     isDisabled: disabled,
   });
 
+  // A decorative icon is not a control, so `disabled` would only announce it as
+  // a disabled button. Must match TouchableRipple's predicate, or an icon with
+  // only `onLongPress` stays pressable on a disabled field.
+  const { onLongPress, onPressIn, onPressOut } = rest;
+  const isInteractive = hasTouchHandler({
+    onPress,
+    onLongPress,
+    onPressIn,
+    onPressOut,
+  });
+
   return (
     <View style={styles.iconWrapper}>
       <IconButton
@@ -86,7 +98,7 @@ const TextInputIcon = ({
         iconColor={color}
         size={iconSize}
         style={[styles.icon, style]}
-        disabled={disabled}
+        disabled={isInteractive ? disabled : undefined}
         onPress={onPress}
       />
     </View>
--- a/src/components/IconButton/IconButton.tsx
+++ b/src/components/IconButton/IconButton.tsx
@@ -202,7 +202,10 @@ const IconButton = ({
         testID={testID}
         {...rest}
       >
-        <View style={{ opacity: iconOpacity }}>
+        <View
+          testID={`${testID}-icon-opacity`}
+          style={{ opacity: iconOpacity }}
+        >
           {loading ? (
             <ActivityIndicator size={size} color={iconColor} />
           ) : (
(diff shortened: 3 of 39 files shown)


src/components/TextInput/TextInputIcon.tsx
```diff
@@ -88,7 +86,8 @@ const TextInputIcon = ({
         iconColor={color}
         size={iconSize}
         style={[styles.icon, style]}
-        onPress={onPressHandler}
+        disabled={disabled}
+        onPress={onPress}
```
MikitasK, line 110: what do you think about making it non-accessible when it has no `onPress`?

```suggestion
        onPress={onPress}
        accessible={Boolean(onPress)}
        role={onPress ? 'button' : 'none'}
```

updated snapshot still has `accessible={true}` & `role="button"`. RN Web maps that role to a native `<button>`, `focusable={false}` only removes it from tab order. 
that's why screen readers may still announce actionable button that does nothing

src/components/IconButton/utils.ts
MikitasK, line 1: what about keeping existing disabled-color precedence here & handling the `TextInput.Icon` case locally?
moving `customIconColor` first changes every `<IconButton disabled iconColor="......" />`.

`Button` still give disabled colors precedence ([link](https://github.com/callstack/react-native-paper/blob/main/src/components/Button/utils.tsx#L41-L59)) while `IconButton` docs list `onSurfaceDisabled` ([link](https://github.com/callstack/react-native-paper/blob/main/docs/6.x/docs/components/IconButton/IconButton.mdx)) 
however, if this global behavior change is intentional, maybe we can at least document it?

src/components/TextInput/TextInputIcon.tsx
lukemorawski: yep, RNW does map role="button" to a real <button>, focusable={false} only adds tabindex="-1". Measured before: web <button tabindex="-1">, Android Button clickable=true. Same on iOS.
but i fixed in TouchableRipple instead of here. IconButton, Drawer.Item and Drawer.CollapsedItem all set role="button" unconditionally with an optional onPress, so fixing it here would leave two others. AppbarContent already gates its role on onPress.
Only the role is dropped, the element stays. accessible={false} is a noop on web anyway (not in RNW's forwardedProps) and on naitve it kills aria-label, which for an icon-only FAB or IconButton is the only thing it has to announce. dropping the role alone is enough on both platforms
Covers button, imagebutton, link, menuitem and tab. checkbox, radio and switch are kept, they still describe a read-only view.
After fix: web <div role="presentation">, Android plain View clickable=false, clear button unchanged on both.

src/components/IconButton/utils.ts
lukemorawski: ok reverted IconButton/utils.ts
The reorder was only collateral from forwarding disabled into IconButton. it's now handled locally instead: TextInput.Icon now only forwards disabled when the icon actually has a touch handler. That also fixes the other bug in this PR - a decorative magnifier glass in a disabled field was still rendering as <button disabled aria-disabled="true">
Default case is unchanged, TextInput's getIconColor already returns onSurface when disabled, same as IconButton's disabled branch.
One real thing: an interactive icon on a disabled field now takes IconButton's disabled colour, so an explicit iconColor or the error colour no longer wins there. matches Button and the docs, so nothing to document

PR #5020, opened 2026-07-01
refactor: searchbar component

MrMuzyk:
<!-- Please provide enough information so that others can review your pull request. -->
<!-- Keep pull requests small and focused on a single change. -->

### Motivation

<!-- Explain the **motivation** for making this change. What existing problem does the pull request solve? -->

Modernizes the Searchbar to the latest MD3 (Expressive) search spec by adopting the shared theme tokens (shape, motion, colors), extracting component-specific tokens, and adding the missing results container - following the token/structure conventions of the already-modernized TextInput/Switch/Checkbox/FAB. Also fixes the long-standing swapped input/placeholder colors and adds the missing focus transition.      

#### Changes                                                                                                                                                                 
                                                                                                                                                                           
#### Colors (tokens.ts / utils.ts)                                                                                                                                           
   - Fixed swapped roles: input text onSurfaceVariant → onSurface; placeholder onSurface → onSurfaceVariant (MD3 order).                                                   
   - Container background resolved from the container token (surfaceContainerHigh) via getSearchbarColors, instead of a hardcoded color.                                   
                                                                                                                                                                           
#### Shape                                                                                                                                                                   
   - contained → corner.extraLarge (28dp), divided → none (0), resolved through the shared resolveCornerRadius util (src/theme/utils/shape.ts).                            
                                                                                                                                                                           
#### Motion - focus transition (new)                                                                                                                                         
  - The contained bar grows a little wider when focused: its side margins shrink from 24dp to 12dp, then expand back on blur. It's a spring animation (Reanimated) and respects the OS "reduce motion" setting.                                                                                                 
  - Note: contained mode (now the default) has side margins, where the old `bar` mode was full-width. Set a horizontal margin in `style` to override it — that also turns the focus animation off.                                                   
                                                                                                                                                                           
#### Structure                                                                                                                                                               
   - Moved Searchbar.tsx into its own src/components/Searchbar/ directory with tokens.ts (sizes/shape/colors), utils.ts (getSearchbarColors, getSearchbarInputFont), and   
   index.tsx (attaches Results as a static sub-component) - mirroring Switch/Checkbox/FAB.                                                                                 
                                                                                                                                                                           
#### New sub-component — Searchbar.Results (SearchbarResults.tsx)                                                                                                            
   - A Surface container for the results/suggestions list (MD3 search anatomy element 6). It provides only the surface; result grouping is left to the consumer. SearchbarResultsProps is exported from the package root.                                                                                                                
                                                                                                                                                                           
[...]

PR #5072, opened 2026-08-25
refactor(dialog): MD3 updates, updated Dialog API

kachmashk:
## Motivation

The goal of this PR is to address Material Design 3 guidelines for the Dialog component based on the [official documentation](https://m3.material.io/components/dialogs/overview) as well as [this issue](https://github.com/callstack/react-native-paper/discussions/4949). This PR also provides ability to avoid using previous compound components implementation in favor of props directly passed to the Dialog component.

#### Material Design 3:

- `Dialog.ScrollArea` divider border color uses `outline`,
- `Dialog` has now restricted minimum and maximum width set to `280dp` and `560dp`,
- Fixed an issue when `DialogIcon` is the first component in the `Dialog` and it had a total top margin of `48dp` instead of `24dp`,
- `Dialog` now detects if `Dialog.Icon` and `Dialog.Title` are present and if so enforcing `Dialog.Title` to be centered.

#### Dialog:

- `Dialog` component now supports `icon`, `title`, `content`, `actions`, `scrollable` props instead of using compound components. `Dialog` still uses compound components underneath to properly comply with Material Design specs as these components enforce proper styling,
- Added new test cases for `Dialog` to ensure the new approach works as intended.

#### Example:

- Updated `Dialog` screen within Example app with the same dialogs as before but using new API to make sure they look and work as expected.

## Related issue

- https://github.com/callstack/react-native-paper/issues/4933

## Test plan

- [x] lint
- [x] typecheck
- [x] test

Reviewers can through new Dialogs added to the Example app. They are duplicated as previous dialogs but they use new approach.


src/components/Dialog/Dialog.tsx
```diff
@@ -46,17 +48,120 @@ export type Props = {
...
+  contentProps?: never;
+  scrollAreaProps?: never;
+  scrollViewProps?: never;
+};
+
+export type Props = CommonProps & (SlotProps | LegacyProps);
+
 const DIALOG_ELEVATION: number = 24;
 
+const renderChildren = (children: React.ReactNode) => {
```
JKobrynski: `React.Children` + `cloneElement` is the composition problem v6 is meant to remove ([#4954](https://github.com/callstack/react-native-paper/discussions/4954)), and this adds `child.type ===` sniffing on top. On the prop path you build the array yourself - pass the margin and centring down as props instead.

src/components/Dialog/Dialog.tsx
```diff
@@ -46,17 +48,120 @@ export type Props = {
...
+   */
+  scrollAreaProps?: Omit<DialogScrollAreaProps, 'children'>;
+  /**
+   * Props passed to the `ScrollView` when `scrollable` is enabled.
+   */
+  scrollViewProps?: Omit<ScrollViewProps, 'children'>;
+  children?: never;
+};
+
+type LegacyProps = {
```
JKobrynski: This leaves v6 with two Dialog APIs - `children` deprecated but kept, plus seven duplicated `NewDialog*` example screens. Which one ships? A major shouldn't carry a deprecation shim, but dropping `children` is a break worth deciding deliberately.

src/components/Dialog/DialogScrollArea.tsx
```diff
@@ -52,7 +60,7 @@ const DialogScrollArea = (props: Props) => {
   const theme = useInternalTheme(props.theme);
   const { colors } = theme;
   const borderStyles = {
-    borderColor: colors.surfaceVariant,
+    borderColor: colors.outline,
```
JKobrynski:
```diff
-    borderColor: colors.outline,
+    borderColor: colors.outlineVariant,
```

`md.comp.dialog.with-divider.divider.color` is deprecated in favour of `md.comp.divider.color` - `outline-variant`, which is what `Divider` already uses.

src/components/Dialog/Dialog.tsx
```diff
@@ -46,17 +48,120 @@ export type Props = {
...
+      const topMarginStyle =
+        i === 0 && child.type !== DialogIcon ? styles.firstChild : undefined;
+      const titleAlignmentStyle =
+        hasIcon && child.type === DialogTitle
+          ? styles.titleWithIcon
+          : undefined;
+
+      if (topMarginStyle || titleAlignmentStyle) {
+        return React.cloneElement(child, {
+          style: [topMarginStyle, child.props.style, titleAlignmentStyle],
```
JKobrynski:
```diff
-          style: [topMarginStyle, child.props.style, titleAlignmentStyle],
+          style: [topMarginStyle, titleAlignmentStyle, child.props.style],
```

As written, `<Dialog.Title style={{ textAlign: 'left' }}>` stops working once there's an icon.

src/components/Dialog/Dialog.tsx
```diff
@@ -46,17 +48,120 @@ export type Props = {
...
+  icon?: IconSource;
+  /**
+   * Title of the dialog.
+   */
+  title?: React.ReactNode;
+  /**
+   * Content of the dialog. Non-empty strings are rendered as Material 3
+   * supporting text.
+   */
+  content: React.ReactNode;
```
JKobrynski: `content` and `actions` are required, so a title-only dialog won't typecheck and `NewDialogWithLoadingIndicator.tsx:35` has to pass `actions={[]}`. Make both optional.

src/components/Dialog/Dialog.tsx
```diff
@@ -46,17 +48,120 @@ export type Props = {
...
+  actions?: never;
+  scrollable?: never;
+  contentProps?: never;
+  scrollAreaProps?: never;
+  scrollViewProps?: never;
+};
+
+export type Props = CommonProps & (SlotProps | LegacyProps);
+
 const DIALOG_ELEVATION: number = 24;
```
JKobrynski: `DIALOG_ELEVATION` never reaches the view - `Surface` applies its own `elevation` prop after the style, so the dialog renders at level 1. `md.comp.dialog.container.elevation` is `level3`. Since this is the MD3 pass, worth fixing or dropping the constant.

src/components/Dialog/Dialog.tsx
```diff
@@ -46,17 +48,120 @@ export type Props = {
   testID?: string;
 };
 
+type DialogActionsProps = Omit<ButtonProps, 'children'> & { label: string };
```
JKobrynski: Why `{ label }[]` rather than `actions?: React.ReactNode`? An array of Button props can't express a custom action component, and it's a second shape to keep in sync with `Button`.

src/components/Dialog/Dialog.tsx
```diff
@@ -97,13 +226,83 @@ const Dialog = ({
...
+            >
+              {label}
+            </Button>
+          )
+        )}
+      </DialogActions>
+    ) : null;
+
+    return [dialogIcon, dialogTitle, dialogContent, dialogActions];
+  }, [children, props, theme.colors.onSurfaceVariant]);
```
JKobrynski: `props` is a fresh rest object every render, so this `useMemo` never hits. Drop it, or list the individual props.

docs/src/data/componentDocs6x.json
```diff
@@ -5206,10 +5206,10 @@
     "Dialog/Dialog": {
```
JKobrynski, line 5212: Run `yarn docs generate` - `componentDocs6x.json` and the `.mdx` pages are generated, and the JSDoc you added to `DialogActions`, `DialogContent`, `DialogScrollArea` and `DialogTitle` hasn't reached either. The hand-typed `### children (depracated…)` heading will be overwritten.

src/components/Dialog/Dialog.tsx
```diff
@@ -97,13 +226,83 @@ const Dialog = ({
...
+        {contentNode}
+      </DialogContent>
+    );
+
+    const dialogActions = actions?.length ? (
+      <DialogActions key="dialogActions">
+        {actions.map(
+          ({ label, onPress: onActionPress, ...buttonProps }, index) => (
+            <Button
+              key={index}
```
JKobrynski: `key={index}` on a consumer-supplied array - the JSDoc note "Keep their order stable between renders" is working around it. Key on `label`.

src/components/Dialog/Dialog.tsx
```diff
@@ -97,13 +226,83 @@ const Dialog = ({
...
+      contentProps,
+      scrollAreaProps,
+      scrollViewProps,
+      title,
+    } = props;
+
+    const dialogIcon = icon ? (
+      <DialogIcon icon={icon} key="dialogIcon" />
+    ) : null;
+    const dialogTitle = title ? <DialogTitle>{title}</DialogTitle> : null;
```
JKobrynski: The new API knows the `title`, but nothing gives the dialog an accessible name - `Modal` sets `aria-modal` with no label. Worth wiring `title` through while the API is being designed.

JKobrynski (changes requested):

src/components/Dialog/Dialog.tsx
kachmashk: Removed legacy approach with `children`. Removed all `NewDialog*` examples and replaced older examples with new implementation.

src/components/Dialog/DialogScrollArea.tsx
kachmashk: Changed!

src/components/Dialog/Dialog.tsx
kachmashk: That's what I wanted to achieve. I can see in the guidelines that `Title` must be always centered when `Icon` is visible. Would you like to drop this restriction?

src/components/Dialog/Dialog.tsx
kachmashk: Guidelines say that `content` and `actions` are not optional. Should we override it and allow title-only dialogs? `DialogWithLoadingIndicator.tsx` uses an empty array for `actions` but in this case it's against the guidelines and shouldn't be used this way. Do we still want `content` and `actions` as optional?

src/components/Dialog/Dialog.tsx
kachmashk: Removed!

src/components/Dialog/Dialog.tsx
kachmashk: This was a proposed example for `actions` and it looked solid but you're right that it cannot express a custom action component.

I changed `actions` into `React.ReactNode[]` type which results in example usage:

```
actions={[
   <Button key="disagree-btn" onPress={close} textColor={Palette.error50}>
      Disagree
   </Button>,
   <Button key="agree-btn" onPress={close}>
      Agree
   </Button>,
]}
```
This way we simplify a composition and let developers decide what action component they would like to use.

I could still change it to `React.ReactNode` instead and we would have these possibilities:

1. Developers to keep using `Dialog.Actions` which does not simplify using `Dialog` component as `title` or `content` props.
```
actions={
   <Dialog.Actions>
      <Button onPress={close} textColor={Palette.error50}>
         Disagree
      </Button>
      <Button onPress={close}>
         Agree
      </Button>
   </Dialog.Actions>
]}
```

2. We would accept a `React.ReactNode` but it would add additional logic to the `DialogActions.tsx` to unwrap one level of children to properly assign a styling to them. Each child is supposed to have `marginRight: 8` expect the last one.
```
actions={
   <>
      <Button onPress={close} textColor={Palette.error50}>
         Disagree
      </Button>
      <Button onPress={close}>
         Agree
      </Button>
   </>
]}
```

What do you think?

src/components/Dialog/Dialog.tsx
kachmashk: Ignoring as `actions` changed a type.

src/components/Dialog/Dialog.tsx
kachmashk: Added `accessibilityLabel` as a `Dialog` prop to enable passing custom `aria-label`. I also added a condition where if `title` is a `string` then it should be taken as `aria-label` by default which is what guidelines suggest. We could also drop it at only pass a provided `accessibilityLabel`. What do you think?

src/components/Dialog/Dialog.tsx
kachmashk: Removed!

src/components/Dialog/Dialog.tsx
kachmashk: Removed and updated!

docs/src/data/componentDocs6x.json
kachmashk: Updated!

src/components/Dialog/Dialog.tsx
```diff
@@ -32,9 +35,46 @@ export type Props = {
    */
   visible: boolean;
   /**
-   * Content of the `Dialog`.
+   * Icon to display above the dialog title.
    */
-  children: React.ReactNode;
+  icon?: IconSource;
```
JKobrynski, line 41: Removing `children` needs an entry in [`migration.md`](https://github.com/callstack/react-native paper/blob/main/docs/6.x/docs/guides/migration.md) - #5020 and #4994 add one for their components. `Dialog.Actions` also stopped injecting `compact` and `uppercase`, which widens every action button.

docs/src/data/componentDocs6x.json
JKobrynski: Still not regenerated. `Dialog.mdx:133` documents a `children` prop that no longer exists, line 71 says passing the sub-components through `children` is "deprecated", and lines 21, 56 and 89 show `actions={[{ label: 'Done', onPress: hideDialog }]}`, which no longer typechecks. Same example in `Dialog.tsx:116` and `DialogActions.tsx:38`.

src/components/Dialog/DialogActions.tsx
```diff
@@ -33,39 +32,41 @@ export type Props = ViewProps & {
...
-            })
-          : child
-      )}
+    <View {...rest} style={[styles.v3Container, style]}>
+      {actions.map((child, index) => (
+        <View
+          key={child.key ?? index}
+          style={[
+            index === actions.length - 1 ? styles.itemLast : styles.item,
+            child.props.style,
```
JKobrynski: `child.props.style` goes on the wrapper while the child still applies it itself, so
`<Button style={{ margin: 10 }}>` gets 10 twice. #5076 spaces the actions with a sibling `View` instead and avoids this — worth aligning, since it rewrites this file too.

src/components/Dialog/Dialog.tsx
```diff
@@ -32,9 +35,46 @@ export type Props = {
...
+  scrollAreaProps?: Omit<DialogScrollAreaProps, 'children'>;
+  /**
+   * Props passed to the `ScrollView` when `scrollable` is enabled.
+   */
+  scrollViewProps?: Omit<ScrollViewProps, 'children'>;
+  /**
+   * Accessibility label for the dialog. You can use it to override the defaults.
+   * By default if a `title` is a string then it's passed as an accessibility label.
+   */
+  accessibilityLabel?: string;
```
JKobrynski: Name it `aria-label`. That's what the library has used since #5005 — see `Button.tsx:83`. Same for `dialogAccessibilityLabel` on `Modal`.

src/components/Dialog/Dialog.tsx
```diff
@@ -121,18 +177,45 @@ const Dialog = ({
...
+          }}
+        >
+          {title}
+        </DialogTitle>
+      ) : null}
+
+      {scrollable ? (
+        <DialogScrollArea
+          {...scrollAreaProps}
+          style={{
```
JKobrynski:
```diff
-        style={{
-          ...(!title ? styles.firstChild : {}),
-          ...scrollAreaProps?.style,
-        }}
+        style={[!title && styles.firstChild, scrollAreaProps?.style]}
```

`style` is a `StyleProp`, so it can be an array - spreading it gives `{0: …, 1: …}` and the consumer's styles vanish. Same at line 209.

src/components/Dialog/DialogTitle.tsx
```diff
@@ -41,6 +42,13 @@ export type Props = React.ComponentPropsWithRef<typeof Text> & {
  *       </Dialog>
  *     </Portal>
  *   );
+ *
+ * 	 // V6 and later
```
JKobrynski: The "V6 and later" example doesn't use `Dialog.Title`. With `children` gone it has no call site left - do `Dialog.Title` and `Dialog.Icon` stay exported?

JKobrynski (changes requested):

src/components/Dialog/Dialog.tsx
kachmashk: Fixed!

src/components/Dialog/DialogTitle.tsx
kachmashk: Fixed.

src/components/Dialog/Dialog.tsx
kachmashk: Fixed as suggested!

src/components/Dialog/Dialog.tsx
kachmashk: Added `Dialog` section to the `migration.md`.

src/components/Dialog/DialogActions.tsx
kachmashk: Fixed it and applied #5076 implementation!

src/components/Modal.tsx
```diff
@@ -220,6 +225,7 @@ function Modal({
           theme={theme}
           style={[{ opacity }, styles.content, contentContainerStyle]}
           container
+          aria-label={ariaLabel}
```
JKobrynski, line 266: Add `role="dialog"`

src/components/Dialog/Dialog.tsx
```diff
@@ -121,18 +175,37 @@ const Dialog = ({
       ]}
       theme={theme}
       testID={testID}
+      aria-label={typeof title === 'string' ? title : ariaLabel}
```
JKobrynski: `aria-label` is ignored whenever `title` is a string, so it can't be overridden — the JSDoc says it is what the screen reader reads. Maybe we should use `ariaLabel ?? (typeof title === 'string' ? title : undefined)` instead?

docs/src/data/componentDocs6x.json
JKobrynski: Regenerated correctly now - all seven pages match the source. One example is still the old shape: `Dialog.tsx:114`, published at `Dialog.mdx:43`, shows `actions={[{ label: 'Done', onPress: hideDialog }]}`.

src/components/Modal.tsx
kachmashk: Added!

src/components/Dialog/Dialog.tsx
kachmashk: Yes, you're right! `aria-label` should be a top priority.

docs/src/data/componentDocs6x.json
kachmashk: Updated!

JKobrynski:
The MD3 corrections here should all land - divider colour, the 280/560 width clamp, the icon margin, the centred headline. What I'd like to reconsider is replacing `children` with `title`/`content`/`actions` props. [#5076](https://github.com/callstack/react-native-paper/issues/5076) solves the same problem while keeping the sub-components, which matches `Card`, `List` and `DataTable`. Could you drop the API change and keep the MD3 fixes? #5076 will need to land first either way, since the two touch the same files. @kachmashk

src/components/Dialog/Dialog.tsx
```diff
@@ -1,19 +1,24 @@
 import * as React from 'react';
-import { Platform, StyleSheet } from 'react-native';
-import type { StyleProp } from 'react-native';
+import { Platform, ScrollView, StyleSheet } from 'react-native';
+import type { ScrollViewProps, StyleProp } from 'react-native';
 
+import type {
+  DialogContentProps,
+  DialogScrollAreaProps,
+} from 'react-native-paper';
```
JKobrynski: Can these come from `./DialogContent` and `./DialogScrollArea` instead? The package importing itself here also lands in the published `Dialog.d.ts`.

docs/6.x/docs/guides/migration.md
```diff
@@ -280,3 +280,55 @@ const theme = {
   style={{ fontSize: 16, color: '#1C1B1F' }}
 />
 ```
+
+### Dialog
```
JKobrynski, line 278: There's already a `### Dialog` section at line 158, under `## Components` - this one landed at the end of the file after `TextInput`. Worth folding the two together?

docs/6.x/docs/guides/migration.md
```diff
@@ -280,3 +280,55 @@ const theme = {
...
+
+The new **`actions`** prop is more customizable because it accepts **`React.ReactNode[]`**. Unlike the previous `Dialog.Actions` children behavior, the dialog no longer injects **`compact`** and **`uppercase`** into action components automatically. When **`actions`** is provided, the passed components are rendered through **`Dialog.Actions`** internally.
+
+#### New props
+
+- **`scrollable`** renders the dialog content inside a scrollable area.
+- **`contentProps`** passes props to the internal `Dialog.Content` when `scrollable` is not enabled.
+- **`scrollAreaProps`** passes props to the internal `Dialog.ScrollArea` when `scrollable` is enabled.
+- **`scrollViewProps`** passes props to the internal `ScrollView` when `scrollable` is enabled.
+- **`aria-label`** provides an accessibility label when the dialog title is not a string.
```
JKobrynski: This is the other way round now - `aria-label` wins even when the title is a string. Maybe "overrides the title as the dialog's accessible name"?

src/components/Modal.tsx
```diff
@@ -259,6 +264,8 @@ function Modal({
           ]}
           elevation={contentElevation}
           transitionDuration={scale * DEFAULT_DURATION}
+          aria-label={ariaLabel}
+          role="dialog"
```
JKobrynski, line 267: Should `aria-modal` move down here too? It sits on the outer wrapper (line 230), which has no role, and [ARIA](https://www.w3.org/TR/wai-aria-1.2/#aria-modal) only defines it on `dialog` / `alertdialog`.

JKobrynski:
Also @kachmashk could you resolve the comments that've been addressed? 🙏

src/components/Dialog/Dialog.tsx
kachmashk: Done!

docs/6.x/docs/guides/migration.md
kachmashk: Done!

docs/6.x/docs/guides/migration.md
kachmashk: Yes, that's better!

src/components/Modal.tsx
kachmashk: Done!

kachmashk:
Waiting for a #5076 to be closed.

JKobrynski:
https://github.com/callstack/react-native-paper/pull/5094 removes docs from the repo, can you remove them from this PR? 🙏

JKobrynski:
@kachmashk ☝️ and we've got conflicts

k0ndee:
Heads up - #5125 is reworking `Modal`'s accessibility props: it replaces `overlayAccessibilityLabel` with `aria-label`, and adds a new configurable `role` prop (defaulting to `dialog`) directly on `Modal`'s `Surface`, rather than a hardcoded value. This PR already forwards `aria-label` from `Dialog` to `Modal`, but its `Modal` usage here still targets the old shape (`overlayAccessibilityLabel` + hardcoded `role="dialog"`), so it'll need a rebase once #5125 merges.

PR #5028, opened 2026-07-14
feat: add ConnectedButtonGroup component

matkoson:
### Motivation

The Material Design 3 spec deprecates segmented buttons in favor of the [connected button group](https://m3.material.io/components/button-groups/overview). This PR adds a `ConnectedButtonGroup` component implementing that spec as part of the ongoing modernization effort:

- Single- and multi-select selection models (same `buttons`/`value`/`onValueChange` API shape as `SegmentedButtons`, easing migration).
- Five sizes per the M3 button-group scale (`extra-small` → `extra-large`: 32/40/56/96/136dp) with per-size component tokens extracted into `tokens.ts`.
- Animated shape morphing driven by the theme motion tokens: selected buttons round to a full pill, the connected inner corners sharpen while pressed, and the group's outer silhouette stays fully rounded. The animation honours the reduce-motion setting.
- MD3 color roles: selected `secondaryContainer`/`onSecondaryContainer`, unselected `surfaceContainer`/`onSurfaceVariant`, disabled `onSurface` at 12% container / 38% content opacity (applied as styles so `PlatformColor` themes work).
- Accessibility: buttons expose the `radio` (single-select, inside a `radiogroup` row) or `checkbox` (multi-select) role with checked/disabled state; per-button `aria-label`, optional selected check icon, `hitSlop`/`background` passthroughs.
- Reuses the shared theme utilities (`resolveCornerRadius`/`ShapeToken`, `toRawSpring`, `TouchableRipple`'s `stateLayerPressed` ripple default).
- `SegmentedButtons` is marked `@deprecated` pointing to this component (non-breaking).
- Wired into the example app, docs config, theme-colors table, and a docs screenshot.

### Related issue

Fixes #4979

### Test plan

- `yarn typecheck`, `yarn lint` — clean.
- `yarn test ConnectedButtonGroup` — 23 tests + 4 render snapshots covering selection (single/multi), sizing, MD3 color roles, disabled behavior, selection check, custom colors, accessibility roles, and shape tokens.
- Example app → "Connected Button Group": verify tapping moves the selection pill (single-select), toggles accumulate (multi-select), sizes render at 32/40/56dp, and the disabled row is dimmed. Verified on an Android emulator (Pixel 9 Pro XL, API 36) and an iOS simulator (iPhone 16 Pro, iOS 18.3):

| Android | iOS |
|---|---|
| android | ios |

PR #5074, opened 2026-08-25
feat: add ConnectedButtonGroup component

oleksandrzavarzin-callstack:
### Motivation

Continues #5028. Addresses @alpharius-ck's review comments, and fixes two corner-animation bugs found while verifying those fixes on device: a button that could get stuck in the wrong shape, and a square flash on the middle button of a group.

### Related issue

Review comments on #5028.

### What changed and why

**Review comments:**
- Fixed the deprecation JSDoc link on `SegmentedButtons` and added a note on
  migrating `density` → `size`.
- Restored the check/icon swap animation to match `SegmentedButtonItem`.
- Extracted position literals into `connectedButtonPositions` and a
  `morphingCorners` lookup table.
- Hoisted `getTestID` out of the component.
- Extracted shared types into `types.ts`.

**Animation bugs:**
- Three separate handlers were animating one shared value, so a single tap ran
  three spring transitions instead of one. Fixed by deriving the radius from
  current props in a single `useDerivedValue`.
- Re-pressing an already-selected button never re-renders the group, so a
  swallowed `onPressOut` could strand the button in its pressed shape. Fixed
  by also clearing the pressed state on press.
- `cornerFull` is a sentinel value of `9999`, not a real radius — animating to
  it caused a large spring overshoot that flashed the corner square. Now
  resolved to `containerHeight / 2` instead.

The press-in morph itself isn't covered by a test: the reanimated jest mock
doesn't track shared-value mutations, so it's verified manually instead,
matching how `Switch.tsx` already handles this.

### Test plan

- `yarn typecheck`, `yarn lint` — clean.
- `yarn test ConnectedButtonGroup` — 28 tests (up from 23).
- Manually verified on an Android emulator: press-and-hold morphs the
  connected corner on both selected and unselected buttons; repeated taps on
  the selected button never strand the shape; the middle button no longer
  flashes square; deselecting a labelled button animates the icon back in.

https://github.com/user-attachments/assets/48a2f381-69be-497b-a7d0-3fb4ceba8c82


src/components/ConnectedButtonGroup/ConnectedButton.tsx
```diff
@@ -0,0 +1,296 @@
...
+    }),
+    [hasLabel]
+  );
+
+  // A gap is needed before the label, and between the check and the leading
+  // icon when an icon-only button shows both.
+  const iconGap =
+    label || (showCheck && showIcon)
+      ? { marginEnd: sizeStyle.iconLabelGap }
+      : null;
```
MikitasK: could we use separate gaps for check & leading icon? 
for a selected icon-only button, both elements render & both receive `iconGap` ([source](https://github.com/callstack/react-native-paper/blob/78a5b512f0db17ad3c591ef0fd00f92eb0f964c2/src/components/ConnectedButtonGroup/ConnectedButton.tsx#L230-L246)), leaving an extra trailing margin & shifting the pair off-center

```suggestion
  const checkGap = label || (showCheck && showIcon) ? { marginEnd: sizeStyle.iconLabelGap } : null;
  const leadingIconGap = label ? { marginEnd: sizeStyle.iconLabelGap } : null;
```

src/components/ConnectedButtonGroup/ConnectedButtonGroup.tsx
```diff
@@ -0,0 +1,269 @@
...
+  style?: StyleProp<ViewStyle>;
+  /**
+   * testID to be used on tests.
+   */
+  testID?: string;
+  /**
+   * @optional
+   */
+  theme?: ThemeProp;
+} & ConditionalValue<T>;
```
MikitasK, line 165: wdyt about making selection props visible in generated docs? because they come from  `ConditionalValue<T>` intersection, the generated data lists only `buttons`, `size`, `style`, `testID`, `theme` ([source](https://github.com/callstack/react-native-paper/blob/78a5b512f0db17ad3c591ef0fd00f92eb0f964c2/docs/src/data/componentDocs6x.json#L4674-L4864)); required `value`/`onValueChange` & `multiSelect` are missing

it would be great to make docs parser flatten intersection props & regenerate the docs. or at least add  generated “Selection props” section documenting these three fields

docs/5.x/docs/components/ConnectedButtonGroup/ConnectedButtonGroup.mdx
MikitasK, line 1: what about keeping this page in the 6.x docs only? `ConnectedButtonGroup` isn't exported by [v5.15.3](https://github.com/callstack/react-native-paper/blob/v5.15.3/src/index.tsx), while 5.x prop tables read `componentDocs5x.json`, which has no entry for it. this advertises unavailable import & renders empty prop details

so could we remove 5.x page/meta entry & exclude `ConnectedButtonGroup` from `getVersionComponentOrder('5.x')` ?

oleksandrzavarzin-callstack:
@MikitasK fixed your comments.
Please let me know if you have any other suggestions.

MikitasK (approved):
LGTM ✅

oleksandrzavarzin-callstack:
@JKobrynski could you please review this Pull Request?

PR #5039, opened 2026-08-03
refactor(appbar): modernize Appbar to latest MD3 spec

matkoson:
## Summary

Modernizes the Appbar components to align with the latest Material Design 3 (MD3) App bar specification.

### Changes
- **Appbar / AppbarHeader / AppbarContent / AppbarAction**: MD3 token-driven layout, elevated/flat modes, and updated spacing/typography behavior
- **New `tokens.ts`**: shared MD3 App bar tokens for consistent theming
- **Example**: updated `AppbarExample` to demonstrate current MD3 patterns
- **Docs (6.x)**: Appbar component docs + `componentDocs6x` / theme color data updates
- **Tests**: expanded unit coverage and snapshot updates for the new API/behavior

### Out of scope
- No private/self-hosted CI workflows (kept on private fork only)

## Test plan
- [ ] Appbar unit tests pass
- [ ] Example app Appbar screens look correct on iOS and Android
- [ ] Docs preview for 6.x Appbar pages
- [ ] Visual check against MD3 App bar spec

Issue #4934, opened 2026-05-11
refactor(appbar): improve MD3 variants and action composition


ruben-rebelo:
Refactor AppBar to improve Material Design 3 compliance, remove problematic children patterns, improve performance characteristics, and modernize the public API.

**Problems**

- Current API relies heavily on children composition in ways that are difficult to optimize
- Internal layout logic is tightly coupled to arbitrary child structures
- MD3 behaviors and visuals are not fully aligned with current Material guidelines
- Rendering performance can degrade in large lists or frequently updated screens
- Type safety and discoverability of the API can be improved
- The component has inconsistent behavior between platforms and edge cases

**Focus Areas**

- Improve support for MD3 top app bar variants
- Simplify action rendering
- Improve scroll behavior integration
- Improve accessibility and semantics

**Proposed API Direction**

`<Appbar
  variant="small"
  title="Inbox"
  actions={[...]}
/>`

**Performance Opportunities**

- Reduce action re-renders
- Improve animated scroll performance
- Reduce style recalculation

adrcotfas:
> [!IMPORTANT]  
See https://github.com/callstack/react-native-paper/discussions/4949 for more details

PR #5075, opened 2026-08-25
feat: Appbar MD3 Expressive improvements

pawelkata:
### Motivation

Replace the legacy compound Appbar API with a prop-driven Material 3 implementation. This adds small, flexible, and search variants; filled actions; scroll-aware colors; safe-area handling; and improved accessibility. Documentation, examples, and downstream usages are updated accordingly.

### Related issue

Related issue: https://github.com/callstack/react-native-paper/issues/4934

Also based on comments on Appbar from here: https://github.com/callstack/react-native-paper/discussions/4949#discussion-10070130 .

### Caveats

The search bar is dropped in verbatim as it needs a separate pool of work to properly handle the "search" variant. Deliberately skipped any animations related work at this stage due to lack of testing device and the PR already being quite big. Can be tackled at a later stage.

NOTE: The PR drops support for the bottom app bar variant, which is now - according to the official docs - part of the `Toolbars`.

### Test plan

- Run `yarn test src/components/__tests__/Appbar`.
- Run `yarn lint` and `yarn typecheck`.
- Verify all Appbar variants in the example app, including search, scrolling, centered headlines, actions, and safe areas.
- Verify accessibility labels, element order, disabled states, and headline interactions with a screen reader.


src/components/Appbar/types.ts
```diff
@@ -0,0 +1,232 @@
...
+  style?: Animated.WithAnimatedValue<StyleProp<ViewStyle>>;
+  ref?: React.Ref<View>;
+  theme?: ThemeProp;
+};
+
+type AppbarHeadlineProps = {
+  /** Headline and subtitle alignment. */
+  headlineAlignment?: AppbarHeadlineAlignment;
+  /** Trailing actions. */
+  trailingActions?: AppbarTrailingActions;
```
JKobrynski, line 186: `trailingActions` takes config objects, so nothing can wrap an action any more. The examples in this PR show the cost - `TooltipExample` drops four `Tooltip`s, and `MenuExample`, `BottomNavigationExample` and `react-navigation.md` all switch `Menu` from an element anchor to raw coordinates. Composition is what #4954 set out to fix.

src/components/Appbar/types.ts
```diff
@@ -0,0 +1,232 @@
...
+  headlineImage?: never;
+  headlineAlignment?: never;
+  headlineProps?: never;
+  subtitleProps?: never;
+  onHeadlinePress?: never;
+  headlinePressableProps?: never;
+  contentStyle?: never;
+};
+
+export type Props = AppbarBaseProps &
```
JKobrynski, line 229: Running the docs generator produces four props for `Appbar` - `headlineAlignment`, `isScrolled`, `testID`, `trailingActions` - with no types and no descriptions. `headline`, `variant`, `leadingButton`, `searchBar` and the rest are missing because react-docgen can't read the union, so the published prop table is effectively empty.

src/components/Appbar/types.ts
```diff
@@ -0,0 +1,232 @@
...
+  | 'style'
+> & {
+  /** Optional leading button. */
+  leadingButton?: AppbarLeadingButton;
+  /** Uses the on-scroll container color when true. */
+  isScrolled?: boolean;
+  /** Override for the automatic top safe-area inset. */
+  statusBarHeight?: number;
+  /** Safe-area inset overrides. Unspecified values use the detected insets. */
+  safeAreaInsets?: {
```
JKobrynski, line 167: `safeAreaInsets.bottom` is gone and `children` is no longer accepted, so the documented bottom app bar can't be built any more. Deliberate? The spec's replacement is the docked toolbar, which isn't here either.

src/components/Appbar/AppbarButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  AppbarButton,
+  (previous, next) =>
+    previous.leading === next.leading &&
+    previous.theme === next.theme &&
+    hasEqualButtonProps(previous.button, next.button)
+);
+
+const styles = StyleSheet.create({
+  button: {
+    margin: 4,
```
JKobrynski: `margin: 4` on every button plus the bar's own `paddingHorizontal: 4` gives 8dp to the edge and 8dp between actions. `md.comp.app-bar.leading-space` and `trailing-space` are 4dp and `icon-button-space` is 0dp, so the icons land 20dp from the edge instead of 16dp.

src/components/Appbar/Appbar.tsx
```diff
@@ -1,340 +1,451 @@
...
-        {
-          height: modeAppbarHeight[mode],
-        },
-        insets,
-        restStyle,
-      ]}
-      elevation={elevation}
+      ref={ref}
+      testID={`${testID}-root-layer`}
+      elevation={0}
```
JKobrynski, line 382: > `md.comp.app-bar.on-scroll.container.elevation` resolves to `level2`, but `elevation` is pinned to `0` - `isScrolled` only changes the container colour. Intentional?

src/components/Appbar/Appbar.tsx
```diff
@@ -1,340 +1,451 @@
...
+    ],
+    [backgroundColor, borderRadius, horizontalInset, topInset]
+  );
+  const appbarStyle = React.useMemo(
+    () => [styles.appbar, { backgroundColor, minHeight }, restStyle],
+    [backgroundColor, minHeight, restStyle]
   );
+  const resolvedSearchInputStyle = React.useMemo(
+    () => [
+      searchBar?.inputStyle,
```
JKobrynski: `searchBar.inputStyle` comes before `styles.searchInput`, so a consumer can't override `textAlign: 'center'`. Put the user's style last. Why is the search text centred at all?

src/components/Appbar/types.ts
```diff
@@ -0,0 +1,232 @@
...
+  headlineProps?: never;
+  subtitleProps?: never;
+};
+
+type AppbarFlexibleImageHeadline = AppbarWrittenHeadline &
+  AppbarHeadlineImage & {
+    variant: Exclude<AppbarHeadlineVariant, 'small'>;
+  };
+
+type AppbarBaseProps = Omit<
```
JKobrynski, line 156: Why omit `role`, `aria-label` and `accessible`? Consumers can't label the app bar region any more.

src/components/Appbar/AppbarContent.tsx
```diff
@@ -1,192 +1,216 @@
...
+    ],
+    [centered, contentStyle, style, variant]
+  );
 
-  const content = (
+  const content = hasHeadlineImage ? (
+    <View
+      testID={`${testID}-headline-image`}
+      aria-hidden
+      importantForAccessibility="no-hide-descendants"
```
JKobrynski: Drop `importantForAccessibility` - `aria-hidden` already covers it, and #5005 migrated the library off the legacy props.

src/components/Appbar/AppbarButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+      selected={mode ? true : undefined}
+      size={24}
+      animated
+      style={[styles.button, width === 'wide' && styles.wideButton, style]}
+      theme={theme}
+    />
+  );
+};
+
+const hasEqualButtonProps = (
```
JKobrynski: Remove `hasEqualButtonProps`. It's a hand-rolled shallow compare with two `as unknown as` casts and no measurement behind it.

src/components/Appbar/Appbar.tsx
```diff
@@ -1,340 +1,451 @@
...
+  isScrolled = false,
+  leadingButton,
+  onHeadlinePress,
   safeAreaInsets,
+  searchBar,
+  statusBarHeight,
+  style,
+  subtitle,
+  subtitleProps,
+  testID = 'appbar',
```
JKobrynski: Drop the default `testID`.

JKobrynski (changes requested):

src/components/Appbar/Appbar.tsx
pawelkata: Dropped.

src/components/Appbar/Appbar.tsx
pawelkata: Great catch. Fixed.

> Why is the search text centred at all?

It should not be. I completely missed the two instances of left aligned search text in the MD3 docs. Removed the style.

src/components/Appbar/Appbar.tsx
pawelkata: I'm probably missing something here, but... according to MD3 guidelines here: https://m3.material.io/components/app-bars/guidelines and the Figma kit, there's no elevation in appbars -- just the color indication that content was scrolled.

JKobrynski:
@pawelkata can you resolve the comments that have been addressed? 🙏

src/components/Appbar/types.ts
pawelkata: Right. Brought back.

src/components/Appbar/AppbarContent.tsx
pawelkata: Dropped.

src/components/Appbar/types.ts
pawelkata: There is a clear separation between top and bottom bars outlined in MD3 docs, hence I decided to drop bottom bar support in the PR. The Toolbar is, in my opinion, out of the scope of these adjustments and should be tackled as a separate task. I'll adjust the documentation to clearly state that.

src/components/Appbar/AppbarButton.tsx
pawelkata: Very true, removed.

adam-sajko:
`docs/6.x/docs/guides/migration.md` has no entry for this yet, and dropping `Appbar.Header`, `Appbar.Content`, `Appbar.Action` and `Appbar.BackAction` is worth mentioning in "before" and "after" there.

src/components/Appbar/AppbarButton.tsx
pawelkata: Yeah, I might have been too early with this one. Removed.

src/components/Appbar/types.ts
pawelkata: Solved this using a decorator prop. Let me know what you think.

src/components/Appbar/types.ts
pawelkata: Docs adjusted.

src/components/Appbar/types.ts
pawelkata: Improved the docs.

pawelkata: Docs are fixed and updated.

src/components/Appbar/Appbar.tsx
JKobrynski: You're right, resolving.

JKobrynski (changes requested):
The docs build fails on this PR, there is also a bunch of conflicts. There are some comments I'd like to post, but can you first address these conflicts and the failing check? I think some of my concerns will be resolved in process.

pawelkata:
> The docs build fails on this PR, there is also a bunch of conflicts. There are some comments I'd like to post, but can you first address these conflicts and the failing check? I think some of my concerns will be resolved in process.

@JKobrynski ~~Rebased~~, docs generate and show up fine on my end.

NVM, this needs another rebase... I'll get to it asap.

JKobrynski:
@matkoson @pawelkata you're both rewriting Appbar, and 11 files overlap, including every component and the snapshot. Could you agree who takes it? Worth noting that #5039 still edits the generated docs that #5094 deleted.

matkoson:
@JKobrynski @pawelkata I've closed https://github.com/callstack/react-native-paper/pull/5039. Unfortunately, I won't be able to work on it in the nearest future. Feel free to borrow any code from there if it turns out to be useful here. Cheers.

PR #5076, opened 2026-08-25
refactor: replace React.Children usage with context and typed props

adam-sajko:
### Motivation

Continues #5018 by @matkoson, with the review feedback applied.

`React.Children` together with `cloneElement` locks consumers into a fixed child
shape. A child wrapped in anything — a `Tooltip`, a layout `View`, a component of
your own — silently stops receiving the props the parent meant to inject. This PR
replaces the pattern with context, typed props and container-owned layout in
`Card`, `Dialog` and `List`.

Appbar is deliberately out of scope: #4934 proposes an `actions` prop API and
#5075 implements it, so a children refactor there would be thrown away.

### Related issue

#4989, continues #5018

### Test plan

Open the Card, Dialog and List examples and check each section renders, including
children wrapped in your own component.


docs/6.x/docs/guides/migration.md
```diff
@@ -126,3 +126,71 @@ const theme = {
...
+
+// After (v6)
+<Dialog.Actions>
+  <Button compact uppercase onPress={hide}>
+    Done
+  </Button>
+</Dialog.Actions>
+```
+
+### List
```
JKobrynski, line 354: No `### ToggleButton` section, but `ToggleButton.Row` changes visibly - divider ownership moves to the row and `alignSelf: 'flex-start'` is new. `Card`, `Dialog` and `List` all got one.

adam-sajko: nice catch, done

src/components/__tests__/Tooltip.test.tsx
```diff
@@ -163,7 +163,7 @@ describe('Tooltip', () => {
       it('hides the tooltip when the user stop pressing the component', async () => {
         const {
           wrapper: { queryByText, getByText, findByText },
-        } = await setup({ enterTouchDelay: 50, leaveTouchDelay: 0 });
+        } = await setup({ enterTouchDelay: 50, leaveTouchDelay: 100 });
```
adam-sajko, line 166: Stolen from #4996 (CI fix)

JKobrynski (approved):
LGTM! @satya164 all yours

JKobrynski:
@adam-sajko we've got conflicts

src/components/Card/CardActions.tsx
```diff
@@ -43,26 +44,20 @@ const CardActions = ({ theme, style, children, ...rest }: Props) => {
...
+        <React.Fragment
+          key={
+            React.isValidElement(child) && child.key != null ? child.key : index
+          }
+        >
+          {index > 0 && <View style={styles.spacer} />}
+          {child}
+        </React.Fragment>
+      ))}
     </View>
```
satya164: this is still using `React.Children` API and inspecting children. why is it necessary? if it's just for spacer, can't it use `columnGap`?

src/components/Card/CardContent.tsx
```diff
@@ -42,59 +32,18 @@ export type Props = ViewProps & {
...
 
 const styles = StyleSheet.create({
   container: {
     paddingHorizontal: 16,
-  },
-  first: {
     paddingTop: 16,
-  },
-  last: {
     paddingBottom: 16,
```
satya164: could just be simplified to `padding`

src/components/Card/Card.tsx
```diff
@@ -102,6 +102,8 @@ export type Props = Omit<ViewProps, 'style'> & {
 
 /**
  * A card is a sheet of material that serves as an entry point to more detailed information.
+ * Card clips its inner content to the card shape and renders children directly;
+ * section spacing is owned by the section components themselves.
```
satya164: remove this comment or rewrite it in human voice. "Card clips its inner content" is very strange phrasing. a more natural sentence would be "The contents of Card are clipped by default".

also generally this JSDoc only describes what the component is. anything else should be a separate paragraph, not combined into the same paragraph as description.

src/components/Card/CardActions.tsx
```diff
@@ -17,6 +16,8 @@ export type Props = ViewProps & {
 
 /**
  * A component to show a list of actions inside a Card.
+ * Actions are rendered directly, so set button `mode`, `compact`, and custom
+ * spacing props explicitly on each action when needed.
```
satya164: "Actions are rendered directly" doesn't tell me much, someone looking at the component the first time won't understand what this means.

"so set button `mode`, `compact`, and custom spacing props explicitly on each action when needed." is also vague, reading this doesn't give a clear idea of what to do.

either remove this sentence or rewrite it from the point of someone who is learning how to use the component, and move to a separate paragraph instead of combining with component description.

src/components/Card/CardContent.tsx
```diff
@@ -7,23 +7,13 @@ export type Props = ViewProps & {
...
-   * @internal
-   */
-  siblings?: Array<string | null>;
   style?: StyleProp<ViewStyle>;
 };
 
 /**
  * A component to show content inside a Card.
+ * Content uses uniform vertical padding and does not depend on neighboring
+ * card sections.
```
satya164: documentation shouldn't be mentioning historical details. if you wrote this component from scratch today, would you add this sentence? if not, then it doesn't belong here.

src/components/Dialog/Dialog.tsx
```diff
@@ -52,6 +51,8 @@ const DIALOG_ELEVATION: Elevation = 3;
 /**
  * Dialogs inform users about a specific task and may contain critical information, require decisions, or involve multiple tasks.
  * To render the `Dialog` above other components, you'll need to wrap it with the [`Portal`](../Portal) component.
+ * Dialog owns the top content inset, so first-slot components render without
+ * adding their own top offset.
```
satya164: try to read this see if you can understand what it says. i don't.

please don't blindly add documentation written by LLMs. they are very bad at it.

remove this sentence, or rewrite to something actionable for the user when using the component, and move to separate paragraph after description.

src/components/Dialog/DialogActions.tsx
```diff
@@ -20,6 +19,8 @@ export type Props = ViewProps & {
 
 /**
  * A component to show a list of actions in a Dialog.
+ * Actions are rendered directly, so configure each action button's props
+ * explicitly when you need non-default behavior.
```
satya164: same comment as card actions

src/components/Dialog/DialogActions.tsx
```diff
@@ -46,26 +47,28 @@ export type Props = ViewProps & {
...
+        <React.Fragment
+          key={
+            React.isValidElement(child) && child.key != null ? child.key : index
+          }
+        >
+          {index > 0 && <View style={styles.spacer} />}
+          {child}
+        </React.Fragment>
+      ))}
     </View>
```
satya164: same as card actions, this is still relying on `React.Children` API

src/components/List/ListAccordion.tsx
```diff
@@ -324,23 +325,11 @@ const ListAccordion = ({
...
-                style: [styles.child, child.props.style],
-                theme,
-              });
-            }
-
-            return child;
-          })
-        : null}
+      {isExpanded ? (
+        <ListAccordionContext.Provider value={{ leftIndent: !!left }}>
```
satya164: Don't inline objects for context provider. Memoize it.

src/components/List/ListAccordionContext.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+
+export type ListAccordionContextType = {
+  /**
+   * Whether descendant items that don't render their own `left`/`right`
+   * element should be indented to align under the accordion's content
+   * (past the leading icon).
+   */
+  leftIndent: boolean;
```
satya164: it'd be more natural to expose "what the accordion does", then the child adjusts itself based on what the accordion does, instead of a parent context having assumptions about how the child looks like and what it should do.

e.g.: accordion has left item, so child can indent itself if necessary based on the boolean

```suggestion
  hasLeft: boolean;
```

src/components/ToggleButton/ToggleButtonRow.tsx
```diff
@@ -44,33 +52,33 @@ export type Props = {
...
+            },
+            style,
+          ]}
+        >
+          {children}
+        </View>
+      </ToggleButtonRowContext.Provider>
     </ToggleButtonGroup>
   );
 };
```
satya164, line 85: this looks broken

**before**

Image

**after**

Image

satya164 (changes requested):
the card spacing also looks broken

Image

when these components were written initially, there was no `gap` style. today i think many of the styling can move to `gap` and keep same visual style without needing to read children API. the card items themselves likely don't need padding inside (unless they contain multiple items inside) and can let the card itself apply gap to space them properly.

same principle for dialog

src/components/Dialog/Dialog.tsx
adam-sajko: Hey @satya164, I'll look into this. However, I just took over the PR, so those changes are not mine and I didn't touch them intentionally :D My commits contain only small changes with no code comments at all. But I'll go through the entire PR and do some cleanup

src/components/Card/CardContent.tsx
adam-sajko: Not my change, but I addressed it, thanks!

src/components/Card/CardActions.tsx
adam-sajko: Not my change, but I addressed it, thanks!

src/components/Card/Card.tsx
adam-sajko: Not my change, but I addressed it, thanks!

src/components/Card/CardContent.tsx
adam-sajko: done

src/components/Card/CardActions.tsx
adam-sajko: done

src/components/Dialog/DialogActions.tsx
adam-sajko: Not my change, but I addressed it, thanks!

src/components/Dialog/DialogActions.tsx
adam-sajko: done

src/components/List/ListAccordion.tsx
adam-sajko: done

src/components/List/ListAccordionContext.tsx
adam-sajko: Not my change, but I addressed it, thanks!

src/components/ToggleButton/ToggleButtonRow.tsx
adam-sajko: Fixed

| Web | iOS | Android |
| --- | --- | --- |
| Screenshot 2026-09-15 at 11 32 20 | Simulator Screenshot - iPhone 17 Pro - 2026-09-15 at 11 35 51 | Screenshot_1789465886 |

satya164: it still has some issues:

- there's a border, but it's not visible at corners
- there shouldn't be a border in general, looking at the vertical example (and previous screenshot)

CleanShot 2026-09-15 at 14 01 39@2x

satya164: toggle button is removed on main, so resolving this

satya164:
I have reworked the Card implementation. Please take a look and see if there are any potential issues.

adam-sajko:
@satya164 
> I have reworked the Card implementation. Please take a look and see if there are any potential issues.

Thanks for picking this up! I think the `gap` direction reads much cleaner than the sibling inspection we had before.

I generated a comparison image for the web (by using a chrome screenshot plugin and an AI prompt to merge both screenshots).

One note on the comparison image below. I saw you added in the migration guide:

`If you have Card.Cover or Card.Actions in positions other than the top and bottom respectively, you will need to adjust their margins.`

So I also added two extra examples in the screenshots, where I put the title on top (before the cover) and the actions before the content, to confirm they need negative margins. I'm not saying this is a valid example, but I assume consumers can arrange it however they want, right? So, they might be surprised. It works for now, but it's a bit of a workaround. And usually people don't read migration guidelines. Especially for people who have just started using it; they have no reason to read it.

Here are two additional examples in the screenshot I attached:
```
        <Card style={styles.card} mode={selectedMode}>
          <Card.Title title="NEW EXAMPLE - Title before the cover" />
          <Card.Cover source={require('../../assets/images/forest.jpg')} />
        </Card>
        <Card style={styles.card} mode={selectedMode}>
          <Card.Actions>
            <Button mode="contained" onPress={() => {}}>
              Action
            </Button>
          </Card.Actions>
          <Card.Content>
            <Text variant="bodyMedium">
              NEW EXAMPLE - Actions before the content. Actions before the
              content. Actions before the content. Actions before the content.
            </Text>
          </Card.Content>
        </Card>
```

And the comparison image (screenshots merged, `main` on the left, `refactor/react-children-takeover` on the right):
card-comparison-main-vs-new

satya164:
> I assume consumers can arrange it however they want, right? So, they might be surprised.

They can, but we can't automatically handle all use cases. So my goal was to handle the common use cases by default and enable users to tweak for less common use case. For example, action buttons are going to be at the bottom 99% of the time.

The UI was also broken before these changes as the spacing between items were too big or too small. So the choice is between having users always need to change things to make it work correctly, or handle common cases automatically when possible, and only require manual work otherwise.

> It works for now, but it's a bit of a workaround.

I wouldn't say workaround to describe a documented approach.

Besides, we can always add APIs or props that read nicer in the future instead of asking the user to specify margins.

Initially I added props for this, but it's better to keep it out of scope from this PR and consider separately.

> And usually people don't read migration guidelines. Especially for people who have just started using it; they have no reason to read it.

All of these will be in the regular documentation. The 6.x docs will need a proper review and update once all the big changes are settled.

adam-sajko:
> All of these will be in the regular documentation. The 6.x docs will need a proper review and update once all the big changes are settled.

@satya164 I'm ok with that. All good then!

PR #5002, opened 2026-06-16
feat(chip): modernize MD3 API

draggie:
### Motivation

  Refactors `Chip` toward the v6/MD3 API and implementation direction.

  This change introduces an explicit `label` prop, extracts Chip-specific tokens, simplifies the internal layout, and removes press-time elevation
  animation overhead. It also updates selected, outlined, disabled, icon/avatar, close icon, touch target, and state layer/ripple handling to better align
  with Material Design 3.

  The example app has been updated to showcase the new API and MD3-oriented Chip states.

  ### Related issue

  Closes #4931

  ### Test plan

  - Verified Chip examples in the example app.
  - Verified Android example build:
    - `cd example/android`
    - `./gradlew app:assembleDebug -x lint -x test --configure-on-demand --build-cache -PreactNativeDevServerPort=8081 -PreactNativeArchitectures=arm64-
    v8a`
  - Ran TypeScript:
    - `yarn typescript`
  - Ran lint:
    - `yarn lint-no-fix`
  - Ran tests:
    - `yarn test --runInBand --watchman=false`
  - Ran docs build:
    - `yarn docs build`

simulator_screenshot_B01F025C-C346-425D-A08F-CB46EEE951BA
Screenshot_1781595724

PR #5078, opened 2026-08-26
refactor: rework Surface and elevation shadows

satya164:
### Motivation

This reworks the `Surface` component to work with Reanimated + updates components that use `Surface`.

Previously, since the component rendered outer and inner views, it flattened styles and picked where to apply what. Style flattening adds challenges when using Reanimated, as Reanimated relies on metadata from `useAnimatedStyle` for UI-thread updates.

After trying a couple of approaches, I arrived at this approach:

- Single `View` for `Surface` so it can accept both position and layout related styles
- Inner `View` for second shadow layer, but instead of wrapping `children` in it (which adds more challenges), we absolutely position it behind the content
- Background color and border radius are added as props since they need to be applied to inner view, and props make it simple without needing to flatten styles

#### Before

appbar-android-before

#### After

appbar-android-after

#### Before

appbar-ios-before

#### After

appbar-ios-after

#### Before

appbar-web-before

#### After

appbar-web-after

#### Before

badge-android-before

#### After

badge-android-after

#### Before

badge-ios-before

#### After

badge-ios-after

#### Before

badge-web-before

#### After

badge-web-after

#### Before

https://github.com/user-attachments/assets/00bd9a3c-3a48-4f2a-908e-63946c998d38

#### After

https://github.com/user-attachments/assets/3cf952f0-20f7-4aea-b28e-51c461219d6e

#### Before

https://github.com/user-attachments/assets/0d8cae72-6c01-49af-b697-0030a2cff1b7

#### After

https://github.com/user-attachments/assets/9d885cf0-2b97-4667-8198-b23ad42ff4e4

#### Before

https://github.com/user-attachments/assets/3e726859-2435-4e2b-b8d6-ebf022ab6bf2

#### After

https://github.com/user-attachments/assets/5f046c51-e892-4e8b-8788-eea975e9e008

#### Before

https://github.com/user-attachments/assets/984a10a4-1007-4fd4-8037-3e998abc3dfb

#### After

https://github.com/user-attachments/assets/15089dff-49ed-41a2-8ba7-c849216f4f66

#### Before

https://github.com/user-attachments/assets/046f7303-4068-4404-9057-96dbdfaa7865

#### After

https://github.com/user-attachments/assets/5406ba4f-c32e-40c0-b8e3-14e6f159edbc

#### Before

https://github.com/user-attachments/assets/3ef3a30e-4cb5-4ee4-a1d0-03760af5257c

#### After

[...]

PR #5081, opened 2026-08-27
feat: implement Chip component

### LINKED ISSUE #4931: refactor(chip): improve selection model and MD3 compliance
Refactor Chip to improve Material Design 3 compliance, remove problematic children patterns, improve performance characteristics, and modernize the public API.

**Problems**

**Focus Areas**

- Improve selected/outlined/filter states
- Improve icon/avatar handling
- Simplify content rendering
- Improve touch target sizing
- Improve state layer implementation

**Proposed API Direction**

Move toward explicit label APIs:

`<Chip label="Filter" selected />`

**Performance Opportunities**

- Reduce animation overhead
- Avoid child inspection
- Improve memoization
- Simplify icon rendering path

k0ndee:
### Motivation

Refactors `Chip` toward the v6/MD3 API and implementation direction.

- Extracts `Chip`-specific sizing, color, and typescale values into a dedicated `tokens.ts` file, following the pattern used by other recently modernized components (e.g. `Badge`).
- Simplifies the internal layout: the ripple/state-layer is now rendered in its own layer behind the content instead of being nested inside it, and content spacing is derived from tokens instead of inline math.
- Removes the press-time elevation animation (`Animated.timing` driven scale/elevation on press) in favor of a static elevation derived from the `elevated` and `disabled` props.
- Reworks selected, outlined, disabled, icon/avatar, and close-icon color resolution (`getChipColors` in `utils.ts`, renamed from `helpers.tsx`) to better align with MD3 state and color roles, including a dedicated ripple color and an avatar selected-overlay color.
- Adds a proper minimum touch target (48dp) via `hitSlop`, and moves the close icon into `TouchableRipple` for consistent ripple/state-layer feedback.
- Removes the `compact` and `showSelectedOverlay` props, which are no longer part of the MD3-aligned API (the selected overlay is now handled automatically for avatars).
- Changes the default `mode` from `flat` to `outlined`.

The example app has been updated to match the new API (e.g. dropping now-removed `showSelectedOverlay` usage) and to showcase MD3-oriented Chip states.

### Related issue

Closes #4931

### Related PRs

Based on the #5002 
Wait for #5084 to be merged to take advantage of the focus indicators in TouchableRipple

### Test plan

- Verified Chip examples in the example app.
- Ran typecheck:
  ```
  yarn typecheck
  ```
- Ran lint:
  ```
  yarn lint
  ```
- Ran tests:
  ```
  yarn test --runInBand --watchman=false
  ```
- Ran docs build:
  ```
  yarn docs build
  ```

Screenshot_1788266249


src/components/Chip/Chip.tsx
```diff
@@ -300,193 +297,184 @@ const Chip = ({
...
-                  ? [
-                      styles.avatar,
-                      styles.avatarSelected,
-                      selected && styles.md3SelectedIcon,
-                    ]
-                  : null,
-              ]}
-            >
-              {icon ? (
+        <View style={styles.rippleContent} />
```
JKobrynski: Try keeping the label inside `TouchableRipple`, or deriving `aria-label` from it - as a `pointerEvents="none"` sibling it never reaches the button, so `getByRole('button', { name: 'Example Chip' })` resolves on `main` and returns `null` here. Worth a test to pin it.

src/components/Chip/tokens.ts
```diff
@@ -0,0 +1,45 @@
...
+  flatContainerColor: 'surfaceContainerLow',
+  selectedContainerColor: 'secondaryContainer',
+  outlinedContainerColor: 'surface',
+  labelColor: 'onSurfaceVariant',
+  selectedLabelColor: 'onSecondaryContainer',
+  leadingIconColor: 'primary',
+  selectedIconColor: 'onSecondaryContainer',
+  trailingIconColor: 'onSurfaceVariant',
+  selectedTrailingIconColor: 'onSecondaryContainer',
+  outlineColor: 'outline',
```
JKobrynski: Was `outline` intentional here? `main` uses `outlineVariant` (`helpers.tsx:48`), and all four live chip namespaces resolve to `outline-variant` — the only token giving `outline` is `md.comp.input-chip.outline.color`, which is deprecated.

src/components/Chip/Chip.tsx
```diff
@@ -300,193 +297,184 @@ const Chip = ({
...
+          background={trailingIconAndroidRipple}
+          rippleColor={rippleColor}
+          style={[
+            styles.closeButton,
+            disabled ? { opacity: contentOpacity } : null,
+          ]}
+          theme={theme}
+        >
+          <Icon
+            source={hasTrailingIcon ? trailingIcon! : (closeIcon ?? 'close')}
```
JKobrynski: Try dropping the `!` - it typechecks without one, and `@typescript-eslint/no-non-null-assertion` is error-level on `main` (`eslint.config.mjs:150`).

src/components/Chip/Chip.tsx
```diff
@@ -300,193 +297,184 @@ const Chip = ({
...
+          borderless
+          centered
+          onPress={hasTrailingIcon ? onTrailingIconPress : onClose}
+          disabled={disabled}
+          role="button"
+          aria-label={
+            hasTrailingIcon
+              ? trailingIconAccessibilityLabel
+              : closeIconAccessibilityLabel
+          }
```
JKobrynski, line 445: Worth giving `trailingIconAccessibilityLabel` a default, or requiring it alongside `onTrailingIconPress` - `<Chip trailingIcon="menu-down" onTrailingIconPress={…}>` currently renders a `role="button"` with no accessible name, where the close button defaults to `'Close'`.

src/components/Chip/Chip.tsx
```diff
@@ -300,193 +297,184 @@ const Chip = ({
...
+        </TouchableRipple>
       ) : null}
     </Surface>
   );
 };
 
 const styles = StyleSheet.create({
   container: {
-    borderWidth: StyleSheet.hairlineWidth,
+    height: ChipTokens.containerHeight,
```
JKobrynski: Try `minHeight` here - with `overflow: 'hidden'` on the content (line 438) a fixed `height` clips the label at larger system font sizes. `main` reached the same 32dp via `lineHeight: 20` plus `marginVertical: 6`, so it scaled.

```diff
-    height: ChipTokens.containerHeight,
+    minHeight: ChipTokens.containerHeight,
```

src/components/Chip/Chip.tsx
```diff
@@ -170,7 +171,7 @@ export type Props = $Omit<React.ComponentProps<typeof Surface>, 'mode'> & {
  * ```
  */
 const Chip = ({
-  mode = 'flat',
+  mode = 'outlined',
```
JKobrynski, line 199: Worth adding a Chip section to [`migration.md`](https://github.com/callstack/react-native-paper/blob/main/docs/6.x/docs/guides/migration.md) - `compact` and `showSelectedOverlay` are gone and this default flip changes every existing `<Chip>`. The `TextInput` entry is a template, and the generated docs are already done.

src/components/Chip/tokens.ts
```diff
@@ -0,0 +1,45 @@
...
+  disabledColor: 'onSurface',
+} as const satisfies Record<string, ColorRole>;
+
+export const ChipTokens = { ...sizes, ...colors };
```
JKobrynski, line 48: Worth settling this one before it lands: @satya164 asked for the constants to be exported directly on [#5002](https://github.com/callstack/react-native-paper/pull/5002#discussion_r3434616306), but `Checkbox`, `Switch` and `FAB` all use the wrapper. Which shape should Chip follow?

JKobrynski:
[#5078](https://github.com/callstack/react-native-paper/pull/5078) landed and reworked `Surface` - the `container` prop is gone, and `backgroundColor` and `borderRadius` are props now rather than style keys. `Chip` on `main` is already ported, so it might be easier to rebase and start from that version than to resolve the conflicts by hand. The custom-colour example will need updating too. @k0ndee

src/components/Chip/Chip.tsx
```diff
@@ -192,287 +193,279 @@ const Chip = ({
...
+  const hasLeading = hasAvatar || showLeadingIcon;
+  const hasTrailingIcon = !!trailingIcon;
+  const hasClose = !hasTrailingIcon && !!onClose;
+  const showTrailingIcon = hasTrailingIcon || hasClose;
 
-  const labelTextStyle = {
-    color: textColor,
-    ...theme.fonts.labelLarge,
+  const leftPadding = hasAvatar
+    ? ChipTokens.avatarLeadingPadding
```
JKobrynski, line 296: This reserves the whole 32dp touch target as padding, so the label ends up ~15dp from the trailing icon where [the specs page](https://m3.material.io/components/chips/specs) says 8dp between elements. Should this count the icon rather than the touch target?

Issue #4710, opened 2025-05-01
Card component injects invalid props into React.Fragment children


adolfogc:
### Current behaviour

The Card component clones its children and injects additional props (index, total, siblings, borderRadiusStyles). This leads to runtime errors when a child is a React.Fragment, which only accepts key and children props. Other components that do not expect these props may also be affected.

![Image](https://github.com/user-attachments/assets/d8f0a199-4e2e-4649-9cb1-62e20472142e)

### Expected behaviour
The Card component should render React.Fragment children without throwing runtime errors by avoiding invalid prop injection.

### How to reproduce?
Pass a React.Fragment as a child of Card.

### What have you tried so far?

Relevant source line: https://github.com/callstack/react-native-paper/blob/main/src/components/Card/Card.tsx#L278

Potential fix:

```tsx
  import { Fragment } from 'react';
  
  // ...

  const content = (
    <View style={[styles.innerContainer, contentStyle]} testID={testID}>
      {React.Children.map(children, (child, index) => {
        if (!React.isValidElement(child)) {
          return child;
        }
        if (child.type === Fragment) {
          return child;
        }

        return React.cloneElement(child as React.ReactElement<any>, {
          index,
          total,
          siblings,
          borderRadiusStyles,
        });
      })}
    </View>
  );
```

adam-kuhn:
+1 
I've been experiencing something similar but for me the error is:

```
 Warning: Invalid prop `compact` supplied to `React.Fragment`. React.Fragment can only have `key` and `children` props.

PR #5082, opened 2026-08-28
feat: structural accessibility for data table

oleksandrzavarzin-callstack:
### Motivation

`DataTable`'s structure was purely visual: bare `View`s and `TouchableRipple`s with no roles, indices or header relationships, so a screen reader conveyed no table context at all. A user heard `"159"` with no way to know it was the Calories column of row 3.

### Changes

- **Web semantics**: `role="table"` with `aria-rowcount`/`aria-colcount`, `role="row"` with `aria-rowindex`, `role="columnheader"` with `aria-sort`, `role="cell"` with `aria-colindex`.
- **Native fallback**: table roles map to no accessibility trait on iOS or Android, so a row is announced as one item: `"Dessert, Frozen yogurt, Calories, 159, row 3 of 6"`. It falls back to per-cell focus when a row holds interactive or non-text content, so nothing becomes unreachable. `nativeFocusMode` overrides the choice.
- **Non-actionable rendering**: rows, cells and titles with no touch handler render a plain `View`.
- **Sort state**: `aria-sort` on web, folded into the accessible name on native, with an announcement on change since focus stays on the header. New `sortAccessibilityLabels` localizes the wording.
- **Reanimated**: the sort indicator moves off legacy `Animated`, matching `Switch` and `Checkbox`. Reduced motion is suppressed via `ReduceMotion.Always` instead of a manual snap, and `150ms` becomes `theme.motion.duration.short3` (same value).
- **Pagination**: real control names, localizable through one `labels` prop. Data-dependent values are functions, so pluralization stays with the app's own i18n.
- **Columns**: optional `columns` describes width and alignment once instead of repeating `style` on a title and every cell. New `align` composes with `numeric`, which is unchanged and not deprecated.
- **Text scaling**: `numberOfLines` is honoured exactly at every font scale; the default is one line at the default scale and unclamped above it.
- **Tokens**: separators move from `surfaceVariant` to `outlineVariant`, matching `Divider`.

`Button` gains `aria-expanded`, needed by the rows-per-page anchor which never reported that it opens a menu.

#### Breaking changes

Rows, cells and titles without a touch handler render a plain `View`. They lose the ripple, the web hover background and keyboard focus, and no longer report a disabled state.

```tsx
// Before (v5): announced as a disabled control
<DataTable.Row>
  <DataTable.Cell>{item.name}</DataTable.Cell>
</DataTable.Row>

// After (v6): pass a handler if the row is meant to be pressable
<DataTable.Row onPress={() => select(item)}>
  <DataTable.Cell>{item.name}</DataTable.Cell>
</DataTable.Row>
```
Other important changes:
- Rows are announced as a single item on native, and need `rowCount` and `firstRowIndex` for correct positions when paginating. 
- Pagination default labels changed, so tests querying the old placeholders need updating. 
- `numeric` now applies tabular figures, the one visual change to existing tables. 

### Related issue

Child of the "Modernize non-standard components" effort; no issue exists yet.

### Test plan

- [x] `yarn typescript`, `yarn lint` clean
- [x] `yarn test` 806 passed, 167 snapshots. The 2828-line snapshot is replaced with 8 small per-component snapshots plus named contract assertions, since the old blob is where the `aria-disabled` regression sat unnoticed.
- [x] Web: verified against the live accessibility tree. Correct roles, counts and indices, `aria-sort` cycling, and no `aria-disabled` on any row, cell or title.
- [x] Android: TalkBack on an emulator, cross-checked with `adb shell uiautomator dump`. Tapping a row focuses that row, each header is its own stop, and a row containing an interactive element (e.g. Checkbox) falls back to per-cell focus.
- [x] iOS: Accessibility Inspector against the Simulator.
- [x] Reduced motion, 200% font scale, and RTL mirroring.

### Screenshots:
Screenshot 2026-08-28 at 16 16 50
Screenshot 2026-08-28 at 16 13 27
Screenshot 2026-08-28 at 16 17 54


src/components/DataTable/DataTableHeader.tsx
```diff
@@ -50,12 +58,48 @@ const DataTableHeader = ({
...
+    });
+
+    return labels;
+  }, [children]);
+
+  const setHeaderLabels = table?.setHeaderLabels;
+  const publish = useLatestCallback(() => setHeaderLabels?.(labels));
+  const signature = labels.join(' ');
+
+  React.useEffect(publish, [publish, signature]);
```
JKobrynski: Publishing the column names through an effect makes the table's ARIA indices a second-render value. Server-rendered output (the docs site is one, `rspress.config.ts:122` aliases to react-native-web) gets `aria-rowcount="6"` instead of 7, and the header row and the first data row both get `aria-rowindex="1"` - [duplicate row indices](https://www.w3.org/TR/wai-aria-1.2/#aria-rowindex). `DataTable` already reads column position from `ColumnIndexContext` without an effect; can the header names go the same way?

src/components/DataTable/DataTableCell.tsx
```diff
@@ -69,46 +89,154 @@ const DataTableCell = ({
...
+    !row?.rowIsFocusUnit &&
+    !row?.header &&
+    !interactive &&
+    // An element child renders verbatim; naming the cell would hide its own
+    // role and state behind this label.
+    !React.isValidElement(children) &&
+    value != null;
+
+  const label =
+    ariaLabel ?? (isWeb ? undefined : composeCellLabel({ columnLabel, value }));
```
JKobrynski: Try `const label = ariaLabel ?? (cellIsFocusUnit ? composeCellLabel({ columnLabel, value }) : undefined)` - a cell wrapping a `Checkbox` currently gets `aria-label="Pick"`, which buries the checkbox's own role and state behind the column name.

src/components/DataTable/DataTableRow.tsx
```diff
@@ -55,25 +77,181 @@ export type Props = $RemoveChildren<typeof TouchableRipple> & {
...
+    // Native maps `button` to a real trait; `row` maps to nothing there.
+    role:
+      Platform.OS === 'web' || !interactive
+        ? ('row' as const)
+        : ('button' as const),
+    ...webAriaProps({
+      'aria-rowindex':
+        index == null ? undefined : index + 1 + (table?.hasHeader ? 1 : 0),
+    }),
+    accessible: accessible ?? (rowIsFocusUnit || undefined),
```
JKobrynski: Try `accessible: accessible ?? rowIsFocusUnit` - the `|| undefined` leaves the prop unset, which is exactly when `Pressable` applies its own default of `true`, so a pressable row swallows the `Checkbox` inside it on iOS.

src/components/DataTable/DataTable.tsx
```diff
@@ -105,11 +159,126 @@ export type Props = ViewProps & {
...
+      if (!isDataTableElement<DataTableRowProps>(child, 'DataTable.Row')) {
+        return child;
+      }
+
+      const offset = rendered++;
+
+      return child.props.index === undefined
+        ? React.cloneElement(child, { index: firstRowIndex + offset })
+        : child;
+    });
```
JKobrynski: Try the provider pattern you already wrote for columns (`DataTableColumnsContext.tsx:36-44`) - `cloneElement` means a consumer's `<NameRow />` wrapper gets no `index` and silently loses its `aria-rowindex`. It is also the `React.Children` problem [#4954](https://github.com/callstack/react-native-paper/discussions/4954) names.

oleksandrzavarzin-callstack:
@JKobrynski Thank you for the review!
I've pushed the fixes to all of your comments. Please review the PR again

src/components/DataTable/utils.ts
```diff
@@ -0,0 +1,126 @@
+import * as React from 'react';
+
+/** Whether a child is a particular `DataTable` sub-component. */
+export const isDataTableElement = <P>(
+  child: React.ReactNode,
+  displayName: string
+): child is React.ReactElement<P> =>
+  React.isValidElement(child) &&
+  typeof child.type !== 'string' &&
+  'displayName' in child.type &&
```
JKobrynski: A fragment child crashes the table - `<DataTable><>{rows}</></DataTable>` throws
`Cannot use 'in' operator to search for 'displayName' in Symbol(react.fragment)` on this line, and so does a fragment inside `DataTable.Row` or `DataTable.Header`. It renders fine on `main`. Same family as [#4710](https://github.com/callstack/react-native-paper/issues/4710).

src/components/DataTable/DataTable.tsx
```diff
@@ -105,11 +163,138 @@ export type Props = ViewProps & {
...
+        }
+
+        if (isDataTableElement(child, 'DataTable.Pagination')) {
+          return child;
+        }
+
+        // Anything else is taken for a row. Virtualized rows have no children
+        // here to number, so those pass `index` themselves.
+        return (
+          <RowIndexContext.Provider value={firstRowIndex + rendered++}>
```
JKobrynski: Wrap the rows in a `View` and they all get the same `aria-rowindex`, because every child that isn't the header or the pagination counts as one row here. A `<Text>` empty state inflates `aria-rowcount` the same way, and the spoken "row 3 of 6" with it. Could rows register themselves instead, so only real rows are counted?

src/components/DataTable/DataTableTitle.tsx
```diff
@@ -81,77 +115,193 @@ const DataTableTitle = ({
...
+      : onPress
+        ? ('button' as const)
+        : undefined;
+
+  const structuralProps = {
+    role,
+    // Native has no column-header semantics, so a title that is not already a
+    // pressable has to opt in, or its text is absorbed by whichever ancestor
+    // happens to be focusable and the columns read as one run-on stop.
+    accessible: Platform.OS === 'web' ? undefined : true,
```
JKobrynski: A header cell holding a select-all checkbox becomes one accessibility element on native, so the checkbox can't be reached - `accessible` is `true` here whatever the title contains. The row has a `describable` check for exactly this case; the title has none.

src/components/DataTable/DataTable.tsx
```diff
@@ -105,11 +163,138 @@ export type Props = ViewProps & {
...
+  layout = 'fluid',
+  rowCount,
+  firstRowIndex = 0,
+  nativeFocusMode = 'row',
+  formatRowPosition = defaultFormatRowPosition,
+  style,
+  ...rest
+}: Props) => {
+  if (
+    __DEV__ &&
```
JKobrynski: This warning throws a `ReferenceError` in any web bundle that doesn't define `__DEV__`, and it is the first use of it in `src/`. `splitStyles.ts:26` guards with `process.env.NODE_ENV !== 'production'` - worth matching that?

src/components/DataTable/DataTable.tsx
oleksandrzavarzin-callstack: Thank you! I fixed your both moments.

The child map is now a recursive walk: rows are numbered wherever they sit, and
anything the table can see into - a fragment, a grouping `View`, a `<Text>`
empty state - is descended into rather than counted.

Not registration because a row can only claim an index by mutating shared state
during render, which double-counts under StrictMode, or in an effect - which
your [DataTableHeader comment](https://github.com/callstack/react-native-paper/pull/5082#discussion_r3902500878) rightly ruled out for the server-rendered docs.

oleksandrzavarzin-callstack:
@JKobrynski, your comments should be resolved now. Please check if that's true, and let me know if you have any additional issues here.

src/components/DataTable/DataTable.tsx
```diff
@@ -105,11 +163,156 @@ export type Props = ViewProps & {
...
+            return child;
+          }
+
+          if (isDataTableElement(child, 'DataTable.Row')) {
+            return number(child);
+          }
+
+          // Anything the table can see into is descended into, not counted:
+          // a wrapper must not make its rows one row, nor an empty state a row.
+          if (nested !== undefined) {
```
JKobrynski: A row component that takes children is now looked through instead of counted, so
`<MyRow>{item.name}</MyRow>` loses its `aria-rowindex` and the table under-counts. One using `React.Children.only` throws outright, where it renders fine on `main`. Both worked before this commit - since registration is out, is there another way to tell a grouping wrapper from a row component?

src/components/DataTable/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+    return false;
+  }
+
+  const type: unknown = child.type;
+
+  return (
+    (typeof type === 'function' ||
+      (typeof type === 'object' && type !== null)) &&
+    'displayName' in type &&
+    (type as { displayName?: unknown }).displayName === displayName
```
JKobrynski: Do we really need to use `as` here?

src/components/DataTable/DataTable.tsx
oleksandrzavarzin-callstack: Thank you for pointing on it. The check is now the element's type plus what the table can see inside it.

The walk descends only into a type it sees through (`Fragment`, `View`, `ScrollView`, `Text`) or an element with a `Header`, `Row` or `Pagination` somewhere inside. Anything else counts as one row, as on main, so `<MyRow>` keeps its `aria-rowindex`.

Still one case, like on main: a wrapper whose rows the table can't see counts as one row. In this case `index` on the row is the way out. I've noted it in the migration doc.

src/components/DataTable/utils.ts
oleksandrzavarzin-callstack: No, I've removed the cast.

src/components/DataTable/DataTable.tsx
```diff
@@ -105,11 +170,166 @@ export type Props = ViewProps & {
...
+          }
+
+          // A container the table sees through, or a wrapper it can see the
+          // table's own parts inside, keeps its rows where they are: descend
+          // into it, so a wrapper does not make its rows one row, nor an empty
+          // state a row.
+          if (isTransparentContainer(child) || containsTableStructure(nested)) {
+            return nested === undefined
+              ? child
+              : React.cloneElement(
```
JKobrynski: Any reason to rebuild every consumer's element here `.github/instructions/code.instructions.md` asks to avoid `cloneElement` and `React.Children`.

oleksandrzavarzin-callstack: Oh, sorry. I haven't seen these instructions. I removed the cloning alltogether.
And now:
- `Fragment` is flatten.
- `Text` is left alone.
- A `View` or a consumer wrapper is now rendered as written. A row that ends up with no position warns in dev. The migration guide notes the change.

PR #5084, opened 2026-08-28
feat: add MD3 keyboard focus indicators

### LINKED ISSUE #5083: No keyboard focus indicator on most components, and the three that have one each did it differently
### Current behaviour

Most interactive components have no MD3 keyboard focus indicator.

| has a ring                              | no ring                                                                                                                                        |
| --------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
| `FAB`, `FAB.Menu`, `Checkbox`, `Switch` | `RadioButton`, `Chip`, `SegmentedButtons`, `List.Item`, pressable `Card`, `Button`, `IconButton`, `Menu.Item`, `Drawer.Item`, `DataTable` rows |

On Android a keyboard user gets nothing. On web you get the browser's default
outline, so not invisible, just not MD3 and different per browser.

The three that do have one each did it their own way. `FAB` has a local
`useFocusRing` hook with a reanimated `SharedValue` and checks
`document.activeElement`. `Checkbox` uses `useState` plus
`src/utils/isKeyboardFocusEvent.ts`. `Switch` uses a `SharedValue` plus the same
util. All three also carry their own copy of

```ts
const webNoOutline = { outline: "none" } as unknown as ViewStyle;
```

`src/theme/tokens/sys/state.ts` already has `focusIndicator: { thickness: 3,
outerOffset: 2 }`. It is just not used in most places.

One thing to know before testing on iOS: `onFocus` never fires there for a `View` or
`Pressable`. `RCTViewComponentView.mm` emits it from `becomeFirstResponder` only
behind the `enableImperativeFocus` flag, which defaults off. Android has no such
gate. So the FAB/Checkbox/Switch rings are already inert on iOS, and this is a web
and Android fix.

### Expected behaviour

A focus indicator on every interactive component, from one shared implementation and
the existing tokens.

Easy to get wrong:

- Keyboard only. A mouse click must not light it. `Pressable`'s own `focused` state
  fires for clicks too, see necolas/react-native-web#1849.
- Placement. MD3 puts the ring outside, hence `outerOffset`. But outside gets
  trimmed by a clipping ancestor sized to its content, and lands on the neighbour for
  things that sit flush. Chips in a horizontal `ScrollView` are the obvious case.
  material-web hits the same thing and has an `inward` variant for it.
- Colour. MD3 tonal palettes are luminance matched by tone, so a `secondary` ring
  drawn on top of another role at the same tone is about 1:1 and disappears. It needs
  to land somewhere with real contrast.

### How to reproduce?

1. `cd example && yarn start`
2. Android, hardware keyboard or `adb shell input keyevent KEYCODE_TAB`. Tab through
   the Chip screen. Nothing. Same on List.Item, Segmented Buttons, Radio Button, a
   pressable Card.
3. FAB and Checkbox do show a ring, so focus itself works.
4. Web, same screens, you get the browser outline instead.

Android drops keyboard focus the moment you touch the screen. Tab again to get back.

### Preview

Nothing to show as a still, it only exists while something is focused.

- Read every component for existing focus handling to get the table, rather than
  trusting the ticket this came from.
- RN has `outlineWidth` / `outlineColor` / `outlineStyle` / `outlineOffset` since
  0.77, implemented on both platforms. Measured on iOS sim, android-35 emulator and
  expo web: paints on all three, costs no layout, takes its radius from the view it
  sits on, survives that view's own `overflow: 'hidden'`, negative offset draws
  inside. An ancestor's `overflow: 'hidden'` does clip it.
- Version floor is fine. The `react-native-reanimated >= 4.3.0` peer dep already
  declares `react-native: "0.81 - 0.85"`.

### Your Environment

| react-native-web   | 0.21.2                 |

lukemorawski:
Fixes #5083

Based on #5080 and contains its two commits, since a fork cannot open a stacked PR.
Worth reviewing after that one merges, at which point I will rebase and this diff
drops to just the focus work.

### What

MD3 keyboard focus indicators, from one shared implementation, on every component
built on `TouchableRipple` plus `Card` and `Chip`'s close button.

The ring is the platform's own `outline`, not a bordered child view. It costs no
layout, takes its corner radius from the view it sits on, and is not clipped by that
view's own `overflow: 'hidden'`, so `borderless` is not a problem. `outline*` landed
in RN 0.77 and Paper's floor is already 0.81 via the reanimated peer dep.

New `src/utils/useFocusRing.ts` holds the hook, the style helper and the one
`webNoOutline`. `FAB`'s local hook is deleted and `FAB`, `Checkbox` and `Switch` are
moved onto the shared one, so the three hand-rolled rings become one.

`TouchableRipple` gets a `focusRing` prop, `'outward' | 'inward' | 'none'`.

rings-web

### Placement

Outward by default, which is what `outerOffset` in
`src/theme/tokens/sys/state.ts` is for. Inward on `Chip`, `List.Item`,
`RadioButton.Item` and `SegmentedButtons`.

Not a style preference. An outward ring is 5dp outside the box, and it gets trimmed
by any clipping ancestor sized to its content. A row of chips in a horizontal
`ScrollView` is the case that shows up in this repo's own example app, same on web
and Android: the ring loses its top and bottom and renders as two arcs. Inward fits.
`List.Item` and `SegmentedButtons` are inward for the neighbour version of the same
problem, full bleed rows and flush segments. material-web does the same split and
uses inward for its list item and tab.

chip-inward-why

Going the other way and making inward the default does not work either. MD3 tonal
palettes are luminance matched by tone, so a `secondary` ring drawn on top of another
role at the same tone is about 1:1 and disappears. On a contained `Button` fill that
is 1.00:1. Outward puts it on the page background instead, 6.13:1 light and 10.91:1
dark. There is a test for this :)

### Notes for review

- Keyboard only, via the existing `isKeyboardFocusEvent`. `Pressable`'s own `focused`
  state fires for clicks, necolas/react-native-web#1849.
- `webNoOutline` stays on `FAB` and `Switch` only, since both ring an inner element.
  Nowhere else sets `outline: none`, so a JS failure still leaves the browser ring.
- ~1k lines of snapshot went bye bye. The old rings were always mounted at `opacity: 0`;
  this one only exists while focused. Index based tree queries against `FAB` and
  `Switch` will shift.
- iOS renders the ring but never focuses it, `enableImperativeFocus` is off by
  default. Pre existing, the old `FAB`/`Checkbox`/`Switch` rings were inert there too.
- Docs not regenerated. `docs/scripts/generate-current-component-docs.ts` deletes its
  output dir before parsing and then crashes on `satisfies`. Pre existing, needs it's own PR.

### Testing

`yarn typecheck`, `yarn lint`, `yarn test` clean

Two tricky things that cost me time:

- A bare `TouchableRipple` import resolves to `.native`, so the web file goes
  untested unless you import it by extension
- RNTL will not dispatch to a `disabled` element, so negative focus assertions pass
  for free

Manual pass per component on web and Android, hardware keyboard: ring on tab, no ring
on mouse click, not clipped. iOS rendering only, for the reason above.


satya164:
closing for https://github.com/callstack/react-native-paper/pull/5107

PR #5087, opened 2026-09-01
refactor: inline Platform.OS checks

michalfedyna:
### Motivation

Inline `Platform.OS` code so minifier can remove dead code.

### Related issue

Inline `Platform.OS` code so minifier can remove dead code.

### Test plan

Look if replacement was done correctly.


JKobrynski:
[#5078](https://github.com/callstack/react-native-paper/pull/5078) landed since you opened this and deleted most of what it touches - the `isWeb` branches in `Button` and `Chip` are gone, and `Menu` on main already inlines the check. That leaves `ProgressBar`, `Tooltip` and the example files. Worth rebasing and narrowing it to those? A sweep like this goes stale fast, so better to land it small and soon. @michalfedyna

satya164 (approved):
Thanks

PR #5088, opened 2026-09-01
fix: remove hardcoded default testIDs from components

k0ndee:
### Motivation

Several components fell back to a hardcoded default `testID` (e.g. `testID = 'button'`, `testID = 'card'`, `testID = 'search-bar'`) whenever a consumer didn't pass one explicitly. This meant every instance of a component rendered without an explicit `testID` still emitted a fixed, non-configurable testID into the tree  leading to duplicate/colliding testIDs when multiple instances of the same component render on screen.

This PR removes all hardcoded default `testID` values so that `testID` is `undefined` unless a consumer explicitly provides one. Suffixed child testIDs (e.g. `` `${testID}-container` ``) are now only rendered when a `testID` is explicitly given, instead of always deriving from the removed default.

Along the way this surfaced two latent bugs that the hardcoded defaults had been masking:
- `IconButton` never forwarded its `testID` down into the inner icon, relying on `CrossFadeIcon`'s own default instead. Now it forwards `${testID}-icon` explicitly.
- `FAB.Menu`'s trigger wrapper put `testID` on a non-interactive positioning `View` instead of the actual pressable `Shell`, which only worked before because `Shell` had its own hardcoded default in the right place.

Affected components: `Button`, `Card`, `Chip`, `Surface`, `Modal`, `Menu`, `Menu.Item`, `IconButton`, `Searchbar`, `CrossFadeIcon`, `ProgressBar`, `Appbar.Header`, `Appbar.Content`, `Drawer.CollapsedItem`, `FAB`, `FAB.Extended`, `FAB.Menu`, `BottomNavigation`, `BottomNavigationBar`.

### Test plan

- `yarn typecheck` passes.
- `yarn lint` passes.
- `yarn jest` passes (55 suites, 732 tests, 169 snapshots — snapshots updated to reflect the removed default testID attributes).
- Updated unit tests that relied on the old default testID strings (`Button`, `Card`, `Menu`, `Menu.Item`, `Drawer.CollapsedItem`, `FAB.Menu`, `ListItem`, `Appbar`, `BottomNavigation`) to pass an explicit `testID` where the assertion actually needs one.
- No behavior change for consumers who already pass an explicit `testID`, only the implicit-default fallback is removed.

### DIFF
--- a/src/components/Appbar/AppbarHeader.tsx
+++ b/src/components/Appbar/AppbarHeader.tsx
@@ -88,7 +88,7 @@ const AppbarHeader = ({
   mode = Platform.OS === 'ios' ? 'center-aligned' : 'small',
   elevated = false,
   theme: themeOverrides,
-  testID = 'appbar-header',
+  testID,
   ...rest
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -141,7 +141,7 @@ const Card = ({
   style,
   contentStyle,
   theme: themeOverrides,
-  testID = 'card',
+  testID,
   accessible,
   disabled,
   ref,
@@ -225,13 +225,13 @@ const Card = ({
       style={[{ borderColor }, style]}
       theme={theme}
       elevation={elevation}
-      testID={`${testID}-container`}
+      testID={testID ? `${testID}-container` : undefined}
       {...rest}
     >
       {isMode('outlined') && (
         <View
           pointerEvents="none"
-          testID={`${testID}-outline`}
+          testID={testID ? `${testID}-outline` : undefined}
           style={[
             {
               borderColor,
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -71,12 +71,61 @@ You can use the component's color prop where available, or override the correspo
 
 ### Test IDs
 
-Some hardcoded and generated test IDs have been removed for the following components:
+Hardcoded default test IDs have been removed for the components listed below. Many of these components also derive test IDs for their internal parts by appending a suffix to the `testID` prop (e.g. `${testID}-container`). Since `testID` is no longer defaulted to a hardcoded value, none of these derived test IDs are set either unless you pass a `testID` explicitly — so all queries by the IDs below will stop matching:
 
-- `Appbar.Header`: `${testID}-root-layer`
-- `Surface`: `surface` and `${testID}-outer-layer`
+- `Appbar.Content`: `appbar-content`
+  - `appbar-content-title-text`
+- `Appbar.Header`: `appbar-header`
+  - `appbar-header-root-layer`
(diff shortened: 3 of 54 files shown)


JKobrynski:
@k0ndee we've got a bunch of conflicts

src/components/Appbar/AppbarHeader.tsx
```diff
@@ -88,7 +88,7 @@ const AppbarHeader = ({
   mode = Platform.OS === 'ios' ? 'center-aligned' : 'small',
   elevated = false,
   theme: themeOverrides,
-  testID = 'appbar-header',
+  testID,
```
JKobrynski, line 91: The removed defaults need a bullet in the migration guide's "Test IDs" section (`docs/6.x/docs/guides/migration.md`) —- it already lists the ones `Surface` and `Appbar.Header` lost. Otherwise anyone selecting `button`, `card` or `search-bar` in their e2e tests loses those matches silently.

src/components/Card/Card.tsx
```diff
@@ -225,13 +225,13 @@ const Card = ({
       style={[{ borderColor }, style]}
       theme={theme}
       elevation={elevation}
-      testID={`${testID}-container`}
+      testID={testID ? `${testID}-container` : undefined}
```
JKobrynski, line 228: Five other components build child testIDs without this guard, so they render a literal
`undefined-text` - visible in the committed snapshots today. `Checkbox/CheckboxItem.tsx:177`, `DataTable/DataTableCell.tsx:111`, `List/ListItem.tsx:245`, `SegmentedButtons/SegmentedButtonItem.tsx:227` and `Snackbar.tsx:341`. Worth including them?

src/components/Appbar/AppbarHeader.tsx
k0ndee: Good point. Added information to migration guide.

src/components/Card/Card.tsx
k0ndee: Missed them, included in the new commit.

satya164 (changes requested):
I think there many test IDs added to views to solely test the styles or props in unit tests, even though they are internal elements. Majority of them are things that add text to `testID`, e.g. `${testID}-icon-wrapper`.

Let's remove them as well. Test IDs should only be present on elements that users need to access for public behavior tests, e.g. tap/interact with an element or assert content of an element.

k0ndee:
> I think there many test IDs added to views to solely test the styles or props in unit tests, even though they are internal elements. Majority of them are things that add text to `testID`, e.g. `${testID}-icon-wrapper`.
> 
> Let's remove them as well. Test IDs should only be present on elements that users need to access for public behavior tests, e.g. tap/interact with an element or assert content of an element.

@satya164 there is a separate task for derived testIDs and I already created a draft PR #5099 that addresses that. If You'd like I can merge those two together so both default and derived testID are handled in one PR.

satya164:
@k0ndee then i'll merge this one and check that one separately

satya164 (approved):
thanks

PR #5089, opened 2026-09-01
refactor: modernize Divider to latest MD3 spec

mikolajadamowicz:
### Motivation

`Divider` was drawing a `StyleSheet.hairlineWidth` line, which is thinner than 1dp on most screens, and the full 1dp line the Material Design 3 spec asks for was only reachable through `bold`. `leftInset` set `marginLeft`, so in RTL the inset stayed on the left instead of moving to the leading edge. MD3 also lists a vertical divider, which we had no way to render.

This is primitive level work for the v6 effort. The consumers (List, Dialog, Drawer, DataTable) pick up the corrected divider in their own tasks.

### Changes

- **Thickness**: `StyleSheet.hairlineWidth` → `1`dp, applied as `height` when horizontal and `width` when vertical
- **Inset**: `leftInset` (`marginLeft: 16`) → `startInset` (`marginStart: 16`), so it follows the writing direction
- **Orientation**: new `orientation="horizontal" | "vertical"`. A vertical divider is 1dp wide and uses `alignSelf: 'stretch'` to take the height of its parent
- **Insets follow the orientation**: on a vertical divider `startInset` insets the top edge and `horizontalInset` insets the top and bottom edges
- **Accessibility**: the divider is `aria-hidden` by default, so screen readers skip it and it stays out of the focus order. React Native maps that to `accessibilityElementsHidden` on iOS and `importantForAccessibility="no-hide-descendants"` on Android. It is spread before `...rest`, so `<Divider accessible aria-hidden={false} role="separator" />` opts back in

#### Breaking change

The `bold` and `leftInset` props have been removed.

```tsx
// Before (v5)
<Divider bold />
<Divider leftInset />

// After (v6)
<Divider />
<Divider startInset />
```

`bold` existed only to reach 1dp, which is now the default. For the old hairline line, pass `style={{ height: StyleSheet.hairlineWidth }}`.

The full entry is in `docs/6.x/docs/guides/migration.md`.

#### Callsites updated

- `src/components/Searchbar.tsx` and `src/components/Drawer/DrawerSection.tsx` dropped `bold`, so they render the same as before
- `example/src/Examples/DividerExample.tsx` now shows full width, start inset, both side inset and vertical dividers
- `example/src/Examples/FABExample.tsx` dropped `bold`
- `src/components/__tests__/Drawer/__snapshots__/DrawerSection.test.tsx.snap` updated

### Related issue

N/A

### Test plan

- [x] `yarn lint` clean
- [x] `yarn typecheck` clean, plus `tsc --noEmit` in `example/`
- [x] `yarn test` 56 suites, 743 passed, including 11 new `Divider` tests in `src/components/__tests__/Divider.test.tsx`
- [x] `yarn docs build` clean, `Divider.mdx` and `componentDocs6x.json` regenerated
- [x] Example app, Divider screen: full width, start inset, both side inset and vertical sections

To check it by hand, open the example app, go to the Divider screen and compare against `main`. The line is visibly thicker, and the vertical section is new.

## Visual verification

| Platform | divider|
| :---: | :---: |
| **Android** (Pixel 9 Pro XL) | divider |
| **iOS** (iPhone 17 Pro) | divider|


docs/6.x/docs/guides/migration.md
```diff
@@ -6,6 +6,64 @@ TBC
 
 ## Components
 
+### Divider
```
JKobrynski, line 83: This conflicts with `main` now - #5078 landed a `## General changes` section at the same insertion point. Both sides only add text, so keeping both resolves it.

JKobrynski (commented):
One minor comment, we've also got conflicts

docs/6.x/docs/guides/migration.md
rjmreis: Rebased on main and resolved the conflicts:

- docs/6.x/docs/guides/migration.md: kept both sections: this PR's new ### Divider entry plus #5078's ## General changes section, as suggested.
- docs/6.x/docs/components/Divider.mdx: removed, following main's deletion. main migrated to auto-generated component docs (from JSDoc in source), and the accessibility note this PR added here already lives in Divider.tsx's JSDoc, so nothing was lost.
- docs/src/data/componentDocs6x.json: removed, also following main's deletion (a generated data file for the old docs system, no longer used).

src/components/Divider.tsx
```diff
@@ -49,41 +56,55 @@ export type Props = $RemoveChildren<typeof View> & {
...
   ...rest
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
 
-  const dividerColor = theme.colors.outlineVariant;
+  const isVertical = orientation === 'vertical';
 
   return (
     <View
+      aria-hidden
```
JKobrynski, line 73: Tests that look up a divider will stop finding it - `aria-hidden` keeps it out of `getByTestId` unless the query passes `{ includeHiddenElements: true }`. Worth a line in the migration guide?

JKobrynski (commented):
LGTM in general, one minor comment + we've got conflicts

src/components/Divider.tsx
rjmreis: Good catch, I've added a note in the migration guide. Also fixed latest conflicts.

JKobrynski (approved):
@satya164 all yours!

PR #5097, opened 2026-09-02
feat: modernize Button to MD3 Expressive

### LINKED ISSUE #4928: refactor(button): improve MD3 compliance and modernize content API

oleksandrzavarzin-callstack:
### Motivation

`Button` was left out of the earlier modernization wave, so it still had one size,
no shape scale, no toggle, MD2-era mode names, and RN `Animated` animations. This
brings it to Material Design 3 Expressive against the gaps listed for Button in the
[MD3 Component Review](https://github.com/callstack/react-native-paper/discussions/4949).

Closes #4928.

This is a takeover of #4943. That branch is rebased onto `main` here, and the
`children` → `label` rename is reverted - the label still goes in `children`.

### What's new

- **`size`** - the XS/S/M/L/XL scale. Defaults to `small`, which is the only size
  5.x had, so existing buttons keep their metrics.
- **`shape`** - `round` (full pill) or `square` (a smaller per-size corner).
  Defaults to `round`.
- **`selected`** - the Expressive toggle. Deliberately nullable: leaving it
  `undefined` is a plain button, because MD3 colors a toggle that is
  unselected differently from a button that is not a toggle at all.
  MD3 defines the toggle for the filled, elevated, tonal and outlined styles only, 
  so selected is inert on mode="text".
- **Corner morph** - the container springs to the pressed corner while held, and
  animates between shapes as `selected` changes. The pressed corner tightens with
  the size per the spec's corner table: 8dp for XS/S, 12dp for M, 16dp for L/XL.
  **`animateShape={false}`** opts out, and the press morph is skipped under
  reduce-motion.
- **`iconPosition`** - `leading` / `trailing`, replacing the undocumented
  `contentStyle={{ flexDirection: 'row-reverse' }}` trick. Flips with `useLocale()`
  under RTL.
- **`rippleColor`** - the ripple / state layer, defaulting to the label color at
  the pressed-state opacity per spec.
- **Component tokens** - every per-size metric now lives in
  `src/components/Button/tokens.ts`, modelled on Compose's `Button{Size}Tokens`
  and resolved through the shared `resolveCornerRadius`.
- **Accessible touch target** - XS and S are shorter than 48dp, so the press area
  is expanded with `hitSlop` without changing the visual size.
- **Reanimated** - the corner morph is a Reanimated spring reading
  `theme.motion.spring.fast.spatial`. Elevation and container color ride
  `Surface`'s CSS transitions, which read `theme.motion.*`.

#### Mode names and default

```diff
- <Button mode="contained">Save</Button>
+ <Button mode="filled">Save</Button>

- <Button mode="contained-tonal">Save</Button>
+ <Button mode="tonal">Save</Button>
```

The default mode is now `filled`, not `text`

```diff
- <Button>Cancel</Button>
+ <Button mode="text">Cancel</Button>
```

#### Removed props

```diff
- <Button uppercase compact contentStyle={{ flexDirection: 'row-reverse' }} icon="chevron-right">
-   Next
- </Button>
+ <Button
+   size="extra-small"
+   iconPosition="trailing"
+   icon="chevron-right"
+   labelStyle={{ textTransform: 'uppercase' }}
+ >
+   Next
+ </Button>
```

- `uppercase` → `textTransform`.
- `compact` → superseded by the size scale.
- `contentStyle={{ flexDirection: 'row-reverse' }}` → `iconPosition`.

#### `style` is a `SurfaceStyle`, so corner radius moves out of it

`Button` renders on `Surface`, so `style` is `StyleProp<SurfaceStyle>` and follows
the same rule the Surface rework introduced in #5078: border radius is not set
through `style`. Use `shape`, or override the corner token that the size maps to.

```diff
- <Button style={{ borderRadius: 8 }}>Press me</Button>
+ <Button shape="square" theme={{ shapes: { corner: { medium: 8 } } }}>
+   Press me
+ </Button>
```

The square corner reads `medium` at XS/S, `large` at M, and `extraLarge` at L/XL.
Being animated, `style` also takes Reanimated styles:

```diff
- <Button style={{ transform: [{ scale: rnAnimatedValue }] }}>Press me</Button>
+ <Button style={reanimatedStyle}>Press me</Button>
```

There is no `elevation` prop; elevation follows `mode="elevated"`.

#### Appearance changes, no code change needed

- `outlined` label and icon use `onSurfaceVariant` instead of `primary`.
- Default icon size is 20dp, up from 18dp.
- `labelStyle={{ fontSize }}` no longer changes the icon size - it follows `size`.
- The resting corner is the full pill radius rather than a fixed 20dp.

#### Card.Actions

Defaults its children to `outlined` for the first and `filled` for the rest, where
it used to default to `contained`. It no longer forwards `compact` or `uppercase`.

### Follow-ups

- **#5084 - MD3 keyboard focus indicators.** That PR gives every 
  `TouchableRipple` consumer a focus ring, so Button gets one with no code here - which is why none
  is implemented locally. One thing to settle there: it defaults to
  `focusRing="outward"` and warns an outward ring is trimmed by "any clipping
  ancestor sized to its content". This PR wraps Button's ripple in a permanent
  `overflow: 'hidden'` clip view (needed so the ripple follows the animated
  corner), so Button will likely need `focusRing="inward"` or the ring on the
  outer view. There is a `TODO` in `Button.tsx` pointing at it.
- **#5080 - 48dp touch targets** overlaps with the `hitSlop` expansion here and also touches Button's snapshot. Whichever lands second will need a rebase; the
  two approaches should be reconciled rather than stacked.
- **#5095 - remove `ToggleButton`.** Split out of this PR and open separately, so
  neither depends on the other. It points icon-only toggles at `IconButton`'s
  existing `selected` prop; the `selected` prop added here is the replacement path
  for a toggle that carries a label.

### Notes for review

- **The outlined border color is unchanged from `main` (`outlineVariant`).**
  @adrcotfas asked for `outline` in the component review and the original branch
  made that change; I reverted it. `OutlinedButtonTokens.OutlineColor` in the
  Compose reference - the implementation the review itself names as the
  cross-check - is `OutlineVariant`.
- `selected` is ignored on `mode="text"`, which MD3 gives no toggle: no toggle
 colors, no shape flip, no `aria-selected`. It is not type-restricted, though. 
 A discriminated union would enforce it, but `react-docgen` silently drops
 union-member props from the generated tables 

### Review threads

All ten from the previous round:

| Thread | Resolution |
| --- | --- |
| Remove `label`, keep `children` | Reverted; `children` is byte-identical to `main` |
| "what makes these *legacy*?" | The unsized path is deleted - `size` defaults to `small`, so the concept is gone |
| Use Reanimated | Corner morph is a Reanimated spring; elevation and container color ride `Surface`'s transitions. Both read motion tokens |
| Drop `uppercase` | Removed, along with its injection from `Dialog.Actions` |
| `contentStyle` doc must not mention `flexDirection` | Rewritten; row-reverse support removed |
| No deprecation warning for `row-reverse` | Support removed outright - Button now has no `console.warn` at all |
| Breaking changes undocumented | This description plus `migration.md` |
| Warnings fire every render | All deprecation warnings are gone |
| `Card.Actions` mode default break | Documented above and in `migration.md` |
| `Animated.timing` never `.start()`ed | Effect deleted. Elevation is now a `Surface` prop derived during render, so a `disabled`/`mode` change moves it with nothing to chase |

### Test plan

- `yarn lint`, `yarn typecheck`, `yarn test` pass. 55 suites, 727 tests, 1 skipped.
- New tests cover: the toggle color table per mode × state, `selected === undefined`
  vs `false`, the per-size token table, shape radii per size, the pressed radius per
  size, the corner morph and its opt-out, the reduce-motion skip, the resting shadow
  per mode, the 48dp target, and that every mode renders at the token height
  including an outlined toggle in both states.
- Tested on Android/iOS simulator and on the web browser:
  - Android dark mode - no square artifact behind `text`/`outlined` labels.
  - Android + iOS - the ripple follows the morphing corner.
  - Every mode the same height at a given size, including an outlined toggle.

Android     |     IOS

https://github.com/user-attachments/assets/46577523-10e4-45e6-b6dd-730fd3a40ef2

---

BREAKING CHANGE: `Button` is redesigned for MD3 Expressive with no deprecation
aliases.

- `mode="contained"` → `mode="filled"`, `mode="contained-tonal"` → `mode="tonal"`.
  The old values are removed.
- The default `mode` is now `filled`, not `text`.
- `uppercase` is removed. Use `labelStyle={{ textTransform: 'uppercase' }}`.
- `compact` is removed. Use `size="extra-small"`.
- `contentStyle={{ flexDirection: 'row-reverse' }}` no longer moves the icon. Use
  `iconPosition="trailing"`.
- `style` is a `StyleProp<SurfaceStyle>`, so it no longer sets border radius and no
  longer accepts RN `Animated` values. Use `shape` or a corner token override for
  the radius, and a Reanimated style for animation. There is no `elevation` prop.
- `outlined` label and icon use `onSurfaceVariant`; the default icon size is 20dp;
  `labelStyle.fontSize` no longer drives the icon size; the resting corner is the
  full pill radius.
- `Card.Actions` defaults its children to `filled` rather than `contained`, and no
  longer forwards `compact` or `uppercase`.


azizbecha:
Hey @oleksandrzavarzin-callstack, thanks for taking time to continue working on Button component, I was away for a little while haha.

Really happy to see you continuing from where I left off and building on top of my previous commits, appreciated!!

Please let me know if there are any tasks we can share or work on together. I’d be happy to help!

JKobrynski:
[#5080](https://github.com/callstack/react-native-paper/issues/5080)
 is solving the same 48dp target problem by growing the container. `hitSlop` here sits on the inner `TouchableRipple`, and RN's own docs say the touch area never extends past the parent bounds - so I don't think this expands anything. Worth coordinating with that PR?

oleksandrzavarzin-callstack:
@JKobrynski, thank you for the mention of it!
`styles.clip` has `overflow: 'hidden'` and the per-size `minHeight` is on the inner content view, so the slop fell outside the touchable's own ancestors and never applied. 
I removed it here, snapshots regenerated.

#5080 applies the same `hitSlop` from inside `TouchableRipple` and doesn't touch `Button.tsx`, so Button's clipping ancestor would still swallow it - Button needs the same treatment `IconButton` got there. Both PRs touch Button's snapshot, so whoever lands second regenerates.

oleksandrzavarzin-callstack:
BTW, this doesn't hold up the PR. Button is 40dp on main today with no 48dp target either, so this just puts it back where it was. I'll unclip and pick up the real target as a follow-up once #5080 lands.

src/components/Button/Button.tsx
```diff
@@ -210,263 +275,323 @@ const Button = ({
...
-    color: textColor,
-    ...font,
-  };
+  const contentBoxStyle = React.useMemo(
+    () => ({
+      minHeight: sizeStyle.minHeight - borderWidth * 2,
+      paddingStart: sizeStyle.paddingStart - borderWidth,
+      paddingEnd: sizeStyle.paddingEnd - borderWidth,
+      gap: sizeStyle.iconGap,
+    }),
```
JKobrynski, line 482: I don't think this actually widens the touch target. React Native's own docs say "the touch area never extends past the parent view bounds", and the touchable's parents here - the clip view and the `Surface` - are both exactly the button's height, so the extra 8dp has nowhere to go. The test only checks the prop is passed through, so it would pass either way. #5080 grows the container instead, which does work — worth taking that approach?

src/components/Button/tokens.ts
```diff
@@ -0,0 +1,147 @@
...
+    selectedContainerShapeRound: 'full',
+    selectedContainerShapeSquare: 'large',
+  },
+  large: {
+    containerHeight: 96,
+    iconSize: 32,
+    iconLabelSpace: 12,
+    leadingSpace: 48,
+    trailingSpace: 48,
+    outlinedOutlineWidth: 1,
```
JKobrynski: The outline gets thicker with the size in MD3 -`md.comp.button.large.outlined.outline.width` is 2dp and `md.comp.button.xlarge.outlined.outline.width` is 3dp, but this table has 1 for all five sizes. https://m3.material.io/components/buttons/specs

JKobrynski:
And we've got conflicts

src/components/Button/Button.tsx
oleksandrzavarzin-callstack: Ok, I found a way to implement the fix without racing with the #5080. 
I used the spec's per-size outer bounds (48/48/64/104/144) rather than a flat 48dp, which only covers XS and S. Verified on device with negative controls on both platforms.

Interaction with the #5080 changes: `shouldExpand` is gated on `hitSlop === undefined`, so Button's explicit per-size slop wins and the auto-measure stays the default everywhere else. 
Web is still a no-op until the PR's absolutely-positioned target lands, then Button gets it for free

oleksandrzavarzin-callstack:
@JKobrynski, I've fixed the conflicts, and pushed the fixes to your comments.

JKobrynski:
@oleksandrzavarzin-callstack unfortunately we've got conflicts again

oleksandrzavarzin-callstack:
@JKobrynski Resolved the conflicts

JKobrynski (approved):
LGTM, all yours @satya164

PR #5098, opened 2026-09-03
refactor: utility types cleanup

michalfedyna:
### Motivation

Removed utilities from `src/types.tsx`. As suggested by @satya164.

### Related issue

Some of them are duplicating what TS already has or can inline using some existing utility. 

### Test plan

Look if replacement was done correctly.


satya164 (approved):
Thanks

PR #5099, opened 2026-09-03
fix: remove derived testIDs

k0ndee:
### Motivation

Many components generated "derived" testIDs by appending a hardcoded suffix to the user-supplied `testID` prop (e.g. `${testID}-container`, `${testID}-outline`, `${testID}-icon`, `${testID}-text`) purely so tests could reach into internal, non-public parts of a component's render tree.

This PR removes every derived testID suffix from the library and fixes up every test that depended on one, choosing between four outcomes on a case-by-case basis (see "Decisions by component" below):

1. **Delete the prop, no test change needed** - nothing queried the derived ID.
2. **Rewrite the test against a public query** (role, accessible label, text, or the component's own `ref`) when the underlying behavior genuinely is public.
3. **Convert to a full-render (`toJSON()`) snapshot** when the check is a legitimate regression guard but has no public equivalent (an internal style value, an icon `source`, an animated value) - this still catches regressions without the test reaching into a specific internal node by ID.
4. **Delete the test** when, once the derived testID was gone, there was nothing left to assert that wasn't pure implementation detail.

One derived testID was deliberately **kept** (`BottomNavigation`'s `${testID}-bar`) because there is currently no public way to reach that node at all - see below.

Affected components: `Surface`, `Button`, `IconButton`, `CrossFadeIcon`, `Appbar` (`AppbarHeader`, `AppbarContent`), `ProgressBar`, `Chip`, `ListItem`, `Card`, `FAB` (`Content`, `Shell`, `Menu`, `Extended`), `SegmentedButtonItem`, `CheckboxItem`, `DataTableCell`, `Searchbar`, `Snackbar`, `Banner`, `ToggleButton`, `Menu`/`MenuItem`, `Modal`, `Dialog`, `DrawerCollapsedItem`, `BottomNavigation`/`BottomNavigationBar`.

The `docs/6.x/docs/guides/migration.md` "Test IDs" section is updated to reflect that these internal testIDs are now gone for good (previously it said they'd come back once you passed `testID` explicitly, which was true right after #5088 but is no longer true after this PR), and calls out the `BottomNavigation` exception explicitly.

### Decisions by component

Grouped by what actually happened, since "removed a testID" covers several different situations:

**Straightforward removal (no test depended on genuine public behavior)**
- `Surface`, `Button`, `Chip`, `ListItem` (`-content`), `FAB.Content`, `FAB.Shell`, `FAB.Menu`, `ProgressBar`, `Snackbar`, `Dialog`, `MenuItem`, `BottomNavigationBar` (`-content`, `-content-wrapper`) - the derived testID was dropped from the source and the corresponding assertions were rewritten to query by role/label/text where one existed, or the check was already covered elsewhere.
- `IconButton`: removed `${testID}-icon` on the inner icon. This one wasn't part of the original PR's scope and had **zero references anywhere in the test suite** - genuinely dead code, unrelated to any test.

**Testid moved to a more meaningful node instead of being deleted outright**
- `CrossFadeIcon`: previously had no testID of its own, only `${testID}-previous`/`${testID}-current` on the two crossfading `Animated.View`s. Since the outer wrapping `View` is the element that represents "this icon" from the outside, it now receives the plain `testID` directly, and the internal previous/current suffixes are gone.
- `Menu`: `${testID}-surface` moved to the `Surface` itself (`testID={testID}`), since the `Surface` *is* the menu's real visible content, not an implementation detail. `${testID}-view` (the animated positioning wrapper) was dropped since positioning is verified via snapshot instead (see below).

**Rewritten to assert public behavior**
- `FABExtended.test.tsx` - `'expands to fit the measured label width'` used to read the animated width off `extended-fab-container`. Rewritten to use `FAB.Extended`'s own `ref` prop (forwards through `Shell` to the `Surface`) with `Reanimated.getAnimatedStyle(ref.current)` - no testID or DOM traversal, just a real forwarded ref any consumer could use.
- `Modal.test.tsx` - backdrop assertions rewritten to `getByLabelText('Close modal')`, a real interactive/accessible element. Surface opacity and inset-margin checks have no accessible-query equivalent, so those became `toJSON()` snapshots.
- `Searchbar.test.tsx` - removed a duplicate clear-icon test, added a negative-case test (`getByLabelText('clear', { includeHiddenElements: true })`), dropped an internal-layout (`marginLeft`) assertion, and converted a `borderRadius` check to a snapshot.
- `Card.test.tsx`, `DrawerCollapsedItem.test.tsx`, `DataTable.test.tsx`, `CheckboxItem.test.tsx`, `SegmentedButton.test.tsx` - style/prop/icon-source checks that read an internal testID (in `DrawerCollapsedItem`'s case, via raw `.props` chains already flagged with an eslint-disable) converted to `toJSON()` snapshots or public queries (`getByText`, sibling `testID`) where one existed. Same regression coverage, no internal-node access.
- `Menu.test.tsx` - position assertions rewritten to check the serialized tree plus a snapshot; shadow-style assertions now query `testID` directly since `Menu`'s testID moved onto the `Surface` (see above).

**Kept as a deliberate, narrow exception**
- `BottomNavigation.tsx`: `testID={testID ? \`${testID}-bar\` : undefined}` on the nested `BottomNavigationBar` is still there. Three tests (`onIndexChange`, `onTabPress`, `onTabLongPress`) need to fire a synthetic `layout` event on `BottomNavigationBar`'s root view before its tabs become pressable (they stay `pointerEvents: 'none'` until the first `onLayout` fires, and `userEvent.press` genuinely respects `pointerEvents`, unlike `fireEvent`). `onLayout` doesn't bubble in React Native, and neither `BottomNavigation` nor `BottomNavigationBar` forwards a `ref`, so there's currently no way to reach that node from outside without this testID. This isn't testing styles/props - it's the only way to get the component into an interactive state for a real interaction test. Flagging this explicitly for review: the clean long-term fix is adding real `ref` forwarding to `BottomNavigationBar` (mirroring what `FAB.Extended` already does), which is a real public-API addition and felt out of scope for a testID cleanup PR - happy to do it here instead if reviewers would rather not carry the exception.

### Related issue

None. This is an internal contributor task.

### Test plan

- `yarn jest` - full suite passes (55 suites, 674 passed + 1 skipped / 675 total, 233 snapshots).
- `yarn eslint src` - no violations (including `testing-library/no-node-access` and the repo's `no-restricted-syntax` rule against raw `.props` access).
- `npx tsc --noEmit -p tsconfig.source.json` (equivalent to `yarn typecheck`'s `tsc -b` for source files) - no type errors.
- Snapshot diffs were reviewed to confirm each one only drops a removed `testID` line (or replaces a targeted assertion with an equivalent full-tree one), with no unrelated structural or style changes.
- No UI/visual changes - this is a test-only and internal-prop-only change; component render output is otherwise identical aside from the removed `testID` attributes themselves.

### DIFF
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -98,7 +98,7 @@ These components used to also derive test IDs for internal, implementation-only 
 
 If you were relying on internal test IDs, update your tests not to rely on internal implementation details and only interact with elements or assert content your users can reach, e.g.: query by role, label, text etc., or `testID` props accepted by the component.
 
-Some components now accept explicit `testID` props for their interactable, internal elements:
+Some components now accept explicit `testID` props for their interactable elements:
 
 - `BottomNavigation`: `barTestID` for the internal `BottomNavigation.Bar`, replacing the previous `${testID}-bar` derivation.
 - `Chip`: `closeIconTestID` for the close icon button.
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -94,9 +94,18 @@ Hardcoded default test IDs have been removed for the components listed below:
 
 You can specify a `testID` explicitly to restore each component's own test ID.
 
-These components used to also derive test IDs for internal, implementation-only elements by appending a suffix to the `testID` prop (e.g. `${testID}-container`, `${testID}-icon`, `${testID}-outline`). Those internal test IDs have been removed entirely — they are no longer set even when you pass `testID` explicitly. Test IDs should only be used to interact with elements or assert content your users can reach, not to reach into a component's internal structure in tests; if you were relying on one of these internal IDs, query by role, label, text, or the component's own `testID`/`ref` instead.
+These components used to also derive test IDs for internal, implementation-only elements by appending a suffix to the `testID` prop (e.g. `${testID}-container`, `${testID}-icon`, `${testID}-outline`). They have been removed entirely.
 
-The one exception is `BottomNavigation`, which still sets `${testID}-bar` on its internal `BottomNavigation.Bar` when you pass `testID`, since there's currently no other way to reference that node from outside the component.
+If you were relying on internal test IDs, update your tests not to rely on internal implementation details and only interact with elements or assert content your users can reach, e.g.: query by role, label, text etc., or `testID` props accepted by the component.
+
+Some components now accept explicit `testID` props for their interactable, internal elements:
+
+- `BottomNavigation`: `barTestID` for the internal `BottomNavigation.Bar`, replacing the previous `${testID}-bar` derivation.
+- `Chip`: `closeIconTestID` for the close icon button.
+- `Dialog` and `Modal`: `overlayTestID` for the overlay displayed behind the content.
+- `Menu`: `overlayTestID` for the overlay displayed behind the menu.
+- `Searchbar`: `searchTestID`, `clearTestID`, and `trailingTestID` for the search, clear, and trailing icon buttons.
+- `Snackbar`: `iconTestID` for the icon button.
 
 ## Components
 
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -71,61 +71,32 @@ You can use the component's color prop where available, or override the correspo
 
 ### Test IDs
 
-Hardcoded default test IDs have been removed for the components listed below. Many of these components also derive test IDs for their internal parts by appending a suffix to the `testID` prop (e.g. `${testID}-container`). Since `testID` is no longer defaulted to a hardcoded value, none of these derived test IDs are set either unless you pass a `testID` explicitly — so all queries by the IDs below will stop matching:
+Hardcoded default test IDs have been removed for the components listed below:
 
 - `Appbar.Content`: `appbar-content`
-  - `appbar-content-title-text`
 - `Appbar.Header`: `appbar-header`
-  - `appbar-header-root-layer`
 - `BottomNavigation`: `bottom-navigation`
-  - `bottom-navigation-bar`
 - `BottomNavigation.Bar`: `bottom-navigation-bar`
-  - `bottom-navigation-bar-content`
-  - `bottom-navigation-bar-content-wrapper`
 - `Button`: `button`
-  - `button-container`
-  - `button-icon-container`
-  - `button-text`
 - `Card`: `card`
-  - `card-container`
-  - `card-outline`
 - `Chip`: `chip`
-  - `chip-container`
 - `Drawer.CollapsedItem`: `drawer-collapsed-item`
-  - `drawer-collapsed-item-outline`
-  - `drawer-collapsed-item-container`
 - `FAB`: `floating-action-button`
-  - `floating-action-button-container`
-  - `floating-action-button-text`
 - `FAB.Extended`: `extended-floating-action-button`
-  - `extended-floating-action-button-container`
-  - `extended-floating-action-button-text`
 - `FAB.Menu`: `floating-action-button-menu`
 - `IconButton`: `icon-button`
-  - `icon-button-container`
-  - `icon-button-icon` (and `icon-button-icon-previous` / `icon-button-icon-current` when `animated`)
 - `Menu`: `menu`
-  - `menu-view`
-  - `menu-surface`
 - `Menu.Item`: `menu-item`
-  - `menu-item-title`
 - `Modal`: `modal`
-  - `modal-backdrop`
-  - `modal-wrapper`
-  - `modal-surface`
 - `ProgressBar`: `progress-bar`
-  - `progress-bar-fill`
 - `Searchbar`: `search-bar`
-  - `search-bar-container`
-  - `search-bar-icon`
-  - `search-bar-icon-wrapper`
-  - `search-bar-clear-icon`
-  - `search-bar-trailering-icon`
-  - `search-bar-divider`
 - `Surface`: `surface`
-  - `surface-outer-layer`
 
-You can specify a `testID` explicitly to restore both the component's own test ID and all of its derived test IDs above, using the same suffixes.
+You can specify a `testID` explicitly to restore each component's own test ID.
+
+These components used to also derive test IDs for internal, implementation-only elements by appending a suffix to the `testID` prop (e.g. `${testID}-container`, `${testID}-icon`, `${testID}-outline`). Those internal test IDs have been removed entirely — they are no longer set even when you pass `testID` explicitly. Test IDs should only be used to interact with elements or assert content your users can reach, not to reach into a component's internal structure in tests; if you were relying on one of these internal IDs, query by role, label, text, or the component's own `testID`/`ref` instead.
+
+The one exception is `BottomNavigation`, which still sets `${testID}-bar` on its internal `BottomNavigation.Bar` when you pass `testID`, since there's currently no other way to reference that node from outside the component.
 
 ## Components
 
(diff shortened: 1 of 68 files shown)


docs/6.x/docs/guides/migration.md
```diff
@@ -71,61 +71,32 @@ You can use the component's color prop where available, or override the correspo
...
-  - `search-bar-divider`
 - `Surface`: `surface`
-  - `surface-outer-layer`
 
-You can specify a `testID` explicitly to restore both the component's own test ID and all of its derived test IDs above, using the same suffixes.
+You can specify a `testID` explicitly to restore each component's own test ID.
+
+These components used to also derive test IDs for internal, implementation-only elements by appending a suffix to the `testID` prop (e.g. `${testID}-container`, `${testID}-icon`, `${testID}-outline`). Those internal test IDs have been removed entirely — they are no longer set even when you pass `testID` explicitly. Test IDs should only be used to interact with elements or assert content your users can reach, not to reach into a component's internal structure in tests; if you were relying on one of these internal IDs, query by role, label, text, or the component's own `testID`/`ref` instead.
+
+The one exception is `BottomNavigation`, which still sets `${testID}-bar` on its internal `BottomNavigation.Bar` when you pass `testID`, since there's currently no other way to reference that node from outside the component.
```
satya164: remove it and add a `barTestID` prop to `BottomNavigation` that sets the test ID for the bar

docs/6.x/docs/guides/migration.md
```diff
@@ -71,61 +71,32 @@ You can use the component's color prop where available, or override the correspo
...
-  - `search-bar-clear-icon`
-  - `search-bar-trailering-icon`
-  - `search-bar-divider`
 - `Surface`: `surface`
-  - `surface-outer-layer`
 
-You can specify a `testID` explicitly to restore both the component's own test ID and all of its derived test IDs above, using the same suffixes.
+You can specify a `testID` explicitly to restore each component's own test ID.
+
+These components used to also derive test IDs for internal, implementation-only elements by appending a suffix to the `testID` prop (e.g. `${testID}-container`, `${testID}-icon`, `${testID}-outline`). Those internal test IDs have been removed entirely — they are no longer set even when you pass `testID` explicitly. Test IDs should only be used to interact with elements or assert content your users can reach, not to reach into a component's internal structure in tests; if you were relying on one of these internal IDs, query by role, label, text, or the component's own `testID`/`ref` instead.
```
satya164:
```suggestion
These components used to also derive test IDs for internal, implementation-only elements by appending a suffix to the `testID` prop (e.g. `${testID}-container`, `${testID}-icon`, `${testID}-outline`). They have been removed entirely.

If you were relying on internal test IDs, update your tests not to rely on internal implementation details and only interact with elements or assert content your users can reach, e.g.: query by role, label, text etc., or `testID` props accepted by the component.
```

src/components/Searchbar.tsx
```diff
@@ -345,7 +340,6 @@ const Searchbar = ({
               backgroundColor: colors.outline,
             },
           ]}
-          testID={testID ? `${testID}-divider` : undefined}
         />
       )}
     </Surface>
```
satya164, line 360: since SearchBar has interactable elements in it, we should expose a way to access them with test IDs. add following props (similar to how we accept accessibility labels):

- `trailingTestID`
- `clearTestID`
- `searchTestID`

also I see it has a prop called `traileringIcon` which seems like a typo, idk a word named "trailering". lets also rename it to "trailing" in a follow-up PR (also check if the same typo is anywhere else), unless i'm missing something, then correct me.

src/components/Snackbar.tsx
```diff
@@ -339,7 +339,6 @@ const Snackbar = ({
                 }
                 aria-label={iconAccessibilityLabel}
                 style={styles.icon}
-                testID={testID ? `${testID}-icon` : undefined}
```
satya164, line 342: Same as `iconAccessibilityLabel`, add `iconTestID`

src/components/Modal.tsx
```diff
@@ -235,7 +235,6 @@ function Modal({
         onPress={dismissable ? onDismissCallback : undefined}
         importantForAccessibility="no"
         style={[styles.backdrop, backdropStyle, backdropTransitionStyle]}
-        testID={testID ? `${testID}-backdrop` : undefined}
```
satya164, line 238: add `overlayTestID`, similar to `overlayAccessibilityLabel`

src/components/Card/Card.tsx
satya164: the location of this `testID` seems wrong. it should be on the interactable element, i.e. the `Pressable` when present, otherwise on the `Surface`.

src/components/Menu/Menu.tsx
satya164, line 688: we also need a `overlayTestID`

satya164 (changes requested):
few more test ID issues we should address:

- `Dialog` needs `overlayTestID`
- `Chip` needs `closeIconTestID`
- `Drawer.Item` needs `testID` on the `TouchableRipple`
- `Card.Title` needs to forward the `testID`

there are also some hardcoded test IDs left:

- `"activity-indicator"` in `Searchbar` - remove it
- `"options-select"` and `"select-page-dropdown-label"` in `DataTablePagination` - remove unless necessary for some reason for testing, then add explicit props.
- `"list-image"` - in `ListImage`
- `"tooltip-container"` in `Tooltip`
- `"touchable-ripple-underlay"` in `TouchableRipple`

I also notice hardcoded a11y labels in some cases. we should create follow ups to handle them. user needs to be able to override a11y labels for i18n.

docs/6.x/docs/guides/migration.md
k0ndee: Done, updated in the new commit.

docs/6.x/docs/guides/migration.md
k0ndee: Done, updated in the new commit.

src/components/Searchbar.tsx
k0ndee: Done, updated in the new commit. I will create a follow-up PR and scan the repo for typos across the board.

src/components/Snackbar.tsx
k0ndee: Done, updated in the new commit.

src/components/Modal.tsx
k0ndee: Done, updated in the new commit.

src/components/Card/Card.tsx
k0ndee: Done, updated in the new commit. I also noticed now that the `CardTitle` siltently dropped the testID and never reached the rendered `View` so I fixed it by destructuring `testID` explicitly and forwarding it to the container `View`, consistent with how `Card.Cover` and `Card.Actions` already support it.

src/components/Menu/Menu.tsx
k0ndee: Done, updated in the new commit.

docs/6.x/docs/guides/migration.md
```diff
@@ -71,61 +71,41 @@ You can use the component's color prop where available, or override the correspo
...
-  - `surface-outer-layer`
 
-You can specify a `testID` explicitly to restore both the component's own test ID and all of its derived test IDs above, using the same suffixes.
+You can specify a `testID` explicitly to restore each component's own test ID.
+
+These components used to also derive test IDs for internal, implementation-only elements by appending a suffix to the `testID` prop (e.g. `${testID}-container`, `${testID}-icon`, `${testID}-outline`). They have been removed entirely.
+
+If you were relying on internal test IDs, update your tests not to rely on internal implementation details and only interact with elements or assert content your users can reach, e.g.: query by role, label, text etc., or `testID` props accepted by the component.
+
+Some components now accept explicit `testID` props for their interactable, internal elements:
```
satya164:
```suggestion
Some components now accept explicit `testID` props for their interactable elements:
```

satya164 (approved):
thanks

PR #5100, opened 2026-09-03
feat(theme): Add medium and high contrast theme

### LINKED ISSUE #5092: Add MD3 medium and high contrast levels
The theme only generates the MD3 `standard` contrast level. MD3 defines three levels, and the higher two exist for accessibility: they raise the contrast between text and background so the UI stays readable for users with low vision and in bright light. 

Add the `medium` and `high` schemes for both light and dark, and a way to select the active level, with `standard` staying the default so existing apps are unaffected.

**Additional context**
- MD3 spec: https://m3.material.io/styles/color/system/overview

hristototov:
### Motivation

The theme only generates the MD3 `standard` contrast level. MD3 defines three levels, and the higher two exist for accessibility: they raise the contrast between text and background so the UI stays readable for users with low vision and in bright light. Without them, apps built with Paper have no supported way to meet the raised contrast ratios.

This adds the `medium` and `high` schemes for light and dark, and a `contrast` prop on `PaperProvider` to select them. `standard` stays the default and its colours do not change. It uses `scripts/generate-contrast-tokens.ts`

Measured against the baseline `#6750A4` seed:

| scheme | `primary` tone | `onPrimaryContainer` tone |
| --- | --- | --- |
| light / standard | 40.1 | 10.1 |
| light / medium | 26.7 | 100.0 |
| light / high | **13.3** | 86.9 |
| dark / high | **96.7** | 23.8 |

**API**

```tsx
<PaperProvider contrast="high">
  <App />
</PaperProvider>
```

### Related issue

Closes #5092

### Test plan

Automated:

- `src/theme/__tests__/contrast.test.ts` checks 17 text/background role pairs across both modes and both new levels, 64 pairs in total, and asserts >= 4.5 for medium and >= 7.0 for high. It also asserts that `LightTheme` and `DarkTheme` colours are unchanged.
- `src/core/__tests__/PaperProvider.test.tsx` covers the `contrast` prop, the `standard` default, prop precedence over `theme.contrast`, and that a provider with only `contrast` set still follows the system color scheme.
- **No component snapshot colours changed.** The only snapshot diff is the new `contrast` field on the serialized theme.

Manual, on an Android 12 emulator (API 31), using the new Contrast control in the example app drawer under Preferences:

1. Open the drawer, switch between Standard, Medium and High. Colours change at each step.
2. Toggle Dark Theme and repeat. All six combinations render correctly with no missing colours.
3. With "Use Dynamic Theme" on, note that Standard shows the wallpaper-based colours while Medium and High fall back to the contrast-correct scheme. This is the intended behaviour described above.
4. Restart the app. The selected level is restored from storage.

https://github.com/user-attachments/assets/850afa25-aa10-48d7-9df1-57567348ba2d

https://github.com/user-attachments/assets/7f5a13d8-6a37-4827-b15a-07d832b4a011


src/theme/tokens/sys/color.ts
```diff
@@ -10,11 +11,11 @@ type PaletteKey = keyof Palette;
 /** Roles that map 1:1 to a palette key. Excludes the computed fields. */
 type MappedRoles = Omit<ThemeColors, 'stateLayerPressed' | 'elevation'>;
 
-type Contrast = 'standard'; // extend with 'medium' | 'high' when those ship
-
+/** Only `standard` uses the reference palette steps. The other levels need
+ *  tones that are not in the palette, so they live in `./contrastSchemes`. */
```
JKobrynski: These don't match MD3's published contrast schemes. On the [color roles page](https://m3.material.io/styles/color/roles), light/high `primary` is `md.ref.palette.primary20` (`#381E72`), but this ships `#312259`; `secondary`, `outline`, `outline-variant` and `on-surface-variant` are off too, and `surface-container` isn't meant to change with contrast at all. Every tone MD3 asks for is already a key of `ref/palette.ts` - could `roleToTone` just gain `medium` and `high` maps, the way the old `// extend with 'medium' | 'high' when those ship` comment planned?

src/core/PaperProvider.tsx
```diff
@@ -13,36 +13,39 @@ import { useSystemColorScheme } from './useSystemColorScheme';
...
+    const level = contrast ?? props.theme?.contrast ?? 'standard';
+    const base = getTheme(isDark, level);
     const scale = resolvedReduceMotion
       ? 0
       : (props.theme?.animation?.scale ?? 1);
 
     return {
       ...base,
       ...props.theme,
+      contrast: level,
```
JKobrynski: Passing a `theme` that has its own `colors` silently overrides the contrast scheme, but `theme.contrast` still reports the level you asked for - so `useTheme().contrast` says `"high"` while the colours on screen are standard. Worth reconciling, given the docs tell people to read the level back off the theme.

example/src/index.tsx
```diff
@@ -98,15 +96,12 @@ export default function PaperExample() {
...
   const theme =
     dynamicThemeSupported && shouldUseDynamicTheme
-      ? isDarkMode
-        ? DynamicDarkTheme
-        : DynamicLightTheme
-      : isDarkMode
-        ? DarkTheme
-        : LightTheme;
+      ? getDynamicTheme(isDarkMode, contrast)
+      : createTheme({ dark: isDarkMode, contrast });
```
JKobrynski: This builds a new theme object on every render, so the provider rebuilds the theme and every themed component re-renders with it. `getTheme(isDarkMode, contrast)` returns the cached one worth using that here?

JKobrynski:
And we've got conflicts

src/core/PaperProvider.tsx
hristototov: Good catch. The behaviour is intentional but was undocumented, which is the actual problem: contrast records which scheme was selected, and anything passed via theme.colors is layered on top, so an override survives a contrast change.

I've added code comments making that explicit.

example/src/index.tsx
hristototov: updated

src/theme/tokens/sys/color.ts
hristototov: updated - I've removed the generator and hand-written the medium and high maps as you suggested.

src/theme/tokens/sys/color.ts
```diff
@@ -67,6 +70,106 @@ const roleToTone: Record<
...
+      onBackground: 'neutral0',
+      outline: 'neutralVariant20',
+      outlineVariant: 'neutralVariant40',
+      inverseSurface: 'neutral20',
+      inverseOnSurface: 'neutral95',
+      inversePrimary: 'primary95',
+      primaryFixed: 'primary90',
+      primaryFixedDim: 'primary80',
+      onPrimaryFixed: 'primary10',
+      onPrimaryFixedVariant: 'primary30',
```
satya164: I get a contrast ratio of 5.47:1 for `onPrimaryFixedVariant` (`#4F378B`) on `primaryFixedDim` (`#D0BCFF`), below the documented 7:1 target for high contrast. The dark/high mapping at lines 311-313 uses the same values. Could we adjust this pair and include it in the contrast assertions? `CONTRAST_PAIRS` currently checks `onPrimaryFixed` on `primaryFixed`, so it misses this combination.

src/theme/tokens/sys/color.ts
```diff
@@ -67,6 +70,106 @@ const roleToTone: Record<
...
+      secondaryFixed: 'secondary90',
+      secondaryFixedDim: 'secondary80',
+      onSecondaryFixed: 'secondary10',
+      onSecondaryFixedVariant: 'secondary30',
```
satya164: I get 5.45:1 for `onSecondaryFixedVariant` (`#4A4458`) on `secondaryFixedDim` (`#CCC2DC`), below the documented 7:1 high-contrast target. The dark/high mapping at lines 315-317 has the same issue. Could we adjust this pair and cover it in the contrast assertions as well?

src/theme/tokens/sys/color.ts
```diff
@@ -67,6 +70,106 @@ const roleToTone: Record<
...
+      tertiaryFixed: 'tertiary90',
+      tertiaryFixedDim: 'tertiary80',
+      onTertiaryFixed: 'tertiary10',
+      onTertiaryFixedVariant: 'tertiary30',
```
satya164: I get 5.49:1 for `onTertiaryFixedVariant` (`#633B48`) on `tertiaryFixedDim` (`#EFB8C8`), below the documented 7:1 high-contrast target. The dark/high mapping at lines 319-321 has the same issue. Could we adjust this pair and cover it in the contrast assertions as well? The test that requires all fixed colors to stay unchanged across contrast levels would also need to be revised.

satya164 (changes requested):
Any additional themes should just be exported theme objects, e.g.:

- `MediumContrastLightTheme`
- `MediumContrastDarkTheme`
- `HighContrastLightTheme`
- `HighContrastDarkTheme`

Similarly there can be variants for dynamic themes.

It's a much simpler concept since we already export themes and accept `theme` prop, and there's no need for additional helpers and utilities. "Building" a theme object with `createTheme({ dark: true, contrast: 'high' })` instead of just importing `HighContrastDarkTheme` is unnecessary indirection.

src/theme/tokens/sys/color.ts
hristototov: updated

src/theme/tokens/sys/color.ts
hristototov: updated

src/theme/tokens/sys/color.ts
hristototov: updated

hristototov:
> Any additional themes should just be exported theme objects

Done. Removed `createTheme`, `getDynamicTheme`, `isDynamicColorSupportedAtContrast`, the `contrast` prop, and the `contrast` field on `Theme`. `PaperProvider` is back to its state on main.

Now exported:

- `MediumContrastLightTheme` / `MediumContrastDarkTheme`
- `HighContrastLightTheme` / `HighContrastDarkTheme`
- `MediumContrastDynamicLightTheme` / `HighContrastDynamicLightTheme` and the dark equivalents

PR #5102, opened 2026-09-04
fix: residual md3 deviations in switch

likevy:
### Motivation

`Switch` is one of the MD3 reference components, but a review against the spec found small residual deviations, mostly accessibility gaps.
The significant one is **breaking**: a switch must now declare how it can be operated — `onValueChange`, `readOnly`, or `disabled`. `readOnly` keeps the enabled appearance and stays in the accessibility tree so its state is announced, and carries `aria-readonly` on web; it is *not* reported as disabled. Enforced by the props type plus a runtime guard for untyped callers.

Two things worth mentioning:

- **Drag is deliberately excluded.** MD3 doesn't specify it
- **The props type shape is deliberate.** The docs generator reads props off the component's *parameter* annotation and drops union members, so the parameter keeps the flat `Props` and the operability union sits on the component's declared type. `SwitchProps` exports the narrowed type, so the public type matches what the component accepts. Commented in place so it isn't "simplified" later.

### Related issue
This closes the Switch review checklist:

- [x] Selected icon color uses `primary`; MD3 spec is `onPrimaryContainer`.
- [x] Vertical touch target is 40dp (the state-layer box); MD3 requires 48dp.
- [x] Keyboard focus applies neither the focus handle color nor a focus state layer.
- [x] Renders an enabled, focusable `role="switch"` with no `onValueChange` (an enabled no-op); require or infer explicit operability (handler / read-only / disabled).
- [x] No drag gesture (tap only), and the delayed handle-growth (100ms) can swallow immediate press feedback. — press feedback fixed; drag **not** added, see above.

### Test plan

`yarn lint`, `yarn typecheck` and `yarn test` pass — 684 tests, 168 snapshots. Each commit is independently green, so the series bisects cleanly.

9 new tests in `Switch.test.tsx` cover the focus indicator, state layer and handle colour, the 52×48 touch target, and operability. The state-layer and handle-colour assertions fail if the focus fix is reverted (mutation-tested).

Manual, on the Switch example screen:

1. Press and hold - the handle grows immediately, no lag.
2. Tap just above/below the track, outside the visible 32dp - still toggles (48dp target); rows are ~8dp taller.
3. "Default with icon when on" - the glyph is dark (`onPrimaryContainer`), not purple.
4. New "Read-only" row - can't be toggled, still looks enabled rather than dimmed.
5. Drawer, FAB and TextInput example toggles still work - now `readOnly`, with the row owning the press.
6. Tab to a switch (web, or a hardware keyboard) - focus ring, state layer, `primaryContainer` handle; a mouse click must *not* light the ring. Read-only is skipped in the tab order and has `aria-readonly="true"`.
7. Screen reader - a read-only switch announces its on/off state, not disabled or actionable.

Run on Android, iOS and web, plus a screen-reader pass. All behaved as described.

### DIFF
--- a/src/components/Switch/Switch.tsx
+++ b/src/components/Switch/Switch.tsx
@@ -68,6 +68,7 @@ const {
   trackHeight: TRACK_HEIGHT,
   trackOutlineWidth: TRACK_OUTLINE_WIDTH,
   stateLayerSize: STATE_LAYER_SIZE,
+  touchTargetSize: TOUCH_TARGET_SIZE,
   selectedHandleSize: SELECTED_HANDLE,
   unselectedHandleSize: UNSELECTED_HANDLE,
   iconHandleSize: ICON_HANDLE,
@@ -86,7 +87,10 @@ const stateOpacity = stateTokens.opacity;
 const { thickness: FOCUS_THICKNESS, outerOffset: FOCUS_OUTER_OFFSET } =
   stateTokens.focusIndicator;
 const FOCUS_RING_INSET = -(FOCUS_OUTER_OFFSET + FOCUS_THICKNESS);
-const OVERLAY_TOP = (STATE_LAYER_SIZE - TRACK_HEIGHT) / 2;
+// Every painted layer is smaller than the touch target and absolutely
+// positioned, so each one is centred against it.
+const centreY = (size: number) => (TOUCH_TARGET_SIZE - size) / 2;
+const TRACK_TOP = centreY(TRACK_HEIGHT);
 
 // Hold-then-grow: a brief delay before snapping to PRESSED_HANDLE so a quick
 // tap doesn't flash the press-grow visual.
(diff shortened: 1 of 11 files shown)


src/components/Switch/Switch.tsx
```diff
@@ -393,6 +460,7 @@ const Switch = ({
       </Pressable>
 
       <Animated.View
+        testID={testID ? `${testID}-state-layer` : undefined}
```
JKobrynski, line 463: #5099 removed every `${testID}-...` from `src/components`

src/components/Switch/Switch.tsx
```diff
@@ -345,30 +402,40 @@ const Switch = ({
...
+        },
+      }
+    : null;
+
+  return (
+    <View style={[styles.wrapper, style]}>
+      <Pressable
+        disabled={disabled}
+        focusable={isInteractive}
+        aria-readonly={isAnnouncedReadOnly}
```
JKobrynski, line 437: Have you tried the read-only switch with VoiceOver or TalkBack? I'm afraid RN has no `aria-readonly` so on native it still announces as operable

PR #5104, opened 2026-09-04
feat: fix residual md3 deviation - checkbox component

likevy:
### Motivation

`Checkbox` is one of the MD3 reference components, but a review against the spec found residual deviations: the state-layer tokens were dead code, so every interaction fell back to `TouchableRipple`'s neutral ripple; the touch target was 40dp; the focus ring sat on the state-layer edge; and a standalone checkbox had no accessible-name contract.

Things worth mentioning:

- **The checkbox paints its own press.** MD3 draws the press as a ripple bounded to the 40dp state layer in the *inverted* role (Compose `Checkbox.kt`: `ripple(bounded = false, radius = StateLayerSize / 2)`; material-web `_checkbox.scss`: `md-ripple` sized to the state layer). `TouchableRipple` can't get there: Android's ripple only accepts a numeric colour, so a dynamic theme's `PlatformColor` roles can never tint it; the ripple and iOS highlight fill the whole pressable, now 48dp; and on web its hover overlay doubles up with the layer. So the platform press is turned off (`rippleColor: 'transparent'`, as `Switch` and `BottomNavigation` already do) and a 40dp ripple is drawn with Reanimated shared values — static colour, only `opacity`/`transform` animate, 200 ms grow with a 200 ms minimum hold so a quick tap still reads (Compose `RippleAnimation`, material-web `MINIMUM_PRESS_MS`). An explicit `rippleColor` or `underlayColor` hands the press back to the platform; `rippleEffectEnabled: false` still disables it.
- **The flat hover/focus layer fades by opacity only**, keeping its colour static, because Reanimated cannot interpolate a `PlatformColor`

Visible behaviour changes, both in the migration guide: the checkbox occupies 48dp instead of 40dp, so rows get ~8dp taller; interaction tints follow selection instead of being neutral. A standalone `Checkbox` with no accessible name now warns.

### Related issue

Checkbox review checklist:

- [x]  Standalone checkbox touch target is 40dp (no default `hitSlop`); MD3 requires 48dp (the 40dp state layer itself already matches).
- [x]  State-layer colours don't follow the per-state spec, which **inverts on press** — hover + focus: selected `primary` / unselected `onSurface`; **pressed: selected `onSurface` / unselected `primary`**. The lib applies one colour per selection state across all three interaction states (`src/components/Checkbox/tokens.ts:24-25`), so hover/focus are correct and pressed is wrong in both directions. Do **not** simply swap the two constants. Error states already match (`error` throughout).
- [x]  Focus ring omits the 2dp outer offset.
- [x]  Standalone Checkbox has no visible-label / whole-row accessible-name contract (`Checkbox.Item` is the labeled wrapper) — document and guarantee the standalone accessible-name path.

### Test plan

`yarn lint`, `yarn typecheck` and `yarn test` pass — 738 tests, 168 snapshots. Each commit is independently green, so the series bisects cleanly.

60 new tests in `Checkbox.test.tsx` and `utils.test.tsx` cover the state-layer colour per interaction (custom colours and the pressed inversion included), the ripple on all three `Platform.OS` values, the minimum-press hold, the fade-out duration, the scale reset on a second press, reduce motion, the `rippleColor`/`underlayColor` escape hatches, `rippleEffectEnabled: false`, `PlatformColor` roles, the 48×48 target, the 50dp ring geometry and the accessible-name warning. Mutation-tested: removing the inversion, the hold, the scale reset, the fade duration or the custom-colour path each fails its test.

Manual, on the Checkbox example screen:

1. Press and hold "Default" - a 40dp ripple grows inside the state layer: `primary` when unchecked, `onSurface` when checked. On Android with dynamic colour on it takes the system accent.
2. Quick tap - the ripple still completes before fading; a rapid double tap doesn't leave one behind.
3. "Error" - `error` ripple. "Custom color (tertiary)" - tertiary ripple, no `primary` halo.
4. Tap just outside the visible 40dp - still toggles (48dp target); rows are ~8dp taller.
5. Web: hover paints the flat layer in the non-inverted colour with no doubled overlay; Tab to a checkbox - 3dp `secondary` ring with a 2dp gap around the 40dp layer; a mouse click must *not* light the ring.
6. Drawer → "Ripple effect" off - no press paint, hover and focus unaffected. Dynamic theme off - baseline `primary`.
7. Screen reader - a standalone checkbox with `aria-label` announces its name and state; `Checkbox.Item` announces once for the row; one without a name logs the warning.

https://github.com/user-attachments/assets/d0fc7713-f4d4-48e9-8669-cb3be445ed33

https://github.com/user-attachments/assets/c3b321e4-23cb-4d58-bd21-85d4e97c9dc1

https://github.com/user-attachments/assets/ee4819ca-45b0-490e-850b-66b70ee23907

### DIFF
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -279,3 +279,9 @@ const theme = {
   style={{ fontSize: 16, color: '#1C1B1F' }}
 />
 ```
+
+### Checkbox
+
+#### Interaction state colors
+
+Hover and focus are now painted as a Material Design 3 state layer, so the tint follows the spec: a flat 40dp layer with `primary` when selected and `onSurface` when not, and `error` throughout an error checkbox. `color` and `uncheckedColor` replace the role they already override on the box, so a custom checkbox no longer picks up a `primary` halo.
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -292,4 +292,4 @@ Interaction states are now painted as a Material Design 3 state layer on every p
 
 #### Custom style target
 
-The `style` prop now applies to the checkbox's outer container rather than the pressable inside it, so layout styles such as `margin`, `position` and `transform` move the whole component, focus ring included. `width` and `height` no longer resize the 48dp tap target, and paint styles land on a different shape: a `backgroundColor` used to fill the circular pressable and now fills the square container around it.
+The `style` prop now applies to the checkbox's outer container rather than the pressable inside it, so layout styles such as `margin`, `position` and `transform` move the whole component, focus ring included. `width` and `height` no longer resize the 48dp tap target, and paint styles land on a different shape: a `backgroundColor` used to fill the circular pressable and now fills the square container around it. Resizing the tap target now lives on the new `tapTargetStyle` prop, which reaches the pressable itself.
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -289,3 +289,7 @@ Interaction states are now painted as a Material Design 3 state layer on every p
 #### Touch target height
 
 `Checkbox` now reserves the 48dp minimum touch target, so it occupies 48dp instead of 40dp. Nothing painted changed size, but rows containing a checkbox may become slightly taller.
+
+#### Custom style target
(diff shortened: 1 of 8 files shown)


JKobrynski (approved):
LGTM, @satya164

PR #5110, opened 2026-09-08
fix(fab): align MD3 variants and interaction states

likevy:
### Motivation

`FAB` is one of the MD3 reference components, but a review against the spec found two deviations: the color presets didn't match the MD3 variant set, and the hover-elevation token was defined but never applied.

The variant rename is **breaking**: `tonalPrimary`, `tonalSecondary` and `tonalTertiary` become `primaryContainer`, `secondaryContainer` and `tertiaryContainer`, with no aliases, and `surface` and `branded` are added. That makes the preset names the MD3 role names, so the mapping is checkable against the spec.

This worth mentioning:

- **The large FAB icon stays 36dp, deliberately.** `md.comp.fab.large.icon.size` is 36dp on every variant, and material-web's generated `_md-comp-fab-branded.scss` emits `icon-size: 36px`. The widely-quoted 32dp comes from a token androidx annotates as wrong in its own source (`LargeIconSize: Dp = 36.dp // TODO: FabLargeTokens.IconSize is incorrect`). Stated explicitly because 32dp is easy to find and looks authoritative.
- **`branded`'s content color is a judgement call.** material-web defines container and elevation for the branded FAB but **no `icon-color` token at all** - brand artwork supplies its own colors. `onSurface` is used as the fallback for the label and a monochrome icon; passing a custom icon source keeps brand artwork in its own colors. `surface` needs no such call: the spec gives it `surface-container-high` with a `primary` icon, which is what it does.
- **Hover elevation is web-only, and the FAB Menu is deliberately excluded.** 
- **A hidden FAB no longer keeps focus or its action.** 

Token values were verified against material-web's generated `_md-comp-fab-surface.scss` and `_md-comp-fab-branded.scss` - the spec site is JS-rendered.

### Related issue

The FAB review checklist:

- [x] Rename the color variants to match the m3 set: `tonalPrimary`/`tonalSecondary`/`tonalTertiary` → `primaryContainer`/`secondaryContainer`/`tertiaryContainer`, and add the missing `surface` and `branded` variants (both `surfaceContainerHigh`). Delete the old names - no alias.
- [x] The large FAB icon size is unchanged at 36dp, and a test pins every variant's resolved colors in both schemes so the set can't drift.
- [x] The hover-elevation token (elevation level 4) is defined but never applied on web — wire it through the interaction state. (FAB Menu behavior/web is a separate task.)

Also included, under the umbrella task's accessibility criterion rather than the checklist: a hidden FAB no longer retains focus, actions, or an accessibility node.

### Test plan

`yarn lint`, `yarn typecheck` and `yarn test` pass - 690 tests, 168 snapshots.

12 new tests (FAB: 54 → 66) cover every variant's container and content role in both light and dark, hover elevation on web for the icon and extended FABs, that an explicit shell elevation (including a flat menu item) is not overridden, that native platforms stay at the resting elevation, that the menu trigger is unaffected, that adding interaction handlers doesn't enable an actionless FAB, and that hiding a FAB removes it from the accessibility tree, disables its action, and resets its elevation.

Manual, on the FAB example screen:

1. Cycle the Color chips - all eight variants render; `surface` and `branded` share `surfaceContainerHigh`, with a `primary` icon and an `onSurface` icon respectively.
2. Web: hover a FAB - the shadow deepens (level 3 → 4); focus and press stay at level 3. Hover the menu trigger - it does not rise.
3. Switch Type to `menu` while `surface` or `branded` is selected - the chip set drops both and the variant falls back to `primaryContainer`.
4. Toggle the FAB hidden - it can't be focused by Tab, pressing does nothing, and a screen reader skips it. Show it again and hover: elevation starts at rest, not raised.
5. Size `large` - the icon measures 36dp, not 32dp.
6. Screen reader - the icon FAB announces its `aria-label`; the hidden FAB is not announced.

Screenshot 2026-09-08 at 09 15 29
Screenshot 2026-09-08 at 09 14 49

Screenshot 2026-09-08 at 09 15 11

### DIFF
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -161,21 +161,30 @@ e.g.:
 
 ### FAB
 
-The color presets for `FAB` and `FAB.Extended` now match the MD3 color roles:
-`primary`, `primaryContainer` (default), `secondary`, `secondaryContainer`,
-`tertiary`, `tertiaryContainer`, `surface`, and `branded`.
+To preserve the v5 FAB color treatment, update the `variant` prop:
 
-If you used an earlier 6.x build, replace `tonalPrimary`, `tonalSecondary`, and
-`tonalTertiary` with `primaryContainer`, `secondaryContainer`, and
-`tertiaryContainer`. The old names have been removed without aliases. This also
-applies to the FAB menu trigger's `variant`.
+| v5 | v6 |
+| --- | --- |
+| `primary` | `primaryContainer` |
+| `secondary` | `secondaryContainer` |
+| `tertiary` | `tertiaryContainer` |
 
-Both `surface` and `branded` use `surfaceContainerHigh`. Surface FAB content uses
-`primary`; branded content defaults to `onSurface`. A custom icon source can
-render brand artwork in its own colors.
+If you omit `variant`, no change is needed. Replace `variant="surface"` with
+one of the supported color variants, such as `primaryContainer`.
 
-On web, hovering a FAB raises its elevation from level 3 to level 4. Focused and
-pressed states use level 3. The large FAB icon remains **36dp**.
+For custom colors, replace `color` with `contentColor` and move
+`style.backgroundColor` to `containerColor`:
+
+```diff
+<FAB
+  icon="plus"
+- color="#ffffff"
+- style={{ backgroundColor: '#6750a4', position: 'absolute', bottom: 16, right: 16 }}
++ contentColor="#ffffff"
++ containerColor="#6750a4"
++ style={{ position: 'absolute', bottom: 16, right: 16 }}
+/>
+```
 
 ### TextInput
 
--- a/docs/6.x/docs/guides/migration.md
+++ b/docs/6.x/docs/guides/migration.md
@@ -159,6 +159,24 @@ e.g.:
 - The default elevation changed from level `1` to level `3`.
 - The `style` prop no longer configures the background color or border radius. You can override `theme.colors.surfaceContainerHigh` and `theme.shapes.corner.extraLarge` using the `theme` prop instead.
 
+### FAB
+
+The color presets for `FAB` and `FAB.Extended` now match the MD3 color roles:
+`primary`, `primaryContainer` (default), `secondary`, `secondaryContainer`,
+`tertiary`, `tertiaryContainer`, `surface`, and `branded`.
+
+If you used an earlier 6.x build, replace `tonalPrimary`, `tonalSecondary`, and
+`tonalTertiary` with `primaryContainer`, `secondaryContainer`, and
+`tertiaryContainer`. The old names have been removed without aliases. This also
+applies to the FAB menu trigger's `variant`.
+
+Both `surface` and `branded` use `surfaceContainerHigh`. Surface FAB content uses
+`primary`; branded content defaults to `onSurface`. A custom icon source can
+render brand artwork in its own colors.
+
+On web, hovering a FAB raises its elevation from level 3 to level 4. Focused and
+pressed states use level 3. The large FAB icon remains **36dp**.
+
 ### TextInput
 
 The Paper 6.x `TextInput` is a complete rewrite with a new API. Import the component the same way, but note that the props and behavior have changed significantly.
--- a/src/components/FAB/Extended.tsx
+++ b/src/components/FAB/Extended.tsx
@@ -38,7 +38,12 @@ export type Props = {
    */
   label: string;
   /**
-   * Role-color preset. Defaults to `tonalPrimary`.
+   * Role-color preset. Defaults to `primaryContainer`.
+   * Choose `primary`, `primaryContainer`, `secondary`, `secondaryContainer`,
+   * `tertiary`, `tertiaryContainer`, `surface`, or `branded`.
+   * `surface` and `branded` use `surfaceContainerHigh`; `surface` uses primary
+   * content, while `branded` uses on-surface content. Pass a custom icon source
+   * to preserve brand artwork colors.
    */
   variant?: Variant;
   /**
@@ -154,7 +159,7 @@ export type Props = {
 const Extended = ({
   icon,
   label,
-  variant = 'tonalPrimary',
+  variant = 'primaryContainer',
   containerColor,
   contentColor,
   size = 'default',
--- a/src/components/FAB/Extended.tsx
+++ b/src/components/FAB/Extended.tsx
@@ -39,11 +39,6 @@ export type Props = {
   label: string;
   /**
    * Role-color preset. Defaults to `primaryContainer`.
-   * Choose `primary`, `primaryContainer`, `secondary`, `secondaryContainer`,
-   * `tertiary`, `tertiaryContainer`, `surface`, or `branded`.
-   * `surface` and `branded` use `surfaceContainerHigh`; `surface` uses primary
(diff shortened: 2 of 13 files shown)


docs/6.x/docs/guides/migration.md
```diff
@@ -208,6 +159,24 @@ e.g.:
...
+`tonalTertiary` with `primaryContainer`, `secondaryContainer`, and
+`tertiaryContainer`. The old names have been removed without aliases. This also
+applies to the FAB menu trigger's `variant`.
+
+Both `surface` and `branded` use `surfaceContainerHigh`. Surface FAB content uses
+`primary`; branded content defaults to `onSurface`. A custom icon source can
+render brand artwork in its own colors.
+
+On web, hovering a FAB raises its elevation from level 3 to level 4. Focused and
+pressed states use level 3. The large FAB icon remains **36dp**.
```
satya164: rewrite this to actual migration instructions. don't include implementation details or what it was in a prerelease version.

src/components/FAB/tokens.ts
```diff
@@ -68,18 +70,20 @@ const stateElevation = {
 } as const satisfies Record<string, Elevation>;
 
 const variants = {
+  surface: { container: 'surfaceContainerHigh', content: 'primary' },
+  branded: { container: 'surfaceContainerHigh', content: 'onSurface' },
```
satya164: is this actually in the official guidelines?

src/components/FAB/tokens.ts
```diff
@@ -68,18 +70,20 @@ const stateElevation = {
 } as const satisfies Record<string, Elevation>;
 
 const variants = {
+  surface: { container: 'surfaceContainerHigh', content: 'primary' },
```
satya164: this looks like it's no longer recommended. remove this

> Surface FAB color styles are still available, but no longer recommended.

src/components/FAB/Shell.tsx
```diff
@@ -186,15 +186,15 @@ export type ShellProps = {
...
+  variant = 'primaryContainer',
   size = 'default',
   containerColor,
   contentColor,
   shape,
   iconSize,
   leading,
   trailing,
-  elevation = Tokens.stateElevation.enabled,
+  elevation,
```
satya164: the `elevation` prop should be removed if `elevation` is supposed to be dynamic.

src/components/FAB/Shell.tsx
```diff
@@ -299,6 +316,18 @@ const Shell = ({
 
   const { focusedSV, onFocus, onBlur } = useFocusRing();
 
+  React.useEffect(() => {
+    if (!visible || !onPress) {
+      if (Platform.OS === 'web') {
+        touchableRef.current?.blur();
+      }
```
satya164: doing this means user's focus is removed from the page and they lose their position. removing focus on toggling `visible` is defensible, though it should move focus back to previous element. but doing it when `onPress` becomes `undefined` seems unnecessary.

src/components/FAB/Shell.tsx
```diff
@@ -212,11 +212,28 @@ const Shell = ({
...
   theme: themeOverrides,
   ref,
 }: ShellProps) => {
   const theme = useInternalTheme(themeOverrides);
+  const [hovered, setHovered] = React.useState(false);
+  const [pressed, setPressed] = React.useState(false);
+  const [focused, setFocused] = React.useState(false);
+  const touchableRef = React.useRef<View>(null);
+
+  // Explicit elevations (including flat menu items) keep their own treatment.
```
satya164: remove unnecessary comment

src/components/FAB/Shell.tsx
```diff
@@ -212,11 +212,28 @@ const Shell = ({
...
+    elevation ??
+    (Platform.OS === 'web' && visible && onPress
+      ? pressed
+        ? Tokens.stateElevation.pressed
+        : focused
+          ? Tokens.stateElevation.focus
+          : hovered
+            ? Tokens.stateElevation.hover
+            : Tokens.stateElevation.enabled
+      : Tokens.stateElevation.enabled);
```
satya164: what's the rationale behind different behavior on web and native?

src/components/FAB/Shell.tsx
```diff
@@ -212,11 +212,28 @@ const Shell = ({
...
+  // Explicit elevations (including flat menu items) keep their own treatment.
+  const resolvedElevation =
+    elevation ??
+    (Platform.OS === 'web' && visible && onPress
+      ? pressed
+        ? Tokens.stateElevation.pressed
+        : focused
+          ? Tokens.stateElevation.focus
+          : hovered
+            ? Tokens.stateElevation.hover
```
satya164: a hovered element can also be focused. focus shouldn't take priority over hover styling.

src/components/FAB/Shell.tsx
```diff
@@ -318,18 +348,30 @@ const Shell = ({
...
         {overlay}
         <TouchableRipple
+          ref={Platform.OS === 'web' ? touchableRef : undefined}
           borderless
           background={background}
-          onPress={onPress}
-          onFocus={onFocus}
-          onBlur={onBlur}
+          onPress={visible ? onPress : undefined}
+          disabled={!onPress || !visible}
```
satya164: why is this needed? `disabled` changes how a11y tree sees the button. a hidden button shouldn't be accessible at all rather than being marked as disabled

src/components/FAB/FAB.tsx
```diff
@@ -21,7 +21,12 @@ export type Props = {
    */
   icon: IconSource;
   /**
-   * Role-color preset. Defaults to `tonalPrimary`.
+   * Role-color preset. Defaults to `primaryContainer`.
+   * Choose `primary`, `primaryContainer`, `secondary`, `secondaryContainer`,
+   * `tertiary`, `tertiaryContainer`, `surface`, or `branded`.
+   * `surface` and `branded` use `surfaceContainerHigh`; `surface` uses primary
+   * content, while `branded` uses on-surface content. Pass a custom icon source
+   * to preserve brand artwork colors.
```
satya164: remove

src/components/FAB/Extended.tsx
```diff
@@ -38,7 +38,12 @@ export type Props = {
    */
   label: string;
   /**
-   * Role-color preset. Defaults to `tonalPrimary`.
+   * Role-color preset. Defaults to `primaryContainer`.
+   * Choose `primary`, `primaryContainer`, `secondary`, `secondaryContainer`,
+   * `tertiary`, `tertiaryContainer`, `surface`, or `branded`.
+   * `surface` and `branded` use `surfaceContainerHigh`; `surface` uses primary
+   * content, while `branded` uses on-surface content. Pass a custom icon source
+   * to preserve brand artwork colors.
```
satya164: remove

satya164 (changes requested):

src/components/FAB/Extended.tsx
likevy: done

src/components/FAB/FAB.tsx
likevy: done

src/components/FAB/Shell.tsx
likevy: fixed

src/components/FAB/Shell.tsx
likevy: fixed

JKobrynski:
We've got conflicts @likevy

PR #5112, opened 2026-09-08
feat(motion): adopt spring tokens and add shared CSS-transition helper

rjmreis:
### Motivation

MD3 Expressive replaces the easing+duration motion model with spring physics where motion is genuinely spring-shaped, and the library's preferred mechanism otherwise is Reanimated CSS transitions. #5015 already handled the library swap (`Animated` → Reanimated) and the CSS-transition conversion for most components; this PR handles what it didn't: spring adoption, plus a shared helper so the CSS tier stops re-deriving duration/easing/reduce-motion by hand in every component.

Concretely:
- Added `getTransition(theme, property, durationToken, easingToken, reduceMotion)` in `theme/tokens/sys/motion.ts` — the shared CSS-transition helper.
- `Badge`: switched to the helper, fixing a hardcoded `150ms` duration (now `theme.motion.duration.short3`), a missing `transitionTimingFunction`, and reduce-motion wired through the legacy `theme.animation.scale` instead of `useReduceMotion()`.
- `Checkbox`: already correct (CSS transitions, `useReduceMotion()`, right tokens) — deduped onto the new helper, no behavior change.
- `TextInput`: converted the label-float and outlined active-border animation from imperative `withTiming`/shared-values/`interpolate()` to CSS transitions. Both only ever animate between two fixed endpoints per boolean state (focus/blur, text present or not) — not a continuous drag — so each derived property becomes an independent CSS transition driven by render state, which also let me delete the effect that used to hand-resync a controlled `value` change.
- `SegmentedButtonItem`: the checkmark's `withSpring` was using a hand-rolled raw spring config (`stiffness: 230.2, damping: 22`) instead of a theme token. Converted to `toRawSpring(theme.motion.spring.slow.spatial)`, which is the actual "spring adoption" this task is about.

**Not included, flagged separately**: while reviewing every `withTiming`/`withSpring` site in `src/`, I found several #5015-touched files that still animate with imperative `withTiming` rather than CSS transitions (`Banner`, `Snackbar`, `Drawer/DrawerCollapsedItem`, `Menu/Menu.tsx`, `CrossFadeIcon`, `DataTable/DataTableTitle`). None look spring-shaped, so they're CSS-transition (mechanism) candidates, not spring-adoption candidates — out of scope here per this task's "splittable" note. I'll file a follow-up issue for these and link it here.

### Related issue

This is part of the ongoing MD3 Expressive motion modernization (spring tokens + CSS-transition helper); there's no existing tracking issue for it yet. The out-of-scope items noted above will get their own follow-up issue, referenced here once filed.

### Test plan

- `yarn typecheck`, `yarn lint`, and `yarn jest` all pass (686 tests, 56 suites), including new coverage:
  - `src/theme/__tests__/motion.test.ts` — unit tests for `getTransition` (property passthrough, duration token lookup, reduce-motion zeroing, cubicBezier easing construction).
  - `src/components/__tests__/TextInput.test.tsx` — asserts the label wrapper's `top`/`fontSize`/`translateX` across focus/blur for both filled and outlined variants, now that they're plain computed values instead of worklet-interpolated ones.
- Manual verification (reviewers, run through this in the example app):
  - **Badge** (`Examples → Badge`): toggle visibility — opacity should fade smoothly at a normal, snappy speed, not pop instantly.
  - **Checkbox / Checkbox Item**: check/uncheck repeatedly, including rapid double-taps mid-transition — this is a dedup-only refactor, so behavior should be unchanged from before.
  - **TextInput** (filled + outlined): focus/blur with empty and pre-filled values, type then clear while focused, change a controlled value programmatically, rapid focus/blur between fields — label float and active-border/outline-slide animations should track correctly and interrupt cleanly with no glitches.
  - **Segmented Buttons** (`Examples → SegmentedButtons → With Selected Check`): selecting a segment should show the checkmark scale in with a slight spring overshoot, not a linear/eased grow.
  - Reduce Motion (OS-level accessibility setting): repeat the Badge and TextInput checks — transitions should be instant, not just faster.

##### Badge
https://github.com/user-attachments/assets/7d757377-aa6b-4277-821b-fa3486aa0399

#####  Checkbox / Checkbox Item
https://github.com/user-attachments/assets/d2315cbc-d357-446b-ab16-d1a7d4ba709e

#####  TextInput
https://github.com/user-attachments/assets/06dc5a96-28d3-47b9-869a-40182d0377a1

#####  Segmented Buttons
https://github.com/user-attachments/assets/f219fb2a-5fc2-481e-8484-f56d697da8eb

### DIFF
--- a/src/components/Badge.tsx
+++ b/src/components/Badge.tsx
@@ -4,6 +4,8 @@ import { StyleSheet } from 'react-native';
 import Animated, { type AnimatedStyle } from 'react-native-reanimated';
 
 import { useInternalTheme } from '../core/theming';
+import { useReduceMotion } from '../theme/accessibility/ReduceMotionContext';
+import { getTransition } from '../theme/tokens/sys/motion';
 import { cornerFull } from '../theme/tokens/sys/shape';
 import type { ThemeProp } from '../theme/types';
 
@@ -56,10 +58,7 @@ const Badge = ({
   ...rest
 }: Props) => {
   const theme = useInternalTheme(themeOverrides);
-
-  const {
-    animation: { scale },
-  } = theme;
+  const reduceMotion = useReduceMotion();
 
   const textColor = theme.colors.onError;
 
@@ -69,8 +68,7 @@ const Badge = ({
 
   const transitionStyle = {
     opacity: visible ? 1 : 0,
-    transitionDuration: 150 * scale,
-    transitionProperty: 'opacity',
+    ...getTransition(theme, 'opacity', 'short3', 'standard', reduceMotion),
   };
 
   return (
--- a/src/components/TextInput/hooks.ts
+++ b/src/components/TextInput/hooks.ts
@@ -1,29 +1,11 @@
-import {
-  useEffect,
-  useImperativeHandle,
-  useRef,
-  useState,
-  type RefObject,
-} from 'react';
+import { useImperativeHandle, useRef, useState, type RefObject } from 'react';
 import { TextInput as NativeTextInput } from 'react-native';
 import type { BlurEvent, FocusEvent } from 'react-native';
 
-import {
-  interpolate,
-  useAnimatedStyle,
-  useSharedValue,
-  withTiming,
-} from 'react-native-reanimated';
-
-import {
-  ACTIVE_LABEL_FONT_SIZE,
-  INACTIVE_LABEL_FONT_SIZE,
-  TIMING_CONFIG,
-} from './constants';
+import { ACTIVE_LABEL_FONT_SIZE, INACTIVE_LABEL_FONT_SIZE } from './constants';
 import type {
   TextInputAnimationState,
   TextInputFlags,
-  TextInputAnimationHandlers,
   TextInputHookReturn,
   TextInputLayoutState,
   TextInputProps,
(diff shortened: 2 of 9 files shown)


src/components/TextInput/hooks.ts
```diff
@@ -38,26 +20,33 @@ import {
...
+import { useReduceMotion } from '../../theme/accessibility/ReduceMotionContext';
+import { getTransition } from '../../theme/tokens/sys/motion';
 import type { InternalTheme } from '../../theme/types';
 
+// The label float (top/translateX/fontSize/opacity) and the outlined active
+// border (scaleX) are both driven by a single boolean state transition
+// (focus/blur, or text appearing/disappearing) — not a continuous or
+// interruptible drag — so each ends up a plain discrete CSS transition
+// between two fixed values per property, with no shared progress value
+// needed. That keeps this on the CSS tier per the motion decision rule.
```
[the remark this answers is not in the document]
rjmreis: Good catch, the code was already correctly scoped to the filled variant, the comment's wording was just wrong.

src/components/Badge.tsx
```diff
@@ -69,8 +68,7 @@ const Badge = ({
 
   const transitionStyle = {
     opacity: visible ? 1 : 0,
-    transitionDuration: 150 * scale,
-    transitionProperty: 'opacity',
+    ...getTransition(theme, 'opacity', 'short3', 'standard', reduceMotion),
   };
```
JKobrynski, line 75: Have you considered typing this style as `CSSStyle<ViewStyle>` like in `Checkbox.tsx`?

rjmreis: Good call, done! Badge renders `Animated.Text`, so I've updated it to `CSSStyle<TextStyle>`.

Issue #4929, opened 2026-05-11
refactor(card): modernize composition model and improve MD3 behavior


ruben-rebelo:
Refactor Card to improve Material Design 3 compliance, remove problematic children patterns, improve performance characteristics, and modernize the public API.

**Problems**

- Current API relies heavily on children composition in ways that are difficult to optimize
- Internal layout logic is tightly coupled to arbitrary child structures
- MD3 behaviors and visuals are not fully aligned with current Material guidelines
- Rendering performance can degrade in large lists or frequently updated screens
- Type safety and discoverability of the API can be improved
- The component has inconsistent behavior between platforms and edge cases

**Focus Areas**

- Reduce dependence on deeply nested children
- Improve Card.Content / Card.Actions architecture
- Improve elevation and surface behavior
- Improve shape and outlined variants
- Simplify internal spacing rules

**Proposed API Direction**

Introduce clearer slot-based APIs:
`<Card
  title="Title"
  subtitle="Subtitle"
  content={<Text>Content</Text>}
  actions={[...]}
/>`

**Performance Opportunities**

- Reduce nested wrappers
- Optimize pressable behavior
- Avoid unnecessary subtree re-renders
- Improve memoization boundaries

adrcotfas:
> [!IMPORTANT]  
See https://github.com/callstack/react-native-paper/discussions/4949 for more details

PR #5116, opened 2026-09-11
feat: Card MD3 Expressive improvements

pawelkata:
### Motivation

`Card` was still on the Paper 5 model: a `mode` prop, arbitrary children composition, and no first-class support for Material 3 expressive states (hover, focus, press, drag) or emphasis tokens. This made it hard to build Cards that match the Material 3 spec — variant-driven container/outline/elevation colors, state layers, focus indicators, and motion — without consumers hand-rolling that behavior on top of `View`/`Surface`.

This PR rewrites `Card` around the Material 3 contract:

- Replaces `mode` + arbitrary children with an explicit `variant` (`filled` | `elevated` | `outlined`) and deterministic slots (`media`, `header`/`title`+`subtitle`+`leading`+`trailing`, `content`, `actions`).
- Adds a single actionable target model: a Card with an interaction handler becomes one accessible control with button semantics; a Card without one stays a neutral grouping container, leaving independent actions to its `actions` slot.
- Introduces Material state layers, keyboard focus indicators, hover/press/drag visuals, and animated elevation/outline transitions driven by a `resolveCardVisuals` token contract, including MD3-correct color tokens (`surfaceContainerHighest` for filled, `outlineVariant` for outlined borders) and real ripple/state-layer feedback via `TouchableRipple` instead of a bare `Pressable`.
- Publishes the v6 Card contract and a Paper 5 → 6 migration guide.

### Related issue

Addresses [#4929](https://github.com/callstack/react-native-paper/issues/4929).

### Screenshots / Videos

See the new "Expressive Card gallery" example (`example/src/Examples/CardExample.tsx`) for variants, states, shapes, and light/dark/platform verification, and `CardRenderCountExample.tsx` for the render-count regression check.

### Test plan

- `yarn jest` — full suite passes (58 suites / 737 tests), including the rewritten `Card`, `Card.tokens`, and `Surface` unit tests covering variants, states (hover/press/focus/drag/disabled), shapes, accessibility, and motion.
- `yarn typecheck` — passes, including the new `@ts-expect-error` cases asserting the removed `mode`/children API is rejected at compile time.
- `yarn lint` — passes.
- Manually exercised the `CardExample` gallery in the example app (variants, custom headers, responsive cover media, disabled/dragged states, light/dark theme, platform-specific pressed/long-press messaging).


src/components/Card/Card.tsx
```diff
@@ -1,263 +1,884 @@
...
+  variant: 'outlined';
+  /**
+   * Outlined Cards do not support custom elevation.
+   */
+  elevation?: never;
+};
+
+type CardVariantProps = FilledCardProps | ElevatedCardProps | OutlinedCardProps;
+
+export type Props = Omit<ViewProps, 'children' | 'style'> &
```
JKobrynski, line 178: #5076 drops `React.Children` using context but keeps `<Card><Card.Content/></Card>`. I'm afraid moving every region to a prop goes further than #4954 asks for - could Card take that route instead?

src/components/Card/Card.tsx
```diff
@@ -1,263 +1,884 @@
...
+  subtitle,
+  leading,
+  trailing,
+  content: cardContent,
+  actions,
   style,
   contentStyle,
   theme: themeOverrides,
-  testID,
+  testID = 'card',
```
JKobrynski, line 370: #5099 just removed the default `testID` and derived ids

src/components/Card/tokens.ts
```diff
@@ -0,0 +1,280 @@
...
+      containerColor: 'surfaceContainerHighest',
+      containerOpacity: 1,
+      outlineColor: 'outlineVariant',
+      outlineOpacity: 0,
+      outlineWidth: 0,
+      elevation: 3,
+      stateLayerOpacity: opacity.dragged,
+    },
+    disabled: {
+      containerColor: 'surfaceVariant',
```
JKobrynski, line 105: Shouldn't the disabled filled Card be grayed out rather than tinted? I mean `on-surface` instead of `surface-variant`?

docs/plugins/component-docs/parser.ts
```diff
@@ -206,22 +206,8 @@ const getProps = (
...
-
-          parent = parent.parent;
-        }
-
-        return false;
-      });
       const type =
-        (hasUnionDeclarations && new Set(declarationTypes).size > 1) ||
-        !declarations.includes(property)
+        new Set(declarationTypes).size > 1 || !declarations.includes(property)
```
JKobrynski, line 210: This changes prop types on every component page. Have you checked the others?

src/components/Card/CardActions.tsx
```diff
@@ -24,59 +22,32 @@ export type Props = ViewProps & {
...
-          ...child.props,
-          compact,
-          mode,
-          style: childStyle,
-        });
-      })}
-    </View>
-  );
-};
+const CardActions = ({ style, children, theme: _theme, ...rest }: Props) => (
```
JKobrynski, line 36: `Card.Actions` no longer sets the button modes, maybe it's worth a migration note?

JKobrynski:
And we've got conflicts

src/components/Card/Card.tsx
pawelkata: True, removed.

src/components/Card/tokens.ts
pawelkata: Nice catch! Should be fixed now.

src/components/Card/CardActions.tsx
pawelkata: Added.

PR #5124, opened 2026-09-15
fix(portal): forward the reduce motion preference and match queued update keys

konstmar:
### Motivation

**1. Reduce Motion is ignored inside a portal**

`Portal` re-provides the settings, locale and theme contexts to its children, but not `ReduceMotionContext`, so `useReduceMotion()` falls back to `false`. The fix forwards it too.

**2.** `PortalHost.update()` **could update the wrong portal**

While operations are queued, `update()` replaces a matching queued `mount` or `update` - but it only compared the key on updates, so it could overwrite a different portal's queued mount. The fix compares the key for both.

### Related issue

Context: [Notion](https://app.notion.com/p/callstack/Add-an-accessible-overlay-Portal-foundation-3ac5d027c0f880668bdfcb6cc4ae7c59)

### Screenshots / Videos

Nothing visual to capture.

### Test plan

- `yarn test src/components/__tests__/Portal.test.tsx` - 5 passed, including a new test asserting `useReduceMotion()` reads `true` inside a `Portal`. Fails on `main`.
- `yarn typecheck`, `yarn lint`, `yarn test` (680 passed) - all pass.


k0ndee (approved):
LGTM! @satya164

satya164 (approved):
thanks

PR #5125, opened 2026-09-15
fix(modal): make Modal accessible and honour its dismissal props

konstmar:
### Motivation

This PR improves some aspects of Paper's Modal accessibility.

- `dismissableBackButton` is now honoured on its own - hardware back read `dismissable || dismissableBackButton`, so `dismissableBackButton={false}` did nothing unless `dismissable` was also `false`
- Adds `aria-label`  prop to Modal. `role` prop should be added in this one instead #5072
- The escape gesture (iOS) no longer closes a modal with `dismissable={false}`
- Hides the backdrop/scrim from assistive technology with `dismissable={false}`

### Related issue

[Notion](https://app.notion.com/p/callstack/Add-an-accessible-overlay-Portal-foundation-3ac5d027c0f880668bdfcb6cc4ae7c59)

### Screenshots / Videos

Screen reader speech is shown on screen in each clip (VoiceOver caption panel / TalkBack "Display speech output").

| | Before (`main`) | After |
| --- | --- | --- |
| Escape gesture on `dismissable={false}` (iOS) | https://github.com/user-attachments/assets/bc547e31-9201-477e-b42c-3c4edf9e3a1c| https://github.com/user-attachments/assets/d90f9022-494b-4643-aa93-47b525f52314 |
| Hardware back on `dismissableBackButton={false}` (Android) | https://github.com/user-attachments/assets/4dcbad7a-d1cc-4d10-9339-8a08482f9c50 | https://github.com/user-attachments/assets/013b9aac-2fa0-45cc-aa30-6b587d9e93ed |

### Test plan

`yarn test`, `yarn lint`, `yarn typecheck` pass. In the example app:

1. Swipe through a dialog with a screen reader - focus skips the backdrop, tapping outside still dismisses.
2. "Undismissable Dialog", escape gesture - stays open.
3. "Undismissable back button" (new example, Android), hardware back - stays open.
4. On web, focus a dialog - `aria-label` and role announced.


src/components/Modal.tsx
k0ndee: dismissable isn't read inside this effect anymore after the fix, You can drop it from the deps array

src/components/Modal.tsx
```diff
@@ -227,18 +233,16 @@ function Modal({
   return (
     <Animated.View
       pointerEvents={visible ? 'auto' : 'none'}
-      aria-modal
       aria-live="polite"
       style={StyleSheet.absoluteFill}
-      onAccessibilityEscape={onDismissCallback}
+      onAccessibilityEscape={dismissable ? onDismissCallback : undefined}
```
k0ndee, line 238: `Modal.tsx` has missing focus-related code that is one of the acceptance criteria for this ticket.

- **Initial focus** - nothing moves focus into the `Surface` when `visible` becomes `true`
- **Focus trap** - nothing intercepts Tab to keep focus inside the modal while open; a keyboard or switch-control user can tab straight into the content behind it
- **Focus restore** - nothing captures the pre-open focus target or restores it in `onDismissCallback` on close

Given the ticket calls this out as the hard, VoiceOver/TalkBack/keyboard/switch-control-verified part of the work, will there be another follow-up PR for this, or should it be included here? @satya164 what do You think about it?

src/components/Modal.tsx
```diff
@@ -227,18 +233,16 @@ function Modal({
...
       testID={testID}
     >
       <AnimatedPressable
-        aria-label={overlayAccessibilityLabel}
         role="button"
```
k0ndee, line 243: The `role="button"` is now a no-op due to the `aria-hidden` prop added to `AnimatedPressable`, it can be removed now.

k0ndee (commented):
Now that `Modal` accepts `aria-label`/`role`, `Dialog` doesn't forward either to it - its `Props` type has no `aria-label`/`role` field at all. So after this merges, `<Dialog>` usage still has no accessible name. Will that be wired up in the Dialog-adoption follow-up task, or should it land here since the plumbing already exists?

src/components/Modal.tsx
konstmar: There will be a follow-up PR with initial focus / restore focus.
Regarding focus trap - https://github.com/callstack/react-native-paper/pull/5126 this one covers it.

src/components/Modal.tsx
konstmar: @k0ndee updated

src/components/Modal.tsx
konstmar: @k0ndee updated

konstmar:
> Now that `Modal` accepts `aria-label`/`role`, `Dialog` doesn't forward either to it - its `Props` type has no `aria-label`/`role` field at all. So after this merges, `<Dialog>` usage still has no accessible name. Will that be wired up in the Dialog-adoption follow-up task, or should it land here since the plumbing already exists?

@k0ndee  I believe there should be an adoption task for Dialog, but I can add this small change here instead.

k0ndee:
> > Now that `Modal` accepts `aria-label`/`role`, `Dialog` doesn't forward either to it - its `Props` type has no `aria-label`/`role` field at all. So after this merges, `<Dialog>` usage still has no accessible name. Will that be wired up in the Dialog-adoption follow-up task, or should it land here since the plumbing already exists?
> 
> @k0ndee I believe there should be an adoption task for Dialog, but I can add this small change here instead.

Actually, it looks like #5072 already forwards `aria-label` from `Dialog` to `Modal`, so that half is covered there. `role` isn't wired up in #5072 though, and its `Modal` usage still predates this PR, so it'll need a rebase against this once it merges regardless. Given that, probably cleanest to leave `role` out of this PR and pick it up as part of that rebase in #5072 instead, so it doesn't end up half-done in one PR and forgotten in the other. @kachmashk, flagging so it's on your radar once this lands.

k0ndee (approved):
LGTM! @satya164

src/components/Modal.tsx
```diff
@@ -227,18 +228,15 @@ function Modal({
...
       style={StyleSheet.absoluteFill}
-      onAccessibilityEscape={onDismissCallback}
+      onAccessibilityEscape={dismissable ? onDismissCallback : undefined}
       testID={testID}
     >
       <AnimatedPressable
-        aria-label={overlayAccessibilityLabel}
-        role="button"
         disabled={!dismissable}
+        aria-hidden
```
satya164: Why is overlay hidden from a11y with aria-hidden?

If a dialog doesn't have a button to close itself, doesn't mean screen reader users and keyboard users won't be able to close it at all as compared to mouse users?

src/components/Modal.tsx
```diff
@@ -251,6 +249,9 @@ function Modal({
         pointerEvents="box-none"
       >
         <Surface
+          role="dialog"
+          aria-modal
+          aria-label={ariaLabel}
```
satya164: i'm not sure I understand the use case for `aria-label` on the entire modal. shouldn't the user read the content of the modal/dialog?

src/components/Modal.tsx
```diff
@@ -33,9 +33,9 @@ export type Props = {
    */
   onDismiss?: () => void;
   /**
-   * Accessibility label for the overlay. This is read by the screen reader when the user taps outside the modal.
+   * Accessible name for the modal.
    */
-  overlayAccessibilityLabel?: string;
```
satya164, line 38: with `overlayAccessibilityLabel` removed, how can user customize aria label for the overlay or access it in tests (there is test id, but better practice is to use a11y, same way your users will)?

src/components/Modal.tsx
```diff
@@ -33,9 +33,9 @@ export type Props = {
...
+  'aria-label'?: string;
```
satya164: other components use the `accessibilityLabel` naming scheme. if it's changed to `aria-*`, it should be a separate ticket and handle across all components.

satya164 (changes requested):

src/components/Modal.tsx
konstmar: @satya164  I made the decision to not make it accessible based on a few things:
1) It was not accessible on android in the first place (via importantForAccessibility)
2) The task suggested removing the label since it was not accessible
3)  https://m3.material.io/components/dialogs/guidelines 's anatomy shows that Dialog should have actions.
4)  https://m3.material.io/components/dialogs/accessibility says the initial focus should land on the first interactive element **within** dialog, not on scrim. And the scrim is just a visual dimmer.
5) https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/ : 
     - "All elements required to operate the dialog are **descendants** of the element that has role dialog."
     - "Tab and Shift + Tab **do not move focus outside** the dialog" 
     - "It is strongly recommended that the tab sequence of all dialogs **include a visible element with role button that closes the dialog**, such as a close icon or cancel button." 



The screen reader users and keyboard users can still use Escape/Back to initiate close.

src/components/Modal.tsx
konstmar: @satya164 please check my comment above. It will not be needed if we move forward with not making it accessible

src/components/Modal.tsx
satya164: > Dialog should have actions.

It should, but material design guidelines are still just guidelines. They are not requirements the users must implement. Also this affects modals in general, not just dialogs.

> says the initial focus should land on the first interactive element within dialog, not on scrim.

What's scrim? Overlay? That sounds right, but it could be achieved without making the overlay inaccessible.

> And the scrim is just a visual dimmer.

But the overlay is not just a visual dimmer here, users can already interact with it using a mouse, which makes it an interactive element.

Every interactive element should have equivalent for screen readers. If it's not the overlay itself, then there needs to be a screen reader specific affordance to close the modal like a mouse user would.

> "All elements required to operate the dialog are descendants of the element that has role dialog."
> "It is strongly recommended that the tab sequence of all dialogs include a visible element with role button that closes the dialog, such as a close icon or cancel button."

Still, the problem is these are guidelines and recommendations. While we can implement the recommendations in the library itself, the user may not. We should still provide an accessible experience by default.

Prior design to this would be what React Aria does. It hides the overlay from a11y, but also provides a screen-reader only affordance.

Following is from Codex:

> When outside-click dismissal is enabled, React Aria automatically adds a visually hidden dismiss button inside the modal. Screen reader users can find and activate it even if the app doesn't provide a close button. It has a localized label and tabIndex={-1}, so it doesn't add an extra Tab stop. Keyboard users can dismiss with Escape. ([Modal implementation](https://github.com/adobe/react-spectrum/blob/be5bcc375f384227e7d048060205a500a1f61556/packages/react-aria-components/src/Modal.tsx#L327), [dismiss button](https://github.com/adobe/react-spectrum/blob/be5bcc375f384227e7d048060205a500a1f61556/packages/react-aria/src/overlays/DismissButton.tsx))
> We could follow the same approach. If we hide the overlay from a11y, we should provide an equivalent dismiss button for screen reader users whenever clicking the overlay can close the modal. We also need Escape handling on web, which is currently missing.

konstmar: @satya164  Got it. I'll revert the changes and make overlay accessible again when modal is dismissable.

src/components/Modal.tsx
konstmar: Restored the label

PR #5126, opened 2026-09-15
feat(portal)!: hide content below an overlay from assistive technology

konstmar:
### Motivation

With a dialog open, a screen reader can still wander into the app content behind it.

Adds an opt-in `modal` prop to `Portal`. A portal marked as a modal hides every layer below it - the app content and any portal mounted earlier - from assistive technology and from the web focus order. Portals mounted on top stay reachable, so a menu or snackbar above a dialog still works.

`Modal` now renders itself in a Portal with `modal` prop tied to visible. Other overlay components will adopt it separately.

### Breaking:
 `Modal` and `Dialog` render themselves in a `Portal`, so they now need a `Portal.Host` above them in the tree.

### Related issue

[Notion](https://app.notion.com/p/callstack/Add-an-accessible-overlay-Portal-foundation-3ac5d027c0f880668bdfcb6cc4ae7c59)

### Screenshots / Videos

https://github.com/user-attachments/assets/42c08ce7-48fc-4fb1-a551-617db38c7e6e

https://github.com/user-attachments/assets/38e0a06e-76ba-4cdf-98bf-66268c55bab4

### Test plan

`yarn test`, `yarn lint`, `yarn typecheck` pass. Six new `Portal` tests.

On a device, open any dialog in the example app and keep swiping past its last button with a screen reader, or hold Tab on web: focus stays inside the dialog, and the screen behind it is reachable again once it closes.


src/components/Portal/PortalManager.tsx
```diff
@@ -1,32 +1,39 @@
 import * as React from 'react';
-import { View, StyleSheet } from 'react-native';
+import { StyleSheet } from 'react-native';
+
+import OverlayLayer from './OverlayLayer';
+
+type Props = {
+  pageContent?: React.ReactNode;
```
satya164: `pageContent` is misleading name as a portal provider doesn't render a page. it should just be `content` or `children`.

it should also not be optional.

```suggestion
  children: React.ReactNode;
```

src/components/Portal/Portal.tsx
```diff
@@ -8,13 +8,24 @@ import {
...
+  /**
+   * Whether this portal hides everything below it -- the app content and any
+   * portal mounted before it -- from screen readers and the focus order.
+   * Portals mounted after it stay reachable.
+   *
+   * Tie it to whether the overlay is open rather than to how long it stays
+   * painted: a layer gives the screen back the moment it starts closing, so
+   * what is underneath is reachable again while the overlay fades out.
+   */
+  overlay?: boolean;
```
satya164: `modal` maybe a better name instead `overlay`

src/components/Portal/PortalManager.tsx
```diff
@@ -38,17 +45,43 @@ export default class PortalManager extends React.PureComponent<{}, State> {
...
-    ));
+    const { portals } = this.state;
+
+    const topmostOverlayIndex = portals.findLastIndex(
+      (portal) => portal.overlay
+    );
+
+    return (
+      <>
+        {/* Need collapsable=false here to clip the elevations, otherwise they appear above Portal components */}
```
satya164: move the comment to `collapsable` like the other one

src/components/Portal/Portal.tsx
```diff
@@ -8,13 +8,24 @@ import {
...
   children: React.ReactNode;
+  /**
+   * Whether this portal hides everything below it -- the app content and any
+   * portal mounted before it -- from screen readers and the focus order.
+   * Portals mounted after it stay reachable.
+   *
+   * Tie it to whether the overlay is open rather than to how long it stays
+   * painted: a layer gives the screen back the moment it starts closing, so
+   * what is underneath is reachable again while the overlay fades out.
+   */
```
satya164, line 23: please write the comment by hand. don't use claude for documentation.


```suggestion
  /**
   * Whether the portal hides items below it from screen readers and focus order.
   *
   * Ensure it's set to true only when the modal is open.
   */
```

example/src/Examples/Dialogs/DialogWithOverlay.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <Button textColor={Palette.tertiary50} disabled>
+          Disagree
+        </Button>
+        <Button onPress={close}>Agree</Button>
+      </Dialog.Actions>
+    </Dialog>
+  </Portal>
+);
+
+export default DialogWithOverlay;
```
satya164: I don't think a separate example is needed. all dialogs need to be used this way. update all the existing examples

satya164 (changes requested):

example/src/Examples/Dialogs/DialogWithOverlay.tsx
konstmar: Updated. Now Modals render the Portal inside them

src/components/Portal/PortalManager.tsx
konstmar: updated

src/components/Portal/Portal.tsx
konstmar: Renamed to `modal`

src/components/Portal/PortalManager.tsx
konstmar: Updated

